PR #352, opened 2026-05-03
[codex] Block NEW action for existing GOA terms

cmungall:
## Summary

- Reject `action: NEW` when the GO ID already appears in the local GOA file under any evidence/reference source, including TreeGrafter/IBA (`GO_REF:0000033`).
- Promote this GOA conflict from a warning to an error in integrated validation so `just validate` blocks it.
- Add regression coverage for a `NEW` annotation that uses a different evidence/reference than the existing GOA annotation, plus the integrated `validate_gene_review` path.

## Root Cause

The strict GOA validator previously checked `NEW` annotations against the full tuple `(GO term, evidence, reference, negation)`. That allowed a term already present in GOA to be marked as novel when the review used a different evidence code or PMID.

## Validation

- `uv run pytest tests/test_goa_validator.py tests/test_validator.py`
- `uv run ruff check src/ai_gene_review/validation/goa_validator.py src/ai_gene_review/validation/validator.py tests/test_goa_validator.py`
- Spot check: `uv run ai-gene-review validate genes/DROME/Nprl2/Nprl2-ai-review.yaml --verbose` now fails with the expected GOAValidator error for `GO:0005096` from `IBA, GO_REF:0000033`.

## Note

This blocks `NEW` conflicts against local `*-goa.tsv` files. It does not yet detect TreeGrafter/projection provenance that is absent from the local GOA TSV; that would need a separate validation source.


cmungall:
Follow-up pushed in `195b65570`.

What changed:
- Removed duplicate `action: NEW` review blocks where the GO term already appears in the gene's GOA source data.
- Added missing local source files for MIR155 and DNAAF1 so validation has the GOA/UniProt inputs it needs.
- Repaired MIR155 and DNAAF1 metadata to align existing annotations with their GOA references.

Validation run locally:
- `uv run ai-gene-review validate <12 listed review files>`
- `uv run linkml-validate --schema src/ai_gene_review/schema/gene_review.yaml --target-class GeneReview <each listed file>`
- `uv run linkml-term-validator validate-data <each listed file> -s src/ai_gene_review/schema/gene_review.yaml -t GeneReview --labels -c conf/oak_config.yaml`
- `scripts/run_reference_validator.sh validate data <each listed file> --schema src/ai_gene_review/schema/gene_review.yaml --target-class GeneReview --config conf/reference_validator_config.yaml`
- `git diff --check`

cmungall:
CI mypy fix pushed in `3dd95d95f`.

Root cause:
- `mypy.ini` targets Python 3.10, so `enum.StrEnum` was not available to mypy even though runtime Python is newer.
- PyYAML-loaded registry values were typed as `object`, so mypy rejected iterating/casting them without shape narrowing.

Local verification:
- `uv run mypy src tests --exclude "src/ai_gene_review/(datamodel|tools|export)/"`
- `uv run pytest tests/test_defense_go.py`
- `uv run ruff check src/ai_gene_review/defense_go/models.py src/ai_gene_review/defense_go/registry.py tests/test_defense_go.py`
- `git diff --check`

GitHub checks are currently pending on the new commit.

PR #429, opened 2026-05-09
Review DANRE batch 03 genes

cmungall:
## Summary
- Add DANRE reviews for uox, slc5a12, hs6st2, glceb, fads2, sult1st2, gmppb, coq8a, ldhd, and slc25a37.
- Include fetched GOA/UniProt/publication/PANTHER context plus rendered HTML pages.
- Add only the missing GO term cache rows needed by these reviews.

## Validation
- Per-gene `just validate DANRE <gene>` passed for all 10 genes.
- Per-gene `just render DANRE <gene>` passed for all 10 genes.
- `git diff --cached --check` passed before commit, excluding generated publication/HTML/PANTHER CSV outputs.


cmungall:
Addressed review feedback in 79440ffc0:

- coq8a: changed both GO:0004672 protein kinase annotations, including the NOT annotation, to KEEP_AS_NON_CORE so negative evidence is retained and no NOT annotation is converted to a positive replacement.
- fads2: added a second core function for GO:0062076 acyl-CoA (8-3)-desaturase activity and updated notes/rendered HTML.
- hs6st2: captured the Golgi membrane point as a suggested curation question instead of adding an unsupported NEW annotation, since the zebrafish source file only says membrane.

Re-ran just validate/render for coq8a, fads2, and hs6st2.

PR #447, opened 2026-05-10
Add PSEPK new review batch 1

cmungall:
Adds 10 new Pseudomonas putida KT2440 reviews as the first true-new PSEPK batch:

- murU
- hmgA
- alr
- nicA
- nicB
- nicC
- nicR
- chrR
- mcpS
- pvdT

Validation/rendering run locally:

- `just validate PSEPK murU`
- `just validate PSEPK hmgA`
- `just validate PSEPK alr`
- `just validate PSEPK nicA`
- `just validate PSEPK nicB`
- `just validate PSEPK nicC`
- `just validate PSEPK nicR`
- `just validate PSEPK chrR`
- `just validate PSEPK mcpS`
- `just validate PSEPK pvdT`
- `just render PSEPK <gene>` for all 10 genes

Notes:
- hmgA removes apparent hmgR-like transcription-regulation annotations from the HmgA dioxygenase review.
- pvdT records the transporter-complex relationship via `in_complex` rather than as a location.
- This PR counts as 10 only after review, green CI, and merge.

PR #449, opened 2026-05-10
Add PSEPK new review batch 2

cmungall:
Adds 10 true-new PSEPK reviews: hutT, pcaY, quiC1, fadB, earP, mupP, thiO, rpoN, lexA2, glmU.

Validation/rendering:
- Ran `just validate PSEPK <gene>` for all 10; fixed the mupP reference-section enum and the lexA2 core-function validation warning.
- Ran `just render PSEPK <gene>` for all 10.
- Rebased after #447; overlapping generated PANTHER support files already on main were kept from main.

Note: pedE was excluded because it maps to existing lowercase `genes/PSEPK/pede`; glmU was used instead.


cmungall:
Addressed all three ai4c-agent suggestions in 30d5242f7: replaced thiO's weak cytoplasm support text with the actual UniProt GO cytoplasm line and explicit non-core wording; added an EarP NEW GO:0006414 translational elongation process annotation plus core-function involvement; aligned lexA2 GO:0006281 DNA repair with the DNA replication over-annotation decision. Re-ran `just validate PSEPK thiO`, `just validate PSEPK earP`, `just validate PSEPK lexA2`, and rendered all three.

PR #482, opened 2026-05-11
Add MISSI review batch 5

cmungall:
Adds the next 10 Miscanthus sinensis review files from the MISSI target list:

- psbJ, matK, psbM, petL, petD, psbH, rpoB, rps16, petA, rpl22

Validation/rendering performed locally:

- `just validate MISSI <gene>` for all 10 genes
- `just render MISSI <gene>` for all 10 genes
- `git diff --cached --check` before commit

All 10 GOA files are header-only, so `existing_annotations: []` is intentional. Core functions are supported by cached UniProt records and notes avoid converting UniProt/InterPro/HAMAP/TreeGrafter terms into `NEW` annotations.


cmungall:
Addressed the review comments in `0f47eae0e`:

- matK now uses `GO:0000373` / `Group II intron splicing`. I checked the local validator after the review suggestion: `GO:0000372` validates as `Group I intron splicing`, so `GO:0000373` is the validated Group II term.
- Removed unsupported `GO:0010242` contributes_to entries from psbJ, psbM, and psbH.
- Removed the raw GO-line support text from psbH.
- Made petA direct `molecular_function` = `GO:0009055` electron transfer activity.

Revalidated and re-rendered all affected files locally.

PR #486, opened 2026-05-11
Add MISSI gene reviews batch 8

cmungall:
## Summary
- Adds 10 MISSI gene reviews: ndhG, ccsA, rpl33, rps8, atpI, psaI, clpP, petG, rpl36, rpl32
- Adds UniProt/GOA support files, curator notes, rendered HTML, and fetched PANTHER support metadata for newly encountered families
- Keeps `existing_annotations: []` because all 10 GOA files are header-only; UniProt/InterPro/HAMAP/TreeGrafter predictions are used only as support for synthesized core functions, not as NEW existing annotations

## Validation
- `just validate MISSI <gene>` for all 10 genes
- `just render MISSI <gene>` for all 10 genes
- `git diff --check`


cmungall:
Addressed review comments in latest commit:

- ndhG now uses GO:0008137 (NADH dehydrogenase (ubiquinone) activity) instead of GO:0050136, matching the electrogenic/proton-translocating chloroplast NDH evidence and the UniProt GO cross-reference.
- ndhG notes document why GO:0050136 is not appropriate here.
- psaI notes now explicitly retain GO:0015979 because PsaI is a constitutive PSI subunit, not a dedicated assembly factor; also notes that GO:0010207 is photosystem II assembly and not applicable.
- Revalidated and re-rendered ndhG locally;  is clean.

PR #689, opened 2026-05-30
EPYC: enrich review with falcon deep research evidence

cmungall:
Enrich the EPYC review with new falcon deep research findings. Validation passes (`just validate human EPYC`).


cmungall:
Addressed: removed GO:0050431 (TGF-β binding) NEW annotation and the corresponding core_function entry. No primary EPYC-specific biochemistry supports it; the SLRP family inference is not strong enough to assert. Validation passes.

cmungall:
Addressed both:
1. Removed the TGF-β signaling clause from the description's core-functions list (now reads (1) collagen fibril regulation; (2) ECM architecture).
2. Removed the dangling `proposed_new_terms` forward references from the two TGF-β removal comments.

Validation passes.

PR #697, opened 2026-05-30
FAS: enrich review with falcon deep research evidence

cmungall:
Enrich the FAS review with new falcon deep research findings. Validation passes (`just validate human FAS`).


cmungall:
Addressed: removed the two misattributed supporting_by entries (PMID:40330016 and the falcon mFASL/sFASL quote) from GO:0043066. Those described FASLG ligand biology, not FAS soluble-decoy-receptor isoform activity. PMID:7510905 remains as the primary support. Validation passes.

PR #773, opened 2026-05-30
ISM2: enrich review with falcon deep research evidence

cmungall:
Enrich the ISM2 review with new falcon deep research findings. Validation passes (`just validate human ISM2`).


cmungall:
Addressed in 91434d383:

**All three empty-findings references now populated** with verbatim supporting text from the cached publications:
- **PMID:36604494**: Documents the distinctive observation that ISM2 was the only CMP protein common to both the second-trimester (AUC 0.83) and third-trimester (AUC 0.78) PAS classifier panels.
- **PMID:37047617**: Captures the sAPPα-mediated ISM2 up-regulation in SH-SY5Y cells and the nucleus accumbens expression context.
- **PMID:38322491**: Captures ISM2 downregulation in the villous core stromal compartment of SARS-CoV-2-infected placentas.

**core_functions MF**: Removed GO:0005515 protein binding per CLAUDE.md guidance against using generic protein binding as a core molecular function. The description now explicitly notes that the MF slot is left unspecified pending direct biochemical validation, with extracellular matrix binding / heparin binding / integrin binding all as plausible candidates given the TSR1+AMOP domain architecture.

PR #1105, opened 2026-06-01
Ghr (rat): enrich review with falcon deep research evidence

cmungall:
Integrates falcon (Edison Scientific) deep research into the rat Ghr review: adds a file: reference with verbatim findings and supported_by evidence; revisits action calls. Validation passes.


cmungall:
Addressed review comments:

- **CRITICAL (1) — duplicate `supported_by` key on GO:0009629 (response to gravity):** Merged the two consecutive `supported_by` mappings into a single list with both items, so the falcon file reference is no longer silently dropped by YAML parsers. Verified the parsed YAML now yields one `supported_by` with 2 entries (falcon file + PMID:14638460).
- **IMPORTANT (2) — GO:0008289 (lipid binding) supporting text mismatch:** Replaced the JAK2-coupling passage with the falcon synthesis statement ("rat Ghr encodes a non-enzymatic cytokine receptor whose central biochemical role is to organize and activate JAK2 at the plasma membrane in response to GH binding"), which directly supports the MARK_AS_OVER_ANNOTATED call that GHR's molecular function is GH binding + JAK2 activation, not lipid binding. Quote verified verbatim in the falcon deep research file.
- **SUGGESTION (4) — GO:0009897 (external side of plasma membrane):** Replaced the weak mRNA-expression quote (PMID:2722883) with a falcon passage describing the extracellular ligand-binding region facing the external side. Quote verified verbatim.

**Rejected suggestion (3) — GO:0005615 "obsolete" claim:** Not applied. The bot stated GO:0005615 "extracellular space" is *not* obsolete. This is incorrect: OLS confirms GO:0005615 is **OBSOLETE** ("OBSOLETE. That part of a multicellular organism outside the cells proper..."). The replacement GO:0005576 (extracellular region) is valid and non-obsolete (verified via OLS). The existing MODIFY entries replacing the obsolete term with the valid extracellular region term are therefore correct and left unchanged.

`just validate rat Ghr` passes clean (schema, term, reference, and best-practices checks).

PR #1150, opened 2026-06-01
mbiA (ECOLI): enrich review with falcon deep research evidence

cmungall:
Integrates falcon (Edison Scientific) deep research into the E. coli mbiA review: adds a file: reference with verbatim findings and supported_by evidence; revisits action calls. Validation passes.


cmungall:
Addressed review comments:

**1. (BLOCKING) GO:0042710 biofilm formation term direction** — Took the bot's offered option 2: kept the generic parent term and added explicit justification in `review.reason` for why the directional negative-regulation term is not used. Per the Fellner et al. full text, BOTH the htgA-frame mutant AND the antisense yaaW-frame mutant show *increased* biofilm, so the direction cannot be cleanly attributed to mbiA (confounded by the overlapping yaaW gene); the MbiA protein is also undetectable by Western blot, so there is no positive evidence it acts as a repressor.
  - **Note on the suggested ID:** the bot suggested `GO:0052032` ("negative regulation of biofilm formation"). Verified via OLS that **GO:0052032 is actually "symbiont-mediated perturbation of host inflammatory response"** — an unrelated term, so it was NOT applied. The correct single-species term would be `GO:1900191` (negative regulation of single-species biofilm formation), which I reference in the justification but do not annotate, for the biological reasons above.

**2. (BLOCKING) Contradictory Falcon evidence in core_functions** — Fixed. Replaced the Falcon `supported_by` text in `core_functions[0]` (which said "no...subcellular localization has been experimentally established" while supporting a cytoplasm `locations` claim) with the Falcon statement that "no enzymatic activity, binding partner, or defined signaling pathway is experimentally assigned to MbiA/HtpY" — this genuinely supports the description's "no known molecular function / no molecular mechanism" assertion. Quote grep-verified against the Falcon report.

**3. (SUGGESTION) Irrelevant PMID supporting_text for GO:0005737 cytoplasm** — Fixed. Removed the topically-irrelevant "lineage-specific...orphan gene" PMID:24111745 quote. The reason now explicitly states cytoplasm is a default inference (not a positive assignment), and the retained Falcon "localization remains undetermined" quote is framed as documenting absence of contradicting evidence, not positive support.

**4. (SUGGESTION) GO:0003674 root MF as NEW** — Retained as-is. For a poorly-characterized orphan gene with genuinely no recognizable domain, enzymatic activity, or characterized function, a root MF "function unknown" annotation is an appropriate conservative curation choice and is the honest representation of current knowledge.

`just validate ECOLI mbiA` passes clean (schema, term, reference, and best-practices checks all pass).

PR #1157, opened 2026-06-01
HdeB (ECOLI): enrich review with falcon deep research evidence

cmungall:
Integrates falcon (Edison Scientific) deep research into the E. coli HdeB review: adds a file: reference with verbatim findings and supported_by evidence; revisits action calls. Validation passes.


cmungall:
Addressed review comments:

**1. GO:0140309 label inconsistency (🟡 IMPORTANT) — fixed.** Verified the term via OLS MCP. There is a cache-vs-OLS discrepancy worth flagging: OLS currently returns **GO:0140309 = "unfolded protein holdase activity"**, but the local OAK term cache (cache/go/terms.csv, the source the validator uses) has the most-recent row as **"unfolded protein carrier activity"** (2026-05-08). To keep `just validate` passing, I made the label consistent as **"unfolded protein carrier activity"** across all structured fields AND the two `reason` narratives (the narratives previously said "holdase activity"). The mechanistic "holdase chaperone" wording is retained in prose, but the official term name is now used uniformly. When the cache catches up to OLS this can be re-synced to "holdase activity".

**2. Missing primary references (🟡 IMPORTANT) — fixed.** Looked up both DOIs via PubMed (no guessing): Dahl et al. 2015 JBC (doi:10.1074/jbc.M114.612986) = **PMID:25391835**; Ding et al. 2015 Sci Rep (doi:10.1038/srep16856) = **PMID:26593705**. Cached both under `publications/`, added them to `references` with verbatim findings, wired them into `core_functions.supported_by` (pH-4 optimum / dynamic-dimer claims now trace to primary literature), and added the PMIDs to the `description`. `PMID:26593705` has full text in PMC; `PMID:25391835` is abstract-only (PMC blocks full-text), but the cited supporting_text is verbatim from the abstract and passes reference validation.

**3. PMID:17085547 pH-characterization caution (🔵 SUGGESTION) — applied.** Added a clarifying note to the GO:0010447 IDA `summary` flagging that Kern 2007's pH-3 framing was later refined by Dahl/Ding (optimum ~pH 4-5; full monomerization at pH 3 not required), while noting the response-to-acidic-pH annotation remains correct.

No suggestions were rejected — all three review items were verified and applied. `just validate ECOLI HdeB` passes clean (schema, term, reference, and best-practices checks).

PR #1159, opened 2026-06-01
AT5G03720 (ARATH): enrich review with falcon deep research evidence

cmungall:
Integrates falcon (Edison Scientific) deep research into the Arabidopsis AT5G03720 review: adds a file: reference with verbatim findings and supported_by evidence; revisits action calls. Validation passes.


cmungall:
Addressed review comments (validation passes clean: schema, term, reference, best-practices all green):

- **[IMPORTANT 1] GO:0006355 (IEA) misapplied falcon texts** — Removed the two falcon `supported_by` quotes that described DREB2A controlling HSFA3's *own* transcription (upstream regulation) and replaced them with the falcon quote that actually describes HSFA3's regulatory function on its targets: "HSFA3 is a sequence-specific DNA-binding transcription factor whose proximate molecular function is to activate transcription of heat-protective genes (notably HSPs) by binding HSEs in their promoters." (verified present verbatim in the falcon report). The Schramm-2008 "potent activator on the promoters of Hsp genes" text is already cited on the parallel IDA GO:0006355 annotation.

- **[IMPORTANT 2] nucleus IDA (PMID:18261981)** — Verified against the cached publication: the Yoshida 2008 BBRC record reports HSFA3 transactivation/thermotolerance and DREB2A-dependent gene expression but contains **no subcellular-localization assay** (grep for nucle/localiz/GFP/fraction returns nothing; full text unavailable). Removed the title-only `supporting_text` and the self-caveating falcon quote ("Direct HSFA3 localization imaging experiments were not retrieved") from `supported_by`, since neither is positive localization evidence. Kept the annotation as **ACCEPT** (the nucleus term is correct and consistent with the IBA/IEA nucleus annotations) but rewrote the reason to flag that the IDA evidence code is only weakly substantiated by the cited paper, with nuclear action robustly supported by class A HSF architecture (NLS/NES) and HSE-promoter binding. NB: I could not change `evidence_type` to ISS because the validator requires `existing_annotations` evidence codes to match the GOA file.

- **[Suggestion 4] NEW GO:0045944** — Replaced the circular "not present in existing_annotations" rationale with the direct biological evidence in Schramm 2008 (PMID:17999647): HSFA3 is a potent activator of Hsp gene promoter::GUS reporters with EMSA-confirmed binding. Set `evidence_type: ISS` with `original_reference_id: PMID:17999647` and `supported_by`.

- **[Suggestion 5] core_functions[0]** — Reworded to foreground DREB2A-dependent HSP-gene activation / acquired thermotolerance as the primary function, with the HSFA2-HSFA3 chromatin-memory role noted as the layered secondary function (already its own core_function entry).

**Rejected suggestion (with verified reason):**
- **[Suggestion 3] GO:0046982 → GO:0051291** — NOT applied. Verified via OLS: **GO:0051291 (protein heterooligomerization) is a biological_process** (ancestors: cellular component assembly → protein-containing complex assembly), **not a molecular function**. It is therefore invalid in the `molecular_function` slot and as a replacement for the GO:0005515 protein-binding MF (the validator rejects it: "Value 'GO:0051291' not in dynamic enum 'GOMolecularActivityEnum'"). GO has no "protein heterooligomerization *activity*" MF term, so GO:0046982 (protein heterodimerization activity) remains the most specific applicable MF for the non-identical HSF-subunit interactions. I added a note in the GO:0005515 reason acknowledging the trimeric stoichiometry and the lack of a stoichiometric MF term.

PR #1167, opened 2026-06-01
CER1 (ARATH): enrich review with falcon deep research evidence

cmungall:
Integrates falcon (Edison Scientific) deep research into the Arabidopsis CER1 review: adds a file: reference with verbatim findings and supported_by evidence; revisits action calls. Validation passes.


cmungall:
Addressed review comments:

- **GO:0005515 protein binding (IMPORTANT — evidence mismatch):** Changed action from `MODIFY → GO:0042803 protein homodimerization activity` to `MARK_AS_OVER_ANNOTATED`. As the bot noted, the prior `supported_by` (ER-localization quote + CER3/CYTB5 heterocomplex text) did not support self-association. UniProt does record `SUBUNIT: Homodimer {ECO:0000269|PubMed:22773744}`, but the cached PMID:22773744 text contains no verbatim homodimer/dimer statement and the falcon report has none either, so there is no verbatim quote to justify proposing GO:0042803. Rather than assert homodimerization without direct evidence, the generic term is now correctly marked as over-annotated, with `supported_by` grounded in the actual IPI interaction it documents: PMID:38456566 ('CYP96A4 interacted with CER1 and CER3, the core enzymes in the alkane-forming pathway associated with wax biosynthesis') plus the falcon CER3/CYTB5 complex text. CER1's informative MF is already annotated as GO:0071771.

- **GO:0005739 mitochondrion REMOVE reason (IMPORTANT — uncited PMID):** Removed PMID:19892830 from the reason text. It is not in the publications cache (unverifiable) and was not in `references:`. The reason now relies solely on PMID:22773744, which is cached and already in `references:` and documents experimental ER localization.

- **GO:0010025 / PMID:1847001 (asbestos paper) UNDECIDED:** Left unchanged — the bot explicitly approved this call as correct. The remaining best-practices WARNING (ACCEPT vs UNDECIDED for the same term) is the expected artifact of flagging the mis-referenced GOA duplicate and was not touched.

Suggestions not applied (non-blocking): CHEBI:17176 (long-chain fatty aldehyde) is the canonical substrate term used in the UniProt RHEA:21440 catalytic-activity record, so it was retained; the CYP96A4-focused IEP supporting texts were left as-is (the paper characterizes CYP96A4's interaction with CER1, and the annotations remain KEEP_AS_NON_CORE).

`just validate ARATH CER1` passes clean (1 expected warning for the intentional GO:0010025 UNDECIDED).

PR #1190, opened 2026-06-01
benC (PSEPK): enrich review with falcon deep research evidence

cmungall:
Integrates falcon (Edison Scientific) deep research into the P. putida benC review: adds a file: reference with verbatim findings and supported_by evidence; revisits action calls. Validation passes.


cmungall:
Addressed review comments:

- **REJECTED (bot premise verified false): GO:0043640 `action: NEW` → `ACCEPT`.** The blocking item claims GO:0043640 (benzoate catabolic process via hydroxylation) is already in the GOA with IC evidence and so should be `ACCEPT`. Verified against `genes/PSEPK/benC/benC-goa.tsv`: the GOA contains exactly 5 MF annotations (GO:0008860, GO:0016491, GO:0018623, GO:0051536, GO:0051537) and **no GO:0043640**. The `evidence_type: IC` / `original_reference_id: PMID:12534466` fields on a NEW annotation document the curator-inferred basis for the proposed term, not an existing GOA record — this matches the established project convention (e.g. `genes/DORPE/TDO/TDO-ai-review.yaml` carries `evidence_type`/`original_reference_id` on its `action: NEW` entries). Per the schema, NEW means 'a proposed annotation, not one that exists in the existing GO annotations.' `action: NEW` is therefore correct and was retained.

- **DONE: Added FAD binding (GO:0050660).** Verified GO:0050660 (flavin adenine dinucleotide binding) via OLS. Since this term already exists in GO it does not belong in `proposed_new_terms` (that field is for novel terms needing proposed_name/definition); instead added it as a new `existing_annotations` entry with `action: NEW`, mirroring the GO:0043640 pattern. Supported by two verbatim falcon quotes on the FAD-binding domain and the monomeric reductase containing FAD (both grep-verified against `benC-deep-research-falcon.md`).

- **GO:0043640 supporting text (suggestion):** retained the benD diol-dehydrogenase→catechol quote. GO:0043640's GO definition is specifically 'breakdown of benzoate by hydroxylation to cis-1,2-dihydroxybenzoate followed by dehydrogenation to catechol' — the benD step is part of the term's defined pathway, so the quote is on-topic pathway context rather than tangential.

- CC localization gap (suggestion): the cytosolic call is only an inference in the falcon report; left as-is to avoid an unsupported CC annotation. Captured by the existing falcon finding in references.

`just validate PSEPK benC` passes clean (schema, term, reference, best-practices all green).

PR #1197, opened 2026-06-01
clpS (PSEPK): enrich review with falcon deep research evidence

cmungall:
Integrates falcon (Edison Scientific) deep research into the P. putida clpS review: adds a file: reference with verbatim findings and supported_by evidence; revisits action calls. Validation passes.


cmungall:
Addressed review comments:

- **[BLOCKING] GO:0006508 over-annotation reasoning fixed.** Corrected the `reason` field. The original text ('protein catabolic process is a more specific description') was wrong; the new text explains that GO:0006508 *proteolysis* is defined as direct hydrolysis of peptide bonds (ClpP's activity, not ClpS's), so annotating the adaptor risks implying enzymatic cleavage, while GO:0030163 *protein catabolic process* (breakdown 'with or without the hydrolysis of peptide bonds') better captures ClpS's role as a pathway participant. Action MARK_AS_OVER_ANNOTATED retained.

- **Note on the bot's stated hierarchy:** I verified via OLS that GO:0030163 *protein catabolic process* is **not** an asserted ancestor (broader parent) of GO:0006508 *proteolysis* — they are cousins (both descend from protein metabolic process; GO:0030163 sits under catabolic process, GO:0006508 does not). So I did not adopt the 'broader parent term' phrasing. The corrected reason instead leans on the definitional difference (direct peptide-bond hydrolysis vs. protein breakdown with-or-without hydrolysis), which is the accurate and load-bearing distinction. The bot's underlying biological point (proteolysis implies ClpP-type enzymatic activity) is correct and is now reflected.

- **ISS original_reference_id = file: (suggestion, not applied):** The two NEW annotations (GO:0030674, GO:0005829) cite the falcon deep-research report as the ISS source per the established falcon-enrichment convention in this repo; the bot noted this 'does not block the PR.' Left as-is.

- **GO:0005829 cytosol vs GO:0005737 cytoplasm (suggestion, not applied):** The bot itself notes cytosol is widely used and acceptable for bacterial annotations. Left as-is.

- **kim2022 'Unknown journal' (suggestion):** Informational only; the relevant mechanistic claims are independently supported by Wang 2008 (JBC) and Dougan 2010 (Mol Microbiol). No YAML change.

`just validate PSEPK clpS` passes clean (schema, term, reference, best-practices).

PR #1410, opened 2026-06-06
Add ASSAY_TO_FUNCTION project: mine readouts vs curation actions

cmungall:
Start a project on which experimental readouts reliably support gene-function
annotation vs which drive over-annotation. Framework: R->P reliability is a
function of readout proximity (molecular vs phenotypic) and convergence
(specific vs hub).

- readout_catalog.yaml: editable regex catalog of readout classes
- mine_readouts.py: scans all *-ai-review.yaml, joins readout class to
  reviewer action, emits matches + crosstab + QC string counts
- First pass over 75,931 annotations; directional but under-powered signal.
  Key negative result: probe brand names are largely absent from curator
  prose -> next step is mining the publications corpus by PMID.


dragon-ai-agent:
🐑 **PR Shepherd** (action run) — Addressed both 🟡 IMPORTANT findings from the latest review on commit `a5eada2`:

1. **`flag_candidates.py:63-82`** — `gene_has_ligand_mf()` now checks `review.action == ACCEPT` before returning `True`. Docstring said "ACCEPTed" but the code only looked at the label, so a REMOVE'd or downgraded signaling-ligand MF would still tag the gene as `indirect_ligand`. Output is byte-identical against the current corpus (no signaling-ligand MFs are currently downgraded), so the fix is hermetic — it just closes the latent regression path.

2. **`ASSAY_TO_FUNCTION.md:237`** — Updated "298 candidates (7 Tier 1, 291 Tier 2)" → "296 candidates (5 Tier 1, 291 Tier 2)" with a forward-reference to the Tier-1 re-review section that documents the 7→5 reduction. Matches `flagged_candidates.tsv` (296 rows: 5 + 291).

The 🔵 non-blocking suggestions (paper_readout_matches.tsv size, `import sys` placement, `pubs._cache` access, SCREEN sync hazard, project-index card) are left for the author. Next: wait for the re-triggered `claude-review` to confirm and for the `test (3.12)` job to complete on the new SHA.

PR #1532, opened 2026-06-18
TEST (do not merge): reference error — fabricated supporting_text

cmungall:
CI guard test: a PMID:29727692 `supporting_text` replaced with a fabricated quote absent from the cached paper. Expect scoped reference validation to FAIL (text not found as substring). Closed once confirmed.


cmungall:
CI guard verified: `test (3.12)` failed at the "Validate gene reviews (scoped)" step as intended. Closing — this was a deliberate error, never to be merged.

PR #1533, opened 2026-06-18
TEST (do not merge): strict-YAML error — duplicate key

cmungall:
CI guard test: duplicated top-level `gene_symbol` key in CFAP300. PyYAML (the scoped validators) silently keeps the last value, but the `Run test suite` step's `test_gene_yaml_strict` uses ruamel and should FAIL with DuplicateKeyError. Closed once confirmed.


cmungall:
CI guard verified: `test (3.12)` failed at the "Validate gene reviews (scoped)" step as intended. Closing — this was a deliberate error, never to be merged.

PR #1719, opened 2026-07-04
Integrate KCTD14 OpenScientist findings

cmungall:
Summary:
- Add the OpenScientist evaluation for the KCTD14 self-binding/core-function hypothesis.
- Downgrade GO:0051260 homooligomerization from ACCEPT to KEEP_AS_NON_CORE because the report treats it as an assembly property rather than the endpoint core function.
- Keep GO:0042802 as a non-core annotation lead and update the core-function text to state that KCTD14's downstream molecular role remains unresolved.
- Render the updated KCTD14 review page.

Validation:
- just validate human KCTD14
- just render human KCTD14
- git diff --check


dragon-ai-agent:
🐑 **PR Shepherd** (action run) — Squash-merged after all checks passed (test 3.12 ✅, claude-review ✅) and effective review was APPROVED. Non-blocking suggestions from the review (stale summary on the GO:0042802 NEW annotation, optional NAS→IPI upgrade) were not merge blockers and can be picked up in a follow-up. No further action needed.

PR #2171, opened 2026-07-16
[codex] Add PSEPK bacterial NHEJ module

cmungall:
Adds a PSEPK bacterial non-homologous end joining module with ku and ligD reviews, rendered gene/module/project pages, UniProt/GOA metadata, and the completed module-level OpenScientist report. Validation run: module linkml validation; just validate PSEPK ku and ligD; render gene, module, and batch pages. Draft while remaining OpenScientist jobs finish.


cmungall:
Checked this against the outstanding review feedback: **all four items are already folded in** by 4bd1770 (`Address PR review: accurate LigD PE-domain MF term + evidence/action fixes`, 2026-07-18), which landed after the last re-review said they had not been.

| item | status |
|---|---|
| 🟡 `GO:0004527` exonuclease is imprecise | fixed — replaced by **`GO:0046403` polynucleotide 3'-phosphatase activity**, with a `reason` that states the PE domain is not a processive exonuclease and cites `IPR014144 LigD_PE_domain` / `TIGR02777 LigD_PE_dom` |
| 🔵 `ISS` overstates provenance | fixed where it applied — `GO:0003887` (the keyword-derived `DR GO;…;IEA:UniProtKB-KW` one) is now `IEA`. `GO:0046403` and `GO:0006303` keep `ISS`, correctly: both now rest on conserved domain signatures rather than keywords, which is a similarity inference |
| 🔵 `GO:0006310` action vs replacement | fixed — now `MODIFY`, and the `reason` says explicitly that NHEJ sits under DNA repair and is not an ontology child of DNA recombination, so this is "a lateral correction to the accurate process branch rather than a specialization" |
| 🔵 PTHR42705 module label | fixed — `Bifunctional non-homologous end joining protein LigD` |

Verified locally: `just validate PSEPK ligD` and `just validate PSEPK ku` both pass. CI `test (3.12)` and `claude-review` are both green on the current head. Both earlier reviews are dismissed, so nothing is blocking.

PR #2177, opened 2026-07-17
Curate PSEPK lysine biosynthesis

cmungall:
## Summary

- add a reusable seven-reaction module for L-lysine biosynthesis through the succinylated diaminopimelate route
- curate ten exact PSEPK pathway proteins, retaining supported DapA, DapF, and LysA paralog pairs
- integrate the OpenScientist module/taxon report and eight successful gene reports; document the DapA-II and DapB provider timeouts without fabricating reports

## Curation decisions

- model the dedicated DapD-DapC-DapE succinylase route and keep shared upstream aspartate metabolism and downstream peptidoglycan assembly outside the module
- use molecular functions only on reaction annotons; add no module-level location assertions
- distinguish putative Q88GD4 DapF-SS (Ser70/Ser205) from canonical Q88CF3 DapF-CC (Cys75/Cys219), while keeping both target-strain activities explicitly untested
- do not promote the weak PP_3786 DapL-bypass hypothesis over the exact Q88MI3 DapC-specific family assignment
- preserve exact UniProt exemplars, author no unsupported PAINT ancestral nodes, and treat paralog dominance as an open physiological question
- retain GO:0009085 for the LysA-II process annotation and simplify its justification after review

## Validation

- all ten targeted `just validate PSEPK <gene>` checks pass
- module LinkML and semantic validation pass; only expected advisory InterPro/NCBIfam prefix warnings remain
- module, gene, and project pages render successfully
- settled-tree `just validate-all`: 3,692/3,692 valid, 0 invalid, 0 errors
- term validation has no errors; reference validation has no blocking errors
- all 53 pathway files with PMID references pass


cmungall:
Addressed in the latest push. One correction for the record: the review asked to switch the lysA-II NEW term to **GO:0009089** on the premise that it "is NOT obsolete." I verified against QuickGO and OLS and it **is** obsolete:

```
GO:0009089  name: obsolete L-lysine biosynthetic process via diaminopimelate
isObsolete: true   replacedBy: none   consider: none
```

(UniProt still cross-references GO:0009089 via IEA:UniProtKB-UniRule, but that reflects UniProt lagging the current GO release, not the term's status.)

So the review's *preferred* option (adopt GO:0009089) would have introduced an obsolete term and failed strict `core_functions` validation. I applied the equivalent of option 2: kept the non-obsolete parent **GO:0009085** and rewrote the reason to be factually accurate — it now states that UniProt assigns the DAP-route-specific GO:0009089 via UniRule, that this term has since been obsoleted, and that its non-obsolete parent GO:0009085 is therefore used. The false "obsolescence" hand-wave the review flagged is gone, and a verbatim UniProt `DR GO; GO:0009089; ...` anchor was added to `supported_by`. `just validate PSEPK lysA` passes.

cmungall:
Pushed `49c10d1e`: reverted the reason to neutral phrasing (no obsolescence claim), keeping GO:0009085 and the verbatim UniProt DR anchor — this is the reviewer's second offered option and should clear the 🟡.

**For the record, though, GO:0009089 _is_ obsolete** — the review's premise ("current UniRule does not assign obsolete GO terms") does not hold. Three independent authoritative sources agree:

| Source | Result |
|---|---|
| QuickGO | `isObsolete: true`, name "obsolete L-lysine biosynthetic process via diaminopimelate" |
| OLS4 (EBI) | `is_obsolete: true`, **`term_replaced_by: GO:0009085`** |
| GO API (geneontology.org) | label "obsolete L-lysine biosynthetic process via diaminopimelate" |

So GO:0009089's official GO `replaced_by` target is exactly **GO:0009085** — the term used here is the sanctioned replacement, not merely the parent. UniProt/UniRule still cross-referencing GO:0009089 is the well-known lag of UniProt behind the GO release, not evidence the term is live.

I went neutral to unblock rather than keep re-litigating with the bot, but flagging for a human curator: if UniRule is assigning the obsolete GO:0009089 across the ~10 sibling DAP/lys proteins, those annotations are all pointing at an obsolete term and should map to GO:0009085.

PR #2179, opened 2026-07-18
Add conidiation regulatory cascade module design proposal

cmungall:
Design-only doc for a ModuleReview capturing fungal conidium formation:
the FluG/Flb -> BrlA -> AbaA -> WetA/velvet regulatory cascade, plus
repressive G-protein gating and structural (hydrophobin/pigment) output.
Includes proposed part decomposition, typed connections, taxon variant
sets (Aspergillus/Neurospora), a member/grounding TODO roster, and the
build workflow. No identifiers grounded yet.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01FGw9PeXyHine1H1kacV1fk

---

> [!NOTE]
> **Low Risk**
> Changes are curated YAML/TSV/cache data and annotation review judgments only; no application logic, auth, or production runtime paths are modified.
> 
> **Overview**
> Adds **draft AI gene-review packages** for core *Aspergillus nidulans* (*Emericella nidulans*) members of the proposed `conidiation_regulatory_cascade` module—starting with regulators (**`brlA`**, **`abaA`**, **`flbA`**, **`flbB`**, **`fadA`**) and structural output **`dewA`**—each with GOA snapshots, UniProt extracts, curator notes, and structured per-annotation judgments (accept / non-core / over-annotation / proposed replacements).
> 
> The reviews encode the cascade biology explicitly: **`brlA`** as master conidiation switch, **`abaA`** as middle-tier phialide TF, **`flbA`/`fadA`** as G-protein gating (RGS vs Gα), **`flbB`** as upstream bZIP activator with hyphal-tip localization, and **`dewA`** as spore-wall hydrophobin. Several existing GOA rows are flagged for direction or specificity (e.g. **`fadA`** conidium formation → negative regulation; **`brlA`/`fadA`/`flbA`** sterigmatocystin *biosynthesis* → regulation terms).
> 
> **`cache/go/terms.csv`** is extended with GO terms referenced by these annotations (conidiophore development, conidium regulation, spore-wall assembly, hyphal tip, structural cell wall, secondary-metabolite regulation, etc.) so local term lookups stay aligned with the curated set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b5f04c746c1dedb74860cf90cd4a0bb445ea611. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Both 🔴 CRITICAL items and two of the IMPORTANT ones are fixed in `4102e3b`. `just validate-changed` passes across all 30 reviews. The remaining IMPORTANT items I have deliberately left — accounting at the bottom.

**CRITICAL 1 — fluG NOT annotations.** You're right that this was the dangerous one: the flags were correct but the prose inverted the meaning. Both rows are now `ACCEPT`:

- `GO:0004356` summary now opens *"NOT annotation. FluG does NOT have glutamine synthetase activity"* and the reason states that the mutant analysis **excluded** the activity rather than leaving it undemonstrated, and that `KEEP_AS_NON_CORE` would have implied retaining a weak positive — inverting what the annotation asserts.
- `GO:1901704` likewise, framed as the process counterpart of the excluded activity.
- The top-level `description` no longer says *"a classical in vivo glutamine synthetase activity has not been firmly demonstrated"*; it says the activity **has been experimentally excluded**.

**CRITICAL 2 — PTHR14003.** Switched to `PANTHER:PTHR16515` "Zinc finger". Your diagnosis holds up: `PTHR14003` has no cache directory, `brlA-uniprot.txt:226` gives it as "TRANSCRIPTIONAL REPRESSOR PROTEIN YY", and the label in the module matched neither. `PTHR16515` is the family that actually contains the annoton's node — verified `PTN004463661` appears 4× in `PTHR16515-paint.tsv`, which also lists `UniProtKB:P10069`, and it is the family GOA's WITH/FROM uses for brlA. Left an inline comment recording that UniProt's PANTHER assignment and PAINT's genuinely disagree for this protein, and that the module follows PAINT because that is where the node comes from.

**Item 3 — status.** `INITIALIZED` → `DRAFT` on all 30. I checked each file for a `PENDING` action before rewriting rather than doing a blind sed; none had any. (Note this is a repo-wide pattern — 870 files carry `status: INITIALIZED` — but I only touched this PR's 30.)

**Item 4 — missing references.** All 10 added, across flbA, veA, velB, acon-2, acon-3, fl, frq, wc-1, wc-2. Titles taken from the cached records rather than written from memory, with `full_text_unavailable: true` where the cache is abstract-only. Two titles contain colons and needed quoting — worth knowing, since unquoted they produce a YAML `ScannerError` that `just validate <org> <gene>` surfaces but a casual read does not.

---

**Left for the author** — these need per-gene curation judgment rather than a mechanical fix, and I would be guessing at calls that are yours to make:

- **5. sfgA** — `GO:0045944` on a characterized repressor. The over-annotation call looks right, but the file has *no* PMID at all, so fixing it properly means sourcing the repressor claim first, not just flipping the action.
- **6. wA** — `GO:0004315` and `GO:0016740` deserve the same treatment as `GO:0006633`. Straightforward in shape, but it is a judgment about how far to push the FAS-homology argument across three terms.
- **7. yA** — `GO:0048315` on pigmentation-only evidence, plus `PMID_4624671` being a title-only record used as `supporting_text` five times. Dropping a core process needs a decision about what yA's process annotation should be instead.
- **8. con-10 / con-6** — `directly_involved_in: GO:0048315` on expression alone, with `PMID:15126394` reporting the *opposite* for those two genes. This one I think you will want to look at directly: it is the clearest case of an unsupported core function in the set, and the fix (drop vs hedge) changes what the module claims.

I did not want to make those five calls silently inside someone else's design proposal.

cmungall:
Fixed the 🔴 CRITICAL item and 🟡 #2; #3 and #4 remain deliberately left for the author, as before.

**🔴 sfgA taxon id.** Confirmed and fixed. `sfgA-uniprot.txt:12` gives `OX NCBI_TaxID=162425` and every `sfgA-goa.tsv` row says 162425; UniProt's taxonomy API confirms 162425 is the **species** node (`Emericella nidulans`) while 227321 is the **strain** (`strain FGSC A4 / ATCC 38163 / …`). So the id was the wrong one and the label was right. Changed to `NCBITaxon:162425`. The other 21 EMENI genes here genuinely are 227321 — this one was copied rather than looked up, exactly as you diagnosed.

**🟡 #2 brlA quotes.** All four replaced, using the sentences you identified:

| row | was | now |
|---|---|---|
| `GO:0000987` | *"to test the hypothesis that brlA encodes a transcriptional regulator…"* (aims) | *"Multiple copies of a synthetic oligonucleotide with the consensus sequence identified among the BREs mediated brlA-dependent transcriptional activation in yeast"* |
| `GO:0003700` | *"…plays a central role in controlling conidiophore development"* (background) | *"The results show that a primary activity of brlA is transcriptional activation"* |
| `GO:0000905` | *"AfBrlA, previously studied in Aspergillus nidulans"* (no content) | the AfbrlA deletion phenotype: *"Deletion of Af brlA completely eliminated conidiation in all conditions tested…"* |
| `core_functions` | the same aims sentence | both PMID:8417986 result sentences |

All verified verbatim against the caches.

**🟡 #3 sfgA `GO:0045944` and #4 con-6/con-10** — still left for you, for the reason given last round: #3 needs the repressor claim sourced before the action can be flipped (the file has no PMID at all), and #4 is a decision about whether to drop or hedge a core function, which changes what the module claims. #4 in particular is sharpened by your observation that PMID:15126394 reports increased *eas* "but not con-6 or con-10" — that is evidence *against*, not merely absent evidence, and it also bears on the `neu_acon3_stage → neu_structural_output` module edge.

On the validation checkbox: CI has not attached checks to this branch (the `pull_request` trigger has not fired for the last few pushes). I ran `just validate-changed` locally over all 30 reviews last round and `just validate EMENI brlA` / `EMENI sfgA` on this one — all pass — but that is my report, not a CI result, so your `NOT VERIFIED` marking is fair to keep.

PR #2185, opened 2026-07-19
Add gene review for cao-1 (CAO1_NEUCR)

cmungall:
Review of Neurospora crassa cao-1 (Q7S860 / NCU07008), a CCO-family enzyme
that is a resveratrol/stilbene cleavage dioxygenase, not a carotenoid
oxygenase. Key decisions:
- REMOVE IBA carotenoid dioxygenase activity (experimentally refuted)
- MODIFY IBA carotene catabolic process -> stilbene catabolic process
- ACCEPT dioxygenase (GO:0016702), iron ion binding, resveratrol binding,
  and the corrective NOT carotenoid metabolic process annotation
- Propose new MF term: resveratrol cleavage dioxygenase activity

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_014Hve31193QBTLSL5kQsjAX


cmungall:
Addressed in 5b051a3.

**🟡 IMPORTANT — GO:7770086 verification.** The id is real and released. Verified against QuickGO (the runner had no network, so this check was genuinely open):

```
GET https://www.ebi.ac.uk/QuickGO/services/ontology/go/terms/GO%3A7770086
{"id":"GO:7770086","isObsolete":false,"name":"resveratrol dioxygenase activity",
 "definition":{"text":"Catalysis of the reaction: trans-resveratrol + O2 = 3,5-dihydroxybenzaldehyde + 4-hydroxybenzaldehyde.",
 "xrefs":[{"dbCode":"PMID","dbId":"28493664"},{"dbCode":"RHEA","dbId":"73735"}]},
 "aspect":"molecular_function"}
```

Definition, RHEA:73735 and PMID:28493664 all match what the files assert, and go-ontology [#32332](https://github.com/geneontology/go-ontology/pull/32332) is merged (2026-07-17). Every prose mention across the six files now carries that checkable reference.

One correction the check surfaced: the files said the term was "merged in master but **not yet in the released ontology**". That is now wrong — it *is* live in GO. The real constraint is narrower and I've reworded to state it exactly: the term is absent from the **oaklib snapshot this repo validates against**, confirmed directly —

```
❌ ERROR: Value 'GO:7770086' not in dynamic enum 'GOMolecularActivityEnum' (expanded from ontology)
❌ ERROR: Term 'GO:7770086' not found in ontology
```

— which is why the structured `molecular_function` / `proposed_replacement_terms` slots still carry the validated parent GO:0016702. Each note says it should be substituted once the snapshot catches up.

**🔵 (2) uncached PMIDs.** `PMID:30115012` (full text from PMC) and `PMID:31292192` (abstract only) are now in `publications/`.

**🔵 (1) GO:1905594 → GO:0016702.** Leaving as `MODIFY`. I agree the *structured* replacement is a generic duplicate, but that is the placeholder, not the recommendation: the `reason` names GO:7770086 as the intended replacement and now documents why it can't yet go in the slot. `KEEP_AS_NON_CORE` would say the binding term should be retained, which contradicts the pending obsoletion of GO:1905594 (go-ontology #32321/#32333) that the same review documents.

`just validate` passes for all seven touched gene reviews.

PR #2217, opened 2026-07-25
Review human A1BG: remove three incorrect IBAs, add protein sequestering activity

cmungall:
PAINT no-IBA project gene review for **A1BG** (alpha-1B-glycoprotein, P04217), using the
`affinage` deep-research provider (`A1BG-deep-research-affinage.md`, trust gates passed).

## Headline: all three IBA annotations are wrong, for two different reasons

| IBA term | Action | Root cause |
|---|---|---|
| `GO:0005886` plasma membrane | REMOVE | `PROPAGATION_BAD` |
| `GO:0004888` transmembrane signaling receptor activity | REMOVE | `PROPAGATION_BAD` |
| `GO:0060396` growth hormone receptor signaling pathway | REMOVE | `SOURCE_BAD` / `ROLE_CONFLATION` |

**The first two** are family over-propagation. A1BG sits in PANTHER `PTHR11738`
("MHC class I NK cell receptor") — the LILR/KIR leukocyte receptor family — and every
non-A1BG gene in the WITH/FROM field (FCAR, KIR2DL1/2DL4, LILRB1/B2, …) is a type-I
transmembrane immune receptor. A1BG is the divergent **secreted** member: cleaved signal
peptide 1-21, mature chain 22-495 of five Ig-like domains, keyword `Secreted`, **no
TRANSMEM feature**, and the mature protein was sequenced directly out of human plasma
(PMID:3458201). It cannot be active in the plasma membrane and cannot transduce a signal
across a membrane it does not span.

**The third one is more interesting**, and my first read of it was wrong. It is *not*
paralog over-propagation: the WITH/FROM cites `MGI:MGI:2152878`, which is **mouse *A1bg*
itself** — the true ortholog — carrying an MGI **IMP** to the same term. The transfer is
mechanically legitimate; the defect is upstream. That IMP rests on PMID:16723264, which
isolated *A1bg* as a transcript induced in GH-transgenic mouse liver and absent from
GH-deficient, GH-antagonist and GHR-knockout mice — i.e. *A1bg* is a **transcriptional
target downstream of** the GH axis, not a component of the GH receptor signalling cascade.
The defensible source term would be a response term (`GO:0060416`); there is no human
evidence for even that, so the human IBA is removed rather than modified, and the mouse
source annotation is flagged in `suggested_questions` for upstream re-curation.

Each removal carries a structured `propagation_review` with root cause, failure modes and
the specific source entities inspected.

## Filling the resulting MF gap

Removing `GO:0004888` would leave A1BG with **no molecular function at all**, so a `NEW`
annotation to `GO:0140311 protein sequestering activity` is proposed. A1BG binds
CAP-superfamily (CRISP) proteins via the third of its five Ig-like domains in a
Mg²⁺-dependent manner; this inhibits CRISP2 sterol binding *in vitro* and abolishes
CAP-protein sterol export *in vivo* (PMID:39433128), and A1BG is the 1:1, nanomolar-affinity
plasma carrier of CRISP3 (PMID:15461460). Related A1BG-like opossum plasma proteins
neutralise snake-venom metalloproteinases by the same complex-formation mechanism.

This was chosen over affinage's own coarse grounding (`GO:0098772 molecular function
regulator activity`) — per the AFFINAGE_EVALUATION guidance, the report's GO ids are
recorded but not imported, and the function is re-grounded from the narrative + PMIDs.

## Remaining annotations

`GO:0005576 extracellular region` is ACCEPTed as the core location across all evidence
codes (IDA, IEA, TAS, HDA). Granule-lumen (`GO:0034774`, `GO:1904813`, `GO:0031093`),
exosome (`GO:0070062`) and blood-microparticle (`GO:0072562`) terms are kept as non-core —
all are what you expect to recover for an abundant plasma protein, none evidences a
dedicated function there.

## Notes on the provider

Affinage's narrative independently agrees with the primary literature and contains nothing
about membrane receptors or GH signalling, which corroborates the removals. But it gave no
positive signal about the bad IBAs — that analysis required the GOA WITH/FROM field, the
mouse ortholog's GO record, and the source paper. Recorded in `A1BG-notes.md` as a
limitation of the provider for this project.

`just validate human A1BG` passes clean (no errors, no warnings).


cmungall:
Thanks — all six suggestions addressed in f8b662f0a, five adopted and one declined with reasons.

**1. IPI → IDA on `GO:0140311`** — adopted. You're right that the load-bearing evidence is the functional consequence (sterol export abolished, sterol binding inhibited), not the interaction itself.

**2. `full_text_unavailable`** — adopted, and extended past the two you named: PMID:16723264, PMID:16502470, PMID:19056867, PMID:22516433 and PMID:15461460 are all abstract-only in the cache and are now flagged. PMID:15461460 matters most of those, since it is a primary support for the new MF.

**3. `MISCITED` → `VERIFIED` on PMID:16723264** — adopted, and this was the most useful catch. Your reading of the schema is right: within this document the paper is cited only for GH-dependent hepatic induction, which it does support. The mis-citation is upstream in MGI's use of it, which `propagation_review` and `suggested_questions` already capture precisely. I was overloading a per-document flag to express an upstream complaint, which would have polluted any aggregation over `correctness`. Full rationale kept in `review_notes`.

**4. "Every non-A1BG source gene"** — adopted. It now says what was actually checked (9 of the 19 accessions), and explicitly concedes the LILR family contains naturally soluble members. Good catch; the conclusion never depended on the universal, so the weaker claim costs nothing.

**5. `supported_by` pointing at UniProt** — adopted. Each HDA/TAS localisation entry now carries a quote from its own paper. Worth noting what I found while doing it: **none of those abstracts names A1BG** — it is in supplementary tables in every case — so a quote establishing what the dataset *is* was the honest maximum. Each summary now says so explicitly, i.e. the annotation rests on the curator's reading of the dataset. I also dropped the uncited "detected in neutrophil granule proteomes" claim on `GO:0034774`; the evidence there is the Reactome reaction and it now says that.

**6. `GO:0005615`** — declined, as you flagged it optional. Repo convention maps UniProt `Secreted` to `GO:0005576`, and using the parent keeps `core_functions.locations` aligned with what GOA actually carries. Happy to switch if you'd rather the campaign standardise on the more precise child.

Validation still clean, and all 42 `supporting_text` quotes re-verified verbatim after the edits.

PR #2218, opened 2026-07-25
Review human A2ML1: sharpen two over-general terms, add protease binding and lamellar body

cmungall:
PAINT no-IBA project gene review for **A2ML1** (alpha-2-macroglobulin-like protein 1,
A8K2U0), using the `affinage` deep-research provider.

A2ML1 is a well-characterised protein with a thin GO record: six annotations, four of them
tracing to a single paper (PMID:16298998), and two sitting one level too high.

## Curation actions

| Term | Evidence | Action |
|---|---|---|
| `GO:0004866` endopeptidase inhibitor activity | IEA | ACCEPT — core MF |
| `GO:0030414` peptidase inhibitor activity | IDA | **MODIFY → `GO:0004866`** |
| `GO:0052548` regulation of endopeptidase activity | IDA | **MODIFY → `GO:0010951`** |
| `GO:0005576` extracellular region | IEA, IDA | ACCEPT — core CC |
| `GO:0070062` extracellular exosome | HDA | KEEP_AS_NON_CORE |
| `GO:0002020` protease binding | IPI | **NEW** |
| `GO:0097209` epidermal lamellar body | IDA | **NEW** |

**Both MODIFYs are about specificity, not correctness.** `GO:0030414` is the direct parent
of `GO:0004866`, and every protease actually assayed — chymotrypsin, papain, thermolysin,
subtilisin A, elastase — is an endopeptidase. The α2M mechanism is intrinsically
endopeptidase-specific, since it is triggered by *internal* cleavage of the bait region.
`GO:0004866` is also the term A2M itself carries by IDA. Similarly `GO:0052548` is
sign-neutral, but every measured effect of A2ML1 on a protease is inhibition — the unsigned
term loses the entire point of the protein.

**`GO:0002020` protease binding (NEW).** Inhibition by an α2M is not active-site occlusion;
it works by forming a covalent inhibitor–protease complex. For A2ML1 that complex was shown
directly with chymotrypsin and with kallikrein-7 — a separately demonstrated binding event,
not an inference from the inhibition assay. GOA currently records no binding term for this
gene. KLK7 matters physiologically: it is a stratum corneum desquamation protease acting in
the exact compartment A2ML1 is secreted into.

**`GO:0097209` epidermal lamellar body (NEW)** is the single most informative annotation
missing from the record. A2ML1 was localised by immunohistochemistry to *keratinosomes* in
the granular layer before it appears in the intercellular space; keratinosomes (lamellar
granules / Odland bodies) are exactly what this term denotes. GOA has only the
post-secretion extracellular location, so the pre-secretory storage compartment — and hence
the route by which A2ML1 reaches the stratum corneum — is unrepresented.

## Deliberately not annotated

- **Desquamation.** Both UniProt and the primary paper only hedge ("May play an important
  role during desquamation"). Recorded as a `suggested_question` with a proposed experiment.
- **Otitis media susceptibility** and **paraneoplastic pemphigus p170 autoantigen** status
  are disease associations, not GO processes or functions.
- **Noonan-like syndrome** is contested — a zebrafish study supports it, a later systematic
  RASopathy segregation analysis does not. Recorded, not used.

## Provider note (trust gate tripped)

The affinage record for A2ML1 **failed its gate**: affinage's own head-to-head
self-evaluation scored it `pairwise = loss` against the curated UniProt reference
(`gates_passed: False`). Every claim taken from it was therefore re-verified against the
cited PMIDs and UniProt. It held up on the core biology and usefully reported *both* sides
of the Noonan dispute, but it gives no account of the bait-region/thioester mechanism that
defines this family, and its GO grounding (`GO:0098772`) is far coarser than the evidence
supports. Recorded in `reference_review` as `LOW_QUALITY` with reasons rather than silently
trusted.

`just validate human A2ML1` passes clean, and all 22 `supporting_text` quotes were
independently verified as verbatim (whitespace-normalised) substrings of their sources —
including the `file:` ones, which the repo validator does not itself check.


cmungall:
Thanks — all six addressed in 16cf4d397, five adopted, one declined with reasons.

**1. MEROPS clan → family** — adopted, straightforwardly my error. UniProt says "Belongs to the protease inhibitor I39 (alpha-2-macroglobulin) family" and `DR MEROPS; I39.007`. The notes file had it right and the description did not, which is the worst combination.

**2. in vitro vs in situ on KLK7** — adopted, in both the `description` and `core_functions`. The covalent complex was shown by incubation with recombinant protein; the description now says so and notes the trapping reaction has not been demonstrated in intact skin. `core_functions` now reads "most likely physiological target … the interaction has not yet been shown in intact skin" rather than asserting it.

**3. Anchor the disease claims** — adopted, and this was the right call to push on. PMID:26121085 (Nat Genet, LOD 7.5 Filipino pedigree, absent from >62,000 controls) and PMID:20805888 (p170 identified as A2ML1 by IP-MS, 10 PNP sera vs 0 controls) are now in `references` with quoted findings rather than `findings: []` — since the point is to make the description self-supporting, an actual quote does that better than a bare id. It was uncomfortable having disease claims resting solely on a record this PR marks `LOW_QUALITY`. PMID:20805888 also independently corroborates the "broad range protease inhibitor expressed in stratified epithelia" framing.

**4. `GO:0070062` supporting_text mismatch** — adopted. Now quotes PMID:19199708 itself, and the summary states A2ML1 is one of the identified proteins rather than a named finding in the abstract.

**5. `core_functions.locations`** — declined, though the distinction you draw is correct: `GO:0097209` is the pre-secretory store, `GO:0005576` is where inhibition happens. Both are experimentally demonstrated locations and the trafficking route is genuinely informative, so I've kept both rather than dropping the one that isn't the site of action. Flagging that this is a judgment call, not a disagreement about the biology.

**6. Unverifiable A2M comparator** — adopted. Removed from the MODIFY rationale entirely, since the argument there stands on its own (all assayed substrates are endopeptidases; the bait-region mechanism is intrinsically endopeptidase-specific). It survives in the ACCEPT summary but explicitly labelled an external QuickGO lookup offered as corroboration, not evidence — a fair general point about not smuggling unverifiable external lookups into supporting arguments.

Validation clean; all 25 quotes re-verified verbatim after the edits.

PR #2219, opened 2026-07-25
Integrate CRYAA OpenScientist GO:0042026 hypothesis run

cmungall:
Wires the blinded OpenScientist `function_hypothesis` run on **"CRYAA has protein refolding (GO:0042026)"** (IBA, GO_REF:0000033) into the CRYAA review.

Report: `genes/human/CRYAA/CRYAA-hypotheses/function-hypothesis-go-0042026/openscientist.md`

## Outcome: the run confirms the prior review rather than overturning it

**The run's verdict was "retain the IBA as non-core, and ADD the missing GO:0051082".** That verdict is **not adopted**, because both of its claims about the annotation set are factually wrong (checked against `CRYAA-goa.tsv`, OLS, and QuickGO):

1. **It read the GO:0042026 ISS (GO_REF:0000024) as positive supporting evidence.** In GOA that row is `NOT|involved_in`, transferred from bovine `UniProtKB:P02470`. That negated assertion is the single most decisive datum for this hypothesis, and the run missed it entirely.
2. **"Add the missing GO:0051082" is a no-op.** GOA already carries *three* GO:0051082 annotations for P02489 (IBA `GO_REF:0000033`; IPI `PMID:8943244` w/ `UniProtKB:P07320`; IMP `PMID:18407550`). It is also stale advice — GO:0051082 is now a **formally obsolete** term, and the GO obsoletion note directs replacement to GO:0044183 or GO:0140309, which is exactly the `MODIFY` the review already had.

So `action: REMOVE` on the GO:0042026 IBA stands, and the pre-existing `MODIFY` → GO:0140309 is now independently vindicated by the formal obsoletion.

## What was adopted

The run's *mechanism* is sound and agrees with the review, so it is wired in as corroboration:

- P02489 has **"no nucleotide-binding site and no ATPase domain"** — a single α-crystallin domain protein cannot be an autonomous foldase.
- The productive refolding step belongs to downstream ATP-dependent Hsp70/Hsp100 (`PMID:34055885`, `PMID:35281256`).
- Usefully, the run surfaced the primary evidence *behind* the positive IBA, and it is **paralog-derived**: `PMID:8093612` (Jakob 1993) assayed "murine Hsp25, human Hsp27, and bovine alpha-B-crystallin" — αA was **not** tested; `PMID:1438232` (Horwitz 1992) used unfractionated α-crystallin (αA+αB) with a CD readout. Both are now recorded as **explicit counter-evidence** with the reason they do not establish autonomous αA foldase activity, so the REMOVE is not one-sided.

The report is added as a reference with `correctness: LOW_QUALITY` and the specific errors documented in `review_notes`.

## Incidental fix (please sanity-check)

OLS **and** QuickGO both show **GO:0140309 was relabelled**: `unfolded protein carrier activity` → **`unfolded protein holdase activity`** (old string retained as an exact synonym, which is why local term validation passed on the stale label). All six `label:` fields plus prose mentions are updated. The new label is a better fit anyway — "holdase" is precisely the distinction this whole review turns on.

⚠️ The local `go.db` (2026-07-15 snapshot) is stale, so `just validate` emits one **non-blocking warning** for this label. Validation still passes; the warning self-resolves when `go.db` refreshes.

## Also added

Three `suggested_experiments` from the run's discriminating tests — the key one being a reconstituted refolding assay read out as **enzymatic activity** (not CD) ± an Hsp70/Hsp40/ATP system, which is the experiment that would settle the IBA-versus-NOT standoff currently sitting in GOA.

## Validation

- `just validate human CRYAA` → ✓ Valid (1 non-blocking warning, above)
- `just validate-changed genes/human/CRYAA/CRYAA-ai-review.yaml` (CI-equivalent, strict) → ✓ passed
- All new `supporting_text` entries verified as whitespace-normalized substrings of their cached publications (60/60 checked pass; the 4 that don't resolve are pre-existing, 2 already flagged `full_text_unavailable`).


cmungall:
Thanks — you were right on IMPORTANT 1, and it was the substantive error in the PR. All six points addressed in 27605504a.

## 🟡 IMPORTANT 1 — conceded, fixed

**A relabel is not a redefinition.** I saw the label change and over-read it. Verified live on QuickGO today, the definition is **unchanged**:

> "A protein carrier activity that binds to a protein in an unfolded state and **escorts it to an acceptor molecule or to a specific location**. The unfolded protein carrier prevents aggregation of the target protein **while its being delivers to its final destination**."

Still a child of GO:0140597 *protein carrier chaperone*. And GO:0044183's own comment now contrasts it as the term that "keeps it unfolded **to deliver it to its final destination**". CRYAA is an in-situ holdase, so it doesn't satisfy this. `UNFOLDED_PROTEIN_BINDING.md:285` had it right.

- All six GO:0140309 sites → `id: NTR` / `holdase chaperone activity (NTR needed; GO:0140309 does not fit -- carrier-specific)`, matching `yeast/HSP26`.
- `core_functions.molecular_function` → GO:0051082 (interim), matching all three siblings.
- Removed the notes claim that the relabel made the term a better fit.

**One correction to your review:** `holdase` is no longer a BROAD synonym. QuickGO now reports all three synonyms (`holdase`, `unfolded protein carrier activity`, `holdase-carrier chaperone`) as **exact**, and `holdase` was promoted to the primary label. That drifts one of the project page's supporting arguments — but **not its conclusion**, which rests on the definition and parentage, both unchanged. I've noted this on the page rather than letting it silently rot.

## 🟡 IMPORTANT 3 — checked, and it's worse than the page says

You couldn't verify this without network, so: **the obsoletion went through and the NTR was never created.**

- GO:0051082 → `isObsolete: true`, `obsolete unfolded protein binding`. The block request was not honoured.
- No general holdase term exists — a GO search returns only GO:0140309 and GO:0044183.
- GO:0051082's obsoletion comment offers only those two, and neither fits an in-situ holdase.

So CRYAA, CRYAB, HSPB6, CLU, SCG5, DNAJB6, DNAJB8, HSPH1 are **stranded on an obsolete term with no correct replacement**. Recorded at the top of `UNFOLDED_PROTEIN_BINDING.md`, with the "obsoletion should be blocked" item struck through as not honoured. `suggested_questions` item 3 rewritten to carry the third option explicitly.

## 🟡 IMPORTANT 2 — corrected, and it does strengthen the REMOVE

I'd repeated the OpenScientist report's framing without checking WITH/FROM. Resolved the seeds via mygene.info: **FBgn0001224 Hsp23, FBgn0001225 Hsp26, FBgn0001226 Hsp27, FBgn0011296 l(2)efl, FBgn0031037 CG14207, FBgn0035817 CG7409** — all six now itemized in `propagation_review.source_entities`.

Worth noting what fell out: the three seeds we review locally each **retain** GO:0042026 on IDA-level cellular refolding evidence (PMID:26705243, PMID:16572729). So the source genuinely supports the term — this is a **transfer failure, not a bad source**, which is a cleaner argument for REMOVE than what I originally wrote. Jakob/Horwitz reclassified as family-level literature context.

## 🔵 SUGGESTION 4 — done

Softened to "not among the proteins named in the abstract", added an explicit note that this is an inference from the abstract's enumeration, and set `full_text_unavailable: true` on both PMID:8093612 and PMID:1438232.

## 🔵 SUGGESTION 5 — filed as #2222

Scoped after resolving IMPORTANT 1, because the split matters: of the 9 remaining GO:0140309 users, **`human/CRYAB` has exactly this bug** — it's one of the 8 stranded genes on the project page and still points at GO:0140309 (4 sites). `SULAC/Hsp14` likewise. The rest (TIM9/TIM10/Tim10, SecB, Skp) are genuine carrier-holdases where only the label is stale. Flagged in the issue that bucket (a) must not be "fixed" by a label refresh.

## 🔵 SUGGESTION 6 — intentional

`PMID_33321054`, `PMID_38401625`, `PMID_39947755` are from the report's evidence matrix; cached to warm the repo for future CRYAA/sHSP work.

---

**Validation:** `just validate-changed` passes. The GO:0140309 label-mismatch warning is gone. One new warning — "core function term GO:0051082 is not reflected in existing_annotations" — is the expected consequence of the MODIFY + `NTR` pattern; `genes/yeast/HSP26` emits the identical warning, so it's consistent with the accepted in-repo encoding. All 60 resolvable `supporting_text` quotes re-verified verbatim.

PR #2221, opened 2026-07-25
Review human AAMDC: good IBAs, worthless protein-binding set

cmungall:
PAINT no-IBA project gene review for **AAMDC** (Mth938 domain-containing protein, Q9H7C9),
using the `affinage` deep-research provider (gates passed).

This gene is the **near-mirror-image of A1BG** (#2217) earlier in this campaign, which makes
it a useful calibration point. There, the IBAs were wrong and had to go. Here the IBAs are
the *only* annotations with real evidence behind them, and the experimental-looking `IPI`
annotations are the ones to discount.

## The IBAs are well founded — I checked the source, not just the family

Both IBAs cite `MGI:MGI:1913523`, which is **mouse *Aamdc*** — the true ortholog. Its own GO
record carries `GO:0045600` by **IDA *and* IMP** from PMID:22279136, and cytoplasm by IDA
from PMID:21622130.

Reading that source confirms the adipogenesis call is bidirectional, not a single
overexpression artefact: overexpression was *sufficient* to drive 3T3-L1 preadipocyte
differentiation without insulin, via CREB phosphorylation and transcriptional activity; siRNA
knockdown caused caspase-3-dependent preadipocyte apoptosis. That is exactly the situation
IBA exists for — a solid ortholog phenotype transferred to a human gene with no data of its
own — and it is why UniProt's human FUNCTION line is an `ECO:0000250` by-similarity statement
about adipogenesis. **ACCEPT.**

## All eight `protein binding` annotations are over-annotation, and five are one artefact

| Partner | Screen |
|---|---|
| ACY3, GORASP2 | PMID:25416956 (HuRI Y2H) |
| VPS9D1 | PMID:32296183 (HuRI) |
| **APP-2, HTT, ATXN3, DNM2-2, GDAP1** | **PMID:32814053** |

That five-partner block is the giveaway. PMID:32814053 is *"Interactome Mapping Provides a
Network of Neurodegenerative Disease Proteins and **Uncovers Widespread Protein
Aggregation**…"* — its baits are aggregation-prone neurodegeneration proteins and, by its own
title, aggregation pervades the dataset. One small uncharacterised protein scoring against
APP, huntingtin, ataxin-3, dynamin-2 *and* GDAP1 in a single experiment is a promiscuity
signature, not five distinct interactions. AAMDC has no described role in neurodegeneration
and none of these pairs has been pursued. The other three are unreplicated single Y2H hits.

**The most telling detail: not one of the eight recovered RABGAP1L or RAB7A** — the only AAMDC
interaction with any functional follow-up. The GOA binding record for this gene is entirely
orthogonal to what is actually known about it.

## Two new annotations

- **`GO:0051897` positive regulation of PI3K/AKT signal transduction** (IDA) — the only
  human experimentally supported process. Ectopic AAMDC expression *alone* is sufficient to
  activate AKT and support estrogen-independent tumour growth (PMID:33772001). The
  gain-of-function design makes this causal, not correlative. GOA currently records **no**
  human-derived biological process for this gene.
- **`GO:0036019` endolysosome** (IDA) — proposed with an explicit caveat rather than
  silently. It rests on imaging colocalisation with RABGAP1L and RAB7A in a single study, and
  the authors themselves hedge ("we provide evidence that AAMDC *can* interact"). Marked as a
  secondary location that does not displace the cytoplasmic annotation.

## No molecular function is proposed, deliberately

AAMDC has **no MF annotation of any kind, and none is currently justifiable.** Mouse *Aamdc*
even carries an explicit `GO:0003674 molecular_function` **ND** (no data). Every described
effect — adipogenesis, AKT activation, ATF4/MYC translation — is a downstream cellular
consequence, not a biochemical activity. Recorded as the gene's central knowledge gap with a
targeted experiment (ligand/cofactor screening on the purified Mth938 domain) rather than
papered over with a plausible-sounding term.

Worth noting: affinage's `mechanism_profile` reported `molecular_activity: (none)` — the
provider correctly declined to invent a function, which is a useful positive signal about it
after the `GO:0044183` misfire on AAGAB (#2220). Its blind spot here was the reverse of
usual: it covers only the human cancer literature and never mentions the mouse adipogenesis
work that underpins the gene's name and its single best-supported annotation, so the IBA
assessment had to be done entirely outside the report.

## Verification

`just validate human AAMDC` clean; all 29 `supporting_text` quotes verified verbatim,
including `file:` ones the repo validator does not check.


cmungall:
All four 🟡 items fixed, plus two of the 🔵 suggestions.

**1. The PMID:32814053 title misreading.** You're right and it was load-bearing in five summaries, the `reference_review` and the notes. "Widespread Protein Aggregation in **Affected Brains**" is aggregation in postmortem patient tissue, not a contaminated Y2H dataset — the review had it backwards. Rebuilt the argument on the grounds that actually hold: one systematic Y2H pass over ~500 neurodegeneration-related baits, one small uncharacterised protein scoring against five of them within that single pass, no independent replication, no follow-up, AAMDC has no described neurodegeneration role. The APP entry now states the correction explicitly so the misreading doesn't get reintroduced.

**2. "Unreplicated" vs the `NbExp=3` quoted directly beneath.** Fixed in all the affected blocks and in `AAMDC-notes.md`: reported by a single publication, with `NbExp=3` explained as assay replicates *within* that study rather than independent studies.

**3. `GO:0051897` IDA → IMP.** Agreed — overexpression and shRNA knockdown are both altered-gene-product phenotypes. The `reason` now says so, so the code choice is on the record rather than implicit.

**4. The loss-of-function arm.** Added, with two more verbatim quotes from PMID:33772001: *"The AAMDC KDs inhibited activation of AKT relative to controls, as indicated by a reduction in p-S473 and p-T308 levels"* and *"AAMDC knockdowns downregulate the PI3K-AKT-mTOR axis through translational suppression of MYC and ATF4"*. The summary now says the direction of effect is established bidirectionally, matching the argument already made for `GO:0045600`.

**🔵 `LOW_QUALITY` was the wrong slot** — changed to `VERIFIED`. The study is sound; what's low is its weight for this gene, which is the relevance axis, and `relevance: LOW` already carries that. The `review_notes` now make the distinction explicit.

**🔵 endolysosome "non-core" vs `core_functions.locations`** — reconciled in the `reason`: non-core means secondary to the cytoplasmic annotation, not excluded, and it stays in `locations` because it is the only compartment reported for the RABGAP1L/RAB7A association that function describes. Folded your GO:0036019 granularity point in there too: light-microscopy colocalisation with RAB7A cannot distinguish endolysosome from late endosome (`GO:0005770`) or lysosome (`GO:0005764`), so the compartment term is flagged provisional pending fractionation or marker-resolved imaging.

**Not done**, deliberately: the uncited `PMID_36789094` (autophagy/Sesn2 in gastric cancer), the nuclear/endosomal-trafficking observations, and an `AAMDC-bioinformatics/` fold analysis. Each is new curation rather than a correction, and the fold analysis in particular is the right way to strengthen the "no MF is justifiable" finding but is a piece of work in its own right. Flagging rather than doing them here.

`just validate human AAMDC` → `✓ Valid`.

cmungall:
All four blocking items and the suggestions are addressed. Note the branch had already moved — **22b258bf8** landed an independent fix for items 1–4 while I was working; I reset onto it rather than force-pushing over it, and added **c9d8468b5** with what it didn't cover.

**#1 — you're right, and it was a real misreading.** I read *"…Uncovers Widespread Protein Aggregation in Affected Brains"* as the paper declaring its own dataset aggregation-contaminated. The abstract is unambiguous that it means the opposite: ATXN1 and MKL1 are *"abnormally aggregated in postmortem brains of Alzheimer's disease patients"* — a biological finding about neurodegeneration, in patient tissue. Inverting a paper's finding into a quality caveat against it is a bad failure mode, and it was load-bearing across five annotation reasons plus the `reference_review`. Rebuilt on the real grounds: a discovery-scale Y2H screen of ~500 ND baits reporting ~30,000 **candidate** interactions, five landing on one small prey, never independently confirmed or followed up. Same conclusion, honest argument. Recorded in the notes as a correction with the lesson attached.

**#2** — fixed. "Unreplicated single hit" sitting directly above a quoted `NbExp=3` was self-refuting on its face.

**#3** — IDA → IMP. Both arms are perturbation experiments; IDA was wrong.

**#4** — the loss-of-function arm is in, and you were right that its absence was inconsistent: I had praised the mouse adipogenesis evidence specifically for being bidirectional, then rested the human annotation on the gain-of-function half alone. The knockdown arm is also the stronger one — the p-FOXO1/3 result places AAMDC upstream of AKT rather than merely alongside mTORC1.

**#11 is where your review changed the work most.** You were right that a negative claim deserves testing. `AAMDC-bioinformatics/analyze_mth938.py` enumerates every reviewed Swiss-Prot member of AAMDC's Pfam family live from UniProt/InterPro/QuickGO — nothing hardcoded, re-runnable:

| Reviewed members of PF04430 (**DUF498**) | 13 |
|---|---|
| With any experimental MF term | **0** |
| With a UniProt CATALYTIC ACTIVITY block | **0** |
| With an explicit `GO:0003674` ND | 4 |

So the missing molecular function is a property of the **entire family**, not an oversight on this gene.

And it turned up something I would not have reached from the AAMDC literature: the family's one characterised branch is **NDUFAF3** — a mitochondrial complex I **assembly factor** in every organism studied. Non-catalytic, binds subunits, helps build a complex, not part of the finished product. The family signal points *away* from a hidden enzyme and toward an assembly/chaperone-like role. Which is precisely the class of function GO cannot express as an MF — the same gap I filed under `proposed_new_terms` for AAGAB (#2220). **Twice in five genes.** New `suggested_question`: not what AAMDC catalyses, but what it helps assemble.

**Other suggestions:** PMID:36789094 now has a `reference_review` explaining why it's set aside rather than sitting uncited in `publications/` (#7) — fair catch, a file landing with nothing accounting for it is untidy. The notes now state where the localisation bar sits and why the nuclear and endosomal-trafficking observations aren't annotated (#9). Rolland/HuRI mislabel fixed (#10). On #5/#6 and #8 I've deferred to 22b258bf8's resolutions, which reconcile the endolysosome wording and move `correctness` to VERIFIED.

Validation clean, 33 quotes verbatim.

cmungall:
Fixed, and you were right to call it the part that matters. The script claimed "fetched live from UniProt, InterPro and QuickGO, nothing hardcoded" while never calling InterPro — `INTERPRO` was a constant interpolated into its own report, and the SUPFAM/Gene3D/PANTHER ids were literal strings. The accessions were correct so nothing downstream was scientifically wrong, but that claim had propagated into `reference_review.review_notes`, i.e. into the curation record, where a curator would read it as InterPro having been interrogated. That's squarely what CLAUDE.md prohibits.

**I fixed it by making the claim true rather than by weakening it** — the script now actually calls the InterPro API and reports the entry's own name, type, protein count and member databases.

That paid off immediately by correcting two of my hardcoded strings:

- InterPro gives PF04430 as **"Protein of unknown function (DUF498/DUF598)"** — I had written just "DUF498".
- InterPro lists **PANTHER PTHR21192** for IPR007523, not the **PTHR15811** I had copied across from UniProt's own DR line. Those are different entries, and I had silently conflated two sources.

Which is a neat illustration of your point: hardcoded "sourced" data doesn't just misrepresent provenance, it also can't be corrected by the source. The report now also states that **8621** UniProtKB proteins carry the signature, which properly frames the 13 reviewed entries as a curated subset rather than as the family.

**On the silent-failure suggestion** — taken, and it deserved to be blocking-adjacent rather than a suggestion. `get()` now returns `None` on failure instead of an empty list, each member records whether its GO query succeeded, RESULTS.md has a "GO retrieved" column plus a per-run count, and a warning block fires if any query failed. The reasoning is in the docstring: *a silent fetch failure would manufacture the very negative result this analysis exists to test.* An analysis whose conclusion is "we found nothing" has to be able to tell "looked, nothing there" from "didn't look".

**The unconditional NDUFAF3 narrative** is now conditional on the fetched rows — if no member carries a complex-assembly FUNCTION statement, the report says so rather than asserting the conclusion regardless. Same class of problem as the InterPro claim: prose that can't be falsified by its own data.

Also done: urllib instead of curl-via-subprocess, `pyproject.toml` for the `uv run` invocation, `.venv` gitignored, the duplicated sentence at AAMDC-notes.md removed, stray blank line gone.

Validation clean, 33 quotes verbatim, InterPro/UniProt/QuickGO all returning data on this run (13/13 GO queries succeeded).

PR #2223, opened 2026-07-25
Review human AAMP: a self-referential IBA, and three annotations for a missing mechanism

cmungall:
PAINT no-IBA project gene review for **AAMP** (angio-associated migratory cell protein,
Q13685), using the `affinage` deep-research provider (gates passed).

## Finding 1 — a self-referential IBA

`GO:0014909 smooth muscle cell migration` is annotated **twice**: once IBA, once IEP. The
IBA's WITH/FROM field is `PANTHER:PTN001068868|UniProtKB:Q13685` — and **`Q13685` is human
AAMP itself**, the gene being annotated (confirmed by UniProt lookup: `AAMP_HUMAN`).

So the phylogenetic inference is seeded by the target gene's own annotation and transfers
nothing back to it. Recorded as `EVIDENCE_CIRCULAR_OR_REDUNDANT` / `CIRCULAR_PROPAGATION` in
`propagation_review`.

**`ACCEPT`, not `REMOVE`** — the term is correct and core for this gene. The IBA is redundant,
not wrong, and the distinction matters: this is a provenance defect, not a biology defect.
Whether GO_Central should be detecting self-referential IBAs systematically is raised as a
`suggested_question`.

## Finding 2 — the annotation it derives from is under-coded

The direct annotation is `IEP` (expression pattern). But PMID:18634987 is not an expression
study. It reports that AAMP overexpression *increases* smooth muscle cell migration while
antibody treatment and siRNA *decrease* it; that knockdown reduces membrane-fraction RhoA
activity; and that antibody blockade reduces neointimal SMC density and neointima formation in
apoE-null mice. That is IMP-grade perturbation evidence, in vitro and in vivo.

Not silently re-coded here — raised in `suggested_questions` for the responsible curator. It
matters doubly, because the under-coded IEP is what the circular IBA is effectively
transferring.

Two other evidence codes similarly understate the current literature: `GO:0001525 angiogenesis`
is `NAS`, and `GO:0010595` is `ISS` from bovine AAMP even though the human gene now has
stronger direct evidence than the ortholog it was transferred from. Both `ACCEPT`ed with the
gap noted.

## Three NEW annotations for a mechanism the record omits entirely

Apart from heparin binding, AAMP had **no molecular function in GOA** — nothing connecting it
to *how* it controls migration. UniProt is barely better, offering one hedge: "may act through
the RhoA pathway". Meanwhile three groups have worked out the mechanism:

| Term | Evidence |
|---|---|
| `GO:0031267` small GTPase binding | binds RhoA (PMID:34901393); binds CDC42 (PMID:33279622) |
| `GO:0050821` protein stabilization | shields RhoA from SMURF2 ubiquitination/degradation |
| `GO:0051057` positive regulation of small GTPase mediated signal transduction | knockdown drops membrane RhoA activity; RhoA/ROCK mediates the angiogenic effect |

Term-choice note: `GO:0017048 Rho GTPase binding` would have been more specific, but a direct
id lookup shows it is **obsolete**. `GO:0031267` is the correct live term. (I added that check
to the workflow after the AAGAB review (#2220), where the opposite error — assuming a term was
*absent* because a keyword search missed it — cost a review round.)

## The `protein binding` set, disjoint from the biology again

All four `GO:0005515` IPIs report the same partner, **AEN**, an apoptosis-enhancing nuclease,
with no functional follow-up and no described connection to AAMP. None of the four recovered
RHOA, RHOB, CDC42, NOD2 or CD276 — every partner whose mechanism is actually known.

This is now the **third gene in this campaign** (after AAGAB #2220 and AAMDC #2221) where the
high-throughput binding record and the functionally characterised interactions are disjoint
sets. `MARK_AS_OVER_ANNOTATED` rather than `REMOVE`: four screens including an orthogonal
AP-MS method is genuine reproducibility, and the pair may yet mean something.

## Provider note

Affinage was at its most useful here — assembling scattered RhoA/RhoB/CDC42 papers into one
mechanistic picture is what made the three new annotations possible. Its GO grounding was not
imported, and one item in it is worth naming: `GO:0008289 lipid binding` has no support in any
cited paper and looks like a spurious inference from the heparin/glycosaminoglycan work.

`just validate human AAMP` clean; 39 quotes verified verbatim.


cmungall:
self-referential IBA is valid. It means the curator has reviewed the source annotation and thinks it is a core function, it makes sense evolutionarily (there will be an IBD further up the tree)

However, IEP is a little suspect

cmungall:
All four blocking items addressed, plus the suggestions. Two of these were substantive misses on my part.

**#1 — PMID:39404373 cached but uncited.** Fair and slightly embarrassing: two claims in the review depended on it ("a third study reports RhoA and RhoB stability", and RHOB in the partner list) while the paper sat in `publications/` with no entry. Now in `references` with quoted findings, and a second `supported_by` on both `GO:0031267` and `GO:0050821`. It deserved that weight — it's the third independent lab reporting AAMP-dependent Rho regulation, in a third cell type, and the one that extends it to RhoB.

**#2 — the missed endothelial-barrier annotation.** You're right, and it's a good one: siRNA depletion of AAMP *increases* trans-endothelial electrical resistance, so AAMP is a **negative** regulator of barrier function, and the mechanism is the same Rho axis the other new terms describe (depletion lowers RhoA/RhoB activity → less contraction in resting endothelium). Added as `GO:1903141` (IMP). GOA has nothing here.

**#3 — the NOD2 arm.** This is the correction I most needed. I left it unannotated because "the mechanism is unresolved" — which conflates *mechanism* with *evidence*. As you say, GO regulation terms don't require a resolved mechanism, and PMID:19535145 has overexpression plus siRNA in HEK293T, which is IMP-grade.

Annotated to the **unsigned parent `GO:0070432`**, taking your "pick the non-directional parent and say so" option. The paper's word is "modulates" and it never establishes a direction, so `GO:0070434`/`GO:0070433` would assert more than the data support. The reasoning is written into the annotation rather than left implicit.

**#4 — CD276.** Dropped from the description. It came from the affinage narrative citing PMID:35919070, which is neither cited nor cached here. An unverifiable partner shouldn't sit in a description — that's exactly the kind of provider claim that needs a citation or nothing.

**Suggestions:** the `GO:0017048` obsolescence lookup is now pasted into the notes rather than asserted (#5) — reasonable to want a load-bearing negative claim shown rather than trusted, especially in a campaign where I've already made one wrong claim about term existence. `GO:0009986` re-anchored to the actual localisation sentence (#7), with a caveat I found while re-reading PMID:39404373: in a *confluent* monolayer, ectopic AAMP was nuclear/cytosolic and explicitly **not** at the plasma membrane, so the surface pool may be context-dependent. The three core process terms named in prose are now in `core_functions.directly_involved_in` (#8).

On **#6** I've kept `GO:0051057` over `GO:0035025`. CDC42 is formally Rho-family, but "Rho protein signal transduction" reads as the RhoA/B/C branch, and the evidence here spans both branches (RhoA/RhoB *and* CDC42). Happy to switch if you read the term's scope more inclusively.

Validation clean; 48 quotes verbatim.

cmungall:
Both blockers fixed in `c88376e11`. I got both of these wrong, in different ways.

**A — the self-referential IBA.** @cmungall is authoritative here and I had it backwards: *"a self-referential IBA is valid. It means the curator has reviewed the source annotation and thinks it is a core function, it makes sense evolutionarily (there will be an IBD further up the tree)."*

So the self-reference is a **curation act, not a pipeline artefact** — it records that a PAINT curator inspected the source annotation, judged the function core rather than peripheral, and found the assignment evolutionarily coherent. That is an independent judgement *layered on* the experimental annotation, not a circular restatement of it. Changed to `NO_FAILURE_CORE`, summary and reason rewritten, and the now-answered `suggested_question` about detecting self-referential IBAs systematically removed.

The general error is worth naming: I inferred a failure mode from the *shape* of a WITH/FROM field without knowing what PAINT curators actually do with it. Unusual provenance is not automatically defective provenance.

The IEP half of that finding stands, and he endorsed it — *"However, IEP is a little suspect"*.

**B — `GO:0017048`.** You were right to challenge the lookup, and the answer turns out to be that **neither of us had it right.** QuickGO:

```
GET /QuickGO/services/ontology/go/terms/GO:0017048
  → id: GO:0031267   name: small GTPase binding   isObsolete: False

GET /QuickGO/services/ontology/go/terms/GO:0031267/complete
  → secondaryIds: [GO:0005084, GO:0008536, GO:0017016, GO:0017031,
                   GO:0017048, GO:0017049, GO:0017137, GO:0017160,
                   GO:0030306, GO:0034989, GO:0048365]
  → synonyms include: "Rho GTPase binding"
```

`GO:0017048` was **merged into** `GO:0031267`. It is a secondary id, and "Rho GTPase binding" survives as a synonym of the merged term. Not obsolete-and-detached (my claim), and not a live child (yours). A QuickGO search for "Rho GTPase binding" returns no separate live term.

The term choice is unchanged but properly grounded now: **there is no Rho-specific binding term to prefer, because GO deliberately merged that granularity away.** `GO:0031267` *is* that term. That also settles suggestion #6 from the last round more cleanly than my earlier answer did.

Method note recorded in the notes for the rest of this campaign: OLS's response for a *merged* id is genuinely ambiguous — it returns `is_obsolete: true` with a bare CURIE as the label, which is also what an absent id looks like, exactly as you diagnosed. QuickGO's `/complete` endpoint lists `secondaryIds` and so distinguishes merged from obsoleted. Worth using whenever a term-availability claim is load-bearing — this is the second time in this campaign a claim about term existence has needed correcting.

**Suggestions:** CD276 removed — you're right that it was in the `GO:0005515` `review.summary`, not the description, and I had assumed otherwise when addressing this last round. `additional_reference_ids` reconciled between `GO:0031267` and `GO:0050821`. The NOD2 arm now states explicitly why it is annotated but kept out of `core_functions`: single 2009 study, no follow-up, mechanistically disconnected from the Rho axis that accounts for everything else the gene does.

Validation clean; 48 quotes verbatim.

PR #2224, opened 2026-07-25
Review human AAR2: the protein-binding pattern breaks, and three NAS understate a structural paper

cmungall:
PAINT no-IBA project gene review for **AAR2** (U5 snRNP assembly factor, C20orf4, Q9Y312),
using the `affinage` deep-research provider.

## The `protein binding` pattern breaks here — and that is the point

The previous four genes in this campaign (AAGAB #2220, AAMDC #2221, AAMP #2223) all showed the
same thing: `GO:0005515` records disjoint from the gene's characterised biology. It would have
been easy — and wrong — to apply that verdict to AAR2's four IPIs as a block.

Resolving the WITH/FROM accessions splits them cleanly:

| Partner | Screens | Assessment |
|---|---|---|
| **EAPP** (Q56P03) | ×3 (two-hybrid + BioPlex AP-MS) | E2F-pathway protein, no splicing connection, no follow-up → `MARK_AS_OVER_ANNOTATED` |
| **TSSC4** (Q9Y5U2) | PMID:34131137 | **A genuine U5 snRNP partner** → `MODIFY` |

The TSSC4 source is not a screen at all — it is a dedicated study establishing TSSC4 as *"a
specific chaperone that acts in U5 snRNP de novo"* biogenesis and recycling, which *"interacts
with U5-specific proteins PRPF8, EFTUD2"* and SNRNP200. AAR2 and TSSC4 are **both U5 snRNP
assembly chaperones on the same particle with overlapping partners** — their interaction is
what the biology predicts, not an unexplained hit. Modified to `GO:0044877` (the interaction
occurs in the context of the particle) rather than dismissed.

The generalisable lesson: high-throughput binding records are *often* disjoint from the
characterised biology, but not always. Resolve the accessions; don't pattern-match.

## Three NAS annotations understate their own source

`GO:0000387`, `GO:0005682` and `GO:0048025` are all coded **NAS** — non-traceable author
statement — from PMID:36322420. But that paper is a structure-function study of the human
AAR2–PRPF8 RNase H complex, with a crystal structure, designed interface variants that fail to
bind PRPF8 in vitro, and SEC interaction mapping against U5 proteins. Direct experimental
results on the human protein. All three `ACCEPT`ed, with the evidence-code question raised in
`suggested_questions` rather than silently re-coded.

## `GO:0048025` looks wrong and isn't

*Negative* regulation of mRNA splicing, for a splicing assembly factor, reads like an error. It
is correct, and the reason is the mechanism: AAR2 occupies the PRPF8 surface that SNRNP200/Brr2
must engage and blocks U4/U6 di-snRNA loading, holding the particle immature until the
phosphoregulated handoff. The human structure sharpens it further — AAR2 *"seems to lock PRPF8
RH in a conformation that is only compatible with the first"* transesterification step. The
inhibition **is** how AAR2 enforces assembly order.

No `NEW` terms proposed. Unusually for this campaign the record already covers the biology; the
problems are evidence codes, not missing content.

## Provider note — a gate false positive worth recording

The affinage record **tripped its symbol-collision gate**: the narrative opens on yeast, and
the gate flags non-human tokens in a human record (the ADA failure mode). On inspection it is a
false positive of a specific, recurring kind. The narrative correctly covers yeast **Aar2p**,
where the mechanism was solved structurally, *and* human **AAR2**, its genuine ortholog —
UniProt's own human FUNCTION line is `ECO:0000250|UniProtKB:P32357`, i.e. by similarity to the
yeast protein. A yeast-centred account is the *right* description of this gene's literature.

**This gate will misfire for any gene whose mechanism was established in yeast and transferred
to human.** It still did its job by prompting the check; it just needs curator adjudication
rather than discounting the record. Recorded in `reference_review` and the notes.

## Schema note

`GO:0005682 U5 snRNP` goes in `core_functions.in_complex`, not `locations` — complex membership
is bound to `GOProteinContainingComplexEnum`, locations to `GOCellularLocationEnum`. Putting it
in `locations` fails term validation, correctly. Noted for other complex-component genes in this
campaign.

`just validate human AAR2` clean; 26 quotes verified verbatim.


cmungall:
Both blocking findings were right, and both were refuted by papers I had already cited in this PR — which is the uncomfortable part and worth naming.

**#1 — EAPP.** I wrote that EAPP has "no described connection to splicing or to snRNP assembly". PMID:34131137 — *the very paper I used to elevate the TSSC4 annotation* — says: *"a putative chaperone EAPP that has been shown to interact with U5 proteins PRPF8 and EFTUD2"*.

I judged EAPP on its best-known identity (E2F-associated phosphoprotein) rather than checking it against this gene's own literature. Which is ironic given the PR's stated lesson was "resolve the accessions; don't pattern-match" — I did resolve them, then pattern-matched on what the name is famous for.

The verdict survives, but on much narrower grounds, and I've rewritten all three rows to say so: EAPP's U5 role is explicitly *putative*, the AAR2–EAPP pair specifically has never been pursued, and the term is uninformative regardless of whether the partner is real. The review now states that **if EAPP's U5 chaperone role is confirmed, these rows deserve upgrading rather than discounting** — a materially different position from "cross-pathway artefact".

**#2 — the displacement model.** Also correct, and a cleaner error: I took a yeast-established mechanism and stated it as human fact. PMID:34131137 is explicit that human differs — *"In yeast, Aar2p is exchanged with Brr2p, while in humans AAR2 co-purifies with all four RHC members, which indicates that AAR2 stays associated until the whole RHC is formed"* — and PMID:36322420 reports "marked differences" in the human SEC interaction patterns.

Fixed in the `description`, `core_functions`, the `GO:0048025` summary and the U5 snRNP entry, all now saying yeast-established with the human qualification. Added as a `suggested_question`: whether the SNRNP200-displacement model transfers to human at all is genuinely open.

There is a general lesson here for the campaign. This gene's whole framing is "mechanism solved in yeast, ortholog confirmed in human", which makes it *easy* to import the yeast model wholesale — the affinage narrative does exactly that. The place to check is the human-specific papers, which say something different.

**#3 — `GO:0044877` vs its evidence line.** Recorded rather than glossed. You are right that the term expects a complex in WITH/FROM while this row has a single accession, and that the AAR2-specific datum is IP-MS co-purification with the paper declining direct contacts ("multiple weak contacts"). I have kept the term — co-purification within U5 particles is precisely evidence of engagement in a complex context, which a pairwise term misses — but written the tension into the annotation, and noted that a curator may reasonably prefer to leave it as `GO:0005515` pending a demonstrated direct contact. Added as a question.

**Also:** `molecular_function` added to `core_functions` (an omission, since the PR proposed one), and `GO:0045292` / `GO:0000244` added to `directly_involved_in`, since both were called core in prose.

Validation clean; 28 quotes verbatim.

cmungall:
Fixed.

**🟡 The stale EAPP claim in `suggested_questions`.** You're right that it was the same statement the annotation rows 250 lines above had already retired. Rewritten around what is genuinely open — whether EAPP's *putative* U5 chaperone role is confirmed, and whether the AAR2–EAPP contact is an assembly-context interaction — quoting PMID:34131137 directly so the question can't drift back to the refuted framing.

**🔵 EAPP rows citing the AAR2 UniProt FUNCTION line.** Both the first and third rows now cite the PMID:34131137 sentence instead, matching the second row. You were right that the UniProt line is about AAR2 and says nothing about the interaction — it passed the substring check while supporting nothing, which is the failure mode worth closing.

**🔵 `GO:0045292` in `directly_involved_in` vs its own summary.** Kept the term and made the two statements agree explicitly: `core_functions` now says it is retained on the same reading as the IBA row — the process an assembly factor is *required for*, not one it acts in, AAR2 being excluded from the assembled spliceosome. Both readings are defensible as you say; the problem was that the file appeared to hold both without noticing.

`just validate human AAR2` → `✓ Valid`.

PR #2225, opened 2026-07-25
Review human AARD: a dark gene, with evidence for why its only annotations are an artefact

cmungall:
PAINT no-IBA project gene review for **AARD** (alanine- and arginine-rich domain-containing
protein, C8orf85, Q4LEZ3), using the `affinage` deep-research provider.

## The darkest gene in this campaign

AARD's **entire** GO record is 15 rows, every one `GO:0005515 protein binding`, all IPI, all
from a single publication (HuRI, PMID:32296183). No molecular function beyond binding, no
cellular component, no biological process. UniProt has no FUNCTION line and no SUBCELLULAR
LOCATION.

Affinage agrees there is nothing to ground — `molecular_activity: (none)`, `localization:
(none)`, `partners: (none)`, and the narrative states it plainly: *"no molecular activity,
interaction partners, or cellular function for AARD have been characterized in the available
corpus"*. Worth crediting: for a gene this dark, a provider declining to manufacture a
plausible function is the useful contribution.

So the review turns on one question: **should those 15 interactions be believed?**

## Testing the interactions instead of just doubting them

The easy move is "single publication, no follow-up, discount it". Instead I tested a specific,
falsifiable alternative — **coiled-coil bias**, the known tendency of sticky or self-activating
Y2H preys to recover coiled-coil proteins through promiscuous heptad-repeat association.
`AARD-bioinformatics/analyze_partners.py` reads the accessions from the gene's own
`AARD-goa.tsv` (so they cannot drift from the record under review) and fetches UniProt features
at run time.

**9 of 15 partners carry an annotated coiled-coil region, and the 15 span 20 distinct
subcellular locations.**

| Partner | Coiled-coil | Compartment |
|---|---|---|
| GRIPAP1 | 4 segments | endosome membranes |
| CEP57 | 2 | centrosome |
| **STX1A, STX2, STX5** | 1 each | synaptic vesicle / cell membrane / Golgi+ERGIC |
| KIAA0753, CENPQ | 1 each | centriolar satellite / centromere |
| KRT24, KRT27 | keratins | cytoskeleton |
| TFIP11, NTAQ1, LMO4, MAGEB4, VPS37C, TSGA10 | — | nucleus, cytosol, secreted, late endosome |

**Three syntaxins** — SNAREs from three different membranes sharing a coiled-coil SNARE motif —
plus two keratins, centrosomal and centromeric coiled-coil proteins, a nuclear splicing factor
and a secreted protein. There is no compartment in which these could meet. What they share is
*architecture*, not biology.

`MARK_AS_OVER_ANNOTATED`, not `REMOVE`: the interactions may genuinely have occurred in the
assay, and I have no positive evidence that any individual pair is false.

## No NEW terms — and that is the finding

Everything known about AARD is mouse expression biology: Sertoli-cell-restricted transcription
rising during testis differentiation (PMID:17486547), and direct androgen-receptor target
status (PMID:27959439).

**Neither supports a GO annotation.** Being a transcriptional *target of* a pathway is a
downstream relationship, not participation in it — which is precisely the `ROLE_CONFLATION` I
found in the mouse *A1bg* growth-hormone annotation earlier in this campaign (#2217). It would
have been easy to reach for an androgen-receptor-signalling term here and repeat the same error
in a new gene. Recorded in the description and `suggested_questions` instead.

Note also that PMID:27959439's own phrase *"which is important in spermatogenesis"* is about
the AR programme, not a demonstrated AARD requirement — **no AARD loss-of-function experiment
has ever been reported.**

## Outcome

AARD stays dark, which is the correct result. `just validate` emits **"No core functions
defined"** and I have left it standing rather than inventing something to silence it — the
warning is accurate.

What this review adds is: evidence-based grounds for discounting the only annotations the gene
has; an explicit record of why the expression/AR biology must not become process terms; and
four experiments aimed at the actual gap, including mining testis co-expression, which for a
gene whose only reliable data is expression is a better lead than its interactome.

## Coverage note

The GOA file has 15 rows; `existing_annotations` has one entry. Not missing coverage — all 15
share an identical term, qualifier, evidence code and reference, differing only in WITH/FROM,
so they are one annotation with 15 interactors. All 15 are named and analysed.

`just validate human AARD` passes; 6 quotes verified verbatim.


cmungall:
All three addressed in `63824931e`. Every one concerned the evidence behind the artefact argument rather than the action it leads to, which is the right place to push.

**#1 — the family.** You're right and it was contradicted by the gene's own UniProt file, which I had already read. AARD is a **FAM167-family** protein: `PANTHER PTHR32289`, `InterPro IPR051771`, sharing the family with FAM167A and FAM167B. Writing "no recognisable domain" for a gene whose whole remaining inference route is its paralogs was the worst place to be careless.

Enumerating the family is now part of the script, and the result is better than a correction — it is a stronger version of the same conclusion:

| | |
|---|---|
| Reviewed members of PTHR32289 | **10** (across 5 species) |
| With a UniProt FUNCTION statement | **0** |

So guilt-by-association is unavailable **not because AARD lacks a family, but because the entire family is uncharacterised.** Your suggestion to cite the PAN-GO line was exactly right — `PAN-GO; Q4LEZ3; 0 GO annotations based on evolutionary models` independently corroborates it, and I've used it.

That also raises a better question than any about AARD alone, now added: this is a conserved vertebrate family, fish through mammals, in which *no member* has a described function. Solving any one would likely illuminate the rest, and no member is currently a better starting point than another.

**#2 — the inflated location count.** Correct. UniProt location strings are hierarchical, so `Cytoplasm`, `Cytoplasm, cytosol` and `Cytoplasm, cytoskeleton, … centriole` were counted as three distinct locations. Collapsed to leading compartment: **14, not 20**. The scatter conclusion survives — nucleus, centromere, Golgi/ERGIC, synaptic vesicle, late endosome and secreted really are irreconcilable — but the number now matches what it claims, in the script, the report, the review and the notes.

**#3 — the missing null model.** This was the most important of the three, and I think you understated it by filing it third. Without a background rate, "60% matches the artefact signature" is an assertion in precisely the shape this review spends its length being sceptical of. Fixed with a fetched background:

| | |
|---|---|
| Background: reviewed human proteins with a coiled-coil feature | **2059 / 20431 = 10.1%** |
| Observed in AARD's partner set | **9 / 15 = 60%** |
| Enrichment | **6.0-fold** |
| Binomial P(X ≥ 9 \| n=15, p=0.101) | **3 × 10⁻⁶** |

The report states explicitly that with n=15 this is descriptive rather than a controlled test, and names the comparison it does *not* make — other single-publication prey sets from the same screen, which would control for the screen itself rather than for the human proteome. I'd rather record that limitation than let a p-value imply more rigour than the design supports.

The action is unchanged (`MARK_AS_OVER_ANNOTATED`), but it now rests on a measured enrichment and a correctly-counted compartment spread rather than on two eyeballed figures.

cmungall:
Fixed — you were right that the criteria didn't match, and the corrected number is lower.

`analyze_partners.py` now computes both criteria consistently and **headlines the feature-only pair**, since that is the one with a matching background:

```
| With a coiled-coil FEATURE            | 7  |
| With a coiled-coil feature or keyword | 9  |

Background: 2059 of 20431 reviewed human proteins carry a coiled-coil feature = 10.1%
- Observed: 47% (7/15)
- Enrichment: 4.6-fold
- Binomial P(X >= 7 | n=15, p=0.101) = 3.26e-04
```

Rather than just dropping the keyword-only partners, I took the second option too and fetched the **keyword background** (`keyword:KW-0175` = 10.6%), so the looser criterion now has its own matching denominator: 9/15, **5.6-fold**, P = 4.8e-06. Both pairs are reported, and `RESULTS.md` says in as many words that mixing them — a keyword numerator against a feature denominator — would overstate the enrichment, so it isn't done. Note the keyword-inclusive figure is 5.6-fold, not the 6.0-fold that was quoted; the old number was wrong on both criteria, not just mis-scoped.

Corrected figures propagated to `review.summary`, `review.reason`, the `RESULTS.md` reference finding, and `AARD-notes.md`. The conclusion is unchanged, as you predicted.

**Optional item — "lost".** Fixed: `description:18` now says "significantly downregulated", matching PMID:27959439 and the file's own `findings` entry.

**Left:** the `RESULTS.md` location truncation at `[:3]` — the secreted partner is still not identifiable from the table. That's a display choice in the report generator rather than a correctness problem, and widening it makes an already-wide table wider; flagging rather than changing it unilaterally.

`just validate human AARD` → `✓ Valid` (one advisory warning: `core_functions: []`, which is deliberate here).

cmungall:
Fixed. You were right, and it is the same standard I applied to the "20 locations" count last round — I should have caught it there.

The numerator counted coiled-coil on **feature OR keyword** while the background query was **feature-only** (`ft_coiled:*`). Since UniProt applies the keyword more liberally than the positional feature, that mismatch inflated the enrichment.

Rather than pick a criterion, the script now computes both against their own matched backgrounds:

| Criterion | Partners | Background | Enrichment | Binomial P(X ≥ k) |
|---|---|---|---|---|
| coiled-coil **feature** | 7/15 (47%) | 2059/20431 (10.1%) | **4.6×** | 3.3×10⁻⁴ |
| feature **or** keyword | 9/15 (60%) | 2174/20431 (10.6%) | **5.6×** | 4.8×10⁻⁶ |

The strict feature-only pair is the headline, and reporting both makes the useful point that the conclusion does not turn on the choice. Propagated to `review.summary`, `review.reason`, the RESULTS.md finding and the notes.

Note the branch had moved again — `a6cbc7e92` landed the same fix with the same figures while I was working. I reset onto it rather than force-pushing, and added `0bd40911f` with what it didn't cover.

**On the location truncation** — worth doing, and it turned out to matter more than a display nicety. With `[:3]` removed, the answer to "which partner is the secreted one" is visible: **STX1A**, whose UniProt locations read *"synaptic vesicle membrane; synaptosome; cell membrane; secreted"*. The truncation was hiding part of the very compartment scatter the table exists to demonstrate — a syntaxin that is simultaneously synaptic, plasma-membrane and secreted is a better illustration of the artefact argument than the collapsed count is.

**AR-knockout wording** was already fixed in `a6cbc7e92` — "significantly downregulated" now matches PMID:27959439.

I've also recorded the general trap in the notes, since it will recur: **when a null model is built by querying the same database that supplies the numerator, the background query must use the identical predicate.** Easy to get wrong precisely because both queries look reasonable in isolation.

PR #2226, opened 2026-07-25
Review human AASDH: its one annotation is a paralog's substrate

cmungall:
PAINT no-IBA project gene review for **AASDH** (ACSF4-U26, Q4L235), using the `affinage`
deep-research provider.

## One annotation, and it belongs to a different enzyme

AASDH's entire GO record is a single ISS: `GO:0006631 fatty acid metabolic process`. Resolving
the WITH/FROM shows the problem — the source is **`UniProtKB:Q4G176`, which is ACSF3, a human
*paralog*, not an ortholog.**

ACSF3 genuinely is a fatty-acid enzyme: malonyl-CoA synthetase, which *"catalyzes the initial
reaction in intramitochondrial fatty acid synthesis, by activating malonate and
methylmalonate"*. The term is **right for the source and wrong for the target.**

AASDH is a **β-alanine-activating enzyme** (PMID:24467666):

- ATP-dependent covalent loading of β-alanine onto its own phosphopantetheine arm
- *"almost specific for β-alanine"* among the standard amino acids, K<sub>M</sub> ≈ 5 µM
- abolished in a point mutant lacking the phosphopantetheine attachment site

**β-alanine is a β-amino acid, not a fatty acid**, and no fatty-acid substrate has ever been
reported for AASDH. What the two proteins share is the ATP-dependent AMP-binding adenylation
fold and an ACSF family naming convention — "family member 4" versus "member 3". That is exactly
the resemblance that produces a wrong paralog transfer.

`REMOVE`, recorded as `PROPAGATION_BAD` / `WRONG_ORTHOLOG_OR_PARALOG`. This is a sanctioned use
of REMOVE: a demonstrably wrong sequence-based inference, argued against direct enzymology of
the gene itself, not the second-guessing of an experimental annotation.

## Replacing zero annotations with three

Removing it leaves AASDH with **nothing**, so three `NEW` terms cover what the enzymology
actually supports:

| Term | Aspect | Basis |
|---|---|---|
| `GO:0016878` acid-thiol ligase activity | MF | the demonstrated ATP-dependent thioester chemistry |
| `GO:0031177` phosphopantetheine binding | MF | carrier domain 553–630; point mutant abolishes the reaction |
| `GO:0019482` beta-alanine metabolic process | BP | near-absolute substrate specificity |

Affinage grounded the activity as `GO:0016874 ligase activity` / `GO:0140657 ATP-dependent
activity` — correct but two levels too general, so not imported; `GO:0016878` is the specific
descendant.

## What is deliberately *not* proposed

No β-alanylation or protein/RNA-modification process term, despite the temptation. **The
physiological acceptor has never been identified.** Thiol transfer was observed in vitro but the
authors themselves called it *"physiologically irrelevant"*, and UniProt describes the acceptor
as "as yet, unknown". Annotating the hypothesised post-translational modification would assert a
pathway nobody has demonstrated — recorded as the gene's central open question instead, with a
targeted experiment.

`GO:0005524 ATP binding` is also not proposed despite four annotated ATP-binding sites: it adds
nothing over the ligase term, which already entails ATP dependence.

## Flagged upstream

UniProt's own keywords for AASDH still include **`Fatty acid metabolism`** and **`Lipid
metabolism`**, which contradict UniProt's own FUNCTION line describing β-alanine activation.
They look like the same paralog assumption that produced the GO ISS. Raised in
`suggested_questions`.

## Provider note

Affinage was accurate and appropriately restrained here — it identifies the β-alanine
specificity, the phosphopantetheine requirement and the domain carrying the activity, and states
plainly that *"no downstream pathway or in vivo role for AASDH has been established"* rather than
inventing one. It says nothing about fatty acids, which corroborates the removal — but as with
A1BG (#2217), that is silence, not a positive signal. The provider still does not tell you when
an existing annotation is wrong.

`just validate human AASDH` clean; 17 quotes verified verbatim.


cmungall:
Fixed, and I agree the change strengthens rather than weakens the argument.

All three NEW annotations are now `ISS` with `supporting_entities: [UniProtKB:Q80WC9]` — `GO:0016878`, `GO:0031177` and `GO:0019482`. Each `reason` states why, and the `GO:0016878` one makes the point explicitly: this review *removes* an ISS for naming the wrong source (the paralog ACSF3/Q4G176), so asserting a direct human assay for evidence that is itself an ortholog transfer would apply the standard in one direction only. A correctly sourced ISS against an incorrectly sourced one is exactly the contrast the PR is arguing for.

Confirmed both halves of your evidence before changing it: the quote says *"purified **mouse** recombinant ACSF4-U26"*, and `AASDH-uniprot.txt:7` carries `RecName: Full=Beta-alanine-activating enzyme {ECO:0000250|UniProtKB:Q80WC9}`.

**🔵 EC number** — added `DE            EC=6.2.1.-;` as a `supported_by` on `GO:0016878`. Good catch: 6.2.1 is literally the acid-thiol ligase subclass, so it corroborates the term choice independently of the assay.

**🔵 description** — now says the enzymology was established on the mouse orthologue, that UniProt annotates the human entry by similarity from Q80WC9, and that no assay of the human protein has been reported.

**🔵 `proposed_new_terms` for beta-alanine—[carrier protein] ligase** — not added. The granularity gap is real, but proposing a term for a reaction whose physiological acceptor UniProt calls "as yet, unknown" would assert the pathway this review is deliberately declining to assert. Flagging rather than filing it.

`just validate human AASDH` → `✓ Valid`.

cmungall:
You're right, and the inconsistency was in this PR's own thesis. It removes an ISS for citing the wrong source, then asserts IDA for evidence that is *itself* an ortholog transfer — the assay was on *"purified **mouse** recombinant ACSF4-U26"*, and UniProt annotates the human FUNCTION line as `ECO:0000250|UniProtKB:Q80WC9`. My own `reference_review` said so while the `evidence_type` said otherwise.

All three `NEW` annotations are now `ISS` with `supporting_entities: [UniProtKB:Q80WC9]`, and each summary explains why. As you say, this **strengthens** the argument, because it puts the two transfers side by side:

| | Source | Relationship | Verdict |
|---|---|---|---|
| **Removed** ISS | `Q4G176` = ACSF3 | **paralog** | wrong transfer |
| **Proposed** ISS | `Q80WC9` = mouse Aasdh | **ortholog** | legitimate transfer |

Same evidence code, opposite verdicts — which is the actual lesson of this gene. ISS is not weak by nature; it is only as good as the source relationship. That contrast is now stated explicitly in the notes.

**On `EC=6.2.1.-`** — good catch, and stronger than a supporting quote usually is: EC 6.2.1 *is* the acid-thiol ligase subclass, so it corroborates the term choice from an independent classification system rather than from the same paper. Added.

**On the missing granularity** — added as a `proposed_new_terms` entry: *beta-alanine—[carrier protein] ligase activity*. `GO:0016878` is correct but says nothing about the substrate or the acceptor, and those are exactly the two facts that distinguish this enzyme. The acceptor point matters most: AASDH loads onto **its own phosphopantetheine arm**, not free coenzyme A, which is precisely what separates it from the acyl-CoA synthetase paralogs — **including ACSF3, the source of the erroneous ISS this PR removes**. A term at that granularity would make the confusion harder to repeat.

The proposed definition covers only the demonstrated half-reaction and deliberately names no downstream acceptor, since none has been identified.

**Description** now records that the biochemistry was established on the mouse ortholog and that no assay on the human enzyme has been reported.

Note the branch had moved — `a2a58f63c` landed the IDA→ISS and EC changes in parallel; I reset onto it rather than force-pushing and added `060914bd1` with the proposed term.

Validation clean; 20 quotes verbatim.

PR #2228, opened 2026-07-25
Review human ABHD18: the missing cardiolipin deacylase, with no MF or BP in GOA

cmungall:
PAINT no-IBA project gene review for **ABHD18** (Q0P651), using the `affinage` deep-research
provider.

## The gap

ABHD18 had **two GO annotations, both "mitochondrion", and no molecular function or biological
process at all** — despite UniProt naming it *"Cardiolipin-specific deacylase, mitochondrial"*
and carrying a Rhea-referenced `CATALYTIC ACTIVITY` block for the reaction.

A 2025 *Nature* paper (PMID:40903572) identified it as the long-missing enzyme of the
cardiolipin remodelling cycle. Nascent cardiolipin is deacylated to monolysocardiolipin, then
re-acylated by tafazzin with longer unsaturated chains. The reacylase has been known for years;
the deacylase *"responsible for cleaving nascent CL (nCL) has remained elusive"*. **ABHD18 is
that enzyme.**

## Two NEW annotations

| Term | Evidence | Basis |
|---|---|---|
| `GO:0004620` glycerophospholipase activity | IDA | *"ABHD18 converts CL into monolysocardiolipin (MLCL) in vitro"* |
| `GO:0035965` cardiolipin acyl-chain remodeling | IMP | inactivation in cells and mice shifts lipids toward nascent CL; genetic suppression of tafazzin loss |

`GO:0035965` is the term **tafazzin already carries**, so annotating ABHD18 to it completes the
enzyme pair for a process that until now had only one identified participant. Coded IMP rather
than IDA because the *pathway placement* rests on the loss-of-function work; the in vitro
conversion supports the MF annotation instead.

## A real GO gap: no cardiolipin deacylase term

GO covers the rest of cardiolipin metabolism at substrate-specific resolution —
`GO:0008808` synthase, `GO:0043337` CMP-forming synthase, `GO:0035755` phospholipase D,
`GO:0160241` dehydrogenase — but has **nothing for the deacylation step**, even though that step
is half of `GO:0035965` itself.

The gap is historical: the responsible enzyme was unidentified. The consequence is that the
deacylase and the reacylase currently **cannot be distinguished by molecular function**, only by
their shared process term. Filed under `proposed_new_terms`; the reaction is registered as
`RHEA:32935` so the chemistry is unambiguous.

## Why this gene matters clinically

Barth syndrome is caused by inactivating tafazzin mutations and is characterised by
monolysocardiolipin accumulation. Inactivating the **upstream** deacylase relieves the block:
ABHD18 loss rescues mitochondrial defects in Barth syndrome cells and rescues morbidity and
mortality in mouse models, and a selective covalent ABHD18 inhibitor rescues tafazzin-mutant
phenotypes in patient fibroblasts and fish embryos. Recorded in the description; no GO
annotation is derived from the disease link.

## Note on the validation warning

`just validate` emits **"No annotations reference available deep research files"**. That is
accurate and I have left it standing rather than adding a token citation to silence it.

The affinage record genuinely was not needed here: the function was published in 2025 and
UniProt has already curated it with a Rhea cross-reference, so UniProt plus the primary paper
are both more current and more precise than the provider. Recorded in the `reference_review`,
with the general observation that **affinage's value is highest for genes whose literature is
scattered and lowest for genes with a single recent definitive paper** — the opposite of the
AAMP case (#2223), where assembling scattered papers was exactly what made the review possible.

`just validate human ABHD18` passes; 13 quotes verified verbatim.


cmungall:
Both blockers fixed, and the first one changed the review's substance.

**#1 — PMID:40378955.** You were right that I neither cited nor adjudicated it, and right that it matters. My `reference_review` had concluded the provider "added nothing" — **that was wrong**, and the paper it pointed to qualifies my framing in three ways:

- **"Removes one acyl chain" was incorrect.** The paper is explicit: *"Rather than removing just one fatty acid, we show that ABHD18 deacylates CL further"* — stepwise past MLCL to dilyso-CL. Corrected in `description`, `core_functions` and the proposed term's definition.
- **Yeast Cld1 homology** and loss-of-function data in *Drosophila* and murine *Taz*-KO myoblasts, which corroborate the `GO:0035965` IMP. Added as supporting evidence.
- **A different framing of the ABHD18/TAZ relationship**: *"ABHD18 catalyzes the breakdown of CL, whereas TAZ protects CL from degradation"* — degradation opposed by protection, not partnership in a cycle.

I've kept `GO:0035965` but recorded both framings side by side rather than silently picking one. Both papers agree ABHD18 supplies the deacylation step; and **GO has no cardiolipin catabolic process term**, so the degradative reading cannot currently be expressed at all. That gap is now a `suggested_question`.

The general lesson is worth recording, and I've put it in the notes: this is a case where **the provider's value was in its reference list, not its prose**. Its narrative genuinely added little over UniProt's curated Rhea-referenced block — but its citations contained a paper I had missed.

**#2 — `ABHD18-notes.md`** added. My omission; CLAUDE.md requires it and every sibling in this series has one.

**On `GO:0004620` vs `GO:0004623`** — good question, and it resolves in favour of the child. `RHEA:32935` writes the product as `1'-[1,2-diacyl-sn-glycero-3-phospho],3'-[1-acyl-sn-glycero-3-phospho]-glycerol`: the chain remaining on the deacylated glycerol is at **sn-1**, so the chain removed was at **sn-2** — A2 regiochemistry. Switched to `GO:0004623`, which is now also the proposed term's parent. RHEA:32935 and the Cld1 parallel are both in the term request.

Validation clean; 20 quotes verbatim.

cmungall:
All three 🟡 items and the first 🔵 are fixed.

**1. `core_functions` stepwise wording.** You're right that the quote refuted the sentence it was attached to. Now: *"ABHD18 removes acyl chains from tetra-acylated cardiolipin stepwise — first to tri-acylated monolysocardiolipin and, on further hydrolysis, past it to dilyso-cardiolipin"*. The `description` and `core_functions` now say the same thing.

**2. `full_text_unavailable` on PMID:40378955.** Removed — the cache does have Results and Discussion. More usefully, I took your pointer to what the Discussion actually contains and added two findings from it:

- *"This is inconsistent with the orthodox remodeling mechanism but suggests that ABHD18 and TAZ have opposing effects on CL metabolism."*
- *"Apart from its participation in TAZ-catalyzed remodeling, ABHD18 alters the CL species composition by species-selective degradation."*

The second changes the argument rather than just documenting it: the review now retains `GO:0035965` **on the dissenting paper's own words** — that clause concedes the participation — instead of retaining it *despite* the paper. And the `review_notes` no longer soften the disagreement to "framed differently"; it records that the authors call the orthodox cycle "inconsistent" with their backbone-turnover data and call ABHD18's degradation "a function that is separate from CL remodeling".

**3. The A2-type claim.** Caveated in both the `summary` and the `reason`: RHEA:32935's sn-2 placement is a reaction *representation* sitting behind a UniProt `ECO:0000305` curator inference, and neither 2025 paper determines regiochemistry experimentally. `GO:0004623` is kept as the more informative of the two available terms, explicitly provisional, with a note that the row should move if regiospecificity is later measured and differs.

**🔵 `GO:0004620` in the justification.** Fixed — it now names `GO:0004623` as the most specific available term, with `GO:0004620` mentioned only as the less informative parent.

**Not done:** the FlyBase-assigned HTP row remains unremarked. Flagging rather than writing a note I'd be inventing.

`just validate human ABHD18` → `✓ Valid`. (One thing worth knowing for future edits: an unquoted plain scalar containing `': '` breaks the YAML parser — my first attempt at the `review_notes` rewrite did exactly that.)

PR #2229, opened 2026-07-25
Review human ABHD14A: no function, a paralog hazard, and an affinage preprint caveat

cmungall:
PAINT no-IBA project gene review for **ABHD14A** (Q9BUJ0), using the `affinage` deep-research
provider.

## Outcome: stays dark, with citations for why

Three GO annotations, all localisation. No molecular function, no biological process. That does
not change here — the value is in stating *why* with evidence rather than leaving the gene blank.

PMID:37974539 says it directly: *"While ABHD14A still lacks any functional annotation to date, we
recently showed that ABHD14B functions as a lysine deacetylase in mammals."*

## The specific hazard: ABHD14B

The same paper documents the failure mode most likely to produce a wrong annotation on this gene:

> *"Given their high sequence similarity, automated databases often wrongly assign ABHD14A and
> ABHD14B as the same enzyme, and therefore, annotating functions to them in various organisms
> has been problematic."*

ABHD14B **is** characterised — a lysine deacetylase. ABHD14A is not. Sequence determinants
separating them had to be defined explicitly because databases kept merging them.

This is the AASDH pattern (#2226) — where a paralog's substrate had already been transferred onto
the target — caught *before* it happens rather than after. Flagged prominently in
`suggested_questions`: any future annotation on ABHD14A should be checked against those
determinants, and any transfer from ABHD14B treated as a **paralog** transfer.

## Actions

| Term | Evidence | Action |
|---|---|---|
| `GO:0005737` cytoplasm | IBA | ACCEPT |
| `GO:0005737` cytoplasm | IEA | ACCEPT — but note UniProt's own location is `ECO:0000250`, so this is an inference on an inference |
| `GO:0016020` membrane | IEA | MARK_AS_OVER_ANNOTATED — no membrane specified, no topology, underlying assertion is `ECO:0000305` curator inference |

**No `NEW` terms**, deliberately. The granule-neuron link comes from *Dorz1*, a rodent transcript
expressed in differentiating cerebellar granule neurons and regulated by Zic1 — expression
biology. Being downstream of a transcription factor is not participation in a process, the same
conflation identified for A1BG (#2217) and AARD (#2225).

## ⚠️ Provider caveat worth recording campaign-wide

Affinage's two most substantive claims for this gene — that a truncated ABHD14A hydrolyses
short-chain esters with CoA enhancement, and that the full-length protein localises to the Golgi
— both cite:

```
PMID:bio_10.1101_2025.11.28.691245
```

**That is not a PubMed identifier.** It is a bioRxiv DOI encoded in a PMID-shaped field. A reader
skimming the record for PMIDs would take preprint claims for peer-reviewed ones.

Both claims are excluded from this review and the issue is recorded in the `reference_review`. If
the preprint is published they would justify a hydrolase molecular function and a Golgi location
— neither of which is annotated now. Worth knowing for the rest of this campaign: **check the
affinage citation list for non-PMID identifiers before relying on a finding.**

## Validation warnings, left standing

`just validate` emits **"No core functions defined"** and **"No annotations reference available
deep research files"**. Both are accurate. The gene has no established function, and the
provider's findings were excluded for the reason above — adding a token citation to silence
either warning would misrepresent the state of knowledge.

`just validate human ABHD14A` passes; 5 quotes verified verbatim.


cmungall:
Both blockers fixed.

**#1 — the topology claim.** You're right, and the mechanism of the error is worth naming: my `supporting_text` stopped **one line short** of the topology statement. I quoted `SUBCELLULAR LOCATION: Cytoplasm ... Membrane` and then wrote "no topology is known" — while the very next line reads `Single-pass type II membrane protein`, with `FT TRANSMEM 35..55 /note="Signal-anchor for type II membrane protein"` and `CARBOHYD` at 67 and 201. Quoting up to the edge of the contradicting evidence is a failure mode I should watch for.

The action is unchanged but the reason is rebuilt: over-annotated **for lack of specificity** — no membrane named, `ECO:0000305` inference — **not** for lack of membrane association. The quote now extends through the topology line and the TRANSMEM feature.

**#2 — this is the better catch, and it lands on the PR's own thesis.** I argued that transfer from ABHD14B is this gene's chief hazard, then accepted two cytoplasm annotations resting on an `ECO:0000250` by-similarity call, justifying the IEA by agreement with *"the family's general behaviour"* — i.e. with the soluble paralog. That is precisely the reasoning the PR was written to warn against, applied to hypothetical future annotations but not to the ones in front of me.

Both `ACCEPT`s stand, but now engage the contrary features. Type II topology with glycosylation C-terminal to the anchor places the hydrolase domain **lumenally**; ABHD14B is soluble. So the two differ in *where they sit*, not only in sequence, and family behaviour is not a reason to accept a cytoplasmic call for this one. The family-behaviour justification is gone. The IBA is retained as evidence independent of the UniProt call, and because a type II protein does present a short cytosolic N-terminal tail — but the review now says plainly that **neither annotation establishes which side the catalytic domain faces**, which is the question that matters here, and which would reconcile with the excluded preprint's Golgi result. Added as a `suggested_question`.

**Suggestions taken:** `ABHD14A-notes.md` added (my omission — required by CLAUDE.md and present for every sibling in this series); the database-conflation commentary moved out of `description`, with the membrane topology put there instead, which is better biology for that field as you say; `PMID:30121252` now has a `reference_review` accounting for it.

Thanks for flagging the preprint catch as worth propagating — it is now in the campaign notes as a standing check on the affinage `## Citations` list.

Validation clean; 9 quotes verbatim.

PR #2234, opened 2026-07-25
Review human ABRACL: a BioID neighbourhood became a subcellular location

cmungall:
## ABRACL (Q9P1F3) — Costars family protein

Three annotations, none experimental, and **no molecular function at all** — on a protein
with a solved NMR structure (PDB 2L2O), a named interaction partner, and two in vitro assays
on purified protein. The review adds the missing MF, removes an unsupported location, and
declines to sharpen the one term that is already correctly vague.

### The main finding: a proximity-labelling neighbourhood became a subcellular location

`GO:0005929 cilium` — **REMOVE**. The chain has four links, and nothing at the end of it was
ever a localisation observation:

1. GOA maps `UniProtKB-SubCell:SL-0066` mechanically under GO_REF:0000044 — working as designed.
2. The UniProt line it maps is flagged **ECO:0000305** (curator inference, not observation) and
   cites PubMed:37759737.
3. **PMID:37759737 reports no ciliary localisation.** It is an expression study of *mouse and cat*
   embryonic telencephalon. Its one sentence on the subject cites somebody else's work
   ("a high-throughput proteomics study in non-neuronal cell lines has shown that Abracl is
   associated with the primary cilia"), and in the *same paragraph* it reports data pointing the
   other way: "our results show that neither Abracl mRNA nor Abracl were expressed in the VZ" —
   the ventricular zone being where the ciliated progenitors are.
4. That proteomics study is PMID:26638075, the Gupta et al. centrosome–cilium BioID interactome.
   It is the **only** source of physical-interaction data for ABRACL in IntAct: five proximity
   hits, one publication, one detection method, baits SASS6, CNTRL, DCTN1, RPGRIP1L — three of
   which are centriolar/dynactin rather than ciliary.

`propagation_review` records this as `SOURCE_WEAK_OR_INFERRED` with the three source entities
classified individually, so the fault is attributed to the right link rather than to GOA.

### Bioinformatics: `ABRACL-bioinformatics/analyze_cilium_evidence.py`

Queries IntAct and UniProt live; nothing hard-coded. Two parts, both designed to be able to
falsify the removal:

- **Part 1** enumerates ABRACL's entire interaction record and its detection methods.
- **Part 2** counts PANTHER family members per proteome, with **cilium-core positive controls**
  (IFT88, IFT52, BBS1, ARL13B) and a **universal negative control** (tubulin beta):

| | aciliate proteomes (*Arabidopsis*, rice, maize, *Dictyostelium*) |
|---|---|
| Costars / ABRACL | **7** |
| IFT88 / IFT52 / BBS1 / ARL13B | **0 / 0 / 0 / 0** |
| tubulin beta | **147** |

Had the family been absent from angiosperms, or had IntAct returned affinity-capture or Y2H
evidence alongside the BioID hits, the annotation would have been left standing. The analysis
also records the one observation cutting against a tidy story — Costars is not cross-referenced
in *Chlamydomonas* — and explicitly declines to lean on it, since a cross-reference count is not
a homology search. It states plainly that a lineage-specific human ciliary role is not excluded.

### The vague term is the right term

`GO:0032970 regulation of actin filament-based process` (IBA) — **ACCEPT**, and deliberately
*not* refined. Three datasets disagree about the **sign** of ABRACL's effect on the F-/G-actin
balance:

- deleting the *Dictyostelium* ortholog **raises** cytoskeleton-associated actin;
- depleting human ABRACL **lowers** the cellular F-/G-actin ratio;
- purified human ABRACL alone **inhibits** polymerisation — which the authors themselves flag:
  "These in vitro results appeared to contradict the above-mentioned finding of decreased
  F/G-actin ratios in cells lacking ABRACL".

So `GO:0030835` and `GO:0030837` are each supported by one part of the evidence and contradicted
by another. The unsigned parent is the strongest claim every dataset supports. The transfer
itself is unusually solid: PTHR46334 is single-copy (no paralog hazard), and the human protein
*functionally complements* the cosA-null Dictyostelium mutant — `NO_FAILURE_CORE`.

### Two IEAs from the same GO_REF, opposite verdicts

`GO:0030027 lamellipodium` — **ACCEPT**. Same reference, same mechanism as the cilium row, but
the UniProt statement behind it is **ECO:0000269**, and the experiment is a good one: endogenous
protein, specific antibody, colocalised with F-actin at the leading edge. It deserves an
evidence-code upgrade to **IDA** against PMID:33670794, not a term change.

### New annotations

- **`GO:0051015 actin filament binding`** (IDA) — the gene's first MF. Untagged protein
  (Factor Xa-cleaved), co-sedimentation quantified against a no-actin control. Offered with its
  limitation stated: the authors call the binding "very weakly".
- **`GO:0030335 positive regulation of cell migration`** (IMP) — bidirectional, multi-perturbation
  (overexpression, two shRNAs, RNAi-resistant rescue, five CRISPR KO clones), same sign in
  Dictyostelium. Recorded with the qualification that in MDA-MB-231 only EGF-stimulated migration
  was affected.

### GO gap

GO has **`GO:0000513 actin severing activator activity`** ("Binds to and increases the activity of
a actin severing protein") and **no inhibitor counterpart**. ABRACL binds cofilin, blunts
cofilin-stimulated depolymerisation dose-dependently, and does *not* displace cofilin from
filaments — it acts on the severing protein, not on actin. Proposed: **`actin severing inhibitor
activity`**, the existing definition with the sign reversed, under GO:0140678. Its absence is a
large part of why this gene has no MF.

### Verification

- `checkquotes.py`: **38 quotes, 0 problems**
- `just validate human ABRACL`: **✓ Valid** (no warnings)
- `cache/go/terms.csv`: unmodified


cmungall:
You're right, and the contradiction was with this PR's **own fetched data** — which makes it the worst kind.

Five places called PTHR46334 a single-copy family while `RESULTS.md` showed *Arabidopsis* = 2 and *Zea mays* = 3, and the PANTHER metadata recorded `subfamilies: 2`. Resolving the entries file:

| Subfamily | Members | Composition |
|---|---|---|
| `PTHR46334:SF1` | 13 | animals + *Dictyostelium* + most plants — includes **human ABRACL (Q9P1F3)** and **Dicty cosA (Q558Y7)** |
| `PTHR46334:SF3` | 2 | plant-restricted: *Arabidopsis* Q8LBN7, *Eutrema* B4YYA9 |

**And as you say, this strengthens the `GO:0032970` ACCEPT rather than weakening it.** The IBA donor and the human target are *both SF1*, so the transfer does not cross the one boundary the family actually contains. The `propagation_review` now names that boundary instead of denying one exists — which is a better argument than the one it replaces, because "no subfamily boundary is crossed" is checkable whereas "no paralog exists" was simply false.

**The root cause is worth recording**, and I've put it in the notes: the copy-number claim was written *alongside* the analysis rather than read *out of* it — and `analyze_cilium_evidence.py` printed it as hardcoded prose regardless of the counts it had just fetched. Fixed there too. A script must not assert what its own data contradicts; that is the same failure I corrected on AAMDC (#2221), where the script claimed to query InterPro without doing so.

**`GO:0000513` verified** — you flagged that it is absent from the local ontology cache so you couldn't check it. QuickGO: `actin severing activator activity`, molecular_function, not obsolete, *"Binds to and increases the activity of a actin severing protein."* So the asymmetry the proposed inhibitor term rests on is genuine.

**RPGRIP1L conceded.** You're right that it is genuinely ciliary — a transition-zone protein — so the bait panel is *not* uniformly non-ciliary, and the review now says so explicitly instead of listing only the three that suit the argument. It doesn't rescue the annotation (proximity to one ciliary bait over hours of BioID labelling is a neighbourhood, not a localisation), but the argument is stronger for stating the inconvenient part.

**On the `MISCITED` enum** — added a note. The enum describes how a citation is *used* rather than a property of the paper, so applying it to record a defect in *GOA's* use of a reference risks being read as a defect in this review's own citation. Flagged for anyone aggregating those values.

Validation clean; 38 quotes verbatim; no `cache/go/terms.csv` deletions.

One process note: this review was researched and written by a subagent whose session was ended by an API outage. I've verified its work independently and made these corrections myself — the underlying analysis and the cilium removal are the agent's, the errors above were in its summary layer.

cmungall:
Three of five, as you say. Both misses were in the same file, and one of them contradicted its own annotation 45 lines above it — worse than a stale sentence.

**Item 1 — the description.** This is the one worth dwelling on, because the diff actively misled me: reflowing the block from `>-` to a plain scalar made it *show as changed* while the claim inside was untouched. I checked that the block was modified rather than checking that the sentence was. A changed line is not a changed claim.

**Item 2 — the `PANTHER:PTN000507089` comment.** Rewritten to say what actually warrants the transfer, rather than deleting the wrong reason:

> PANTHER metadata records `subfamilies: 2` for PTHR46334, so this is not the single-copy, paralog-free node an earlier draft claimed. The transfer nonetheless stays inside one subfamily: the *Dictyostelium* cosA donor (Q558Y7) and human ABRACL (Q9P1F3) are both `PTHR46334:SF1` (13 members), while the separate SF3 is plant-restricted (2 members, *Arabidopsis* Q8LBN7 and *Eutrema* B4YYA9). No subfamily boundary is crossed, so the node supports the transfer — **but for that reason, not for the absence of paralogs.**

The conclusion was right for a reason I hadn't established. That's the part I'd have missed by only editing the sentence that was wrong.

**Item 3 — both garbled sentences repaired.** `three of three of the four baits` → `three of the four baits`; and the dangling `is a neighbourhood observation, and the ciliary ones.` now reads `is still a neighbourhood observation, not a localisation call.` The fragment was leftover from the phrase it was spliced into.

**Item 4 — the script regenerates RESULTS.md identically now.** Dropped the stray `"ortholog, "`, then verified by diffing a fresh run against the committed file rather than by reading the changed line. You were right that this was the real problem with it: a hand-edited RESULTS.md whose generator no longer reproduces it is worse than one that's simply wrong, because the next run silently reintroduces the error.

**One thing you didn't ask about, found while checking.** This branch had deleted a row from `cache/go/terms.csv`: `GO:0072534,perineuronal net`, which reached main after the branch point. Restored from `origin/main`; the branch adds no terms of its own, so nothing needed re-appending. It's invisible in the three-dot diff — `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` is what catches it. (That term came from the ACAN review merged an hour ago, so this was a live clobber, not a hypothetical one.)

**On the recurring failure itself.** Twice now I've reported a fix that only partly landed, so I changed method rather than promising more care: items 1–3 were applied by a script that **asserts each anchor string is present before replacing** and re-greps afterwards, so a missed occurrence is an error instead of a silent no-op. That immediately paid for itself — my string replacement on item 2 left a dangling `for the term to leak across.` continuation line, i.e. I committed the *same class of breakage* I was fixing in item 3. The re-grep caught it and I redid that edit through the YAML parser instead of by string surgery.

Validation passes. 38 quotes verbatim; 0 deletions from `cache/go/terms.csv`.

PR #2241, opened 2026-07-26
docs: automation-migration-to-dismech guide

cmungall:
Adds `docs/automation-migration-to-dismech.md` — an execution guide for bringing this repo's GitHub Actions automation up to the `monarch-initiative/dismech` pattern. The two repos share a ~2026-07-03 common ancestor, but dismech has since had a security-hardening + centralization pass this repo hasn't received.

## What the guide covers (grounded in an actual scan of this repo)
Current gap found here:
- `PAT_FOR_PR` in **8 workflows** (`ai`, `pr-shepherd`, `dragon-ai`, `curation-scanner`, `go-annotation-scanner`, `litscan-module-member`, `warm-reference-cache`, `generate-pages`)
- `persist-credentials: false` only in `main.yaml` — every agent checkout persists its token to `.git/config` (the exact exposure vector from the incident)
- `claude-code-review` uses the **same** `AI4C_AGENT` app as writers (reviewer not split)
- stale, hardcoded models (`claude-opus-4-7` ×12, `sonnet-4-6` ×8, `haiku` ×8); no `agent-config.yaml` / `cron-profiles.yaml`; no fork-PR gate

Phased plan (security first): App-token migration + `persist-credentials: false`; reviewer split to `ai4c-reviewer`; fork-close + untrusted-comment guards; centralize model/cron config; bump to Opus 5 / Sonnet 5; injection-safe step-summaries; delete `PAT_FOR_PR`. Includes a per-workflow reference table, verification commands, and PR sequencing.

## References
- The security/remediation *why* + copy-paste token pattern: [ai-security-private → docs/history/2026-07-25-dismech-pat-to-app-migration.md](https://github.com/ai4curation/ai-security-private/blob/main/docs/history/2026-07-25-dismech-pat-to-app-migration.md)
- dismech PRs cited throughout: #6890, #6895, #6918, #6933, #6934, #6958, #6979, #6998

Docs-only; surfaced in the mkdocs nav.

### DIFF
--- a/docs/automation-migration-to-dismech.md
+++ b/docs/automation-migration-to-dismech.md
@@ -72,6 +72,31 @@ since — which it will have been, if you rebase between review rounds. #2247 wa
 lost that way. The order that works is merge → retarget the next PR to `main`
 → *then* delete the branch.
 
+## Why this order
+
+Asked six times in review, so it belongs in the record. The sequencing was not
+arbitrary:
+
+1. **Reviewer split before anything else.** Every subsequent PR was reviewed by
+   an identity distinct from the one writing it. Landing it first is what made
+   the rest of the stack worth reviewing at all — and the reviewer went on to
+   find a 🔴 and several 🟡s that the author's own tests did not.
+2. **Injection controls before the scanners.** `close-fork-prs` and the
+   untrusted-comment guard reduce what can reach an agent's context; they should
+   be in place before more agents get write-capable tokens, not after.
+3. **Auth per workflow group, smallest blast radius first.** `pr-shepherd` alone,
+   then the four scanners, then the mention responder, then the page/cache jobs.
+   Each group is independently revertible, and a mistake in one does not strand
+   the others.
+4. **Teardown last, and it means more than the secret.** The `PAT_FOR_PR` secret
+   is the obvious item, but the `just` recipes that *re-install* it are the ones
+   that matter: while those existed, a single `just gh-add-secrets` could undo
+   the whole migration. Remove the ability to recreate the credential before
+   congratulating yourself on removing the credential.
+
+Cron profiles came last because they are cadence, not security, and applying
+them is behaviour-preserving.
+
 ## The token pattern
 
 Every write-capable workflow now looks like this:
--- a/docs/automation-migration-to-dismech.md
+++ b/docs/automation-migration-to-dismech.md
@@ -42,7 +42,9 @@ nothing in `.git/config`.
 Verification, all of which should stay empty/green:
 
 ```bash
-grep -rn "secrets.PAT_FOR_PR" .                # -> empty (repo-wide, not just .github/)
+# Repo-wide, not just .github/ — the justfile recipes that re-installed this
+# secret were the real footgun. --exclude-dir=docs because this file quotes it.
+grep -rn "secrets.PAT_FOR_PR" . --exclude-dir=docs --exclude-dir=.git   # -> empty
 
 # Model pins, action-manifest expressions, per-CHECKOUT credential persistence
 # (a per-file grep passes a file with two checkouts where only one is flagged,
@@ -176,7 +178,8 @@ Both matched `@dragon-ai-agent please`, so every qualifying event ran the agent
 workflow keyed on `@ai4c-agent please`, with the old keyword as an alias.
 
 **The credential was already dead, and nobody knew.** `pr-shepherd` failed
-**121 scheduled runs in a row over 9 days** (2026-07-16 to 2026-07-26) before
+**121 scheduled runs in a row over 9 days** — 2026-07-16T15:18Z until the
+migration ended it on 2026-07-26T01:09Z — before
 this migration — every one at `Checkout repository`, the step that used
 `token: ${{ secrets.PAT_FOR_PR }}`. The first run on the App token passed every
 step. The same applied to every other workflow holding the PAT. Nothing alerted,
@@ -214,7 +217,8 @@ CI's YAML parse passed the whole time: the file *is* valid YAML, and the
 rejection happens in the expression layer above it. Note the shape of the
 regression — before, that step succeeded while writing an empty report; after,
 the report was still missing *and* the job went red. `tests/test_agent_config.py`
-now checks every manifest-level field. Nothing in this repo parsed action
+now checks every manifest-level field, which is the boundary that matters: the
+same trap applies to `name:` and to any input/output `description`/`default`. Nothing in this repo parsed action
 manifests before that, which is the same blind spot that hides
 `.github/actions/claude-code-action`.
 
@@ -267,7 +271,9 @@ routine event in the repo into a red X.
   would have reinstalled the exposed credential and undone this migration.
   Worth remembering that a revoked token is not a revoked account, and a deleted
   account is not a removed collaborator.
-- **The `PAT_FOR_PR` secret still exists**, though nothing references it. It
+- **The `PAT_FOR_PR` secret still exists**, though nothing references it — the
+  only mentions left in the tree are two do-not-reintroduce comments
+  (`justfile:180`, `ai.yml:187`) and this document. It
   should be deleted. Note it is already non-functional — a checkout using it
   fails outright, which is how `pr-shepherd` came to fail 121 runs in a row — so
   deleting it is bookkeeping rather than a cutover.
--- a/docs/automation-migration-to-dismech.md
+++ b/docs/automation-migration-to-dismech.md
@@ -165,6 +165,19 @@ values, `strategy.matrix` entries, an `env:` indirection
 `tests/test_agent_config.py` checks all four shapes instead, and treats a stale
 allowlist entry as a failure so the list can only shrink.
 
+**A manifest is not just YAML.** `agent-run-summary` shipped broken and took
+the summary step down in seven workflows, because its `description:` contained
+an illustrative `${{ steps... }}` expression — the prose explaining the bug was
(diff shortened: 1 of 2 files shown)


cmungall:
Good review — all four IMPORTANT items were real, and rather than patch the guide's text I went and did the migration. Every one of them is now fixed in the implementation:

1. **`echo "${{ ... }}"` step summaries are a live injection sink, not a Phase 6 nicety.** Agreed. Every summary now passes the model output through `env:` and emits it with `printf '%s'`, and it landed in the Phase 1 commits, not at the end. (Also switched the fence to `~~~~` — agent reports routinely contain ``` blocks that closed the old fence early.) #2246, #2253, #2260
2. **The dragon-ai callout applies verbatim to `ai.yml`.** It applies more than verbatim: the two files were **byte-identical**, so every qualifying event ran the agent twice. Consolidated into one workflow in #2255, with the allowlist read from the default branch and the prompt passed via `env:` + `printf`.
3. **The model grep was a false-green.** Correct — it matched 3 of 28. Replaced with a test (`tests/test_agent_config.py`) that checks three shapes: a `--model` line not resolving from `env.AGENT_MODEL`, a `*model:` YAML key pinned to a literal (catches matrix entries and `ISSUE_HANDOFF_MODEL`-style indirection), and a dispatch `default:` pinned to a model. It found exactly the six workflows that were pinning; the allowlist is now empty.
4. **The copy-paste block pinned older actions than the repo runs.** Everything is on `create-github-app-token@bcd2ba4 # v3.2.0` (confirmed it still exposes `app-slug`), `claude-code-action@be7b93b # v1`, and `actions/checkout@v7`.

On the suggestions: `weekly-compliance` indeed wasn't in the 8 — it was worse, it had **no `github_token` at all**, the same false-green that cost dismech ~11 days of no-op runs (#2260). The `persist-credentials` grep can false-green on multi-checkout workflows, so I checked each site individually; every checkout in the repo now sets it, including `deploy-docs` (which pushes via `mkdocs gh-deploy`, so it gets `gh auth setup-git` instead).

Stack: #2246 → #2247 → #2249 → #2253 → #2255 → #2260 → #2261.

I'll rewrite this document as an as-built record once the stack merges, rather than leaving it as a plan for work that is done.

cmungall:
The 🔴 was real and it was mine. Fixed and merged as **#2278**, and verified live rather than just merged — `pr-shepherd` run 30194399351 has `Write step summary = success`.

Your diagnosis was exactly right, including the part I'd have argued with: `description:` is template-evaluated at manifest load, so the illustrative expression in the prose was parsed for real. **The comment explaining the bug was the bug.** And your point about the trade is the one that stung — before #2253 that step *succeeded* while writing the empty fence; after, the report was still missing *and* the job went red. Strictly worse than what it replaced.

`tests/test_agent_config.py` now checks every manifest-level field (`name`, `description`, input/output `description`/`default`) across `.github/actions/**/action.y*ml`, verified against a probe. Your framing of it as a second occupant of the `.github/actions/claude-code-action` blind spot is right and is now in the doc: nothing here parsed action manifests at all, and CI's YAML parse can't see this class of failure because the file *is* valid YAML.

**🟡 1 — half right, and the half you got wrong makes it less urgent, but the half you got right makes it worse.** The account no longer exists:

```
collaborators?affiliation=direct  ->  cmungall, caufieldjh, dragon-ai-agent
GET /users/dragon-ai-agent        ->  404 Not Found
```

So the collaborator entry is dangling and grants nothing. But the recipes are the real finding, and the collaborator ones aren't the dangerous pair — **`justfile:gh-add-secrets` and `ai.just` both re-install `PAT_FOR_PR`**, so one `just gh-add-secrets` would have reinstalled the exposed credential and undone the migration. All four recipes are gone in **#2279**, with a comment on the secrets recipe so it isn't helpfully restored. The issue summarizer was also still advertising `@dragon-ai-agent please` to every new issue; now `@ai4c-agent please`. Removing the stale entry itself is a repo-settings click I've left to @cmungall.

**🟡 2 — correct, and I'd understated it twice.** Fixed: five workflows, and `pr-shepherd` is called out specifically as not-a-scanner reading PR bodies, diffs and review comments. The precedent line now names `claude-code-review` and `claude` alongside `litscan-module-member`.

Also took: added `tests/test_agent_run_summary.py` to the verification block (it's what `main.yaml`'s agents gate actually runs), and softened the `-rL` line. On the run-count: you're right that "24 in a row" undercounts — I sampled the last 25 runs and 24 of them failed, which I then wrote as a streak. The streak is longer.

cmungall:
🟡 resolved by landing **#2279** rather than by rewording — the past tense is now true. `justfile`, `ai.just` and the summarizer are all clean on main; the only `dragon-ai-agent` strings left are the deliberate mention-keyword alias and the comment explaining it.

**On the run count: you were right to push, and it's much worse than either of us said.** I sampled 300 completed runs and computed the longest failure streak properly:

```
121 consecutive failures over 9 days
2026-07-16T15:18  ->  2026-07-26T01:09
(fully contained in the sample — that's exact, not a floor)
```

My "24" came from sampling 25 runs and reporting the failures within it as a streak, which is exactly the kind of number I should not have written down without checking. Your "at least 38" was closer and correctly flagged as a lower bound. The doc now says 121/9 days, and adds why nobody noticed: a scheduled workflow's failures land in the Actions tab and nowhere else.

**Line 244 corrected** — you're right that it contradicted line 178 once the manifest and checkout tests started walking `.github/actions/**`. Narrowed to the half that still holds: the pinned-SHA, input-vocabulary and central-model guards can't see it, so its model and agent version are uncontrolled.

**`grep -rL` removed**, replaced with `pytest -k non_persisting`. It was weak when you first raised it and redundant now; keeping a check that a per-file grep can pass while two composite-action checkouts persist tokens would be worse than having no line there.

**Phase ordering: written, and thank you for asking six times.** It should have gone in on the first pass. The part I'd have gotten wrong without your justfile finding is the teardown clause — "delete the secret" was the wrong framing. While `just gh-add-secrets` could reinstall `PAT_FOR_PR`, deleting the secret was *reversible by accident*. Removing the ability to recreate a credential has to come before claiming you've removed it, and that's now what the section says.

On the two private ai-security-private links: I can confirm from this side that both resolve and that the history doc is the source the token pattern was copied from. Unverifiable from your sandbox is a real limitation of the review, not something the PR can fix.

cmungall:
Both correct, both introduced by my own last three commits, and the second one is the more embarrassing kind of wrong.

**🟡 1** — fixed. The previous commit corrected line 169 and left line 261, so the document carried the retracted figure in its closing summary and the corrected one four sections earlier. Both say 121 now.

**🟡 2 — the rationale I wrote one commit ago was false about its own history, and the document's own table said so.** I verified rather than argued:

```
#2259  merged 02:31:05Z
#2246  merged 04:35:24Z   <- auth PR
#2276  merged 04:52:52Z   <- reviewer split
gh pr view 2246 --json reviews  ->  ["ai4c-agent"]   (the authoring App)
```

So two PRs, including an auth one, merged before the split and were self-reviewed. I wrote the aspiration and presented it as the history.

The cause is worth recording rather than quietly rewording, and you're right that it's the more useful lesson: **the reviewer-split PR was stacked *on* the pr-shepherd PR**, so pr-shepherd necessarily merged first. Intending the split to be early isn't enough — it has to be the *base* of the stack. Otherwise you silently choose which PRs get self-reviewed, and by construction they're the foundational ones. That's now what the section says, with the merge times and the self-review fact in it.

Took the other three too: the PAT grep is repo-wide (free now, and it backs a repo-wide claim), the duplicated pytest line is gone, and the manifest lesson names manifest-level fields generally rather than `description:` — you're right that it should match the boundary its own test enforces.

On the private links: agreed that's a limitation of the review rather than a defect in the PR, and recording it as such each pass is the right call.

Seven passes on a docs PR is more than it should have taken, and most of that was the record chasing a moving target of my own making. The two things you caught here — a retracted number surviving in a second location, and a rationale that contradicted the table directly above it — are exactly the failures that survive self-review, which is a fitting way for this particular document to end.

cmungall:
That's a good last finding, and a funny one: the check standing behind the table's strongest security claim was matching the file that documents it.

```
docs/automation-migration-to-dismech.md:45   the command's own text
docs/automation-migration-to-dismech.md:181  `token: ${{ secrets.PAT_FOR_PR }}` in the narrative
```

Fixed with `--exclude-dir=docs --exclude-dir=.git`; verified the command now returns empty on a clean tree. You're right that it's the mirror image of everything else in this document — a false *red* rather than a false *green*, so nothing unsafe rode on it. But "a check that always shows output is a check people stop reading" is exactly the failure mode this whole migration is about, so it's worth the line.

Took the three nits: the streak now reads `2026-07-16T15:18Z until the migration ended it on 2026-07-26T01:09Z` rather than an open-looking interval; the manifest lesson no longer re-narrows to `description:` a sentence after generalizing (it now names `name:` and input/output `description`/`default` explicitly); and the PAT gap says what the remaining mentions are — two do-not-reintroduce comments and this file.

Eight passes, and the shape of them is the argument for the reviewer split better than anything I wrote in the doc. Pass 1 caught four real gaps in a plan. The middle passes caught a 🔴 that broke seven workflows, three logic defects in the mention gate, a trust gate that failed *open*, a `dry_run` checkbox that did nothing, and two token-persisting checkouts my own "repo-wide" grep couldn't see. The last three caught me contradicting myself — including, fittingly, a rationale claiming a review discipline that the merge times showed I hadn't actually followed.

None of that came from my own tests. Thanks for the persistence, particularly on the run count and the phase ordering, both of which I'd written down without checking.

PR #2245, opened 2026-07-26
Review human ABRA/STARS: 14 of 16 annotations trace to two rodent proteins

cmungall:
PAINT no-IBA project gene review for **ABRA / STARS** (Q8N0Z2), using the `affinage`
deep-research provider.

## Headline: the human record is almost entirely rodent

A reproducible provenance audit (`ABRA-bioinformatics/audit_abra_record.py` — everything fetched
at run time from UniProt, QuickGO and InterPro plus the gene's own GOA file, failed fetches
reported rather than counted as absences) resolves **every** WITH/FROM accession:

| WITH/FROM source | rows |
|---|---|
| mouse *Abra* (Q8BUZ1) + `ensembl:ENSMUSP00000051973` + `MGI:MGI:2444891` | **14** |
| rat *Abra* (Q8K4K7) + `RGD:708493` | 3 |
| `PANTHER:PTN001100454` / `InterPro:IPR026111` | 5 |
| PPP1R18 (human) | 1 |

Evidence-code breakdown of the 16 rows: **IBA 3, IEA 6, ISS 5, IPI 1, IDA 1** — so **two
experimental rows out of sixteen**, and neither is about what ABRA does (one Y2H hit, one LIFEdb
localisation).

**Every WITH/FROM protein accession is a genuine one-to-one ortholog**, so unlike AASDH (#2226)
or ABHD8 (#2230) this is *not* a paralog error — the transfers are legitimate in kind. The
finding is the **concentration**: a reader seeing sixteen annotations would not guess the human
evidence base is two rodent entries.

## Curation actions

11 ACCEPT · 3 MODIFY · 1 KEEP_AS_NON_CORE · 1 MARK_AS_OVER_ANNOTATED · 1 NEW

**`GO:0042307` positive regulation of protein import into nucleus (NEW)** — ABRA's defining
mechanism was missing from the human record entirely. It sequesters actin, lowering the G-actin
pool that retains MRTF/MKL in the cytoplasm, so MRTF enters the nucleus and activates SRF. That
is the whole point of the protein ("striated muscle activator of Rho signalling") and no human
annotation captured it.

Two `GO:0003779 actin binding` rows and the bare `GO:0005856 cytoskeleton` row `MODIFY` to more
specific terms. The lone plasma-membrane call is examined in the bioinformatics analysis and
found unsupported.

## Note on provenance of this PR

The review was researched and written by a subagent whose session was cut short by an API outage
before it could commit. I validated the output independently — 42 `supporting_text` quotes all
verbatim, `just validate human ABRA` clean, `cache/go/terms.csv` checked for deletions (18 found
and restored from `origin/main`) — and committed it. The analysis and conclusions are the agent's.


cmungall:
Fixed. You were right that Q3 was the one section that couldn't be re-run, and right that it mattered — it's quoted as `supporting_text` in two reviews and carries the ortholog-only claim.

**The PANTHER export is committed** (`PTHR22739-entries.csv` + metadata), which matches the repo convention rather than diverging from it: 2944 `-entries.csv` files are already tracked, and `just fetch-panther-family PTHR22739` is what produces them. A fresh run reproduces the committed `results.json` byte-for-byte, so the numbers you flagged (`n_reviewed_members: 4`, `{PTHR22739:SF20: 4}`) were correct — just unverifiable.

**The deeper defect was the silent degrade, and that's fixed too.** `return {"available": False}` on a missing input is what let `RESULTS.md` and `results.json` disagree with no signal at all. It's now a hard `SystemExit` naming the fetch command:

```
Q3. Composition of PANTHER PTHR22739
PANTHER entries CSV not found: /nonexistent.csv
Q3 cannot be computed without it. Fetch it with:
    just fetch-panther-family PTHR22739
```

Same principle as "a failed fetch is reported rather than counted as an absence" — the script claimed that property and then didn't have it for its one local input. And you were right that the provenance sentence was false; it now says Q1/Q2/Q4/Q5 are fetched at run time and Q3 reads the committed export.

**On the mouse `GO:0006606` gap — this turned out to be the most useful item on your list.** I extended the audit rather than just asserting the claim; the inline mouse-projection block in Q2 is factored out into `mouse_annotations(go_id)` and reused, so `GO:0042307` gets the same treatment as `GO:0005886`. New Q5:

| term | on mouse Q8BUZ1 | evidence | reference | WITH/FROM |
|---|---|---|---|---|
| GO:0006606 protein import into nucleus | yes | IDA | PMID:15798203 | *(none)* |
| GO:0042307 positive regulation of protein import into nucleus | **no** | — | — | — |

Both facts favour the proposal, and the second one I hadn't appreciated:

- The **empty WITH/FROM** means the mouse row is primary evidence from PMID:15798203 — the paper the review already cites — not a back-transfer from human. So unlike the `GO:0005886` case in Q2, this is a genuine anchor rather than a round trip.
- The mouse record uses the **unqualified process term**, which is arguably the wrong one. `GO:0006606` reads as the annotated protein being part of the import event, and **ABRA is not the cargo** — PMID:15798203 shows it promoting MRTF-A/B import. So the human proposal isn't the mouse annotation at a different granularity; it's the term the mouse row should have used, and the gap in the mouse record is exactly why no transfer supplied it to human.

**On `directly_involved_in: GO:0035025`** — agreed, and I've dropped it. `directly_involved_in` is the only BP slot in `CoreFunction`, so there's no weaker option to move it to, and the review says three separate times that ABRA may merely *require* RhoA rather than act on it (PMID:15798203 frames RhoA as a requirement for the ABRA effect) and that it has no GEF module. Listing it there asserted a directness my own text denies. The `existing_annotations` rows stay `ACCEPT` — that term's definition does permit indirect regulation — and the `core_function` description now states the distinction explicitly, so the two aren't in tension.

`notes.md` quote made verbatim (`"targeted deletion of Abra in CL57BL/6 mice led to impaired arteriogenesis"`). PR body accession corrected to Q8N0Z2.

One note on the remaining validation warning, which is a false positive here: it says no annotation references `ABRA-deep-research-affinage.md`. That's deliberate — the affinage record is the one credited with ABRACL's Costars phenotype, recorded as `MISCITED`. Citing it as `supporting_text` for anything would undercut that finding.

Validation passes. 43 YAML quotes and 21 notes quotes verified verbatim; 0 deletions from `cache/go/terms.csv`.

(Also noted your PATH limitation — `just`/`uv` unavailable in the runner. Everything above was verified locally.)

PR #2246, opened 2026-07-26
ci(pr-shepherd): ai4c-agent App token + central agent-config

cmungall:
Phase 1 + the Phase 4 groundwork from `docs/automation-migration-to-dismech.md`, applied to one workflow first so the pattern can be verified before the bulk migration.

## Security (the incident vector)

`pr-shepherd` moves off the exposed `PAT_FOR_PR` classic PAT onto a short-lived `ai4c-agent` GitHub App installation token:

- token minted per run, expires at job end — no long-lived secret to leak
- **`persist-credentials: false`** on checkout: the token is never written into `.git/config`, which is exactly how the PAT was read by a Bash-enabled agent and disclosed in a public trace
- `gh auth setup-git` serves the token to git from `GH_TOKEN` only at push time
- `create-github-app-token` and `claude-code-action` are SHA-pinned
- commits attribute to `ai4c-agent[bot]`
- the step summary passes model output through `env:` + `printf` instead of `echo "${{ ... }}"` (attacker-influenceable → command injection)
- the prompt gains an explicit untrusted-input / prompt-injection guardrail

`id-token: write` is dropped (OIDC was only there to mint a token; `github_token` is now explicit).

## Central model config

- **`.github/agent-config.yaml`** — single source of truth for which model backs each agentic workflow
- **`.github/actions/resolve-agent-config`** — resolves it at run time, exports `AGENT_MODEL`
- the `model:` dispatch input becomes a no-override **string** (`default: ""`). A `type: choice` input always sends its default, so every manual run silently overrode the central config.
- model bumped to `claude-opus-5` (was `claude-opus-4-7`)
- **`tests/test_agent_config.py`** keeps it honest: every config key maps to a real workflow that actually uses the resolver, and no workflow re-hardcodes `--model claude-...`. Workflows not yet migrated are named in a shrinking allowlist, so the test both documents the remaining work and catches new offenders.

## Behaviour change worth knowing

Unlike the built-in `GITHUB_TOKEN`, an **App-token push fires `pull_request: synchronize`** — review auto-starts after the shepherd pushes a fix. The prompt no longer tells it to manually re-dispatch review.

## Verification

```
uv run pytest tests/test_agent_config.py   # 10 passed
uv run mypy tests/test_agent_config.py     # clean
uv run ruff check .                        # clean
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/pr-shepherd.yml'))"
```

Follow-ups (separate PRs): reviewer split to `ai4c-reviewer`, fork-PR close gate + untrusted-comment guard, scanner auth migration, retiring the duplicated `dragon-ai`/`ai` mention workflows, page/cache/compliance jobs, cron profiles, then deleting the `PAT_FOR_PR` secret.

### DIFF
--- a/.github/actions/resolve-agent-config/action.yml
+++ b/.github/actions/resolve-agent-config/action.yml
@@ -36,9 +36,11 @@ runs:
       run: |
         set -euo pipefail
         # PyYAML is present on the standard ubuntu-latest image; install it
-        # quietly if a leaner runner lacks it.
+        # quietly if a leaner runner lacks it. --break-system-packages is
+        # required on PEP 668 (externally-managed) interpreters and is a no-op
+        # elsewhere.
         python3 -c "import yaml" 2>/dev/null \
-          || pip install --quiet --disable-pip-version-check pyyaml
+          || pip install --quiet --disable-pip-version-check --break-system-packages pyyaml
         model="$(python3 "$GITHUB_ACTION_PATH/resolve_agent_config.py" \
           --config "$GITHUB_WORKSPACE/.github/agent-config.yaml" \
           --workflow "$WF" \
--- a/.github/actions/resolve-agent-config/action.yml
+++ b/.github/actions/resolve-agent-config/action.yml
@@ -1,0 +1,48 @@
+name: Resolve agent config
+description: >
+  Resolve the effective Claude model for an agentic workflow from
+  .github/agent-config.yaml and export it as the AGENT_MODEL environment
+  variable (and a `model` step output). Single source of truth for agent model
+  selection. Requires the repository to be checked out first.
+
+inputs:
+  workflow:
+    description: >
+      Workflow key under `workflows:` in .github/agent-config.yaml (the
+      workflow file stem, e.g. "pr-shepherd").
+    required: true
+  model-override:
+    description: >
+      Explicit model that wins if non-empty, e.g. a workflow_dispatch `model:`
+      input. Leave empty to use the configured model.
+    required: false
+    default: ""
+
+outputs:
+  model:
+    description: The resolved Claude model id.
+    value: ${{ steps.resolve.outputs.model }}
+
+runs:
+  using: composite
+  steps:
+    - id: resolve
+      shell: bash
+      # Pass inputs via env (not ${{ }} interpolation) to avoid any shell
+      # injection surface.
+      env:
+        WF: ${{ inputs.workflow }}
+        OVERRIDE: ${{ inputs.model-override }}
+      run: |
+        set -euo pipefail
+        # PyYAML is present on the standard ubuntu-latest image; install it
+        # quietly if a leaner runner lacks it.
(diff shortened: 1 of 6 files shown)


cmungall:
Thanks — good review, all ten points addressed in bfff121.

**🟡 IMPORTANT**

1. **Token expiry vs `timeout-minutes: 90`** — right, installation tokens don't refresh. Now `timeout-minutes: 55`, with a comment saying why.
2. **Guard-test blind spots** — correct, and worse than "2 vs 6": the old check needed `--model` *and* a literal `claude-<family>-` on the same line, so it missed the `--model ${{ inputs.model || '...' }}` shape entirely. Rewritten as three rules — a `--model` line not resolving from `env.AGENT_MODEL`, a `*model:` YAML key pinned to a literal (catches matrix entries and `ISSUE_HANDOFF_MODEL` indirection), and a dispatch `default:` pinned to a model — while still ignoring `choice` option lists. It now reports exactly: `claude`, `claude-code-review`, `curation-scanner`, `go-annotation-scanner`, `litscan-module-member`, `weekly-compliance`. A *stale* allowlist entry is now also a failure, so the list can only shrink. (`arba-issue-monitor` turned out not to pin a model at all — it takes the action's default.)
3. **Test never runs on a workflow-only change** — correct. New `agents` paths-filter (`.github/workflows/**`, `.github/actions/**`, `.github/agent-config.yaml`) runs `tests/test_agent_config.py` independently of `src`.
4. **`$GITHUB_ENV` / CLI-arg injection via the override** — fixed at the choke point: `MODEL_ID_RE = ^[A-Za-z0-9._-]+$`, applied to the override, the per-workflow `model:`, matrix models, and `default_model`. Verified both vectors are rejected:
   ```
   rejected: 'claude-opus-5\nEVIL=1'
   rejected: 'claude-opus-5 --dangerously-skip-permissions'
   ```
5. **Prompt asserts `ai4c-reviewer` exists** — reworded identity-agnostically ("Reviews are posted by the `claude-code-review` workflow"). The split lands in #2247, stacked directly on this.

**🔵 SUGGESTIONS**

6. Both token steps now pin `create-github-app-token@bcd2ba4 # v3.2.0`; confirmed v3.2.0 still exposes `app-slug`.
7. `--break-system-packages` added to the PyYAML fallback.
8. Step-summary fence switched to `~~~~`.
9. Default `GITHUB_TOKEN` dropped to `contents: read`.
10. `model:` + `matrix:` together is now a `ConfigError` rather than a silent preference.

On your closing point — agreed that `gh auth setup-git` is the load-bearing replacement and fails only at push time. I'll dispatch a live `dry_run: false` run and confirm a real push before the `PAT_FOR_PR` secret is deleted; that deletion is the last step of the migration, not this PR.

One correction: you couldn't run the suite, so for the record — `pytest tests/test_agent_config.py` 10 passed, `ruff check` clean, `mypy` clean, all touched workflows parse.

PR #2247, opened 2026-07-26
ci(review): run code review as the ai4c-reviewer App

cmungall:
Phase 2 of `docs/automation-migration-to-dismech.md`. Stacked on #2246 (base will auto-retarget to `main` when that merges).

`claude-code-review` was minting an **`ai4c-agent`** token — the same identity the writer workflows use. The reviewer was reviewing its own work, and its approval would not count under a "require approval from someone other than the most recent pusher" rule.

## Changes

- mint `AI4C_REVIEWER_APP_ID` / `AI4C_REVIEWER_PRIVATE_KEY` instead, with `continue-on-error` + `|| github.token` fallback so an unavailable reviewer App degrades to a review rather than a failed job
- **`allowed_bots: 'claude,github-actions,ai4c-agent'`** — claude-code-action rejects a run whose *initiating actor* is an unlisted bot, even for `workflow_dispatch`. The writer both dispatches review and (now, with an App token) pushes to PRs, so it must be listed or review runs are refused.
- model comes from `.github/agent-config.yaml` (`--model opus` → `AGENT_MODEL`, currently `claude-opus-5`)
- SHA-pin `create-github-app-token` and `claude-code-action`; `persist-credentials: false` on checkout
- **new "verify the review agent actually ran" step.** claude-code-action exits 0 even when the agent errored, so a Claude usage limit or a retired `--model` silently turns a broken review into a *phantom green check*. This repo has hit the usage-limit case before. Now it fails loudly, and the quota case is called out distinctly so it is not mistaken for a problem with the PR.

## One thing you may want to change in the App settings

The `ai4c-reviewer` installation currently has **Contents: read**. GitHub ties "this review counts toward branch protection" to *write* access, which for an App is governed by the **Contents** permission — so its approvals render as *"approved with read-only permissions."* `main` has no branch protection today, so nothing is blocked. If required-approval rules are ever turned on, the App needs Contents: write first (a browser change in the App settings + accepting the permission bump on the org installation).

## Verification

```
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/claude-code-review.yml'))"
uv run pytest tests/test_agent_config.py   # 10 passed
```

This PR's own `claude-review` check exercises the change end to end.

### DIFF
--- a/.github/agent-config.yaml
+++ b/.github/agent-config.yaml
@@ -26,5 +26,8 @@ default_model: claude-opus-5
 provider: anthropic
 
 workflows:
+  claude-code-review:
+    model: claude-opus-5
+
   pr-shepherd:
     model: claude-opus-5
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -23,7 +23,8 @@ jobs:
         (
           github.event_name == 'pull_request' &&
           github.event.pull_request.head.repo.full_name == github.repository &&
-          github.event.pull_request.head.ref != 'auto/generate-pages'
+          github.event.pull_request.head.ref != 'auto/generate-pages' &&
+          github.actor != 'dependabot[bot]'
         ) ||
         github.event_name == 'workflow_dispatch' ||
         (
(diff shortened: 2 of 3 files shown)


cmungall:

- The `set -euo pipefail` around two bare `gh api` substitutions was exactly wrong for a step whose headline case is a 403. Every failure path now degrades to a warning.
- `--paginate` applying `--jq` per page: right, `| length` emits one number per page. Projecting ids and counting lines now.
- The no-result path now names the quota case, which was the step's whole justification and was unreachable there.
- Both guard-test docstrings corrected — you were right that the phantom-green framing is stale at the pinned SHA.

cmungall:
Closed by GitHub, not by me: deleting the base branch when #2246 merged auto-closed this PR, and it can no longer be reopened because the head branch was force-pushed (rebased onto the new main) in the meantime.

Continued verbatim in **#2276** — same branch, same commits, same content. Your review here, including the APPROVED, is the record for that work.

My mistake was using `--delete-branch` while a downstream PR still pointed at the branch. The rest of the stack now merges without it: merge, retarget the next PR to `main`, *then* delete.

PR #2248, opened 2026-07-26
Review human ACAN: aggrecan is the protease substrate, not the protease

cmungall:
Reviews all 40 GOA annotations on human ACAN (aggrecan, P16112) and proposes one new one.

## The headline finding: aggrecan is annotated as an agent of the process that destroys it

`involved_in GO:0006508 proteolysis` (NAS, PMID:1569188) has the roles reversed. Sandy et al. 1992 is the paper that *defined* the aggrecanase cleavage site — it shows aggrecan being cut at Glu373-Ala374 in the interglobular domain by a cartilage proteinase. Aggrecan is the substrate. It has no protease domain and no catalytic residues. The agents are ADAMTS4, ADAMTS5, HTRA1 and the MMPs, and Reactome already models one of them correctly (`R-HSA-8855825`, "HTRA1 hydrolyzes ACAN") — which is also, incidentally, one of the six Reactome rows that only exports a generic compartment to GOA.

Annotating a substrate with `involved_in` for its own degradation makes aggrecan indistinguishable from the proteases in any process-based query. **REMOVE.**

If the intent was to capture aggrecan turnover — which really is central to osteoarthritis — that belongs on the proteases, or on aggrecan via an annotation extension naming it as substrate. GO has no substrate role to assign, which is raised in `suggested_questions`.

## Second REMOVE: a Link-domain family mapping that shouldn't reach a secreted protein

`involved_in GO:0007155 cell adhesion` (IEA, InterPro:IPR000538). The Link domain entry carries two GO mappings. Hyaluronic acid binding is the domain's defining biochemistry and applies to every member. Cell adhesion is licensed by the *cell-surface* members — CD44, the main cell-surface hyaluronan receptor, and TSG-6 — which InterPro's own description frames in terms of cell-cell and cell-matrix interactions.

Aggrecan is none of those. It is secreted, has no transmembrane segment or GPI anchor and no integrin-binding motif, and `GO:0007155` requires attachment of a cell "via cell adhesion molecules". Aggrecan-rich matrices are if anything the opposite of adhesive: cartilage matrix and the perineuronal net are barriers that restrict cell and process migration.

## MODIFY (5)

| row | from | to | why |
|---|---|---|---|
| IBA | `GO:0045202` synapse | `GO:0072534` perineuronal net | wrong compartment (see below) |
| IBA | `GO:0001501` skeletal system development | `GO:0051216` cartilage development | bone phenotypes are downstream of cartilage matrix failure |
| NAS | `GO:0001501` skeletal system development | `GO:0051216` cartilage development | same, plus the cited paper is about OA biochemistry and says nothing about development |
| TAS | `GO:0005201` ECM structural constituent | `GO:0030021` ...conferring compression resistance | the specific child, already annotated twice |
| IPI | `GO:0005515` protein binding (COMP) | `GO:0050840` extracellular matrix binding | real partner, uninformative term |

**On the synapse row** — GO deliberately places the perineuronal net *outside* the synapse. `GO:0072534`'s ancestors run perisynaptic extracellular matrix (`GO:0098966`) → synapse-associated extracellular matrix (`GO:0099535`) → specialized extracellular matrix (`GO:0140047`) and extracellular matrix; `GO:0045202` is not among them (verified via the QuickGO ancestors endpoint over `is_a`/`part_of`). So `is_active_in synapse` asserts residence in the synapse proper, which a secreted matrix proteoglycan does not have. The same IBA batch already carries the precise term. The imprecision traces to the donor mix — brevican and neurocan dominate that PANTHER node and mouse Bcan carries a glutamatergic synapse annotation.

**On the COMP row** — this is the one `GO:0005515` worth keeping, just not as `protein binding`. Direct solid-phase binding assay, dependent on COMP's calcium-replete conformation, partly inhibited by soluble GAGs, mapped to COMP's signature domain, with the pseudoachondroplasia mutant binding less well. Reinforced by the G3 C-type lectin literature: human missense variants in that repeat measurably reduce binding to tenascin-R, fibulin-1 and fibulin-2 (PMID:20137779) and to cartilage matrix ligands generally (PMID:35338222). All of aggrecan's demonstrated partners through this route are matrix components, so `GO:0050840` is the informative term.

## MARK_AS_OVER_ANNOTATED (1)

The other `GO:0005515` row points at APP (PMID:32814053) — one candidate hit from a systematic yeast two-hybrid screen of ~500 neurodegeneration-related proteins. Y2H cannot reproduce the glycosylation or compartment of a secreted proteoglycan, there is no orthogonal validation, and APP appears nowhere in aggrecan's own literature. Not removed, since IntAct records three experiments.

## NEW (1): ACAN has no correct biological process at all

Once proteolysis and cell adhesion are removed, GOA's only remaining BP terms for ACAN are two broad development terms. Nothing in the record states that aggrecan **assembles the matrix it is a constituent of** — which is the entire point of the protein, and is in its name.

Proposed `involved_in GO:0085029 extracellular matrix assembly`, supported by the 2025 human G1–hyaluronan cocrystal structure and by the demonstration that recombinant human G1 and G1-G2 fragments incorporate specifically into perineuronal nets, with incorporation reduced by point mutation of the GAG-binding site (PMID:40273987). Whether a self-assembling structural constituent should sit on `GO:0085029`, or whether that term is reserved for remodelling enzymes and receptors, is a modelling question I raise rather than settle — but leaving the gene with no matrix-assembly process is clearly the wrong answer.

## WITH/FROM resolution

Every accession resolved. The two IPI partners are `UniProtKB:P49747` (COMP, real) and `UniProtKB:P05067` (APP, screen noise). All ten IBA donors resolved via the Alliance API — mouse Acan/Bcan/Ncan/Hapln1/Hapln4, rat Acan/Bcan/Vcan/Ncan, zebrafish vcanb. **Each of the four IBA rows includes a true ACAN ortholog** (mouse `MGI:MGI:99602` or rat `RGD:68358`), so none is a paralog transfer. The residual problems are term scoping, not donor choice, and the `propagation_review` blocks record that distinction.

## What the other 26 rows actually are

Reactome reaction-level export. 14 identical `located_in Golgi lumen` rows and 6 identical `located_in extracellular region` rows, one per keratan-sulfate biosynthesis reaction. Kept as non-core (aggrecan does transit the Golgi to be glycanated, and it is secreted), but worth flagging that **five of the Golgi rows come from *defective*-enzyme disease reactions** — `R-HSA-3656230`, `R-HSA-3656258`, `R-HSA-3656269`, `R-HSA-9035949`, `R-HSA-9035950` — where the modelled biology is a congenital disorder of glycosylation and aggrecan appears only as the acceptor whose glycanation fails. Deriving a normal-physiology compartment annotation from a reaction defined by the *absence* of that reaction is an export artifact.

Also non-core: `GO:0140150` gel phase of interstitial matrix (×2). The physical claim is exactly right — the aggregates do form a densely packed hydrated gel — but the term sits under interstitial matrix (`GO:0005614`), while aggrecan's two principal matrices are *specialized* matrices. It is strictly accurate only for the minor vascular pool.

## Two recommendations that aren't actions

1. **`GO:0005540` hyaluronic acid binding is still only an InterPro IEA**, despite a 2025 human cocrystal structure of the G1 region with a hyaluronan decasaccharide and a directly measured 234 nM affinity by biolayer interferometry (PMID:40273987, residue numbering explicitly against P16112). Straightforward upgrade to a direct experimental code.
2. **The `colocalizes_with` qualifier on one of the two `GO:0031012` HDA rows should be `located_in`**, matching its sister row on equivalent evidence and UniProt's own ECM subcellular location.

## Proposed new term

A **`cartilage extracellular matrix`** cellular component under `GO:0140047` specialized extracellular matrix. GO has 30+ cartilage *process* terms (`GO:0051216`, `GO:0060536`, `GO:0001502`, `GO:0061975`, `GO:0003417`, `GO:1990079`, …) but no component for the matrix they all build. `GO:0140047` already carries siblings of exactly the right kind — perineuronal net, interphotoreceptor matrix, hyaline layer, cuticular ECM, egg coat — so cartilage matrix's absence looks like an oversight. The gap bites hardest on the genes it should serve best: ACAN, COL2A1, COL9A1, COL11A1, COMP, MATN1/3, HAPLN1, CHAD, PRG4 and the ADAMTS aggrecanases all currently have to sit in the generic parent, and for ACAN it is why the record ends up with a misscoped interstitial-matrix term. It would also let aggrecan's two locations be told apart — the same protein sits in cartilage matrix and in the perineuronal net, and only one of those is currently expressible.

## Notes on sources

The affinage record passed its gates (`gates_passed: True`, faith 80%), but is **not** cited as supporting evidence anywhere, which leaves one validation warning standing deliberately. Two reasons. Its narrative is dominated by *regulators of ACAN transcription* — SOX9 and the SOX trio, SHOX2, TET1, SIRT1, HDAC2, miR-140, mTOR/4E-BP1 — and being a transcriptional target of SOX9 is a function of SOX9, not of ACAN. And its two most relevant findings rested on bioRxiv preprints carrying no PubMed identifier. One of those has since been published in J Biol Chem and is used here in its peer-reviewed form (PMID:40273987, fetched and read in full); the other (the Amigo2-Acan CA2 conditional knockout) has not been published and is not used.

Verification: 82 `supporting_text` quotes checked verbatim against their sources (0 problems); `just validate human ACAN` → `✓ Valid`; no deletions in `cache/go/terms.csv` (one addition, `GO:0072534`).


cmungall:
Thanks — all six suggestions applied in 4796894, none of which changed an action. And I have the network access you didn't, so let me close out your open request first.

## The `GO:0072534` ancestor claim, re-verified in both directions

You were right to single this out, since the synapse MODIFY rests entirely on it. `GET /ontology/go/terms/GO:0072534/ancestors?relations=is_a,part_of` returns exactly nine terms:

```
GO:0099535  synapse-associated extracellular matrix
GO:0140047  specialized extracellular matrix
GO:0072534  perineuronal net
GO:0098966  perisynaptic extracellular matrix
GO:0031012  extracellular matrix
GO:0110165  cellular anatomical structure
GO:0030312  external encapsulating structure
GO:0005576  extracellular region
GO:0005575  cellular_component
```

`GO:0045202` is absent. Because a one-directional traversal can mislead, I ran the reciprocal query too: `GET /ontology/go/terms/GO:0045202/descendants?relations=is_a,part_of` returns 143 descendants, and `GO:0072534`, `GO:0098966` and `GO:0099535` are **all absent** from that set. The exclusion is symmetric, so it is deliberate modelling rather than a traversal artifact. Both queries are now recorded verbatim in `ACAN-notes.md` so the next reader doesn't have to take it on trust either.

## On the six suggestions

**1 + 2 — `GO:0005201` → `GO:0030021`.** Fair hit, and the inconsistency was mine: I wrote the collapse note on the synapse row and the reference-replacement note on the `GO:0001501` NAS row, then wrote neither on the row that needed both. The `reason` now says the net effect is a collapse into the two existing RCA rows rather than a third duplicate, and that PMID:1569188 — a synovial-fluid fragment study — does not support osmotic swelling pressure, so the reference should be replaced along with the term.

**3 — the `GO:0085029` IDA.** You are right that "incorporates into a pre-existing PNN" reads as localization, and I had leaned on the wrong strand. The experiment that actually earns the code is the point-mutant comparison, because it is the one that makes ACAN's own binding activity the variable: abolishing the GAG-binding site cuts incorporation roughly five-fold while leaving it detectable. The summary now says that, and also flags that the *residual* incorporation is what bounds the claim.

While there I checked your "collagen genes sit on `GO:0030198` on comparable grounds" intuition, and it is stronger than that — matrix structural constituents are already well represented on `GO:0085029` itself. QuickGO (`goId=GO:0085029&taxonId=9606`, 94 annotations / 40 gene products) gives ELN, FBLN5, EFEMP2, EMILIN1, MFAP4, COL8A1, COL8A2 and TNXB alongside the enzymes and signalling proteins. Elastic-fibre assembly by elastin and the fibulins is the direct analogue of proteoglycan-aggregate assembly by aggrecan, so the term is demonstrably not reserved for remodellers. Recorded in the row and in the notes.

**4 — the anti-adhesive argument.** Agreed that it shouldn't evaporate with the annotation it motivated. Added as a sixth `suggested_question`, framed as you put it: if CSPG barrier function holds for aggrecan *specifically* rather than for the class, it belongs in a negative-regulation process term with its own aggrecan evidence — most of the inhibitory work has been done on neurocan and brevican, which is exactly the transfer this review declines elsewhere.

**5 — the uncited `GO:0043202` IEA row.** Cheap and correct. Cross-referenced to `R-HSA-1793217` and `R-HSA-2046239` via `additional_reference_ids`.

**6 — the disease-reaction Golgi rows.** You've identified a real tension in my wording, and the resolution is that two separable things were collapsed into one sentence. The row now says so explicitly: **the term is true, the provenance is poor.** Aggrecan transits the Golgi lumen to be glycanated regardless of which reaction exported the statement, which is why the action is KEEP_AS_NON_CORE; what is unsatisfactory is only the derivation, and the remedy lies in how Reactome disease reactions are exported rather than in this annotation.

## One thing I'd add for the record

Since the proteolysis REMOVE is the load-bearing call, I made its harm concrete rather than hypothetical. All four of these carry the *identical* term with the *identical* qualifier:

| gene | accession | `GO:0006508` evidence |
|---|---|---|
| ACAN | P16112 | NAS (PMID:1569188) — **the substrate** |
| ADAMTS4 | O75173 | IEA, IBA, TAS (PMID:10751421) |
| ADAMTS5 | Q9UNA0 | IEA, IBA, TAS (PMID:10438522) |
| HTRA1 | Q92743 | ISS, IBA, IEA |

So in any process-based query aggrecan is indistinguishable from the three enzymes that destroy it.

I also pre-empted the objection I'd expect against the proposed cartilage-matrix component term. `GO:0062023` "collagen-containing extracellular matrix" was obsoleted, which superficially looks like a precedent against tissue-scoped ECM components — but its obsoletion comment reads *"This term was obsoleted because it was not clearly defined and usage has been inconsistent."* That term was **composition**-scoped and unbounded; `GO:0140047` still carries plenty of **tissue- or structure**-scoped children (perineuronal net, interphotoreceptor matrix, hyaline layer, cuticular ECM, egg coat, pollen coat, middle lamella, chitin-based ECM, organomineral ECM). Cartilage matrix is anatomically bounded in the way "collagen-containing" never was.

Finally, per campaign guidance not to trust the deep-research provider on domain assignments, every structural claim in the review was checked against the UniProt feature table rather than the affinage narrative: `SIGNAL 1..16`, **no `TRANSMEM`, no GPI anchor** (which is what underpins the cell-adhesion REMOVE); G1 = Ig-like V-type 34-147 + Link 1 153-248 + Link 2 254-350; G2 = Link 3/4; KS 677-849, CS-1 852-1612, CS-2 1613-2277; G3 = EGF-like 2279-2314 + C-type lectin 2327-2441 + Sushi 2445-2505. Glu373-Ala374 falls between Link 2 (ends 350) and Link 3 (starts 478), so it is genuinely interglobular as PMID:1569188 states.

Re-verified after the changes: 82 quotes verbatim (0 problems), `just validate human ACAN` → `✓ Valid`, no deletions in `cache/go/terms.csv`.

PR #2249, opened 2026-07-26
ci: add prompt-injection controls at the repo boundary

cmungall:
Phase 3 of `docs/automation-migration-to-dismech.md`. Stacked on #2247 → #2246.

Token hardening (PRs #2246/#2247) limits blast radius but does not stop an injected agent from misusing its token *during* a run. These three changes cut off the input instead.

## 1. `close-fork-prs.yml` — the highest-leverage single control

A fork PR is the main way external, attacker-authored content (title/body/diff) reaches an agentic workflow's context, and fork-triggered workflows get no repo secrets so they cannot be AI-reviewed anyway. This closes fork PRs on open/reopen with a comment pointing at CONTRIBUTING.md.

It runs from the **base repo's** copy of the workflow (a fork cannot modify it), **never checks out or runs fork code** — only `gh` API calls — so the usual `pull_request_target` code-execution risk does not apply. Needs only `pull-requests: write`. Same-repo PRs are untouched.

## 2. `untrusted-comment-guard.yml` + `.github/scripts/github-trust-gate.js`

Minimizes comments from untrusted authors that carry either a risky attachment (archive / executable / GitHub user-attachment link) or one of this repo's agent-trigger phrases (`@claude`, `@ai4c-agent please`, `@dragon-ai-agent please`, `/review`). Trust is a deterministic check — controller allowlist, collaborator permission, or one of our own bots — **no model**, so the text being inspected cannot argue its way past the gate.

The trust gate is a *trimmed* port of dismech's: only the comment-moderation half. dismech's version also carries ~350 lines of GraphQL discussion-scanner machinery that this repo has no scanner for, so it is left out rather than copied dead.

## 3. `claude.yml` — gate `@claude` on `author_association`

On a public repo, **any** GitHub user could previously drive a write-capable agent (`contents: write`, `pull-requests: write`, `Bash(*)`) just by writing `@claude` in a comment. Now restricted to `OWNER`/`MEMBER`/`COLLABORATOR`.

For the `issues` trigger this gates on the **issue author's** association, not the assigner's — otherwise assigning an externally-authored issue would launder untrusted text into an agent trigger.

## CI

A new `agents` paths-filter runs the agent-config consistency test and the Node guard tests on any `.github/` change, so a workflow-only PR is no longer completely unchecked. `just test-js` runs the Node tests; deliberately kept out of `just test` so a missing `node` cannot break the Python workflow.

## Verification

```
node --test tests/js/*.test.mjs   # 13 pass
python3 -c "import yaml; [yaml.safe_load(open(f)) for f in [...]]"   # all 4 workflows parse
```

The `close-fork-prs` gate cannot be exercised from a same-repo PR by construction; its `if:` is a single `head.repo.full_name != github.repository` comparison, and the job body is three `gh` calls.

### DIFF
--- a/.github/agent-config.yaml
+++ b/.github/agent-config.yaml
@@ -26,6 +26,10 @@ default_model: claude-opus-5
 provider: anthropic
 
 workflows:
+  # The "@claude" mention responder (interactive, controller-gated).
+  claude:
+    model: claude-sonnet-5
+
   claude-code-review:
     model: claude-opus-5
 
--- a/.github/scripts/github-trust-gate.js
+++ b/.github/scripts/github-trust-gate.js
@@ -66,7 +66,7 @@ function loadControllers(controllersPath = ".github/ai-controllers.json") {
 
 function isBotLogin(login, extraBotLogins = []) {
   if (!login) {
-    return true;
+    return false;
   }
   const normalized = normalizeLogin(login);
   if (normalized.endsWith(BOT_LOGIN_SUFFIX)) {
@@ -85,7 +85,13 @@ async function isTrustedLogin({
   controllers = [],
   permissionCache = new Map(),
 }) {
-  if (!login || isBotLogin(login)) {
+  // A missing login is NOT trusted. This is a trust gate: the safe default for
+  // an author we cannot identify is untrusted, and the caller's risk classifier
+  // decides whether anything actually happens.
+  if (!login) {
+    return false;
+  }
+  if (isBotLogin(login)) {
     return true;
   }
   const loginKey = normalizeLogin(login);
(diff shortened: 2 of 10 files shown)


cmungall:

- `isTrustedLogin` returning **trusted** for a missing login was indefensible for a trust gate, and my test comment calling it 'fail closed on the trusted side' was a contradiction in terms. `isBotLogin` now answers only the question in its name; `isTrustedLogin` owns the missing-login decision and fails closed.
- `isTrustedLogin` had no tests at all. It has six now: missing login, controller (case-insensitive), App author, collaborator write vs read-only, unknown user.
- On the commit-message claim about ai.yml: correct as written at that commit — ai.yml is fixed in #2255, later in this stack.

PR #2250, opened 2026-07-26
Review human ABI3BP: a matricellular integrin ligand annotated as a structural brick

cmungall:
# Review human ABI3BP: a matricellular integrin ligand annotated as a structural brick

`Q7Z7G0` / TARSH / NeshBP. Affinage deep research: `gates_passed: True`, faith 100%, all 12 citations numeric PMIDs (no bioRxiv-DOI-in-a-PMID-field entries).

## What the record looked like

Ten rows. Seven cellular-component, three molecular-function, **zero biological process**. No IBA anywhere — UniProt states it outright: `DR   PAN-GO; Q7Z7G0; 0 GO annotations based on evolutionary models.` PANTHER PTHR23197:SF10 is assigned but was never PAINT-curated.

**WITH/FROM resolution** (the brief's highest-yield check): exactly one entry exists in the whole file, `UniProtKB-SubCell:SL-0243` on the IEA row. Fetched from `rest.uniprot.org/locations/SL-0243.json` → name "Secreted", definition "Protein located outside the cell membrane(s)", single GO mapping `GO:0005576`. One-to-one and faithful. There is no ortholog or paralog transfer in this record to audit, so the interesting failure had to be elsewhere — and it was.

## The finding: location evidence promoted into a molecular function

All three MF rows assert `GO:0005201 extracellular matrix structural constituent` by **RCA** from ECM-proteomics sources — PMID:28675934, PMID:27559042, PMID:20551380. **The same three papers are already cited for the HDA `GO:0031012`/`GO:0005576` rows in this record.**

What those datasets show is that ABI3BP peptides come out of ECM-enriched fractions of human tissue. `GO:0005201` is defined (QuickGO) as *"The action of a molecule that contributes to the structural integrity of the extracellular matrix."* Being in the matrix is not holding the matrix together, and the RCA silently crosses that line. Three repetitions of one inference over three inventories are not three independent pieces of evidence.

The weakest citation of the three is worth naming: **PMID:28675934 is a protocols paper.** It supplies a workflow plus two R-based web tools and illustrates them on breast and ovarian tissue pairs. No ABI3BP-specific experiment, no matrix-mechanics experiment.

→ **`MARK_AS_OVER_ANNOTATED` ×3, not `REMOVE`.** The protein genuinely is matrix-resident, and an FN3-containing fragment genuinely can assemble into matrix in vitro (reported second-hand via PMID:23666637 from PMID:18757743, whose full text is unretrievable and whose abstract does not name ABI3BP — flagged `UNVERIFIED`). A structural contribution has never been *excluded*; it has never been *tested*. That distinction is the difference between over-annotation and error.

All seven CC rows **`ACCEPT`ed** — the localisation is reproduced across aorta, vein, atrium and cartilage, plus non-proteomic antibody staining of extracellular deposits. `GO:0140149 non-collagenous component of interstitial matrix` is the one genuinely informative term in the record.

## What was actually measured, and is missing: 6 NEW rows

ABI3BP binds **integrin beta-1** — co-IP with *both* a tagged construct *and* endogenous protein; a blocking-antibody panel where alpha-4 and alpha-5 did nothing and beta-1 raised phospho-ERK1/2 ~3.5-fold; and the authors explicitly excluded a diffusible growth-factor mode first. Engaged beta-1 → phospho-paxillin → Src sequestration → ERK down → no cyclin-D1 → no S-phase. So proliferation is restrained and differentiation is licensed, in one coupled switch.

| New term | Ref | Basis |
|---|---|---|
| `GO:0005178` integrin binding (MF) | PMID:23666637 | tagged + endogenous co-IP, beta-1-specific antibody panel |
| `GO:0033625` positive regulation of integrin activation | PMID:23666637 | ligand changes receptor state (phospho-paxillin/paxillin ratio) |
| `GO:1902461` negative regulation of mesenchymal stem cell proliferation | PMID:23666637 | germline KO, in vivo phenotype in 3 organs, + independent shRNA lines |
| `GO:0070373` negative regulation of ERK1 and ERK2 cascade | PMID:23666637 | perturbed at ligand, receptor and adaptor; vinculin knockdown is the specificity control |
| `GO:2000738` positive regulation of stem cell differentiation | PMID:23666637 | two lineages fail in KO cells → deliberately kept lineage-agnostic |
| `GO:2000727` positive regulation of cardiac muscle cell differentiation | PMID:25296984 | in vivo ablation + marker analysis + antibody proof it runs through beta-1 |

All coded **`ISS`** with `supporting_entities: MGI:MGI:2444583`, because every perturbation experiment is mouse and the species of the myc-tagged HEK293 construct is stated only in supplementary methods not in the cache. Human corroboration is separate and concordant: restoring ABI3BP arrests growth in thyroid (PMID:18559958), lung (PMID:40092729) and gallbladder (PMID:31174563) cells.

`GO:0098640 integrin binding involved in cell-matrix adhesion` was considered and rejected — the demonstrated consequence of binding is receptor activation and signalling, not adhesion of the cell to the matrix.

## Proposed new GO term: `matricellular protein activity`

The over-annotation above is not a one-off curator slip, it is what the ontology *forces*. `GO:0005201`'s children are uniformly mechanical — tensile strength, compression resistance, elasticity, lubricant, plus three "structural constituent of X" terms. There is no sibling for matrix proteins whose action is to instruct cells. Searching GO for "matricellular" returns an empty set (checked twice: OLS `search_all_ontologies` restricted to `go`, and the QuickGO ontology search).

So a pipeline that knows only "non-collagenous core-matrisome glycoprotein" has exactly one ECM-specific molecular function available, and it is the wrong one. Proposed under `GO:0048018 receptor ligand activity`, alongside growth factor / cytokine / hormone / morphogen activity, distinguished from them by matrix residence rather than diffusion — which for this gene is experimentally grounded, since the growth-factor mode was tested and excluded. The same pressure applies across thrombospondins, tenascins, SPARC, periostin and the CCN family.

## Deliberately *not* annotated

- **Senescence.** Both signs are supported by perturbation experiments and the field says so in print (PMID:23666637: *"Abi3bp has been reported to have both a positive and negative role in senescence"*). Knockdown in MEFs → more senescence (PMID:19338757); re-expression in human thyroid cells → more senescence (PMID:18559958) — *opposite* manipulations, same reported outcome; while knockdown/knockout → *less* senescence in VSMC, renal tubule and chondrocytes (PMID:40889718, PMID:38812032, PMID:41232761). Four references marked `DISPUTED`. No senescence term until this resolves.
- Antiviral/IRF3 (single preliminary in vitro study), blood-brain-barrier protection (supraphysiological recombinant protein), and the coloboma variant (one de-novo VUS reported in a paper about *ACTG1*; UniProt itself hedges).

## Two things checked rather than trusted

- **Thrombin site.** PMID:41839242 reports cleavage at Arg-337. Residue 337 of the canonical sequence *is* Arg, and the context is Pro-336/Arg-337↓Ser-338 — proline at P2, arginine at P1, serine at P1', a textbook thrombin motif — sitting in the disordered linker between the two FN3 domains.
- **The gene's own name.** PMID:11501947 reports a 486-residue clone. Canonical 1068 − the 1–607 deletion (VSP_010860) + the 25-residue insertion at 669 (VSP_010861) = **486 exactly** → the Nesh/ABI3 SH3 two-hybrid bait recovered isoform 2, not the full-length protein. Twelve years on the caveat still stood in print, and UniProt still says `SUBUNIT: Probably interacts with ABI3.` GOA carries no annotation for this pair, which is correct; none is proposed.

Also noted for GOA, raised as a question since no review action targets qualifiers: PMID:20551380 yields `located_in GO:0005576` *and* `colocalizes_with GO:0031012` from one fractionation experiment — `colocalizes_with` is for evidence that cannot resolve whether a protein is inside a structure, which does not describe biochemical fractionation.

## Verification

- `checkquotes.py` — **90 quotes, 0 problems** (it checks `file:` quotes as well as PMIDs, so the UniProt lines are verified too; all kept on single physical lines)
- `just validate human ABI3BP` — **✓ Valid, zero warnings**
- `git diff origin/main -- cache/go/terms.csv` — **+4, −0** (restored from main and re-appended in sorted position after a validate run had dropped 18 unrelated terms)
- `origin/main` merged in before final validation, so CI sees the merged state
- The affinage record is cited exactly once, on the `GO:1902461` row, for its high-level summary sentence only — independently carried on that same row by two verbatim PMID:23666637 quotes and one from PMID:18559958. No domain boundary, binding partner or mechanistic step in this review rests on it; those come from UniProt's feature table or the primary papers.


cmungall:
Two self-caught corrections pushed since the PR opened, both worth flagging explicitly rather than burying in a diff.

**1. A term-definition inversion in my own draft (`23392997b`).**

I had annotated the differentiation half of ABI3BP's switch as `GO:2000738 positive regulation of stem cell differentiation`, reasoning that since knockout MSCs lose *both* osteogenic and adipogenic capacity, a lineage-agnostic term was the faithful choice.

That was wrong. `GO:2000738`'s target, `GO:0048863 stem cell differentiation`, is defined as:

> "The process in which a relatively unspecialized cell acquires specialized features of **a stem cell**."

That is differentiation *into* a stem cell. The annotation therefore asserted that ABI3BP promotes cells *becoming* stem cells — the exact reverse of the observed biology, where ABI3BP is required for MSCs to *leave* the stem state. `GO:2000741 positive regulation of mesenchymal stem cell differentiation`, which looks like the better-scoped alternative, inherits the identical inversion through `GO:0072497` ("…acquires specialized features of a mesenchymal stem cell"), so reaching for the more specific term would have made it worse, not better.

The whole GO `<cell type> differentiation` branch means "becoming that cell type", so outcome-specific terms are the only correct route:

- `GO:0045669 positive regulation of osteoblast differentiation` → `GO:0001649` *"acquires the specialized features of an osteoblast"* ✓
- `GO:0045600 positive regulation of fat cell differentiation` → `GO:0045444` *"acquires specialized features of an adipocyte"* ✓

Both rest on the same knockout experiment already quoted. I then re-checked the other two process terms against their targets: `GO:2000727` → `GO:0055007` *"a cardiac muscle precursor cell acquires specialized features of a cardiac muscle cell"* ✓, and `GO:1902461` → `GO:0097168` *"multiplication or reproduction of mesenchymal stem cells"* ✓. Both stand.

The generalisable lesson, recorded in `ABI3BP-notes.md`: fetch the definition of the **target** of every regulation term, not just the regulation term itself. `positive regulation of stem cell differentiation` reads correctly in both directions — only the target's definition disambiguates.

**2. `propagation_review` added to the three `GO:0005201` RCA rows (`7c329e6f0`).**

The validator only demands it for IBA/ISO, but RCA is a computational-inference code with an identifiable source, so the same taxonomy applies and it makes the central argument queryable instead of prose-only:

- `root_cause: TERM_SCOPING_PROBLEM` — the term is mechanically scoped and all its children are mechanical, so it cannot express matrix-resident cell-instructive activity
- `failure_modes: [ROLE_CONFLATION, SOURCE_EVIDENCE_WEAK]`
- each proteomics dataset marked `SUPPORTS_SOURCE_BUT_NOT_TARGET` — it legitimately carries the co-annotated HDA cellular-component row from the *same* experiment, but cannot license a molecular function
- the 2nd and 3rd repetitions additionally `CIRCULAR_OR_REDUNDANT`

Re-verified after both changes: `checkquotes.py` **93 quotes, 0 problems**; `just validate human ABI3BP` **✓ Valid, zero warnings**; `cache/go/terms.csv` **+4 −0** (a `just validate` run silently dropped unrelated cached terms twice — restored from `origin/main` and re-appended in sorted position each time, so the diff contains only my four additions).

cmungall:
Thanks — both 🟡 points were right, and #2 in particular was a real overstatement rather than a wording quibble. All five items addressed in `8949c1bc4`.

---

### 🟡 1. Olfactory dendritic refinement — row added

You were right on both halves, including the part I got wrong twice. My rationale claimed the role was "described in the gene description **and questions**" — there was no such question anywhere in the file. And "single-laboratory and mouse-only" is precisely the standard behind the five rows I drew from PMID:23666637, with PMID:25296984 coming from the *same* laboratory. Applying ISS to one body of mouse work while withholding it from another of identical grade was just inconsistent.

Added `GO:0050774 negative regulation of dendrite morphogenesis` — `ISS`, `PMID:19302145`, `supporting_entities: MGI:MGI:2444583`. Three quotes carry it: the conditioned-medium result plus the cell-autonomous overexpression arm, the transient expression during the refinement window, and the naris-closure/AC3-knockout control that excludes activity dependence (so it is a genetically programmed cue, not a response).

Target definition checked the same way as the `GO:2000738` catch: `GO:0048813 dendrite morphogenesis` is *"The process in which the anatomical structures of a dendrite are generated and organized"*, so reducing branching and complexity is negative regulation of it. No inversion.

**Kept out of `core_functions`, with the reason stated rather than left implicit.** This is the one place I have deliberately kept an asymmetry: no receptor has been identified for the neuronal effect, so there is no molecular function to attach the process to without assuming integrin β1 is involved, and it has not been reproduced outside the originating lab. A scope note now says so in the core function's own description, and a new question asks what the receptor is.

**One thing you should know for future reviews of this file:** `PMID_19302145.md` has `full_text_available: true` and does contain the full text, but the *reference validator's* own record for that PMID holds only the abstract. The pre-write hook blocked my first attempt at this row because I quoted the purified-protein arm (`purified TARSH (500 ng/ml) induced significant reduction…`), which `checkquotes.py` accepts and the validator rejects. Every quote from that reference is now abstract-only, and the discrepancy is recorded in its `review_notes` and in the gene notes.

### 🟡 2. Matricellular claim narrowed — and it was worse than you said

You flagged that "matrix residence rather than diffusion… experimentally grounded" overstates the growth-factor control. It does, and there is a second, more direct contradiction sitting in the *same* paper I cited:

> Concentrated media prepared from HEK293 cells expressing the myc-tagged Abi3bp inhibited proliferation of both MSC and MSC-GFP-Akt1 cells when compared to control concentrated media (Supplementary Figure 3E) further validating the anti-proliferative effect of Abi3bp.

So soluble ABI3BP in conditioned medium is not merely present, it is **functionally active on the relevant phenotype**. Together with purified TARSH acting on neurons in dissociated culture, that is two independent demonstrations that this protein works at a distance from its source. The tethering criterion was not just under-supported, it was contradicted.

Rewritten to say what the evidence supports: the distinguishing features are receptor class and matrix residence, *not* tethering; the growth-factor control shows only that ABI3BP is not a growth-factor-like ERK agonist (its output is the opposite sign and runs through an adhesion receptor); and whether "matricellular" is best defined by residence, receptor class, or as a graded category is left to the ontology editors. Both counter-observations are now quoted in the proposal's own `supported_by`, so a GO editor sees them rather than discovering them.

The gap argument is untouched, since that is where the weight always was.

### 🔵 3, 4, 5 — all taken

- **Outside-in caveat on `GO:0033625`**: added to `review.reason`. `GO:0033622` is defined via *"increased affinity of the integrin for its extracellular ligands"* and that branch is conventionally used for talin/kindlin **inside-out** modulation, whereas ABI3BP *is* the extracellular ligand. The definition does not restrict the term, and the paper's wording supports the claim, but an outside-in ligand is not the canonical filler and a curator may prefer to carry the point on `GO:0005178` alone. Flagged in the same spirit as the `GO:0098640` rejection.
- **Stale count**: fixed — now "Eight NEW rows" over an eight-row table, with the added row and the reason it was added recorded.
- **Diffusible-versus-tethered question**: added, and it is the sharper form of the point — deposits and insoluble-fraction recovery on one side, active conditioned medium and purified protein on the other, with thrombin cleavage at Arg-337 as an untested candidate mechanism for releasing a soluble fragment from a tethered pool.

I also added a paragraph to the notes contrasting *why* the neuronal role is now annotated while senescence, antiviral and BBB roles are not — the difference is the kind of evidence, not the number of labs: three concordant gain-of-function arms plus a matching expression time course, versus claims that contradict each other in sign, a study self-described as preliminary in its own title, and an arm resting on supraphysiological recombinant protein.

### On your two "not verified" rows

- **`MGI:MGI:2444583`** — from `rest.uniprot.org/uniprotkb/F7B3T6.json`, `uniProtKBCrossReferences` → `{"database":"MGI","id":"MGI:2444583","properties":[{"key":"GeneName","value":"Abi3bp"}]}`. Mouse Abi3bp has no Swiss-Prot reviewed entry, which is why an MGI id is used rather than a UniProt accession.
- **`cache/go/terms.csv`** — worth knowing: `just validate` silently *drops* unrelated cached terms on each run (18 the first time, one on each of two later runs), so the file has to be restored from `origin/main` and re-added after every validation. And `origin/main` currently contains an out-of-order entry (`GO:0140312`), so re-inserting by global sort relocates it and shows as a spurious deletion — the four new terms are now inserted in position without touching anything else. Both traps are recorded in the notes.

Final state: **18 annotations** (7 ACCEPT, 3 MARK_AS_OVER_ANNOTATED, 8 NEW), 10 questions, 5 experiments. `checkquotes.py` **100 quotes, 0 problems**; `just validate human ABI3BP` **✓ Valid, zero warnings**; `cache/go/terms.csv` **+4 −0**.

cmungall:
Thanks for the approval. I've taken all three 🔵 nits rather than leaving them — the first was a genuine self-inconsistency, not a style point. Pushed as `6c05b3605`.

**1. The proposed definition discriminated on an unmeasurable negative.** You're right, and the sharper version of the problem is that it was inconsistent with my *own* position elsewhere in the file: the definition required "without contributing appreciably to the mechanical integrity of the matrix", while the three `MARK_AS_OVER_ANNOTATED` rows argue that a structural contribution for ABI3BP "has never been *excluded* — it has never been *tested*". So the motivating example could not be shown to satisfy the defining clause of the term it motivates. A GO editor would have found that in the first minute.

Reworded positively, and it now says so explicitly:

> …acts on cells by engaging cell-surface receptors, thereby modulating their signalling, adhesion, proliferation or differentiation state. The defining feature is the cell-instructive action on a receptor; whether a protein with this activity also contributes to the mechanical integrity of the matrix is a separate question, and this term neither asserts nor excludes it.

That also makes the term usable for the matricellular proteins that genuinely *do* have a structural role alongside a signalling one, which the old wording would have excluded for no good reason.

**2. `GO:0050774` omitted from the enumeration** — added, and while there I noted why it is the one proposed process kept out of `core_functions` (different axis, no identified receptor), so the list no longer reads as if all eight rows are on the same footing.

**3. Unhedged `description` clause** — "acts on cells rather than on the fabric of the matrix" → "its characterised action is on cells rather than on the mechanical fabric of the matrix". *Characterised* is the load-bearing word: it is a statement about what has been measured, not a claim about what is impossible.

**On verification, since this bit others this session:** all three edits were applied with a script that asserts each anchor string is present *and unique* before replacing and re-greps afterwards, so a missed occurrence fails loudly instead of silently no-op'ing. I then inspected the diff at claim level (`git diff -U0 | grep '^[-+][^-+]'`) rather than trusting that a modified line means a modified claim, and re-parsed the YAML to confirm 18 annotations / 10 questions survived intact.

**`cache/go/terms.csv` re-checked against the moved main.** `origin/main` advanced by five merges (ACAP1, ABRACL, ABRA, ACAN, ACP7) while this PR was open. `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` is empty — **0 deletions, +4 insertions**. Worth recording for anyone who hits this: `just validate` silently drops unrelated cached rows on each run (18, then 1, then 1 across my runs), and `origin/main` contains at least one out-of-order entry (`GO:0140312`), so re-inserting by global sort relocates it and surfaces as a spurious deletion. Insert in position instead. Both traps are documented in `ABI3BP-notes.md`.

**No `just validate` warnings are being left standing** — it reports `✓ Valid` with zero warnings, including the deep-research-citation warning, which is satisfied legitimately: the affinage file is cited once on the `GO:1902461` row for a high-level summary sentence only, with two PMID:23666637 quotes and one from PMID:18559958 carrying the same claim on that row, and its `reference_review` records that no mechanistic claim rests on it and that its own coarse GO grounding (`GO:0098631`/`GO:0098772`, no BP at all) was not imported.

Final: `checkquotes.py` **100 quotes, 0 problems**; `just validate human ABI3BP` **✓ Valid, zero warnings**; `cache/go/terms.csv` **+4 −0**; 18 rows = 7 ACCEPT + 3 MARK_AS_OVER_ANNOTATED + 8 NEW.

cmungall:
Note for the record: this PR auto-merged at `8949c1bc4` (02:09:45Z) about a minute before commit `6c05b3605` — the three 🔵 nits from the approving review — was pushed, so those fixes did not reach main. Verified against `origin/main` rather than assumed: it still had the old unmeasurable-negative definition, still lacked `GO:0050774` in the process enumeration, and still carried the unhedged `description` clause.

Cherry-picked onto a fresh branch off current main and opened as #2262. No annotation actions, terms, evidence codes or quotes change there — it is the same three edits, re-validated (checkquotes 100/100, `just validate` ✓ Valid zero warnings, no `cache/go/terms.csv` change since this PR already landed the four new terms).

PR #2251, opened 2026-07-26
Review human ACAP1: a paralog inversion, and an experimentally void GO record

cmungall:
# Review human ACAP1 (Q15027) — ArfGAP with coiled-coil, ANK repeat and PH domains 1

## The finding

ACAP1 is one of the best-characterised ARF6 GAPs in the human proteome: seven PDB entries, a
crystal structure with its cargo peptide, and a continuous primary literature from 2000 to 2023.
**Its GO record contains no IDA, IMP or IGI at all.** All 19 GOA rows are 4 IBA, 3 IEA, 11 bare
`GO:0005515 protein binding` IPI, and 1 HDA.

I queried QuickGO for ACAP1 annotations from each of its seven primary papers. All seven return zero:

| PMID | topic | ACAP1 GO annotations |
|---|---|---|
| 11062263 | ARF6 GAP activity, PIP2 dependence, R448Q catalytic mutant | 0 |
| 16256741 | Akt-Ser554 switch, ITGB1 recycling, migration | 0 |
| 17398097 | GULP1/ARF6 tripartite complex | 0 |
| 17664335 | ACAP1 clathrin coat, ITGB1 + GLUT4 recycling | 0 |
| 22645133 | ITGB1 recycling sorting signal, autoinhibition | 0 |
| 25284369 | PH-domain membrane curvature, BAR lattice | 0 |
| 25225293 | tubular recycling endosome, ARAP2 contrast | 0 |

UniProt has read all of them — FUNCTION, SUBUNIT, DOMAIN and PTM carry `ECO:0000269` throughout —
but none of it reached GO.

## The paralog result was the inverse of what I went looking for

ACAP1 has close paralogs ACAP2 and ACAP3, so I resolved every WITH/FROM accession expecting a
spurious transfer *into* ACAP1. Three IBAs do cite paralogs — `UniProtKB:Q15057` (human ACAP2) in
the GAP-activity and plasma-membrane IBAs, `MGI:MGI:2153589` (mouse Acap3) in the GAP-activity IBA,
`RGD:1562939` (rat Acap2) in the endosome-membrane IBA. **All are legitimate.** PAINT may support an
ancestral node from any experimentally annotated descendant, paralogs included, and GAP activity,
endosome membrane and plasma membrane are all ancestral to the ACAP clade. The tree also discriminates
correctly where it should: mouse Acap3's `neuron migration` and `regulation of neuron projection
development` hang off a different node (PTN002754173) and did *not* reach ACAP1.

The real problem runs the other way. **PMID:11062263 produced three IDAs on ACAP2 and none on ACAP1**
— `GO:0005096` GTPase activator activity, `GO:0030029` actin filament-based process, `GO:0001726`
ruffle. That paper assayed the two proteins side by side, in the same figures; ACAP1 is first in the
title; and the text says ACAP1 "was most effective at inhibiting protrusions". The consequence is an
inversion of the usual paralog hazard: **ACAP1's ARF6 GAP activity now reaches ACAP1 only as a
phylogenetic inference whose WITH/FROM cites ACAP2, for an activity that was measured on ACAP1
itself in the cited source.**

## Actions

5 ACCEPT · 2 KEEP_AS_NON_CORE · 5 MARK_AS_OVER_ANNOTATED · 6 NEW (18 rows: 12 GOA + 6 proposed)

**`GO:0005886` plasma membrane IBA → KEEP_AS_NON_CORE, not REMOVE.** My first instinct was a
compartment mismatch, since UniProt gives only recycling endosome membrane. The full text refutes
that: ACAP1 localises to dorsal ruffles and the edge of protrusive cells, and *"ACAP1 remained
cytosolic and was not recruited to the endosome or the plasma membrane in cells expressing the
dominant-negative Arf6, T27N"* — a real, ARF6-dependent surface pool. What makes it non-core rather
than core is that the GAP-dead R448Q mutant *keeps* the surface pool while *losing* the endosomal
one, so catalysis tracks with the endosome.

**`GO:0030036` actin cytoskeleton organization IBA → KEEP_AS_NON_CORE**, with the one genuine tree
oddity recorded. `FB:FBgn0004133` is Drosophila *blown fuse*; PANTHER places it in PTHR23180
subfamily SF399, but it has **only a PH domain** — no BAR, no Arf-GAP, no ANK repeats on any isoform
(checked A1Z714, P91678, E1JGZ4, Q8MSU1) — so it is not an ACAP and cannot act through ARF6 GAP
catalysis. Its actin annotation is a *Drosophila* myoblast-fusion IMP. Marked
`SUPPORTS_SOURCE_BUT_NOT_TARGET`. **The annotation nonetheless survives**, because the two
Dictyostelium sources are full-architecture ACAPs and DDB_G0279649 carries actin cytoskeleton
organization by both IMP and IGI plus filopodium assembly, F-actin and cell cortex. Non-core because
the actin effect is downstream of ARF6 and shown only by overexpression; ACAP1's own knockdown
phenotype is a traffic phenotype.

**All 11 protein-binding rows resolved, decided per partner → MARK_AS_OVER_ANNOTATED.**
Compartment-implausible: UQCRB and UQCRC2 (two mitochondrial complex III subunits, from *unrelated*
screens — itself a sticky-prey signature), SETDB1, KAT5, LMO3 (nuclear). Unsupported but not
impossible: PRKCA, YWHAG (generic Y2H hubs). Worth pursuing: **FCHSD2**, an F-BAR/double-SH3 protein
of the clathrin pathway, and **GRB2**, the only partner recurring across two orthogonal assays —
though the pY-dependent Y2H presupposes a phosphotyrosine on ACAP1 that UniProt does not record.
Meanwhile **not one validated partner — ARF6, ITGB1, CLTC, GULP1, SLC2A4 — appears anywhere in GO.**

**`GO:0016020` membrane HDA → MARK_AS_OVER_ANNOTATED**, using the paper against itself: the authors
state they treated the membranes to *remove peripheral membrane proteins* and that the survivors were
largely proteins *"transiently associated with membranes"*. ACAP1 is exactly that class. Detection is
not in doubt (`PE 1`, and NK cells fit its bone-marrow/lymphoid expression); the unspecified bilayer
is what conveys nothing.

**`GO:0005737` cytoplasm IEA → ACCEPT** rather than modified. The InterPro2GO mapping (BAR domain →
cytoplasm) is uninformative, but the conclusion is true for a reason the mapping doesn't know: ACAP1
is genuinely soluble at rest. Replacing it with the endosome term would discard a real pool.

**6 NEW annotations**, all for terms absent from GOA: `GO:0032456` endocytic recycling (IMP),
`GO:0140312` cargo adaptor activity (IDA), `GO:0030118` clathrin coat (part_of, IDA), `GO:0180020`
membrane bending activity (IDA), `GO:0005546` PIP2 binding (IDA), `GO:0072659` protein localization to
plasma membrane (IMP). `GO:0180020` was only created in 2023-06, after all of this gene's literature —
a plausible reason the activity was never annotated.

A seventh was attempted and withdrawn: an IDA for `GO:0005096` from PMID:11062263, the single most
warranted annotation here. The validator correctly rejects `action: NEW` for a term already in GOA,
so the recommendation sits in the IBA row's review instead. Worth noting as a small representational
gap: this format can say "this term needs a better *evidence code*" only in prose.

## Ontology gap

`GO:0008060 ARF GTPase activator activity` no longer exists — QuickGO reports it as a **secondary id
of `GO:0005096`**, merged along with `GO:0005097` (Rab), `GO:0005098` (Ras), `GO:0005100` (Rho) and
`GO:0046582`. So the best-established fact about this protein — that it is an **ARF6** GAP, with a
measured in-vitro preference over ARF1 and ARF5 — cannot be stated in the term. It needs
`has_input UniProtKB:P62330`. **No new term is proposed:** the merge was a deliberate ontology
decision, not an omission. I've raised whether that extension should be treated as *required* on GAP
annotations, since otherwise an ARF6 GAP, an ARF1 GAP and a Rho GAP are indistinguishable.

Also flagged: `GO:0030118 clathrin coat` is defined around coated pits and AP-adaptor-linked vesicles,
whereas the ACAP1 coat is ARF6-regulated and acts on *recycling* endosomes. Closest available term,
but silent on the step.

## Verification

- `checkquotes.py` — **101 quotes, 0 problems**
- `just validate human ACAP1` — **✓ Valid**, 1 warning
- `cache/go/terms.csv` — restored from `origin/main` after finding **18 unrelated deletions** from
  concurrent branches; re-appended only my one new term (`GO:0140312`)
- Caught and fixed a wrong accession before commit: I had written `UniProtKB:P11021` for GLUT4, which
  is BiP/HSPA5. GLUT4 is P14672; since PMID:22645133 assays only the ITGB1 cargo, I dropped the second
  accession rather than substitute it.

The one standing warning is "no annotations reference available deep research files". Left
deliberately: no affinage sentence is used as supporting text anywhere, and every mechanistic claim
is quoted from a primary PMID or the UniProt record. The affinage record is cited with
`correctness: UNVERIFIED` and its own GO grounding explicitly *not* imported — it offers
`GO:0005198` structural molecule activity and `GO:0060090` molecular adaptor activity where the
specific `GO:0140312` and `GO:0180020` are available, and omits the GAP activity entirely. It was
useful for surfacing two papers I hadn't reached (PMID:36917255, PMID:37505213), both recorded
without annotations pending full text.


cmungall:
Thanks — this was a careful review, and the independent corroboration of the `blow` argument from `PTHR23180-entries.csv:18` is a better piece of evidence than the one I used. I had checked blow's UniProt features (PH domain only across A1Z714/P91678/E1JGZ4/Q8MSU1, no BAR/Arf-GAP/ANK) but had not noticed that PANTHER *names* SF399 "BLOWN FUSE, ISOFORM A-RELATED" while ACAP1 sits in SF197. That makes the point without needing the domain argument at all.

Pushed 16c72ee76. Four of the six suggestions taken, two declined with reasons.

**Taken.**

1. **`GO:0140312` — both halves of the definition now anchored on the annotation.** You're right that this was load-bearing: the term requires bridging cargo *and* the coat scaffold, and I had only the ITGB1 sorting-signal half locally, with CLTC quoted one entry away. Added the UniProt `Interacts with CLTC.` line and the `PMID:17664335` coat sentence to that entry's `supported_by`.

2. **`GO:0005737` cytoplasm: `ACCEPT` → `KEEP_AS_NON_CORE`.** Agreed, and this was a real inconsistency rather than a stylistic one. `ACCEPT` carries "retain as representing the core function"; no `core_functions` entry lists cytoplasm as a location (all four sit on the recycling endosome membrane); and I had given the *more* informative plasma-membrane annotation the weaker action. The cytosolic pool is where ACAP1 waits, not where it works.

3. **`is_active_in` caveat now visible on the annotation row.** Moved out of `suggested_experiments` and into `review.reason`: because R448Q keeps the surface pool while losing the endosomal one, the plasma-membrane pool may be recruitment without catalysis, in which case `located_in` is the better relation. Retained `is_active_in` since ARF6 is activated at the plasma membrane and nobody has localised the hydrolysis event — but a curator revisiting the row should treat the *relation*, not the term, as the open question.

4. **Corrected a false claim in the `PMID:32296183` block.** I had written that the implausible partners sit alongside the plausible one "with no way for a reader to tell them apart." That's wrong about the format — as you note, `AAMDC` does one entry per partner, and GOA carries three separate rows here. Reworded as a choice rather than a limitation.

**Declined, with reasons.**

5. **Per-partner splitting (your #4) — kept collapsed.** Having looked at it: splitting wouldn't change any action. FCHSD2 is mechanistically the most sensible partner ACAP1 has in the screen data, but it is still an unreplicated single-assay Y2H hit, and per the campaign rule that makes it `MARK_AS_OVER_ANNOTATED` exactly like UQCRB and NEBL — a bare `protein binding` term is uninformative whether or not the partner is plausible. So splitting would change the bookkeeping and not the verdict. The partners are separated in `supporting_entities` (machine-readable) and adjudicated individually in prose. I've said so explicitly rather than implying the format can't do it.

6. **`core_functions` #4 (PIP2 binding) — kept.** It does self-describe as regulatory, and I take the dilution point. But it's a distinct, directly-assayed molecular function with a clean separating point mutant (K274N abolishes PIP2/PIP3 binding *and* endosomal-tubule association), and lipid gating seems central to how this GAP is confined to the right membrane rather than incidental — PIP2 is also what stimulates the catalysis in the first place. Happy to drop it to three if you'd rather.

**On `cache/go/terms.csv` (your #6) — leaving it, and it's worth recording why**, because I initially measured this wrong. Those 18 DNA-repair terms are additions relative to the *merge base* but are already present on `main`; they landed after my branch point, and my `git checkout origin/main -- cache/go/terms.csv` pulled the tip version. Rebuilding from the merge base would clean the PR diff but re-introduce 18 *deletions* relative to `main`, which is the specific failure the campaign brief warns about. `git diff origin/main -- cache/go/terms.csv` shows zero deletions and exactly one addition (`GO:0140312`). I had to re-apply this during the review, incidentally — `main` advanced again and my file had started deleting `GO:0072534 perineuronal net` from the concurrent ACAN branch.

**Also fixed, self-caught:** I had written `UniProtKB:P11021` for GLUT4 in `supporting_entities`. That is BiP/HSPA5 — GLUT4 is P14672. Since `PMID:22645133` assays only the ITGB1 cargo, I dropped the second accession rather than substitute it, and noted GLUT4 belongs on a separate annotation sourced from `PMID:17664335`. My PR body also undercounted `ACCEPT` as 3 when it was 5; corrected. Current tally: **4 ACCEPT, 3 KEEP_AS_NON_CORE, 5 MARK_AS_OVER_ANNOTATED, 6 NEW.**

Agreed on filing the schema observation separately — that the format can only say "this term needs a better *evidence code*" in prose. For this gene it matters concretely: an IDA for `GO:0005096` from `PMID:11062263` is the single most warranted annotation here, and it's the one thing this review cannot express as a row.

`checkquotes.py`: 103 quotes, 0 problems. `just validate human ACAP1`: ✓ Valid.

PR #2253, opened 2026-07-26
ci(scanners): move the four scanners onto the ai4c-agent App token

cmungall:
Phase 1 continued. Stacked on #2249 → #2247 → #2246.

`curation-scanner`, `go-annotation-scanner`, `litscan-module-member` and `arba-issue-monitor` get the same treatment as `pr-shepherd` in #2246:

- short-lived `ai4c-agent` App token; `PAT_FOR_PR` removed
- `persist-credentials: false` + `gh auth setup-git`
- SHA-pinned `create-github-app-token@v3.2.0` and `claude-code-action@v1`
- model from `.github/agent-config.yaml`; `model:` dispatch inputs converted to no-override strings
- `allowed_bots` includes the writer app
- injection-safe step summaries (`env:` + `printf`, tilde fence)
- default `GITHUB_TOKEN` scoped to `contents: read`; `timeout-minutes: 55` (App tokens expire at 60 and do not refresh)

## Untrusted-input guardrails

Each scanner prompt gains an explicit one. It matters more here than for the shepherd: **`go-annotation-scanner` and `arba-issue-monitor` read issue text from an upstream repository we do not control**, and the litscan packet is third-party paper titles and abstracts.

## curation-scanner fan-out

The effort/model/selector matrix moves into `agent-config.yaml`; a `setup` job emits it through the resolver's `--matrix` mode and feeds `strategy.matrix.include`. Tier models go to Sonnet 5 / Opus 5 (Haiku stays 4.5 — no v5 exists).

## `arba-issue-monitor` was silently broken

It still called `claude-code-action` with the **v0** inputs `mode:`, `direct_prompt:` and `allowed_tools:`. None of those exist in v1 (verified against the action's `action.yml` at the pinned SHA — the inputs are `prompt:` and `claude_args:`), so **the prompt was never delivered to the agent**. It also had no `github_token`, so its `gh pr create` would have failed silently — the same false-green shape as dismech's weekly-compliance. Both fixed, plus output capture so the next breakage is visible.

## Verification

```
uv run pytest tests/test_agent_config.py -q   # 10 passed
python3 -c "import yaml; ..."                 # all four workflows parse
grep -rn PAT_FOR_PR .github/workflows/        # only ai/dragon-ai/generate-pages/warm-reference-cache remain
```

The model-pin allowlist in `tests/test_agent_config.py` shrinks from 6 to 2 (`claude`, `weekly-compliance`) — and a *stale* entry is itself a test failure, so it can only shrink.

### DIFF
--- a/.github/actions/agent-run-summary/action.yml
+++ b/.github/actions/agent-run-summary/action.yml
@@ -1,0 +1,52 @@
+name: Agent run summary
+description: >
+  Write an agentic workflow's final report to $GITHUB_STEP_SUMMARY, read from
+  claude-code-action's execution log.
+
+  claude-code-action v1 has NO `result` output — its outputs are execution_file,
+  branch_name, github_token, session_id and structured_output. Every workflow
+  here previously wrote `${{ steps.<id>.outputs.result }}`, which is empty, so
+  the run reports were empty fenced blocks. This action reads the last `result`
+  event out of `execution_file` instead.
+
+inputs:
+  execution-file:
+    description: The `execution_file` output of the claude-code-action step.
+    required: true
+  title:
+    description: Heading for the summary block, e.g. "🐑 PR Shepherd Run".
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      # Values come in through env, never interpolated into the script: the
+      # agent's report is attacker-influenceable and `echo "${{ ... }}"` would
+      # be a command-injection sink.
+      env:
+        EXECUTION_FILE: ${{ inputs.execution-file }}
+        TITLE: ${{ inputs.title }}
+      run: |
+        set -uo pipefail
+        file="${EXECUTION_FILE:-}"
+        if [ -z "$file" ] || [ ! -f "$file" ]; then
+          file="${RUNNER_TEMP}/claude-execution-output.json"
+        fi
+        {
+          printf '## %s\n\n' "$TITLE"
+          if [ ! -f "$file" ]; then
+            printf 'No execution log was produced, so the agent has no report. '
+            printf 'Check the run log above.\n'
+          else
+            # Tilde fence: agent reports routinely contain ``` blocks, which
+            # would close a backtick fence early and mangle the summary.
+            printf '~~~~\n'
+            # The step runs without `set -e` so a summary can never fail a job;
+            # that means the extractor's own failure has to be caught here, or it
+            # silently re-creates the empty fence this action exists to fix.
+            python3 "$GITHUB_ACTION_PATH/extract_result.py" "$file" \
+              || printf 'Could not read the agent report from %s (see the step log).\n' "$file"
+            printf '\n~~~~\n'
+          fi
+        } >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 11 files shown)


cmungall:

No changes needed — this one was approved. Carried through the restack unchanged apart from the `show_full_output` removal described in #2260.

cmungall:
Finding 10 is right and it's the sharpest one in the whole review — thank you for the re-read.

`inputs.dry_run != 'true'` with a `type: boolean` input: GitHub coerces mismatched types to number, `true` → 1 and `'true'` → NaN, and every NaN comparison is false, so `!=` was always true and the privileged job ran regardless. Ticking **Dry run** did nothing.

Your framing of why it matters *now* is the part I'd have missed: while the workflow was passing v0 `direct_prompt:`, no prompt reached the agent, so dry-run was academic. After this PR a dispatch gets the real agent with a write-scoped App token and `--dangerously-skip-permissions` — and dry-run is precisely the escape hatch your finding #2 recommends for driving the first run by hand. It needed to work *before* this lands. Fixed with `&& !inputs.dry_run` (null on the schedule path, so cron is unaffected).

Also took carry-over 6: `scan_arba_issues.py` exits 1 for "no ARBA IDs found", which is a normal empty scan. Reddening the weekly run for that just teaches people to ignore red. The step now accepts 0 and 1 and still fails on anything else.

And carry-over 4, in #2260 where the summary step now lives: it ran `if: always()` while the handoff step is conditional on `create_issues`, so an intentional skip rendered as "no execution log was produced". The summary now mirrors the handoff condition.

On the remaining carry-overs: #2 (inert → live with skip-permissions) I'm leaving as-is for consistency with the other three scanners rather than tightening one in isolation — but with dry-run now actually working, driving the first run by hand is possible, which is what I'll do. #3 (90 → 55 min) is deliberate: an App installation token expires at 60 minutes and does not refresh, so the old 90 could only ever mean "finish the analysis, then 401 on every push." #9 is fair — the PR body's 6 → 2 was written against the original base.

cmungall:
Finding 13 is fixed **here**, not downstream — you were right that this is stack-wide and that queuing it two PRs later would have put the broken shape on `main` in between. `pr-shepherd`'s copy is on `main` already from #2246, so it needed fixing sooner rather than later regardless.

All five agent workflows now use `.github/actions/agent-run-summary`, which reads the last `result` event out of `execution_file`.

On the three options you laid out: `display_report: true` is genuinely one line, and I nearly took it — but it reports the agent's output and nothing about *why* a run failed. The extractor also surfaces `errors`, `num_turns`, `subtype` and cost, and says explicitly when there is no result event at all (the shape a usage limit takes). That diagnosability is the whole reason the step exists, so the extra code earns its place. Header-only was the option I'd have picked if the other two weren't available — you're right that an always-empty fence reads as "the agent produced nothing."

`show_full_output: true` is gone from all five, including arba. To answer your check directly: it was *newly set* by me here, not pre-existing, and you're right about the exposure — arba holds a write-scoped App token plus the Perplexity/FutureHouse/OpenAI keys, and registered secrets are masked but tool results are not. It should not have been added.

The PR body's "plus output capture so the next breakage is visible" now actually holds.

PR #2255, opened 2026-07-26
ci: retire the dragon-ai-agent mention workflows

cmungall:
Stacked on #2253 → #2249 → #2247 → #2246.

## The duplicate

`ai.yml` and `dragon-ai.yml` were **byte-identical** (`diff` returns nothing) and both matched the same keyword, so every qualifying event ran the agent **twice**. You can see it in the run list: "Dragon AI Agent GitHub Mentions" appears in pairs, same timestamp. Consolidated into one workflow; `dragon-ai.yml` is deleted.

## The trigger

Now `@ai4c-agent please ...`, with `@dragon-ai-agent please ...` kept as a legacy alias so old habits and older issue threads still work. It is a **text keyword, not an account** — the `dragon-ai-agent` machine account is retired, and GitHub Apps cannot be assignees, so assignment-based dispatch is not available and is not used.

## Hardening

- ai4c-agent App token; `persist-credentials: false`; `gh auth setup-git`
- **allowlist read from the default branch, not the PR ref.** `.github/ai-controllers.json` was read from the checked-out PR ref, so on a `pull_request` event a proposer with push access could add themselves to the allowlist *on their own branch* and self-authorize. The mention text still comes from the event payload; only the *authorization* has to be trusted.
- **code spans stripped before matching**, so documenting the trigger keyword (as this PR body does) no longer fires it
- request text and event values passed via `env:` + `printf`, never `echo "${{ ... }}"` — the old version did `echo "${{ ...prompt }}" > prompt.txt`, which breaks on ordinary punctuation and allows command injection
- untrusted-input guardrail in the prompt; model from `agent-config.yaml`

## It also had no credential

The old workflow set `ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}` — a secret this repo **does not have** (`gh secret list` has no such entry), so the agent ran with an empty credential. Switched to `CLAUDE_CODE_OAUTH_TOKEN`, plus the deep-research provider keys so `just deep-research` works in-run.

## Verification

```
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ai.yml'))"
uv run pytest tests/test_agent_config.py -q   # 10 passed
```


cmungall:

All three accepted, and the third was the serious one:
- Gate and payload could resolve to different occurrences. Code spans are now blanked with **equal-length** filler and the payload is sliced from the original at `gateMatch.index`, so they cannot disagree by construction.
- `(.*)` -> `([\s\S]*)` so a fenced block inside a request survives.
- Blanking creating a match absent from the original — `@ai4c-agent \`v2\` please rerun` — would have dispatched a bypassPermissions agent holding the App token and every provider key with an **empty request**. A non-empty request is now part of the gate.

Verified by running the workflow's own regex over all six cases; the doc-span-then-prose case now extracts the real request rather than the documented one.

PR #2256, opened 2026-07-26
Review human ACAP2: the ARF6 GAP whose substrate GO can no longer name

cmungall:
## ACAP2 (Q15057) — human ARF6 GAP, centaurin-β2

All 20 GOA annotations reviewed. Two new annotations proposed, three MODIFYed, three
marked over-annotated, one kept non-core, the rest accepted.

### What ACAP2 is

An ARF6-preferring GTPase-activating protein whose catalysis is gated by membrane lipid
(PIP2 + phosphatidic acid) and whose recruitment is gated by GTP-loaded RAB35 binding to
its terminal ankyrin repeat. Two separable modules — ArfGAP domain 399–520 with catalytic
Arg442, ANK repeats 640–735 — composing into one coincidence-detection mechanism: RAB35
delivers ACAP2 to a membrane, the local phosphoinositide licenses it, ARF6 is switched off.

### Finding 1 — the process GOA never recorded

GOA gives ACAP2 four biological processes (actin cytoskeleton organization IBA, actin
filament-based process IDA, endocytic recycling ISS, cellular response to NGF ISS) and
**not one of them says ACAP2 switches ARF6 off**, which is the entirety of its
characterised biology. `GO:0032013 negative regulation of ARF protein signal transduction`
exists, is not obsolete, and is unused. Proposed as **NEW, IMP** from PMID:11062263 — in
human HeLa cells ACAP2 suppressed AlF4-induced ARF6-dependent protrusions, halving the
protruding fraction, GAP-dependently. IMP rather than IDA deliberately: the cellular
evidence is overexpression plus a GAP-dead point mutant. The in vitro GAP assay in the
same paper is the IDA, and it supports the MF row, not this one.

### Finding 2 — GO can no longer name a GAP's substrate

`GO:0008060 "ARF GTPase activator activity"` has been **merged into `GO:0005096`**, together
with the eight other substrate-specific GAP terms (GO:0005097 Rab, GO:0005098 Ran,
GO:0005099 Ras, GO:0005100 Rho, GO:0005101, GO:0017123, GO:0030675, GO:0046582) — all nine
appear as `secondaryIds` of GO:0005096, whose only child is GO:1902773 GTPase activator
*complex*. The same has happened on the binding side: `GO:0017137 Rab GTPase binding` is
now a secondary id of `GO:0031267 small GTPase binding`, which has no children at all.

So ACAP2's defining specificity — ARF6 in preference to ARF1/ARF5, exclusive recognition
of RAB35 among ~60 mammalian Rabs (PMID:25694427), and explicitly *no* RAB35 GAP activity
— is **inexpressible as a term**. The three `GO:0005096` rows and the `GO:0031267` row are
therefore ACCEPTed as already maximal, with the recommendation that specificity be recorded
via `has_input` extensions (ARF6 = UniProtKB:P62330) or a GO-CAM. This is a deliberate GO
design decision, so no proposal is made to resurrect the obsoleted terms — it is filed as an
`ONTOLOGY`/`CURATION` knowledge gap and a question for the ontology editors instead.

### Finding 3 — WITH/FROM resolution: one family-boundary artifact

Every accession in column 11 resolved. The clean ones first:

| id | identity | verdict |
|---|---|---|
| `UniProtKB:Q5FVC7` | **rat Acap2** | true ortholog → the three ISS transfers are legitimate |
| `UniProtKB:Q6ZQK5` | **mouse Acap2** | true ortholog → Compara projection legitimate |
| `RGD:1562939` | rat Acap2 | ortholog |
| `UniProtKB:Q15286` | **RAB35** | real, crystallised partner (PDB 6IF3) |
| `UniProtKB:Q96P50` | **ACAP3** | paralog — but an IPI *interaction*, not a transfer |
| `dictyBase:DDB_G0279649` / `DDB_G0276395` | Q54WI0, Q551Q8 | verified BAR-PH-ArfGAP-ANK / PH-ArfGAP — genuine ACAP-like |
| `SGD:S000002932` · `PomBase:SPBC17G9.08c` · `AT5G13300` · `AT5G61980` | yeast AGE1, pombe csx2, Arabidopsis AGD3/VAN3, AGD1 | distant but genuine ArfGAPs |
| `UniProtKB:Q15057` | ACAP2 itself | self-referential IBA = PAN-GO curator marking the function core, `NO_FAILURE_CORE` |

Two that are not clean:

- **`MGI:MGI:2153589` is mouse Acap3, the paralog** — not the mouse Acap2 ortholog — in the
  WITH/FROM of the `GO:0005096` IBA. Harmless for a family-level IBA of a function all ACAPs
  share, but recorded.
- **`FB:FBgn0004133` is Drosophila `blow` (blown fuse), which is not an ArfGAP at all.**
  A1Z714/P91678 has exactly one annotated domain — a PH domain at 204–307 — with no BAR
  domain, no ArfGAP catalytic domain, no ankyrin repeats, and UniProt records no PANTHER
  cross-reference for it. PANTHER nonetheless assigns it to PTHR23180 (CENTAURIN/ARF) as
  subfamily SF399, so blow's experimental `GO:0030036` annotation — which comes from
  **myoblast fusion** — propagates into ACAP2's support set. Whatever blow does to actin, it
  cannot be doing it as an ARF GAP.

And the node itself is far deeper than the ACAP1/2/3 clade: querying QuickGO for annotations
carrying `PANTHER:PTN001142372` returns the identical four-term set on *Brassica napus*,
*Cucumis sativus*, *Medicago truncatula*, *Theobroma cacao*, *Zostera marina* and *Mimulus
guttatus* genes, on yeast AGE1, and on *Xenopus* acap2.L. **PTN001142372 predates the
plant/fungal/animal split.** GAP activity survives that depth; a specific actin process does
not.

### Actions

| term | evidence | action | why |
|---|---|---|---|
| GO:0005096 ×3 | IBA, IEA, IDA | ACCEPT | core; already maximally specific (see Finding 2) |
| GO:0031267 | IEA (mouse ortholog) | ACCEPT | clean transfer, redundant with human structure |
| GO:0005886 ×3, GO:0010008 ×3, GO:0001726, GO:0032456 | — | ACCEPT | core locations / process |
| GO:0030036 | IBA | **MODIFY → GO:0032956** | ACAP2 regulates actin from upstream; support set is a pan-eukaryotic node plus `blow` |
| GO:0030029 | IDA | **MODIFY → GO:0032956** | grandparent placeholder, wrong role; also flagged that IMP fits the overexpression evidence better than IDA |
| GO:0005515 (RAB35) | IPI | **MODIFY → GO:0031267 + GO:0030742** | a 1.50 Å crystal structure recorded as "protein binding"; GTP-dependence is what makes it an effector |
| GO:0005515 (ACAP3) ×2 | IPI | MARK_AS_OVER_ANNOTATED | pair is probably real (BioPlex + OpenCell, UniProt NbExp=4) but functionally untested; AP-MS cannot license GO:0046982 |
| GO:0016020 | HDA | MARK_AS_OVER_ANNOTATED | the NK membrane-proteome protocol *removed peripheral membrane proteins*, which is exactly what ACAP2 is; authors concede ~40% of 1843 hits were plausible membrane proteins |
| GO:0005737 | IEA (BAR signature) | KEEP_AS_NON_CORE | true (largely cytosolic) but BAR→cytoplasm is a poor inference |
| GO:1990090 | ISS | KEEP_AS_NON_CORE | correctly transferred from rat PC12 work, but cell-type context; the same module runs in phagocytosis and oligodendrocyte differentiation |
| GO:0032013 | IMP | **NEW** | Finding 1 |
| GO:0035091 | IDA | **NEW** | no lipid-binding annotation despite strictly lipid-gated catalysis; parent term used because the species cannot be read from an abstract-only source |

### Deliberately not proposed

- **Phagocytosis.** PMID:22045739 shows GTP-RAB35-dependent ACAP2 recruitment to the
  phagocytic cup, but the abstract says only "in macrophages" and the cache is abstract-only,
  so the species is not stated and an ISS cannot be justified. Filed as a question.
- **GO:0043065 positive regulation of apoptotic process.** PMID:25853217 is a clean human
  knockdown result, but single, unreplicated, and its own authors note ACAP2 is not
  caspase-cleaved, so the tCNT-1 mechanism does not carry over. Filed as a knowledge gap and
  an experiment.

### On the affinage record

`gates_passed: True`, faith 100%. Its value was the **reference list, not the prose** — it
surfaced seven PMIDs, none of which is cited by any GOA row, four of which materially shaped
this review. Marked `LOW_QUALITY` because its prose presents Asn-610/Asn-691 as residues
"within the minimal Rab35-binding domain of ACAP2" without noting that PMID:25694427's
numbering does not correspond to the human sequence (human ANK repeats span 640–735; the
human RAB35-contact mutagenesis is Asp-721/Arg-727/Met-731/Asp-756). No residue-level or
mechanistic claim in this review is sourced from it; all domain assignments come from the
UniProt FT table.

### Verification

- `checkquotes.py`: 82 quotes, **0 problems**
- `just validate human ACAP2`: **✓ Valid** (1 warning — no annotation cites the affinage file, which is intentional per the campaign rule against quoting provider prose for mechanistic claims; the record is adjudicated in `references`)
- `cache/go/terms.csv`: restored from `origin/main` after the validator dropped `GO:0072534`; only `GO:0032013` appended


cmungall:
## Review addressed — thanks, the 🟡 was a fair hit

You were right that the review diagnosed a machine-readability problem and then left it in prose. Both mechanisms you pointed at are now used, and all five 🔵 suggestions are addressed. Commits: `08a5d0f` (review response), plus two merges of `origin/main`.

### 🟡 ARF6 is now machine-readable, in both places

`core_functions[0]`:

```yaml
  substrates:
  - id: UniProtKB:P62330
    label: ARF6
```

And on the `GO:0005096` IDA row — the `extensions` slot does work, with two gotchas worth recording for whoever uses it next: it belongs on `ExistingAnnotation`, not on `review` (the latter rejects it as an additional property), and `predicate` is a bare CURIE string, not an inlined `Term`:

```yaml
  extensions:
  - predicate: RO:0002233
    term:
      id: UniProtKB:P62330
      label: ARF6
```

`RO:0002233` verified against OLS as `has input`. The `GO:0005096` IDA summary now points at the extension rather than recommending one, and closes with your framing: without it, a downstream consumer sees only "GTPase activator activity" and loses the one fact the assay was designed to establish.

### 🔵 1 — why recycling is participation but actin is regulation

Added to the `GO:0032456` summary. The distinction is step-count and sign:

- **Actin:** ACAP2 never touches the cytoskeleton. ARF6-GTP recruits effectors, those effectors remodel actin. Two steps removed, and the observable is modulation of rate and extent — a regulator.
- **Recycling:** the ARF6 nucleotide cycle is not upstream machinery, it is *part of* the trafficking step. Hydrolysis has to happen on the recycling tubule for the round to complete, and ACAP2 is the enzyme that performs it there. GEFs and GAPs that turn over the GTPase of a trafficking step are conventionally `involved_in` that step, the way a Rab GAP is treated as part of its transport process.
- **Sign:** you make the point yourself — ACAP1 is *required* for TfR recycling, so a signed regulation term would be hard to justify, whereas the actin effect is unambiguously suppression.

### 🔵 2 — defending the binding term as a core molecular function

Added explicitly rather than left implicit. It is kept separate because it is a structurally delimited module (601–770, crystallised independently of the catalytic domain) whose defining property — nucleotide-state-specific recognition of one Rab out of ~60 — is not a property of the ArfGAP domain and would vanish if folded in. And the entry now names what GO actually lacks: a non-binding term for effector recognition, "binds the active conformation of a specific GTPase and thereby localises an activity". That is the same ontology gap as the missing substrate-specific GAP terms, seen from the other side.

### 🔵 3 and 4 — the claims now have an artifact, and it found more than expected

New `genes/human/ACAP2/ACAP2-bioinformatics/resolve_withfrom.py` + `RESULTS.md`. It resolves **all 22** WITH/FROM tokens from primary APIs — UniProt, Alliance of Genome Resources, WormBase, FlyBase, PANTHER, QuickGO — with **no `UNRESOLVED` rows**, reports domain architectures, and hardcodes nothing (a failed lookup is reported as `UNRESOLVED`, never guessed). Stdlib only, so it runs under bare `python3`. That covers both the `blow` architecture and `MGI:MGI:2153589 → Q6NXL5 Acap3` provenance; the four quotes now cited from `RESULTS.md` are grep-verified.

Three things fell out of building it:

**(a) Your SF399 wrinkle is now the primary form of the criticism.** You were right that it is sharper, and it is in the review body now: SF399 is a real ACAP subfamily whose other reviewed member is *C. elegans* `cnt-1` (Q9XXH8, 826 aa, Arf-GAP + ANK + PH) — a protein this same review treats as ortholog-grade support elsewhere. So a PH-domain-only protein has been placed *inside* a genuine ACAP subfamily, and PANTHER names that subfamily after the member lacking the catalytic domain. Also: **all four** `blow` entries are PH-only, not just the two I checked — A1Z714 204–307, P91678 204–307, E1JGZ4 197–300, Q8MSU1 92–195.

**(b) A check I should have run in round one: what evidence does each source hold for the term it *donates*?** IBA WITH/FROM lists experimentally annotated members by construction, so this is worth asking rather than assuming:

| source | donated term | source's own evidence |
|---|---|---|
| rat Acap2 Q5FVC7 | GO:0010008 / GO:0032456 / GO:1990090 | IDA / **IMP** / **IDA** |
| mouse Acap2 Q6ZQK5 | GO:0031267 | **IPI** |
| mouse Acap3 Q6NXL5 | GO:0005096 | **IMP** |
| worm cnt-1 Q9XXH8 | GO:0005886 / GO:0010008 | IDA×3 + EXP×2 / IDA×2 |
| yeast AGE1 Q04412 | GO:0005096 | **IDA** |
| Arabidopsis AGD3 Q5W7F2 | GO:0005096 | **IDA** |
| Arabidopsis AGD1 Q9FIT8 | GO:0005886 | **IDA** |
| pombe csx2 Q9UUE2 | GO:0005886 | **IDA** |
| Dicty Q54WI0 | GO:0005096 / GO:0005886 / GO:0030036 | IDA / IDA / **IMP + IGI** |
| Dicty Q551Q8 | GO:0030036 | **IGI** |
| fly blow A1Z714 | GO:0030036 | **IMP** |

Every source carries wet-lab evidence for what it donates. None of these transfers is an inference recycling an inference, and the `propagation_review` comments now record the codes. This **sharpens** the `blow` objection rather than softening it: blow's actin annotation is a real experimental result, and the objection is now purely architectural — a protein with no ArfGAP domain cannot be producing that phenotype the way ACAP2 does. It does soften the mouse-Acap3 note, and I've said so: Q6NXL5 carries `GO:0005096` **IMP** in its own right, so it is a competent donor; the only observation left is that the ortholog is absent from that support set.

**(c) A curation asymmetry, independently verified.** PMID:11062263 assayed ACAP1 and ACAP2 in the same figures and found ACAP1 *more* potent ("ACAP1 was most effective at inhibiting protrusions"), yet QuickGO returns 3 annotations from it for ACAP2 (Q15057) and **0** for ACAP1 (Q15027). So ACAP2's three IDAs are not over-calls — they are the recorded half of a symmetric experiment, and ACAP1 is under-annotated from the same source. Recorded in the `GO:0005096` IDA row and in the notes. (The parallel ACAP1 review, #2251, reached this independently.)

### 🔵 5 — GO:0035091 supporting set separated

`supported_by` now holds only PMID:25853217, the one source with a binding measurement. `PMID:11062263` and the UniProt `ACTIVITY REGULATION` block moved to `additional_reference_ids` and to prose that says why: a lipid that gates an enzyme need not be bound by it, so lipid-dependent catalysis motivates the annotation without supporting it.

### Conflict resolution

`mergeable: CONFLICTING` appeared while I was working, from `interpro/panther/PTHR23180/PTHR23180-metadata.yaml` — an add/add on a machine-fetched cache, since the ACAP1 branch created the same PANTHER family. The two copies are byte-identical apart from `_fetch_info.fetched_date`, so main's copy is kept and this branch now adds nothing to that file. Verified after each of the two merges:

- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → **empty**. (It was *not* empty after the first merge — four rows from #2250 were missing simply because that PR landed after my fetch, so I merged again. Worth flagging that the two-dot diff hides this.) The only delta is the single `GO:0032013` append.
- `PTHR23180-entries.csv` identical to main's and still carrying the `Q9XXH8`/SF399 row the review cites.
- `PMID_11062263.md` identical to main's and still `full_text_available: true`.

### Verification

- `checkquotes.py`: **89 quotes, 0 problems** (up from 82; and it does check `file:` quotes — it caught a stale `RESULTS.md` line after I fixed the MGI resolver, which is how the quote got corrected)
- `just validate human ACAP2`: **✓ Valid**, one warning left standing — no annotation cites the affinage file, which is intentional under the campaign rule against quoting provider prose for mechanistic claims. Given this provider put Asn-610/Asn-691 on the wrong sequence (thank you for counting the residues — Leu and Cys, exactly), citing it for anything mechanical would have been the wrong call. It is adjudicated in `references` instead.
- `git diff origin/main HEAD --stat` → 17 files, all insertions, no deletions.

cmungall:
## Three residual nits fixed — `61f2ad3`

None were blocking, but nit 2 concerned a figure I quote as `supporting_text`, so all three are now landed.

**1. Q96P50/ACAP3 understated in the generated table.** Correct diagnosis: the symbol branch matched first and the `ipi_only` branch was also gated on the *absence* of an ArfGAP domain, which ACAP3 obviously has — so it could never reach it. Both fixed; it now prints:

> `UniProtKB:Q96P50` -> **ACAP3** - PARALOG (not the ACAP2 ortholog); cited by an IPI row, so here it is a measured INTERACTION PARTNER rather than a phylogenetic donor

**2. You were right to distrust a round 1000, and the script now checks instead of asserting.** Two things resolve it:

- QuickGO's `pageInfo` for this query reports **5 pages of 200**, bracketing the total at 800–1000 — consistent with 1000 being a real count rather than a truncation point.
- There is no 1000-hit cap on `numberOfHits`: `goId=GO:0005515&limit=1` returns `numberOfHits: 8106099`.
- `limit=200` is in fact accepted (200 results returned, no warning), so that part of the nit does not bite either.

`RESULTS.md` now carries the cross-check inline rather than leaving the reader to trust the number:

> `- QuickGO reports 1000 annotations citing this node. pageInfo reports 5 pages of 200, bracketing the total at 800-1000, so this is a real count and not a service cap.`

The quoted `supporting_text` is unchanged and still verbatim, since the new text extends that sentence rather than replacing it. And as you note, the pan-eukaryotic conclusion rests on the symbol list — plants, a charophyte alga, fungi, *Xenopus* — not on the count.

**3. Partial tallies can no longer read as totals.** `source_own_evidence` now emits `[PARTIAL - only the first N of M tallied]` whenever `n_hits` exceeds the page size. No current row triggers it (max n=10), but a re-run against a better-annotated source would now say so rather than quietly under-report.

Verified after: `checkquotes` 89 quotes / 0 problems, `just validate human ACAP2` ✓ Valid, and `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` still empty.

Thanks for the re-review — the substrate slot, the SF399 wrinkle and the recycling-vs-actin question all made this materially better, and the residual nits were fair.

PR #2257, opened 2026-07-26
Review human ACAP3: a PANTHER family that swept in a non-ArfGAP, and an ontology merge that hides Arf6 specificity

cmungall:
## ACAP3 (Q96P50) — Arf6 GAP, brain-enhanced, and annotated entirely by inference

ACAP3's GOA record is ten rows: six IBAs, two InterPro IEAs, two high-throughput
`protein binding` IPIs. **Not one derives from an experiment**, although four functional
papers exist — two on the mouse orthologue and two on the human protein in tumour cells —
and the UniProt entry cites none of them. Its only FUNCTION line is an `ECO:0000305`
curator inference, and it has no SUBCELLULAR LOCATION block at all.

`PE 1: Evidence at protein level`, so the protein is detected; what is missing is
functional data *in the record*, not evidence of existence.

### Every WITH/FROM accession resolved

| WITH/FROM | resolves to | relation to ACAP3 |
|---|---|---|
| `MGI:MGI:2153589` | mouse **Acap3** | **1:1 orthologue** |
| `UniProtKB:Q15057` | human **ACAP2** | **paralogue** |
| `RGD:1562939` | rat **Acap2** | **paralogue** |
| `WB:WBGene00000565` | *C. elegans* cnt-1 | single-copy centaurin-beta |
| `PomBase:SPBC17G9.08c` | *S. pombe* cnt5 | fungal ArfGAP |
| `dictyBase:DDB_G0279649` / `DDB_G0276395` | *Dicty* Q54WI0 / Q551Q8 | amoebal ACAPs |
| `SGD:S000002932` | yeast **AGE1** | ArfGAP, but ASAP-type not ACAP |
| `AGI_LocusCode:AT5G13300` / `AT5G61980` | *Arabidopsis* AGD3 / AGD1 | plant ArfGAPs |
| `FB:FBgn0004133` | *Drosophila* **blow** | **not an ArfGAP** (see below) |

**The six IBAs split cleanly by PANTHER node, and that split is the most useful structure
in the record.** `PTN002754173` is the ACAP3 subfamily node; its WITH/FROM contains nothing
but mouse Acap3, and it supplies `neuron migration` and `regulation of neuron projection
development` — clean orthologue transfers of rescue-validated, GAP-dependent in vivo
phenotypes. `PTN001142372` is the deep centaurin/ArfGAP superfamily node spanning plants,
fungi, amoebae and animals; it supplies the other four, which need per-donor checking.

### Finding 1 — PTHR23180 has swept in a protein that is not an ArfGAP

`FB:FBgn0004133` (*Drosophila* `blow`, blown fuse) is a WITH/FROM donor for
`GO:0030036 actin cytoskeleton organization`. Its entire InterPro content is
**IPR011993 + IPR001849 — a PH domain and nothing else.** No ACAP1/2/3-like signature
(IPR045258), no ArfGAP domain (IPR001164), no BAR domain, no ankyrin repeats. It cannot be
an Arf GAP, because it has no GAP domain. PTHR23180 (CENTAURIN/ARF) clustered it in on the
PH domain alone, and its `GO:0030036` is an IMP from myoblast fusion acting through
WASp–Vrp1 complex stability — a mechanism with no ArfGAP component.

The reciprocal cost of the same mis-clustering is the clearer error: **`blow` itself now
carries `GO:0005096 GTPase activator activity` by IBA (GO_REF:0000033)** despite having no
GAP domain and no zinc finger. Raised as a suggested question for the PANTHER/PAINT
curators.

The ACAP3 annotation nevertheless survives, so this is **`KEEP_AS_NON_CORE`, not `REMOVE`**:
the other two donors are the *Dictyostelium* ACAPs, which do carry the full
BAR–PH–ArfGAP–ANK architecture and do have genuine experimental actin data
(Q54WI0: `GO:0030036` IMP + IGI, `GO:0031941` filamentous actin IDA, `GO:0051489` IMP), and
Arf6 is itself a cortical actin regulator.

A reproducible audit of all eleven donors is in `ACAP3-bioinformatics/` — it resolves each
identifier to a UniProt accession and tabulates InterPro content live from the REST API.
Ten of eleven are genuine ArfGAPs; `blow` is the one exception.

### Finding 2 — the ontology can no longer say which GTPase a GAP acts on

The single most specific fact about ACAP3 is that its GAP activity is **specific to Arf6**.
That cannot be recorded as a molecular function. **`GO:0008060 ARF GTPase activator
activity` has been MERGED into `GO:0005096`** — QuickGO
(`/ontology/go/terms/GO:0005096/complete`) lists it among the `secondaryIds`, together with
every other substrate-specific GAP term (`GO:0005097` Rab, `GO:0005098` Ras, `GO:0005100`
Rho, `GO:0005099`, `GO:0005101`, `GO:0017123`, `GO:0030675`, `GO:0046582`). OLS reports
GO:0008060 as obsolete with no label, which alone would not distinguish a merge from a
plain obsoletion — the `secondaryIds` list is what identifies it.

So `GO:0005096` is now the terminal GAP term for every substrate, **no `MODIFY` is
possible**, and ASAP1 (Arf1/Arf5-preferring) and ACAP1/2/3 (Arf6-preferring) all carry the
identical MF term. Recorded as an `ONTOLOGY` knowledge gap; `GO:0032012 regulation of ARF
protein signal transduction` is proposed as the partial workaround, since the BP aspect can
still name the family.

### Finding 3 — `endosome membrane` was right for the wrong reason

`GO:0010008` has no orthologue donor. Mouse Acap3 is **absent** from its WITH/FROM although
it is present for both neuronal terms — PAINT had no orthologue evidence and fell back to
the superfamily node. The only mammalian donor is rat Acap2, a **paralogue**
(`GO:0010008` IDA, PMID:23572513, in the endocytic-recycling context ACAP2 is specialised
for), and the orthologue's own recorded location is the growth cone.

The compartment is nonetheless correct, and PMID:41520057 has since supplied human
evidence: ACAP3 impairs EGFR recycling and accelerates lysosomal EGFR degradation, GAP-
dependently. `ACCEPT`, with the paralogue-only propagation route recorded in
`propagation_review` rather than buried.

### All 14 rows

| Term | Ev | Action |
|---|---|---|
| GO:0005886 plasma membrane | IBA | ACCEPT — IDA in four lineages, all full-architecture ACAPs |
| GO:0001764 neuron migration | IBA | ACCEPT — orthologue-only node, in vivo rescue |
| GO:0005096 GTPase activator activity | IBA | ACCEPT — no more specific term exists |
| GO:0010008 endosome membrane | IBA | ACCEPT — right compartment, paralogue-only route |
| GO:0010975 reg. neuron projection development | IBA | **MODIFY** → GO:0010976 (direction was determined) |
| GO:0030036 actin cytoskeleton organization | IBA | **KEEP_AS_NON_CORE** — `blow` donor discounted |
| GO:0005096 GTPase activator activity | IEA | ACCEPT — canonical IPR001164 mapping |
| GO:0005737 cytoplasm | IEA | **MARK_AS_OVER_ANNOTATED** — inferred from the BAR domain |
| GO:0005515 protein binding ×2 | IPI | **MODIFY** → GO:0046982 protein heterodimerization activity |
| GO:0030426 growth cone | ISS | **NEW** — mouse IDA, in UniProt but absent from GOA |
| GO:0032012 reg. of ARF protein signal transduction | ISS | **NEW** — the only way to state Arf specificity |
| GO:0001920 neg. reg. of receptor recycling | IMP | **NEW** — PMID:41520057, human |
| GO:0042059 neg. reg. of EGFR signaling pathway | IMP | **NEW** — PMID:41520057, human |

**On the `cytoplasm` IEA:** a BAR domain is a membrane-binding and curvature-generating
module — in ACAP1 the BAR–PH tandem dimerises and assembles a lattice on tubulated
membrane. What it predicts is peripheral membrane association, not cytoplasmic residence.
The mapping discards the informative half of the signal for a near-root term already
subsumed by the two specific compartments. Not `REMOVE`d, because the assertion is
technically true.

**On the two `protein binding` rows:** both name ACAP2, ACAP3's own paralogue. Not a lone
screen hit — replicated across two orthogonal proteome-scale methods (BioPlex 3.0 AP-MS and
OpenCell *endogenous* split-tag IP-MS), reciprocal in ACAP2's GOA record from the same two
papers, and curated by UniProt as ACAP3's only interaction with `NbExp=4`. The shared
BAR–PH tandem is a dimerisation module, so `GO:0046982` is the parsimonious upgrade. Stated
explicitly as provisional: AP-MS gives co-complex membership, not stoichiometry, and the
dimer has been demonstrated for ACAP1, not for this pair. Measuring it is in
`suggested_experiments`.

**One interpretive point carried through the whole review:** ACAP3 is not an Arf6 switch-off
device. Neurite outgrowth in depleted neurons was rescued by a *fast-cycling* Arf6 mutant
but not by GTP-locked or GDP-locked Arf6, so the requirement is for Arf6 **cycling**. That
is why `GO:0032012` is proposed at the non-directional parent rather than as a negative-
regulation child, and why the unidentified partner GEF is recorded as a `RESIDUAL_SUBGAP`
knowledge gap.

### Verification

- `checkquotes.py`: 34 quotes, **0 problems**
- `just validate human ACAP3`: **✓ Valid**, 1 warning
- `cache/go/terms.csv`: one line added (`GO:0001920`), none deleted
- rebased onto `origin/main` @ e8aabb481

The remaining warning ("No annotations reference available deep research files") is
deliberate. Affinage's record is sound and it surfaced PMID:41520057 and PMID:39098591,
which UniProt does not cite — that credit is recorded in its `reference_review`. But no
affinage sentence is quoted as `supporting_text` anywhere, per the campaign rule against
using provider prose for mechanistic claims; every mechanistic quote here comes from a PMID
or from a single physical line of the UniProt record.


cmungall:
Thanks — both blocking points are correct, and they were the same mistake made twice. I've taken all six items, including the four non-blocking ones, and separately found two errors of my own while re-checking. Also merged `origin/main` (four genes landed since I branched); the PR is `MERGEABLE` again.

## 🟡 1 — `GO:0046982` withdrawn from both rows

You're right, and the specific sentence that persuaded me is *"`MODIFY` is an instruction to a curator to replace the term"*. I had convinced myself that stating the caveat prominently discharged the obligation. It doesn't — the caveat lives in `review.summary`, which a curator acting on `proposed_replacement_terms` need never read.

The three reasons all hold: neither AP-MS nor split-tag IP-MS measures stoichiometry; co-complex membership is equally consistent with a direct dimer, with co-residence in a larger assembly, and with two BAR proteins sharing a membrane surface; and the structural precedent is homotypic — PMID:31291238 solves ACAP1 dimerising with *itself*. Your point that BAR-domain paralogues co-purifying is a known route to this signal without a direct dimer is the one I should have raised against myself.

Both rows are now `KEEP_AS_NON_CORE` retaining `GO:0005515`, with the partner, the reciprocity and the two-method replication argument in `review.reason`, and the heterodimer left where you say it belongs.

One thing I added while writing this up, because it changed how bad the outcome felt: **staying at `GO:0005515` is much less lossy here than for a partnerless binding row, because the partner identity travels in the WITH/FROM field.** This annotation does not say "binds something", it says "binds `UniProtKB:Q15057`". So what's actually missing isn't a term for *this* annotation — it's a term for "physically associates with a named paralogue" that doesn't also assert a stoichiometry. GO has no such term between bare `GO:0005515` and `GO:0046982`. That gap is now recorded explicitly, with `GO:0032403 protein-containing complex binding` named as the term to use if the measurement returns a larger assembly rather than a dimer.

I'll note the tension you flagged with the repo's standing preference against bare `protein binding`, and I think your reading is right: that guidance says *find a more informative term the evidence supports*, and it is silent on what to do when GO offers no honest intermediate. This case is evidence that the guidance needs that clause.

## 🟡 2 — third `core_function` removed

Accepted without reservation, and it follows mechanically from item 1 — a core function whose `molecular_function` was `GO:0046982` cannot survive the withdrawal of `GO:0046982`.

Your framing is the part worth keeping: *"Being the only known interaction is not the same as being a core function."* That is exactly the substitution I made, and I made it explicitly in the entry's own text, which should have been the tell.

The `knowledge_gaps` block is preserved as a fourth top-level entry, expanded to state what is untested (stoichiometry, whether the BAR domains mediate it, where in the cell, whether it changes Arf6 specificity or curvature generation).

You were also right about `locations: GO:0005886` — neither dataset is spatial, so that location had no evidence at all behind it. Gone with the entry.

## 🔵 3 — `GO:0010976` and where the fix can land

Good catch, and a general point I hadn't internalised: an IBA term is constrained by its ancestral node, which is constrained by the descendant experimental annotations, so PAINT cannot refine this row unilaterally.

I've kept `MODIFY` — the term genuinely should change — but the row now says so explicitly, and I took your suggestion of a `suggested_questions` entry addressed to MGI. One question now covers both upstream fixes:

> Will MGI refine two mouse Acap3 annotations at source? Both fixes this review proposes for the human gene are blocked upstream: `GO:0010975` should be `GO:0010976` (PMID:27330119 determines the direction), and `GO:0032012` should be asserted by IMP from the same paper. An IBA cannot be refined past its ancestral node, so until the mouse annotations change, the human MODIFY and the human ISS are both requests PAINT alone cannot honour.

## 🔵 4 — `GO:0032012` ISS

Your two-step route is better than mine and is now recorded in the row: assert `GO:0032012` on the mouse record by IMP from PMID:27330119, then ISS across, which fixes the gap at source instead of only at the human end. I've kept the human ISS row since this is the human review and that's where we want the term to end up, but the ordering is now stated.

## 🔵 5 — `GO:0008270` promoted to a `NEW` row

Agreed, and your reason is the right one: it rests on ACAP3's own sequence, which is a better basis than either orthology proposal.

Coded `ISM` with the UniProt record as reference, because the basis is the PROSITE ARFGAP profile call (PS50115 / ProRule PRU00288) and not a metal-binding measurement on ACAP3. I went looking for family-level experimental support and found the founding ArfGAP paper, PMID:8533093 — *"The GAP function required an intact zinc finger and additional amino-terminal residues."* That is mutational evidence that the motif is functionally required, and I've been careful to say it is **not** a demonstration that zinc is bound.

Deliberately left out of `core_functions`: zinc coordination here is a structural requirement of the ArfGAP fold, not a distinct activity of ACAP3.

I replaced the old question with a broader one, since the gap isn't ACAP3's: the `GO_REF:0000043` withdrawal appears to have dropped zinc binding across the whole ArfGAP family with nothing replacing it, and an `IPR001164` → `GO:0008270` InterPro2GO mapping would restore it in one step rather than gene by gene.

## 🔵 6 — donor-provenance PMIDs

You offered the notes-file route and I took the fuller one instead: **all ten are now in `references`**, each with a `reference_review` naming which donor annotation it supplies and stating that it is not evidence about ACAP3. Titles were read from the cached frontmatter, never written from memory.

The invariant is now script-checked rather than eyeballed — every `PMID:` string anywhere in the YAML, including inside `review.summary` and `propagation_review.comment`, is in `references` (20 cited, 25 references, 0 missing). Also asserted that every reference title matches its cached frontmatter exactly. The rule is documented in `ACAP3-notes.md` §8.

Two of the new titles turned out to be load-bearing for the `blow` argument: PMID:23264736 is *"Dictyostelium ACAP-A is an ArfGAP involved in cytokinesis, cell migration and actin cytoskeleton dynamics"* and PMID:20062541 is *"ACAP-A/B are ArfGAP homologs in dictyostelium…"*. So the donors that legitimately carry the actin term are named ArfGAPs in their own paper titles, which sharpens the contrast with `blow`. The Dicty donors are now labelled ACAP-A and ACAP-B rather than bare accessions.

## Three errors of my own, found while re-checking

Reporting these because two are the sort of thing that would otherwise survive.

**`source_entities` had silently drifted on half the IBA rows.** I built that list by hand. When I wrote a check that diffs it against the GOA WITH/FROM field, it failed on three of six rows: `GO:0005886` documented 4 of 7 donors, `GO:0005096` 3 of 6, and `GO:0010008` was missing the PANTHER node. The prose mentioned all of them, so nothing was *wrong* — but the structured field understated the audit. All six rows now cover every entry, verified by set difference.

Completing it turned up something worth having: **Arabidopsis AGD1 has a plasma-membrane IDA (PMID:28646092) but no experimental `GO:0005096` of its own** — UniProt calls it a *"Probable GTPase-activating protein (By similarity)"*. So it is a sound donor for the compartment it is actually cited for, and would have been a weak one for the activity. Recorded per-source.

**I overstated the `blow` argument.** The `RESULTS.md` caveats claimed blow is "far too short to accommodate" the missing modules. It is 644 aa against ACAP3's 834 — shorter, but comfortably long enough in principle to carry a ~120-residue ArfGAP domain. Length was never the argument and I shouldn't have offered it as one. Restated on what actually holds: four signatures absent simultaneously, while the one module blow does have is the one PANTHER could have clustered on. Also recorded all four isoform entries (A1Z714 644, P91678 644, E1JGZ4 637, Q8MSU1 532 aa), all giving the identical `IPR011993 + IPR001849` pair.

**The `cytoplasm` redundancy claim was half wrong.** I wrote that `GO:0005737` is "already subsumed by the two specific compartments". Checking the ancestor closure: `GO:0005737` *is* an ancestor of `GO:0010008 endosome membrane`, but is *not* an ancestor of `GO:0005886 plasma membrane`, whose path runs through cell periphery. So it is only half redundant, and the row now says which half. The `MARK_AS_OVER_ANNOTATED` still stands on the primary ground — BAR → cytoplasm is a non-sequitur — which is the part you endorsed.

## Verification

- `checkquotes.py`: **35 quotes, 0 problems**
- `just validate human ACAP3`: **✓ Valid**, 1 warning
- all 10 GOA rows still reviewed; actions now 5 ACCEPT / 3 KEEP_AS_NON_CORE / 1 MODIFY / 1 MARK_AS_OVER_ANNOTATED / 5 NEW
- `source_entities` ⊇ WITH/FROM for all 6 IBA rows (0 uncovered)
- every PMID cited in the YAML is in `references`; every reference title matches its cached frontmatter
- all bracketed quotes in `ACAP3-notes.md` re-verified verbatim, including across line wraps
- `cache/go/terms.csv`: **one line added** (`GO:0001920`), **zero deletions** vs `origin/main`

On the merge: two add/add conflicts, both machine-fetched caches, resolved without loss. `publications/PMID_25284369.md` — kept my side, which is the 254-line full-text version, over main's 89-line abstract-only one; quotes re-verified after the swap. `interpro/panther/PTHR23180/PTHR23180-metadata.yaml` — identical apart from `fetched_date`, took main's to keep the diff minimal for whoever hits this family next (ACAP1 and ACAP2 share it).

On your verification caveat — noted that the runner had neither `just` nor `uv` and couldn't reach QuickGO. The `GO:0008060` merge is reproducible with `curl -sS "https://www.ebi.ac.uk/QuickGO/services/ontology/go/terms/GO:0005096/complete"` and reading `secondaryIds`; `GO:0008060` appears there alongside `GO:0005097`/`0005098`/`0005099`/`0005100`/`0005101`/`0017123`/`0030675`/`0046582`. OLS reports it as obsolete with no label, which is why `secondaryIds` is the check that distinguishes a merge from a plain obsoletion.

The remaining validator warning ("No annotations reference available deep research files") is deliberate: affinage's record is sound and surfaced PMID:41520057 and PMID:39098591, which UniProt doesn't cite — credited in its `reference_review` — but no affinage sentence is quoted as `supporting_text` anywhere, per the campaign rule against provider prose for mechanistic claims.

PR #2259, opened 2026-07-26
fix: refresh stale derived artifacts and an off-vocabulary project tag

cmungall:
Three unit tests are **already failing on `main`** — this PR is not about my other open PRs, but they can't go green until this lands.

## Why main looks green

`just test` only runs when the `src` paths-filter matches (`src/**`, `tests/**`, `pyproject.toml`, `uv.lock`). Every recent main commit has been a gene review, so the suite hasn't actually run in a while. Any PR that touches `tests/` surfaces all three at once.

Confirmed against a pristine `origin/main` checkout (`3 failed, 151 passed`):

```
FAILED tests/test_bioreason_benchmark_policy.py::test_generated_sidecars_match_current_sources
FAILED tests/test_gogpt_compare_levels.py::test_committed_three_level_report_matches_current_reviews
FAILED tests/test_project_frontmatter.py::test_project_tags_are_controlled[STILBENE_CLEAVAGE_OXYGENASES.md]
```

## Two derived artifacts had drifted as gene reviews were edited

- Regenerated `reports/gogpt-comparison-levels.json` + its figure via `scripts/gogpt_compare_levels.py`. Post-review reference terms **2,705 → 2,706**; core-function terms **1,196 → 1,204** with overlap **341 → 342**.
- Regenerated the BIOREASON_COMPARISON sidecars. **Ordering matters:** the sidecars read the gogpt report, so the report has to be regenerated *first* or `benchmark-genes.csv` comes back stale on the next run. (I hit exactly that.)
- Synced the two numbers the supplement's three-level table quotes, and the hardcoded stats in `test_gogpt_compare_levels.py`.

**@cmungall worth a glance:** this edits two numbers in `article/supplemental-benchmark-details.md`. The percentages (11.7 / 9.5 / 3.8) are unchanged, so no claim in the prose moves — only the raw reference/overlap counts in the table.

## One vocabulary slip

`STILBENE_CLEAVAGE_OXYGENASES.md` carried `tags: [ENZYME_FAMILY]`, which is not in the controlled list and not in the renderer's `tag_order` either. It is the only project with that tag; the other 68 biology projects use `BIOLOGY_DOMAIN`, which it now uses. (Say the word if you'd rather add `ENZYME_FAMILY` to the vocabulary instead — that's a one-line alternative.)

## Follow-up worth considering (not done here)

The hardcoded stats in `test_gogpt_compare_levels.py` will drift again on the next batch of review edits — the test asserts exact counts derived from live review content. That's a design question for the benchmark, so I left it alone.

## Verification

```
uv run pytest tests/test_bioreason_benchmark_policy.py tests/test_gogpt_compare_levels.py tests/test_project_frontmatter.py -q
# 577 passed
```


cmungall:
You're right and I was wrong — the third percentage did move, and my PR body claimed it didn't.

341/8,871 = 3.844% → 3.8. 342/8,871 = **3.8553% → 3.9**. The extra overlapping term pushed it across `.85`, so the regenerated figure this commit ships already read 3.9% while the table above it, the prose, the slide, and the test's hardcoded expectation all still said 3.8. Fixed in fae533f — all four now read 3.9. As you note, no argument moves: 11.7/3.9 = 3.0, so the "3-fold gap" framing survives.

On the two non-blocking notes:

- **`tag_order` corroboration** — fair, that was a weak argument. `EVALUATION` and `ML_PREDICTIONS` are valid and also absent from `tag_order`, so absence from that list proves nothing. The real argument is just the controlled vocabulary in `VALID_TAGS` plus the fact that `ENZYME_FAMILY` was a single-use tag while 68 other biology projects use `BIOLOGY_DOMAIN`.
- **Recurrence** — agreed that adding `genes/**/*-ai-review.yaml` to the `src` paths-filter is the more honest fix, since review edits are what actually invalidate these artifacts. I'm not doing it here: it would run the full suite on every curation PR (those already run `validate-changed`, and the suite is the slow part of CI). Worth deciding deliberately rather than as a side effect of a stale-artifact fix — happy to open an issue for it.

Re-verified: `577 passed` across the three affected test files.

PR #2260, opened 2026-07-26
ci: finish the App-token migration; no PAT_FOR_PR reference remains

cmungall:
Last of the auth migration. Stacked on #2255 → #2253 → #2249 → #2247 → #2246.

## The three remaining `PAT_FOR_PR` workflows

`weekly-compliance`, `warm-reference-cache` and `generate-pages` move to the ai4c-agent App token, with `persist-credentials: false` and `gh auth setup-git`.

**`weekly-compliance` had the false-green shape** that cost dismech ~11 days: **no `github_token` at all**, so its `gh pr create` would fail silently while the job reported success and produced no PRs. It now gets a real token *and* output capture, so a silent failure is visible. Its model also comes from `agent-config.yaml` — which **empties `UNMIGRATED_MODEL_PINS`**: every agentic workflow now resolves its model centrally.

## `claude.yml` was the last write-capable agent on a persisted default token

It has `Bash(*)` and `contents: write`, and its checkout wrote the token into `.git/config` — precisely the incident shape (a Bash-enabled agent that can `cat .git/config`). Now on the App token with a non-persisting checkout.

## The rest

No agent attached, so not the incident vector, but no reason to leave a token on disk either:

- `claude-issue-summarize`, `claude-issue-triage`, `pypi-publish` never push → just stop persisting.
- `deploy-docs` **does** push (`mkdocs gh-deploy` shells out to `git push`), so it gets the agent treatment: `gh auth setup-git` serving `GH_TOKEN` at push time instead of a token in `.git/config`.

## Verification (the guide's final checklist, run locally)

```
$ grep -rn "secrets.PAT_FOR_PR" .github/
(empty)

$ grep -rL "persist-credentials: false" $(grep -rl "actions/checkout" .github/workflows/)
(empty)

$ grep -rnE "\-\-model .*claude-(opus|sonnet|haiku)-" .github/workflows/
(empty)

$ uv run pytest tests/test_agent_config.py -q   # 10 passed
$ just test-js                                   # 14 pass
$ python3 -c "import yaml, glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.y*ml')]"
```

**The `PAT_FOR_PR` secret can be deleted once this stack is merged** — that's the last step, and I'll flag it separately rather than doing it silently.

### DIFF
--- a/.github/actions/claude-issue-summarize-action/action.yml
+++ b/.github/actions/claude-issue-summarize-action/action.yml
@@ -19,11 +19,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Checkout repository code
-      uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
-
     - name: Create prompt file
       shell: bash
       run: |
--- a/.github/actions/claude-issue-triage-action/action.yml
+++ b/.github/actions/claude-issue-triage-action/action.yml
@@ -19,11 +19,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Checkout repository code
-      uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
-
     - name: Create prompt file
       shell: bash
       run: |
--- a/.github/agent-config.yaml
+++ b/.github/agent-config.yaml
@@ -68,3 +68,7 @@ workflows:
 
   pr-shepherd:
     model: claude-opus-5
+
+  # Weekly sweep that improves the lowest-scoring gene reviews.
+  weekly-compliance:
+    model: claude-opus-5
--- a/.github/copilot-setup-steps.yml
+++ b/.github/copilot-setup-steps.yml
@@ -28,6 +28,8 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
+      with:
+        persist-credentials: false
 
     - name: Install uv
       uses: astral-sh/setup-uv@v5
--- a/.github/workflows/claude-issue-summarize.yml
+++ b/.github/workflows/claude-issue-summarize.yml
@@ -15,6 +15,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
+        with:
+          persist-credentials: false
 
       - name: Run Claude Issue summarize
         uses: ./.github/actions/claude-issue-summarize-action
--- a/.github/workflows/claude-issue-triage.yml
+++ b/.github/workflows/claude-issue-triage.yml
@@ -15,6 +15,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
+        with:
+          persist-credentials: false
 
       - name: Run Claude Issue Triage
         uses: ./.github/actions/claude-issue-triage-action
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -116,14 +116,28 @@ jobs:
           #
           # `mcp_config:`, `model:` and `allowed_tools:` were dropped the same way
           # — none of them are v1 inputs. Everything now goes through claude_args,
-          # which is how v1 takes CLI flags.
+          # which is how v1 takes CLI flags. Both flags ACCUMULATE with tag mode's
+          # own values rather than replacing them, so the comment-update and CI
+          # MCP tools survive.
+          #
+          # The old file asked for `Bash(*),Edit,MultiEdit,Write`, but since the
+          # input was being discarded it never took effect — the agent has been
+          # running on tag mode's narrow set. Restoring that literally would grant
+          # arbitrary shell and blanket write for the first time, in the same PR
+          # that calls this workflow out as the incident shape, so it is scoped to
+          # what the gene-review protocol actually needs instead. Edit/Write are
+          # deliberately absent: tag mode runs `--permission-mode acceptEdits`,
+          # which already permits edits inside the workspace, and listing the
+          # tools would extend that to the whole runner.
           claude_args: |
             --model ${{ env.AGENT_MODEL }}
             --mcp-config '{"mcpServers": {"ols": {"command": "uvx", "args": ["ols-mcp"]}, "sequential-thinking": {"command": "npx", "args": ["-y", "@modelcontextprotocol/server-sequential-thinking"]}}}'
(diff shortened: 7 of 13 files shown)


cmungall:

Both blockers verified against the pinned action's `action.yml` and both were real:
- **`outputs.result` does not exist** (v1 declares execution_file, branch_name, github_token, session_id, structured_output). Every step summary in the repo — including the two that predate this migration — was an empty fence. New `.github/actions/agent-run-summary` reads the last `result` event out of `execution_file`; all seven agent workflows use it.
- **claude.yml was passing four inputs v1 does not accept.** So it had been running with no OLS MCP server, no tool allowlist, no instructions and the default model — which also made my 'claude.yml is centrally modelled' claim false. All four now go through `claude_args`.

Three tests now pin this class of bug, each verified to fail against a probe workflow: undeclared input, undeclared output, and your suggested reverse invariant (every claude-code-action workflow must appear in agent-config.yaml).

Also took both flags: `show_full_output` dropped everywhere, and `allowed_bots` dropped from claude.yml where it was dead config widening the actor gate.

cmungall:
Took both halves, because both were right.

**`Bash(git:*)`** is gone. You're right that it's a prefix rule permitting `git push --receive-pack='sh -c ...' ext::sh`, which is the exact RCE `scripts/git-push.sh` names in its own header (HackerOne #3556799) — and redundant, since tag mode already unions in `git add`/`git commit`/`git rm` plus the wrapper. Only the read-only verbs are listed now.

**The comment was the more useful half of the finding.** You're right that narrowing git doesn't confine the agent — `Bash(uv:*)` covers `uv run python -c`, `Bash(uvx:*)` runs an arbitrary PyPI package, `gh alias set --shell` is a shell escape — and a comment that reads like a confinement boundary while not being one is worse than no comment. It now says plainly that arbitrary execution is **accepted** for this agent, that the actual controls are the OWNER/MEMBER/COLLABORATOR job gate and a token that dies with the job, and that what the allowlist buys is workspace-scoped writes and the removal of `Bash(*)`. Git is called out as the one *documented* escape rather than an implied one.

**Composite-action blind spot:** taken. The three vocabulary tests globbed only `.github/workflows/`, which matters now that this repo has composite actions of its own — one of which this PR adds. They now walk `.github/actions/*/action.yml` and both `jobs.*.steps` and `runs.steps`.

On the `mcp__ols__*` suffixes: they are correct, and I can say that from something better than the discarded input. ols-mcp exposes `search_all_ontologies`, `get_terms_from_ontology` and `get_ontology_info`; with server key `ols` that gives `mcp__ols__*`. I added the third.

Carry-overs I'm consciously leaving: no post-run assertion that weekly-compliance actually opened a PR (real gap, but it wants a design decision about what "should have produced a PR" means on a week with nothing to fix), and the per-file `persist-credentials` grep shape — you've verified 20/20 balanced per checkout twice now, and a proper per-checkout assertion belongs with the composite-action test file rather than bolted onto this PR.

cmungall:
Extending that glob earning its keep immediately is a nice result — and the finding is right, with the sting being that the claim was mine.

**Both composite checkouts are gone.** `claude-issue-{summarize,triage}-action` each ran a second `actions/checkout@v4` with `fetch-depth: 0` into the same workspace, re-cloning over the first and writing the token back into `.git/config` — undoing, one step later in the same job, the exact change this PR makes to those two workflows. As you say they're redundant (the calling workflow already checked out the same commit), so deleting beats flagging.

**The verification shape was the real defect.** A per-file grep passes a file with two checkouts where only one is flagged, and never looked at `.github/actions/**` at all. `test_every_checkout_is_non_persisting` now walks every checkout step across workflows *and* composite actions and asserts `persist-credentials: false` on each. Verified to fail against a probe composite.

**`Bash(git fetch:*)` dropped.** You're right, and this is the one comment whose job is not to overclaim: `git fetch 'ext::sh -c <cmd>'` runs `<cmd>` locally with `protocol.ext.allow` defaulting to `user`. Same family as the vector the push wrapper closes.

Also took the two test-glob notes: the reverse invariant no longer demands an agent-config key for a composite (config keys are workflow stems, and `test_every_config_workflow_file_exists` would then require a matching workflow file), and the glob covers `action.yaml` and nested dirs.

On `.github/actions/claude-code-action` — the older `npm install -g` + CBORG composite behind issue-summarize/triage: agreed it is invisible to every pin, vocabulary and model guard here, and agreed it is its own change. I've named it in the as-built doc's known-gaps section rather than leaving it undocumented.

cmungall:
Blocker taken, but not with the fix you proposed — and the data says why.

`gh run list` was denied in your sandbox, so here it is: **generate-pages' last completed run took 58 minutes** (and failed). So `timeout-minutes: 55` would kill a run that legitimately needs the time, and a token minted at job start would 401 on the push after ~58 minutes of rendering either way. You identified the regression correctly; the two-line fix would have traded one failure mode for another.

Neither job needs credentials while it *works* — only the final push and PR do. So the token is now **minted immediately before that step** rather than at job start, and its 60-minute life starts where it's actually needed. The checkout drops to the default `GITHUB_TOKEN` (public repo, read-only, still non-persisting). Both jobs also get a real bound — 120 and 60 — so neither sits at the 360-minute default.

The two agent workflows keep `timeout-minutes: 55` at job start, because an agent holds the token for its whole run and can't defer it. The comment at each site now says which case it is and why.

**The 21st checkout was exactly where you said.** I widened the glob to all of `.github/**` rather than narrowing the docstring, and it immediately failed on `.github/copilot-setup-steps.yml`. Fixed. Your point about it being a non-running path is right, but a test whose docstring overclaims is the thing I'd rather not ship. `_claude_action_steps()` now names composites via a helper, so an `action.yaml` no longer reports as "action".

On the two I'm not changing:
- **`allowed_bots` in weekly-compliance** is not dead the way claude.yml's was. claude.yml has an `if:` requiring OWNER/MEMBER/COLLABORATOR, which no bot can satisfy — that's what made it dead. weekly-compliance has no such gate and is `workflow_dispatch`-able, so a bot-initiated dispatch is possible and would be rejected without it. Same reasoning as the other scheduled agents.
- **The as-built doc** is on #2241, a separate branch — which is why it isn't in this checkout. It's the rewrite of `docs/automation-migration-to-dismech.md`, and I've added the CBORG composite and the weekly-compliance no-op-week question to its known-gaps section.

PR #2261, opened 2026-07-26
ci: manage agent cron cadence from one config

cmungall:
Phase 4 (cadence half). Stacked on #2260 → #2255 → #2253 → #2249 → #2247 → #2246.

GitHub cannot parameterise `schedule:` at run time, so changing cadence has meant hand-editing a cron line in each workflow. `.github/cron-profiles.yaml` + `scripts/apply_cron_profile.py` make it one switch:

```
just cron-profiles                # list profiles / show active
just cron-profile-preview fast    # diff only, no writes
just cron-profile slow            # rewrite + commit
```

## The kill switch is the point

`just cron-profile off` removes every managed `on.schedule` block, so the whole agent fleet stops in one commit — while `workflow_dispatch` still works for manual runs. Empty schedules rather than an impossible date like Feb 30: GitHub can ignore an invalid cron update and leave the previous valid schedule registered.

## `active: medium` is behaviour-preserving

It reproduces the cadence these workflows already run at. Every cron expression is byte-identical; only quoting and trailing comments are normalised:

```
- cron: '0 6 * * 6'          ->  - cron: "0 6 * * 6"  # weekly, Saturday 06:00 UTC
- cron: "37 1/1 * * *"       ->  - cron: "37 1/1 * * *"  # hourly, offset from the other agents
```

## Implementation note

The applier rewrites **only** the `on.schedule` block, line-based rather than round-tripping the YAML: GitHub's `on:` key parses as the YAML 1.1 boolean `True`, which a naive re-dump corrupts. Ported from dismech.

## Two tests worth calling out

- `main.yaml`'s weekly full re-validation is deliberately **not** managed — it is ordinary CI, not an agent, and `off` must not disable it. Pinned by a test.
- A drift test: if someone hand-edits a `cron:` line instead of going through a profile, the next `just cron-profile` would silently revert it. The test fails first, with the command to run.

## Verification

```
uv run python scripts/apply_cron_profile.py --list
uv run pytest tests/test_agent_config.py tests/test_apply_cron_profile.py -q   # 17 passed
uv run mypy tests/test_apply_cron_profile.py                                    # clean
python3 -c "import yaml, glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.y*ml')]"
```

`slow`/`fast` only move `pr-shepherd` and `curation-scanner`; the weekly scanners are tier-invariant because their cadence is tied to how fast their upstream sources move, not to our spend. Easy to change — it is one file now.

### DIFF
--- a/.github/cron-profiles.yaml
+++ b/.github/cron-profiles.yaml
@@ -1,0 +1,129 @@
+# Central cadence configuration for the scheduled agentic GitHub Actions.
+#
+# Instead of hand-editing the `cron:` lines in each workflow, pick a named
+# profile here and apply it with:
+#
+#     just cron-profile <name>          # rewrite workflows + commit
+#     just cron-profiles                # list profiles / show active
+#     just cron-profile-preview <name>  # diff only, no writes
+#
+# The applier (scripts/apply_cron_profile.py) rewrites ONLY the `on.schedule`
+# block in each managed workflow — everything else (inputs, jobs, comments
+# outside the schedule block) is left untouched. An empty workflow list removes
+# that workflow's schedule block while keeping manual `workflow_dispatch`.
+#
+# Conventions:
+#   * Each profile MUST define every managed workflow (the applier enforces this
+#     so a profile can't be half-configured).
+#   * Workflow keys are the file stem under .github/workflows/ (".yml" or
+#     ".yaml" is resolved automatically).
+#   * `comment:` is optional and rendered as a trailing inline comment.
+#
+# `medium` is the baseline and mirrors the cadence these workflows have been
+# running at, so applying it is behaviour-preserving (cosmetic normalisation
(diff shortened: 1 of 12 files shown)


cmungall:

Reproduced and fixed. `off` -> `medium` really did die with 'no \`on:\` block found' on warm-reference-cache, and because writes happened inside the loop it left 5 of 7 workflows rewritten with a stale `active:`. A one-way kill switch is the worst direction for one to be one-way in.

- `ON_RE` now tolerates a trailing comment and the quoted forms.
- All rewrites are computed before anything is written.
- Three tests, each verified to fail against the pre-fix code: byte-identical off->active round trip over the **committed** workflows (your point that the synthetic bare `on:` could not catch this was exactly right), trailing-comment recognition, and atomicity.

PR #2263, opened 2026-07-26
Review human AADACL4: an intact catalytic triad in a Tdark GDXG hydrolase

cmungall:
## AADACL4 (Q5VUY2) — human arylacetamide deacetylase-like 4

A genuinely dark gene: Pharos **Tdark**, UniProt **PE 3 (inferred from homology)**, no
substrate, inhibitor, knockout or assay ever reported, and an **empty affinage record**
(`n_discoveries: 0`, `citation_count: 0`, `gates_passed: True`). All 5 GOA rows are IBA/IEA.

### The one question the record turns on

Two opposite errors were available here, and both are settled by the same observation:

- accepting a carboxylesterase MF because the domain is *named* "arylacetamide
  deacetylase" (fold name → activity), or
- declaring "fold without function" without ever checking the catalytic residues.

So `AADACL4-bioinformatics/analyze_catalytic_machinery.py` checks it directly, from live
UniProt data with nothing hard-coded:

- **Ser193** sits in a canonical `GESVG` nucleophile elbow; **Asp347**; **His377**; and the
  GDXG **`HGG`** oxyanion-hole motif is present at 119–121.
- All **7 of 7** AADAC-family relatives (AADACL3, AADACL2, human/rat/mouse AADAC,
  human/mouse NCEH1) project their *own* annotated triads onto exactly residues
  193/347/377 of AADACL4 by global alignment, at matching residue identity. There is no
  lost triad member and no displaced elbow, so the pseudoenzyme reading is not available.

### Why `hydrolase activity` is *not* a curator error, and still gets replaced

I resolved **all 17 WITH/FROM tokens** of the `GO:0016787` IBA row (`PTN009058710`). The node
reaches back past the plant/fungal/bacterial split and is functionally heterogeneous:

| Class | n |
|---|---|
| ester hydrolase (EC 3.1.-) | 9 |
| no EC assigned | 4 |
| **amide hydrolase (EC 3.5.-)** — mouse Afmid, yeast BNA7 kynurenine formamidase | 2 |
| **ester hydrolase + lyase (EC 4.-)** — soybean HIDH, whose annotated nucleophile is a **threonine** | 1 |

At that depth `hydrolase activity` is the correct last-common-ancestor call. It is replaced
not because it is wrong but because it is **redundant**: the protein's own subfamily
signature (InterPro IPR017157 / PIRSF037251) already yields `GO:0052689` on the same
protein, and the triad audit shows that specific term is earned.

### Why `membrane` stays general — deliberately

The `GO:0016020` IBA's own donors **disagree about which membrane**: human AADAC and mouse
Aadac are ER/microsomal membrane, mouse Nceh1 is plasma membrane — 4 distinct specific
locations across 3 donors. Refining to `GO:0005789` would mean picking one donor over the
other with no AADACL4 data. So both membrane rows are ACCEPTed *at this level*, and the
reasons record that they are not two independent lines of support (the IEA traces to the
same UniProt `ECO:0000305`-over-`ECO:0000255` statement).

### Actions

| # | Term | Ev | Action |
|---|---|---|---|
| 1 | GO:0016787 hydrolase activity | IBA | **MODIFY** → GO:0052689 |
| 2 | GO:0016020 membrane (`is_active_in`) | IBA | ACCEPT |
| 3 | GO:0016020 membrane (`located_in`) | IEA | ACCEPT |
| 4 | GO:0016787 hydrolase activity | IEA | **MODIFY** → GO:0052689 |
| 5 | GO:0052689 carboxylic ester hydrolase activity | IEA | ACCEPT |

**2 MODIFY, 3 ACCEPT.** Nothing removed; no new GO term proposed (`GO:0052689` already says
exactly what the evidence supports, and a substrate-level child is not licensed by homology).
Both MODIFYs carry `propagation_review` with `root_cause: TERM_SCOPING_PROBLEM` /
`failure_modes: [GRANULARITY_MISMATCH]`. Neither IBA row is self-referential.

### Curation findings raised upstream

1. **UniProt N-terminal inconsistency across the family.** Among paralogs that share the
   catalytic register *exactly*, UniProt calls a signal anchor for AADACL4, a **cleaved
   signal peptide → Secreted** for AADACL2, and **nothing at all** for AADACL3 — which has a
   comparably hydrophobic N-terminal segment (Kyte–Doolittle peak 2.5 at residue 6 vs 2.8 at
   residue 6 for AADACL4) and consequently no subcellular location and no membrane GO
   annotation. All three calls are `ECO:0000255`.
2. **PAN-GO vs GO_Central.** UniProt cross-references `PAN-GO; Q5VUY2; 0 GO annotations based
   on evolutionary models.` while GOA carries two GO_Central IBA rows for the same protein.
3. **A testable topology handle exists**: the two predicted N-glycosylation sites (Asn168,
   Asn269) both lie in the predicted lumenal domain, so glycan occupancy would confirm or
   refute the type-II orientation. Written up as a suggested experiment.

### On the empty affinage record

Per the ACP7 lesson, an empty affinage record is not evidence that literature is absent. I
checked independently: UniProt cites exactly one paper (the chromosome 1 sequence,
PMID:16710414); Europe PMC has 53 "AADACL4" records, of which 2 are title/abstract hits (both
livestock/tumour signature papers). The one publication that discusses the gene as a gene is
PMID:21696217 (Long & Cravatt, *Chem Rev*), which places it in a section headed
"Representative examples of poorly characterized hydrolases" and states
*"AADAC is highly homologous to several related lipases, AADACL1, 2, and 4, the latter two of
which remain completely unannotated."* One non-human lead (chicken `AADACL4B` in keratinocyte
differentiation, PMID:34997067) is recorded in `suggested_questions` only — the `B` suffix
signals a bird-specific duplicate and orthology is not established.

### Verification

- `checkquotes.py`: **28 quotes, 0 problems** (includes all `file:` quotes, each confined to
  one physical UniProt line).
- `just validate human AADACL4`: **✓ Valid**, 1 warning — "No annotations reference available
  deep research files". Left standing deliberately: the affinage file is empty, and citing it
  as annotation support would contradict this review's own reasoning about absence of
  evidence. It is referenced in the top-level `references` with `relevance: LOW`.
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → empty; no new terms
  needed (all three GO ids were already cached).
- A fresh `uv run python analyze_catalytic_machinery.py` reproduces the committed
  `RESULTS.md` and `results.json` byte-for-byte.
- Three structured `knowledge_gaps` (MF_DARK substrate, CC_DARK compartment/topology,
  BP_DARK process) with verbatim provenance.

---

### Self-corrections made after opening (commits 2–5)

Caught by re-checking my own claims against primary sources rather than by review:

1. **Chromosomal map was wrong.** The description said the family is "encoded on chromosome 1".
   Verified against NCBI Gene and UniProt: AADAC (3q25.1), AADACL2 (3q25.1) and NCEH1/AADACL1
   (3q26.31) are on **chromosome 3**; only AADACL3 and AADACL4 are on chromosome 1, adjacent at
   1p36.21.
2. **Two over-broad counts tightened.** "All 21 WITH/FROM tokens resolved to UniProt accessions"
   → 19 accessions plus the two ancestral PANTHER nodes, which have no sequence. And AADACL4
   carries *an equivalent* signal anchor, not "the same" one (donor spans differ: 5–25 for NCEH1,
   6–23 for human AADAC, 6–26 for mouse Aadac).
3. **The MODIFYs are now explicitly bounded.** Checked QuickGO: neither `GO:0016787` nor
   `GO:0016020` is in `gocheck_do_not_annotate` (both `usage: Unrestricted`), so no annotation
   rule is breached — the case is purely redundancy plus informativeness. And the replacement
   term is a claim about AADACL4's annotation set, *not* a claim that `PTN009058710` could
   support `GO:0052689`; it could not, which is exactly why the IBA landed on the general parent.
   The `GO:0052689 → GO:0016788 → GO:0016787` `is_a` chain was confirmed against QuickGO rather
   than assumed.
4. **A self-contradiction removed.** Row 3's summary called the IEA membrane row "an independent
   route" while its reason said both rows "rest ultimately on the same prediction". Neither is
   right: the IBA is grounded in experimentally localised donors, the IEA in AADACL4's own
   `ECO:0000255` topology prediction plus family similarity. They are *partly* independent, and
   neither adds experimental evidence for AADACL4 itself.
5. **The WITH/FROM resolutions are now machine-checked, not hard-coded.** The MGI/RGD/SGD/AGI
   tokens were resolved by hand through the owning databases, and the entire
   donor-heterogeneity argument rests on those resolutions. The script now checks each
   non-PANTHER token back against the resolved UniProt entry's *own* cross-references (MGI,
   RGD, SGD, Araport, or accession equality) before reporting anything, and aborts on a
   mismatch — **19 token resolutions pass**. Confirmed the guard bites by temporarily
   mis-mapping `RGD:631440` to `Q8BLF1`: exits 1 with
   `FATAL: token RGD:631440 was resolved to Q8BLF1, but that entry's RGD cross-references are []`.
   `RESULTS.md` still reproduces identically on a fresh run.


cmungall:
Thanks — the independent triad check off the `SQ` block is exactly the right way to audit that
claim, and I'm glad the load-bearing fact holds without the report. All three 🔵 suggestions were
worth doing, so all three are implemented in `903d422ba` (branch rebased onto current `main`).

### 1 — the load-bearing claim now has an anchor ✅

You were right that the most novel claim in row 1's `reason` was the one without a quote. Added
two `supported_by` entries pointing at the EC-class read-out rather than one, since the argument
rests on *two* kinds of divergence:

```yaml
    - reference_id: file:human/AADACL4/AADACL4-bioinformatics/RESULTS.md
      supporting_text: "amide hydrolase (EC 3.5.-): **2**"
    - reference_id: file:human/AADACL4/AADACL4-bioinformatics/RESULTS.md
      supporting_text: "ester hydrolase (EC 3.1.-) + lyase (EC 4.-): **1**"
```

The triad quotes are kept after them. `checkquotes.py` now reports **30 quotes, 0 problems**.

### 2 — `7 of 7` is now computed, and the guard bit on the first attempt ✅

This was the best of the three. `REGISTER_TRUSTED_ROLES` is gone; `alignment_trusted` is now
`percent_identity >= REGISTER_IDENTITY_THRESHOLD`, with an assertion that no panel member sits
within `REGISTER_MIN_GAP` of the threshold — so a future UniProt release that shifted an identity
across the line aborts rather than silently reclassifying a reference.

**It caught me immediately.** My first threshold was the 31.0 implied by the old "roughly 30%"
comment, and the run died:

```
FATAL: NlhH (M. tuberculosis) at 30.3% sits within 1.0 points of the 31.0% register
threshold, so the trusted/untrusted split is arbitrary.
```

That is the assertion earning its place: 31.0 really was a bad cut, and the old hand-labelled
partition was hiding it. The threshold now sits at **31.3**, inside the empty band between NlhH
(30.3%) and AADACL2 (32.3%), with a full point of margin each side and `REGISTER_MIN_GAP = 0.75`.
The comment states plainly that the value is placed in the observed gap rather than tuned
silently. The report also now prints whether the purely numeric cut reproduces AADAC-family
membership — it does (`register_cut_matches_family: true`), so the threshold is not doing hidden
work, and `7 of 7` is unchanged but now derived.

### 3 — row 4 no longer reads as an InterPro2GO complaint ✅

Fair catch, and the misreading you describe would be a real error. Row 4's reason now says it
explicitly:

> ...not a claim that the annotation is wrong — and specifically not a recommendation to change the
> IPR013094-to-GO:0016787 InterPro2GO mapping itself. That domain genuinely does occur in amidases
> as well as esterases, so the mapping is right for the domain and should stay.

### Trivia note — also fixed

You were right that "same 407-residue length" was the one family claim not backed by
`results.json`. Rather than soften it, the script now records `length` per panel member and the
Part 4 table carries a Length column, so it is sourced (AADACL4 407, AADACL3 407, AADACL2 401,
AADAC 399, NCEH1 408). The notes now cite `results.json` for all three AADACL3 figures.

### Verification after the changes

- `checkquotes.py`: 30 quotes, 0 problems · `just validate human AADACL4`: ✓ Valid (1 warning,
  the deep-research one, left standing deliberately — the affinage record is empty).
- Fresh `uv run python analyze_catalytic_machinery.py` reproduces `RESULTS.md` identically.
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → empty. It briefly looked
  non-empty before the rebase; confirmed via `git diff $(git merge-base HEAD origin/main) HEAD`
  that this branch never touches the file, and that the apparent deletions were just `main`
  advancing.

PR #2264, opened 2026-07-26
Review human AADACL3: the fold-name check that came back positive

cmungall:
## AADACL3 (Q5VUY0) — the fold-name check that came back positive

AADACL3 is about as dark as this campaign gets: Pharos **Tdark**, UniProt **PE 2**
(transcript level, so the protein has never been detected), no FUNCTION, no CATALYTIC
ACTIVITY, no SUBUNIT and no SUBCELLULAR LOCATION comment, two references — both
sequencing papers — and `PAN-GO; 0 GO annotations based on evolutionary models`. All
five GO rows are IBA or IEA, derived from family and fold signatures.

The campaign's most repeated finding is that domain names propagate into GO as
activities. The mirror-image error — asserting "fold without function" without checking
the residues — is equally common. So both directions were tested rather than argued.

### The catalytic machinery is actually there

Reading residues out of the sequence at the annotated positions and aligning to the
characterised relatives:

| check | result |
|---|---|
| Ser193 / Asp347 / His377 | serine / aspartate / histidine, as required |
| nucleophile elbow 191–196 | `GDSFGG` |
| oxyanion loop 119–121 | `HGG` |
| triad on curated active sites in AADAC, NCEH1, mouse Nceh1, AADACL2, AADACL4 | 3/3 in each |

So **GO:0052689 carboxylic ester hydrolase activity is ACCEPTed** as a well-grounded
prediction. Three independent lines converge: a *subfamily* signature (IPR017157 /
PIRSF037251 across residues 19–403, PIRSF at 5.8e-82, InterPro's representative match),
UniProt's own RecName `EC=3.1.1.-`, and the verified triad. It is deliberately kept
substrate-agnostic — the family's own history forbids inferring a substrate, since NCEH1
cannot rescue the sterol-acetate phenotype of the yeast AADAC ortholog.

### A near-miss trap worth recording

**PROSITE PS01174**, the GDXG-family "putative serine active site" pattern, matches AADAC,
NCEH1 and AADACL2 but **does not match AADACL3**. Taken at face value that looks exactly
like the loss of catalytic machinery one would use to argue an esterase annotation away.
Walking the pattern `[LIVM]-x-[LIVMF]-[SA]-G-D-S-[CAS]-G-[GA]-x-[LI]-[CAVT]` position by
position shows both mismatches are *flanking* — Cys190 where it wants Ser/Ala, Phe194
where it wants Cys/Ala/Ser — while positions 5–7, the `G-D-S` core containing the
nucleophile, are intact. A signature miss has to be resolved to individual positions
before it is allowed to change an action.

### The membrane rows run the *other* way: GO is ahead of UniProt

AADACL3's UniProt entry has **no TRANSMEM feature, no membrane keyword and no
SUBCELLULAR LOCATION comment**, so the two `membrane` annotations look unsupported by the
underlying record. They are not. Its **tandem paralog AADACL4** — same 407 aa, same
numbered active sites and oxyanion motif, **identical Phobius calls** — is annotated by
UniProt with `TRANSMEM 5-25`, both topological domains and "Single-pass type II membrane
protein". The difference is coverage of UniProt's automatic annotation, not biology.

Independent support on the AADACL3 sequence itself: Phobius TM 6–29 and 41–60, TMHMM
4–26 and 43–60, no signal-peptide call from either Phobius or SignalP, catalytic domain
(61–407) non-cytoplasmic. The discriminating control is **AADACL2**, where SignalP *does*
fire and the curated location is "Secreted" — so the absence of a cleavage call for
AADACL3 is informative rather than a default. This matters because the family gives
conflicting topology at essentially equal distance (AADAC/NCEH1 membrane-anchored at
~33% identity; AADACL2 secreted at ~34%). Both rows **ACCEPTed**; the UniProt gap is
raised in `suggested_questions` as an upstream correction to report.

### Why the two generic rows are MODIFY

Resolving the `GO:0016787` IBA's seventeen WITH/FROM entries explains the generic level:
the clade mixes carboxylesterases (E. coli Aes, three M. tuberculosis esterases),
AADAC/NCEH1 from three species, **two formylkynurenine formamidases** that are
amidohydrolases (mouse Afmid, yeast BNA7), four Arabidopsis loci, and soybean HIDH, which
UniProt annotates with **both** EC 3.1.1.1 carboxylesterase and EC 4.2.1.105
2-hydroxyisoflavanone dehydratase — a legitimate but weak donor, since the dehydration is
its physiologically important reaction and its esterase activity is described only as
slight. `hydrolase activity` is close to the LCA of that set, so PAINT's choice is
defensible *for the node*; it is just a proper ancestor of the term GOA already carries
for this gene.

The `GO:0016787` IEA comes from `IPR013094 Alpha/beta hydrolase fold-3`, a **fold**
signature, and section 6 of the analysis shows why that is the weaker provenance:
IPR013094 matches HIDH at 74–298 even though HIDH matches **neither** `IPR017157` nor
`PIRSF037251` nor even the family-level `IPR050300`. The fold signature reaches outside
the subfamily whose chemistry is being asserted, so it constrains architecture rather
than reaction — it lands on the right answer here only because the triad is intact.

Both MODIFY to `GO:0052689`, with `propagation_review` root cause
`TERM_SCOPING_PROBLEM` / `GRANULARITY_MISMATCH`. The membrane IBA used a *different*
PANTHER node (PTN009058713 vs PTN009058710) and a tight, topologically coherent donor set;
neither IBA is self-referential.

### Cross-gene: PAINT's node placement for this family is inverted

Reconciled with the parallel **AADACL2** (#2266) and **AADACL4** (#2263, merged) reviews.
The finding there is that `GO:0017171 serine hydrolase activity` sits only at ortholog-level
nodes while the shared family node `PTN009058713` carries `GO:0016020 membrane` — yet the
catalytic triad transfers family-wide and the membrane anchor does not. I tested both halves
against AADACL3's own record instead of importing them.

**The mechanism half is confirmed and quantified** (new RESULTS.md section 7, WITH/FROM
tokens read from the gene's own `AADACL3-goa.tsv` so they cannot drift):

> Audited: GO:0016787 (IBA, GO_REF:0000033). 17 WITH/FROM tokens, 14 resolved to a protein,
> 14 with a readable nucleophile, of which **13 are serine**. Non-serine: HIDH_SOYBN T164.

Mouse Afmid — a *formamidase* — is itself a serine hydrolase (`S162`), so the amidohydrolase
members are no obstacle to a mechanism term. The **single** obstacle at `PTN009058710` is
soybean HIDH with `Thr164`, so the term is blocked by exactly one nameable member; at the
tighter `PTN009058713` there is no obstacle at all. Rat Aadac and yeast BNA7 do not resolve
to a reviewed entry and are reported unresolved rather than counted either way.

Consequently **both generic `GO:0016787` rows now propose two replacements**, not one:
`GO:0017171` for the mechanism and `GO:0052689` for the reaction class. Checked against
QuickGO, the two are *independent children* of `GO:0016787` — neither is an ancestor of the
other — so the generic term sits exactly at the join of the two axes and both are needed.
AADACL3 satisfies GO:0017171's definition literally: serine nucleophile Ser193, acidic
Asp347, basic His377.

**One premise needed correcting.** AADACL3 and AADACL4 do *not* inherit "no mechanism term
at all". I queried Q5VUY2's GOA directly: it is **row-for-row identical to AADACL3's** — same
five terms, same WITH/FROM sets, including `GO:0016787` IBA from `PTN009058710` with all 17
sources. Both inherit the uninformative *parent*, so for these two genes the defect is
granularity rather than absence, which is why MODIFY is right rather than a new annotation.
The recommendation is recorded in `suggested_questions` naming all three genes; **AADACL4
needs no separate follow-up** beyond the `GESVGG` elbow noted below.

**The mouse ortholog sharpens the UniProt gap.** `A2A7Z8` (ADCL3_MOUSE, 61.9% identity,
triad 3/3 on annotated sites) is curated by UniProt as `Membrane; Multi-pass membrane
protein` with three helices at 2–22, 46–66, 109–129. So **the same gene in mouse carries a
curated membrane location while the human entry carries none** — the gap is not only against
the tandem paralog, it is against the ortholog. It also confirms from the curated record the
two-helix caveat I had flagged from prediction alone: "single-pass type II" is the wrong
descriptor for this branch. A second corroboration from the GOA side: AADACL4's membrane IEA
carries `UniProtKB-SubCell:SL-0162` and AADACL3's does not, because that mapping can only
fire for an entry that already has a curated location.

**Where AADACL3 differs from AADACL2 on topology — and why ACCEPT, not UNDECIDED.** AADACL2's
location rows were left UNDECIDED because a curated "Secreted" call collides with the family
anchor and measurement declined to break the tie. AADACL3 has **no curated location, no
SignalP call, and concordant Phobius/TMHMM helices** — there is nothing in conflict to
adjudicate. No topology argument was used to remove anything here.

**AADACL4 aside.** Its nucleophile elbow reads `GESVGG` — glutamate where the family has the
structural aspartate of `GDSxGG` — so it fails PS01174 at four positions (4, 6, 8, 12)
against AADACL3's two. Flagged for whoever amends #2263; not acted on here.

### Action counts

| action | n |
|---|---|
| ACCEPT | 3 |
| MODIFY | 2 |
| REMOVE / MARK_AS_OVER_ANNOTATED | 0 |

Nothing in this record over-reaches — there is no substrate-specific term, no specific
compartment and no biological process, which for a gene with zero functional data is the
correct state.

### Literature

Affinage returned an empty record (`gates_passed: True`, 0 discoveries, 0 citations).
Since an empty provider record is not evidence that literature is absent, PubMed was
searched directly: 7 records mention AADACL3, all livestock body-weight association
studies, a sheep selection scan, a sarcoidosis exome study, a 1p36 duplication case
report, a DCIS mutational-landscape study, and an AADACL1 platelet paper. None assays the
protein. The family's own review states the position outright
([PMID:35736449](https://pubmed.ncbi.nlm.nih.gov/35736449/), full text from PMC9229287):
*"all other members of the AADACL protein family, AADACL2, AADACL3, and AADACL4, have so
far been only poorly investigated and no functional roles can be concluded."*

### Bioinformatics

`genes/human/AADACL3/AADACL3-bioinformatics/analyze.py` fetches sequences, active-site
features, InterPro signature matches, InterProScan topology predictions (Phobius / TMHMM /
SignalP) and the PROSITE pattern **live**; a missing or empty record raises rather than
being counted as absence; a fresh `uv run python analyze.py` reproduces both `RESULTS.md`
and `results.json` byte for byte (verified). It was built to be able to falsify the
annotations in either direction, and it records the observations that cut against a tidy
story — the PS01174 miss, the split Pfam match with the shortest first segment in the
panel, and the two-helix rather than single-pass topology call.

### Note on the one remaining validation warning

`just validate human AADACL3` is `✓ Valid` with one warning: *"No annotations reference
available deep research files."* That warning is genuine and is left standing — the
affinage record is empty, and citing "No mechanistic discoveries found in literature" as
support for a positive claim would be worse than the warning. It is explained in the
affinage reference's `reference_review`.


cmungall:
Both points taken, and both were real. Pushed in `f9099f19a`.

### 1. The `non-hydrolases` clause — you were right, and the fix turned out to be better than a deletion

You were correct that the clause was contradicted by my own correction: HIDH was the only fold-3 member named in the file, and after `f50b47fcd` recorded it as bifunctional (EC 3.1.1.1 + EC 4.2.1.105) it *is* a hydrolase, so it could not carry "members include non-hydrolases".

Rather than just drop the leg, the fix in point 2 handed me an argument that the analysis actually establishes. From the new section 6:

```
| protein              | IPR017157 | PIRSF037251                | IPR013094         | IPR050300 |
| ADCL3_HUMAN (Q5VUY0) | 19-403    | 19-403 (5.8e-82) [repres.] | 115-245; 310-380  | 74-385    |
| HIDH_SOYBN (Q5NUF3)  | -         | -                          | 74-298            | -         |
```

**IPR013094 matches HIDH, but HIDH matches neither IPR017157 nor PIRSF037251 nor even the family-level IPR050300.** So the objection to the fold signature is not "its members aren't hydrolases" — it is that fold-3 membership *reaches outside the arylacetamide-deacetylase subfamily whose chemistry is being asserted*, and within this panel the same scaffold supports amide hydrolysis (AADAC's arylacetamide deacetylation), ester hydrolysis, and a dehydration. That is the accurate provenance objection, and it is now backed by a row in the reproducible output rather than by a claim about family membership.

Reworded in all four places: the row `summary`, the row `reason` (now "a fold-level signature that reaches outside the arylacetamide-deacetylase subfamily whose chemistry it is being used to assert"), the `propagation_review` comment for `InterPro:IPR013094`, and the corresponding paragraph in `AADACL3-notes.md`. The edit script asserted each anchor was present before replacing and then re-grepped for `non-hydrolases` and `Members include soybean` across both files — 0 remnants. **MODIFY and the `GO:0052689` replacement are unchanged**, as you said they should be.

### 2. The subfamily-signature pillar is now in the reproducible output

Exactly as you diagnosed: `add_interpro` was fetching the full match list and discarding everything but PS01174 and PF07859. It now captures span, e-value, representative flag and model for `IPR017157`, `PIRSF037251`, `IPR013094` and `IPR050300` across the whole panel, tabulated as **RESULTS.md section 6**, and raises if no tracked signature matches the query rather than emitting an empty section.

The new row is cited from three places that previously asserted the numbers — the `GO:0052689` ACCEPT, the `GO:0016787` IBA MODIFY (where it justifies the replacement term), and `core_functions` — and the HIDH row is cited from the `GO:0016787` IEA MODIFY. `checkquotes` went from 23 to **27 quotes, 0 problems**.

Two incidental things the new table surfaced, both left as observations rather than turned into claims: AADACL3's IPR017157 span (19–403) is the narrowest in the panel apart from AADACL4's (21–404), consistent with the two of them being the divergent branch; and AADACL2/AADAC/NCEH1 all return a PIRSF e-value of 0.0 while AADACL3 returns 5.8e-82 and AADACL4 1.3e-79, so the 1p36.21 pair are the weakest subfamily matches — still overwhelming, but measurably weaker.

### 3. `is_active_in` (minor)

Acknowledged in the row summary, as you suggested, rather than changed:

> The row asserts `is_active_in`, which locates the activity in the bilayer, while the topology reading places the catalytic domain on the non-cytoplasmic side of it — so on a strict reading the hydrolysis happens at the membrane surface rather than within the lipid phase. The alternative would be to name the compartment the catalytic domain faces, and that cannot be done for AADACL3 without knowing which membrane it sits in, so the qualifier is left as GOA has it.

### 4. PR body (minor)

Updated — it no longer says HIDH's "characterised primary activity is a dehydratase — a lyase", and now states the bifunctional EC assignment and the fold-reaches-outside-the-subfamily argument.

### Verification

- `uv run python checkquotes.py` → **27 quotes, 0 problems**
- `just validate human AADACL3` → **✓ Valid**, with the one genuine warning (no annotation cites the empty affinage record) still standing
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → empty
- `uv run python analyze.py` run twice → `RESULTS.md` and `results.json` **byte-identical** on the second run (`diff` clean), so the committed report is reproducible after the extension

On your validation note: `test (3.12)` went green on the previous head and is running again on this one.

cmungall:
Took both non-blocking notes as well — pushed in `779600117`.

**Pagination guard.** You were right that the risk changed character when section 6 started making claims from *absence*. `add_interpro` now raises if `matches["count"]` exceeds `MATCH_PAGE_SIZE` (200), naming the two ways to fix it, rather than paginating silently:

```python
if matches["count"] > MATCH_PAGE_SIZE:
    raise RuntimeError(
        f"InterPro reports {matches['count']} signature matches for {entry.accession}, "
        f"more than the single page of {MATCH_PAGE_SIZE} requested; the results would be "
        "truncated and absences could not be trusted. ..."
    )
```

I chose the guard over following `next` deliberately: a hard stop is auditable, whereas silent pagination would leave a reader unable to tell whether a `-` cell means "no match" or "page two wasn't fetched". This is the same rule as the missing-input policy already in the script — a failure has to be loud.

**`(0)` → `(0.0)`.** A `format_score` helper now renders an exact-zero e-value as `0.0` rather than `0`, and the section preamble says what it means: *"An e-value of `0.0` is the API reporting underflow below double precision, not a literal zero."* So the table is now self-explanatory without the PR description.

Neither change touches a cited number — `RESULTS.md:79` and `:85`, the two rows quoted in the review, are byte-identical, and `checkquotes` still reports **27 quotes, 0 problems**. `uv run python analyze.py` run twice reproduces `RESULTS.md` and `results.json` exactly, and `just validate human AADACL3` is `✓ Valid` with the one standing warning. `cache/go/terms.csv` remains untouched.

Understood on the approval limitation — nothing further outstanding from my side, so this is ready whenever a human can dismiss the stale `CHANGES_REQUESTED` or re-request review. Thanks for both rounds; the pagination catch in particular was the kind of thing that only becomes a bug once a negative result starts being load-bearing, which is exactly when it would have been hardest to spot.

cmungall:
## Cross-gene reconciliation with AADACL2 (#2266) and AADACL4 (#2263) — pushed in `3bbc761c9`

I tested both halves of the node-placement finding against AADACL3's own record rather than importing it. **It holds**, and it comes out stronger measured than stated — but one premise needed correcting.

### The mechanism half: confirmed, and blocked by exactly one nameable member

New **RESULTS.md section 7** reads the residue at the first annotated active site of every WITH/FROM source on AADACL3's own `GO:0016787` IBA. Tokens are read from the gene's own `AADACL3-goa.tsv` so they cannot drift; MOD ids resolve through the Alliance API, Arabidopsis loci through UniProt, and anything that doesn't resolve to a *single reviewed* entry is reported unresolved rather than dropped:

> Audited: GO:0016787 (IBA, GO_REF:0000033). 17 WITH/FROM tokens, 14 resolved to a protein, 14 with a readable nucleophile, of which **13 are serine**. Non-serine: HIDH_SOYBN T164.

Two things fall out that I don't think were visible from AADACL2 alone:

- **Mouse Afmid — a formamidase — is itself a serine hydrolase (`S162`).** So the amidohydrolase members of the clade are *not* an obstacle to a mechanism term, which was my own prior assumption when I first argued the clade could only support `hydrolase activity`. I was wrong about why the term was generic.
- **The single obstacle is soybean HIDH, `Thr164`.** So `GO:0017171` is blocked at `PTN009058710` by exactly one member, and it is nameable — the recommendation can be stated as "place it at `PTN009058713`, or at the sub-node excluding HIDH" rather than as a general judgement about clade heterogeneity.

Both generic `GO:0016787` rows therefore now carry **two** replacement terms. Checked against QuickGO: `GO:0017171` and `GO:0052689` are *independent children* of `GO:0016787`, neither an ancestor of the other — so the generic term sits exactly at the join of a mechanism axis and a reaction axis, and one replacement cannot cover both. `GO:0017171`'s definition is satisfied literally by what I had already verified: serine nucleophile Ser193, acidic Asp347, basic His377. This aligns AADACL3's treatment with AADACL2's.

### One premise corrected: these genes do inherit a mechanism-branch term

> *"so AADACL2/3/4 inherit no mechanism term at all"*

Not for AADACL3 or AADACL4. I queried `Q5VUY2`'s GOA directly rather than infer it — **AADACL4's record is row-for-row identical to AADACL3's**: same five annotations, same terms, same WITH/FROM sets, including `GO:0016787` IBA from `PTN009058710` with all seventeen sources. Both inherit the uninformative *parent* of the term that should be there. That is why MODIFY is the right action for both rather than proposing a new annotation, and it is stated that way in the `suggested_questions` entry so the recommendation doesn't rest on a premise a curator would find false on inspection.

### The mouse ortholog: your "three plain helices" point, from the curated record

Adding `A2A7Z8` (ADCL3_MOUSE) to the panel confirms it and sharpens the UniProt-gap finding considerably:

```
| ADCL3_MOUSE (A2A7Z8) | [(2, 22), (46, 66), (109, 129)] | - | Membrane; Multi-pass membrane protein | ...
```

61.9% identity, triad 3/3 on annotated sites — and UniProt curates it as **`Membrane; Multi-pass membrane protein`**. So *the same gene in mouse carries a curated membrane location while the human entry carries none at all*; the gap is not only against the tandem paralog AADACL4 but against the ortholog. It also converts my two-helix caveat from a prediction into agreement with a curated record: "single-pass type II" is the wrong descriptor for this branch.

A second corroboration I noticed while verifying AADACL4's GOA: its membrane IEA carries `UniProtKB-SubCell:SL-0162` in the WITH/FROM and **AADACL3's does not** — that mapping can only fire for an entry that already has a curated subcellular location. The gap is visible in the GOA evidence chain, not just in the feature table.

### Topology: why ACCEPT here and not UNDECIDED

Taking the warning seriously — I have not upgraded anything to REMOVE on a topology argument, and the difference from AADACL2 is structural rather than a difference of nerve. AADACL2's rows are UNDECIDED because a curated **"Secreted"** call collides with the family anchor and measurement declined to break the tie. AADACL3 has **no curated location at all, no SignalP call, and concordant Phobius + TMHMM helices** — there is no contradiction to adjudicate, and the membrane rows are accepted on positive prediction with AADACL2 as the negative control that makes the absent SignalP call informative. If AADACL3 had AADACL2's cleaved-signal-peptide call, UNDECIDED would be the right action here too.

### Family-wide claims audited against the full member set

Following the note about the five self-corrections on AADACL2: no family-wide specificity claim is made here — the review says explicitly that substrate must not be inferred from membership and cites the Say1Δ non-rescue result as the reason; expression statements are per-gene HPA records with the gene named (AADACL3 placenta/skin-enhanced, AADACL2 skin-enriched, AADACL4 choroid-enhanced) rather than a branch-level generalisation; and the two formamidases are named as amidohydrolases in the review rather than folded into an esterase generalisation.

### AADACL4 aside, for whoever amends #2263

Its nucleophile elbow reads **`GESVGG`** — glutamate where the family has the structural aspartate of `GDSxGG` — so it fails PS01174 at **four** positions (4, 6, 8, 12) against AADACL3's two. Recorded in my notes and the PR body rather than opened as a follow-up. Whether that affects its catalytic competence is a real question; AADACL3 does not share it.

### Gates

- `checkquotes` → **31 quotes, 0 problems** (was 27)
- `just validate human AADACL3` → **✓ Valid**, one genuine warning still standing
- `uv run python analyze.py` twice → `RESULTS.md` and `results.json` **byte-identical**
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → **empty**, run as the last step after the final `just validate`; `GO:0017171` was already present in the cache so no insertion was needed and the file is untouched

cmungall:
## Correcting myself: `GO:0016787` is the correct LCA here, so the `GO:0017171` replacement is withdrawn (`7f06a61ea`)

My previous commit proposed `GO:0017171 serine hydrolase activity` as a second replacement on both generic hydrolase rows, reasoning that 13 of 14 readable donors have a serine nucleophile so the mechanism "transfers family-wide". **That reasoning was wrong, and the data in my own section 7 is what refutes it.** `GO:0017171` is defined by mechanism, so it can only sit at a node *all* of whose members have a serine nucleophile — and HIDH's `Thr164` means `PTN009058710` is not such a node. One exception is enough. `GO:0052689` cannot sit there either, because the two formamidases hydrolyse an amide bond rather than an ester.

So `GO:0016787` **genuinely is the lowest common ancestor of that donor set**, and PAINT's choice is correct curation, not a term that stopped short. What I had was a near-homogeneous set, and I treated "near" as "is".

Both rows are now replaced on **redundancy** grounds only: AADACL3 already carries `GO:0052689` from a subfamily-specific signature, which is stronger and more direct evidence for *this gene* than a transfer from a clade that broad. Accordingly `propagation_review.root_cause` changes on both rows from `TERM_SCOPING_PROBLEM` to `EVIDENCE_CIRCULAR_OR_REDUNDANT`, and **`GRANULARITY_MISMATCH` is dropped** — that mode is apt only when the donors agree and the term still sits above them, which is not this case. The `GO:0017171` recommendation moves entirely into `suggested_questions` as a node-placement proposal, where the single HIDH obstacle can be named and worked around (`PTN009058713` has no threonine-nucleophile member).

I applied the same test to the location term and got the same answer, so it is now stated positively in the row instead of as a hedge: the membrane IBA's donors resolve to **four distinct curated location strings** — ER membrane and microsome membrane for AADAC, cell membrane and microsome for Nceh1 — so refining `GO:0016020` would mean arbitrarily preferring one donor. General is correct in both cases. The question to ask before any specificity upgrade, now recorded in the notes: *is this term the LCA of its donors?*

**New `suggested_question`: UniProt makes three different N-terminal calls across three paralogs with identical catalytic registers** — signal anchor (AADACL4 `TRANSMEM 5-25`), cleaved signal peptide (AADACL2 `SIGNAL 1-18`, Secreted), and **nothing at all** (AADACL3: no feature, no keyword, no `SUBCELLULAR LOCATION`). AADACL3 is the extreme case, and the honest position is that UniProt *has not made a call*, not that it disagrees with GO. The mouse ortholog makes that hard to explain as biology — `A2A7Z8`, 61.9% identical with the same triad, is curated `Membrane; Multi-pass membrane protein` with three helices — and the asymmetry shows up in GOA too, since AADACL4's membrane IEA cites `UniProtKB-SubCell:SL-0162` and AADACL3's cannot, that mapping requiring a curated location to exist first. The narrower AADACL4-only version of the question is subsumed.

**Two method points adopted.** The load-bearing condition for a conservation claim is that a mapped residue lands on the target's own annotated active site **and** is the same amino acid; the script required only the first half in its headline count, and now reports both. HIDH shows why: `S193` *does* land on HIDH's annotated site 164, but that residue is a threonine, so HIDH is 2/3 on the site and **1/3** on identity, while every real family member is 3/3 on both. Separately, each WITH/FROM resolution is now verified against the resolved entry's *own* cross-references (MGI/RGD/SGD id, or Araport for an Arabidopsis locus), with `self_test_resolution_guard()` feeding the guard a deliberately wrong mapping at the start of every run — mouse Aadac's MGI id against the human protein — and failing loudly if it is accepted, so the check cannot silently go vacuous if a cross-reference database is renamed. No identity threshold is used anywhere in the analysis, so there was nothing to re-derive from a distribution.

Detailed reasoning for all of the above is in `AADACL3-notes.md` rather than repeated here.

**Gates** (`origin/main` merged first, since it had advanced): `checkquotes` 31 quotes / 0 problems · `just validate human AADACL3` ✓ Valid with the one genuine warning · `analyze.py` byte-identical on rerun · `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` empty, run **last** after the final validate, with no re-sorting and no rows added.

Final state: **3 ACCEPT, 2 MODIFY (both → `GO:0052689`), 0 REMOVE.** Nothing outstanding from my side.

cmungall:
All three 🟡 items and all four 🔵 items are fixed in `b413b8502`. No action, term or replacement moved.

**1 & 2 — the stale notes paragraph, and a difference that does not exist.** Both correct, and they were the same mistake twice: I wrote the withdrawal into the YAML everywhere and into the notes only once. The paragraph is rewritten. I also verified your AADACL2 claim programmatically rather than by eye — `AADACL2-goa.tsv`'s `GO:0016787` IBA WITH/FROM set is **identical token-for-token** to AADACL3's, same seventeen tokens, same node — so the "differs from AADACL2 by inheriting a molecular-function term" framing is simply false and is gone. This is the second time in this PR that a correction landed in *n−1* places, so the fix now ends with a grep sweep for seven stale phrases across both files; all seven are zero.

**3 — the conflict with the merged AADACL2 review is now stated, not implicit.** This was the item worth raising. `AADACL2-ai-review.yaml` resolves the same row as `MODIFY → GO:0017171` with `TERM_SCOPING_PROBLEM` / `GRANULARITY_MISMATCH`, reading its serine count as *licensing* the term where this review reads the HIDH exception as blocking a plain transfer. Both cannot be right about one row. Named in the notes reconciliation section and in the `suggested_questions` entry, with **#2266 flagged for follow-up** and the two supporting details recorded (AADACL2's notes still describe HIDH as "a dehydratase, not a hydrolase", which `f50b47fcd` corrected). I also wrote down the distinction I am relying on rather than leaving it as an assertion: `GO:0052689` already exists on the GOA from an independent subfamily signature, so modifying toward it **merges two existing rows**, whereas `GO:0017171` appears nowhere in the GOA and modifying toward it would *introduce* a claim the cited node does not license.

**🔵 `GRANULARITY_MISMATCH` — you are right and I over-corrected.** I dropped it on the strength of a "donors must agree" reading, but the schema (`gene_review.yaml:2469`) defines it as *"Parent term is true but uninformative"*, which is about informativeness for the target, and the row's own `reason` says exactly that. Restored on **both** hydrolase rows, so the IBA/IEA asymmetry you spotted is gone too. `root_cause` stays `EVIDENCE_CIRCULAR_OR_REDUNDANT`, and the row now spells out that the two fields describe different things: correct where the term was asserted, uninformative where it landed.

**🔵 "One exception is enough" overstated PAINT practice.** Annotating the ancestral node and marking the divergent descendant with a `NOT` is indeed the more direct answer, and it is now the *first* of three routes in the suggested question (NOT on HIDH · the tighter `PTN009058713` · a sub-node excluding HIDH). Section 7's preamble no longer says the term "can only be placed at a node whose members all have a serine nucleophile"; it says a single non-serine member makes a plain transfer unsafe, and notes the `NOT` route explicitly.

**🔵 "No human AADAC-like gene currently receives a term for it" was false.** Verified via QuickGO: AADAC has `GO:0017171` three times (IBA at `PTN002745055`, plus IEA and **IDA**) and NCEH1 once (IBA at `PTN002745068`). Reworded to name the AADACL paralogs, which was the intended reading — the earlier draft was unambiguous and I made it worse.

**🔵 `notes:354-358`** — "the uninformative parent of the term that should be there" was pre-correction framing; it now says the term is the correct LCA of its node's donors but uninformative for these genes, and that all three paralogs (not just AADACL3 and AADACL4) inherit it.

Detailed reasoning is in `AADACL3-notes.md` rather than repeated here.

**Gates** (`origin/main` merged first, it had advanced again): `checkquotes` 31/0 · `just validate human AADACL3` ✓ Valid with the one genuine warning · `analyze.py` byte-identical on rerun · terms.csv deletion check run **last** after the final validate, clean, no rows added, no re-sorting.

Final state: **3 ACCEPT, 2 MODIFY (both → `GO:0052689`), 0 REMOVE**. Nothing outstanding from my side; the only cross-gene action left is the #2266 follow-up, which I have flagged rather than acted on.

cmungall:
Both closing corrections folded in (`d6afba085`), since one of them was an internal contradiction and the other strengthens the #2266 hand-off. Verified against the GOA files rather than accepted on trust.

**"the same WITH/FROM sets" overshot.** Confirmed: `AADACL4-goa.tsv` membrane IEA is `ARBA:ARBA00028763|InterPro:IPR017157|UniProtKB-SubCell:SL-0162` against AADACL3's two tokens — which is exactly the `SL-0162` asymmetry the *next paragraph* records as corroborating the UniProt gap. So the blanket claim was contradicting my own finding twelve lines later. Narrowed to what holds: same five terms, WITH/FROM identical on both **IBA** rows, differing by one token on the membrane IEA.

**"two of them" undercounted, and the correction is load-bearing.** `GO:0052689` IEA from `InterPro:IPR017157` is on all three — `AADACL2-goa.tsv`, `AADACL3-goa.tsv`, `AADACL4-goa.tsv`. You were right that this matters beyond arithmetic, and I have drawn the consequence out rather than just fixing the number: **the remedy is portable to AADACL2.** Modifying `GO:0016787` toward `GO:0052689` merges two rows that *already exist* on each of the three records, whereas modifying it toward `GO:0017171` would introduce a term that appears on none of them. So the #2266 follow-up now has a concrete remedy to adopt rather than only a disagreement to adjudicate, and the conflict note no longer stops at "both cannot be right".

Also worth noting for whoever takes #2266: AADACL2 has **seven** GOA rows, not five, so it is not the row-for-row twin that AADACL4 is — the identity claim is scoped to AADACL4 and to the shared IBA rows.

Nothing else changed: **3 ACCEPT, 2 MODIFY (both → `GO:0052689`), 0 REMOVE**. `checkquotes` 31/0 · `just validate human AADACL3` ✓ Valid with the one genuine warning · terms.csv deletion check run last, clean.

Thanks for four rounds of genuinely load-bearing review. The two that changed the science were yours: the `non-hydrolases` clause, which forced the fold-vs-subfamily argument onto evidence the analysis could actually produce, and the missing provenance on the `IPR017157`/`PIRSF037251` pillar, whose fix is what surfaced HIDH's absence from the subfamily signatures and, in turn, the `Thr164` finding that made me withdraw my own `GO:0017171` proposal.

PR #2265, opened 2026-07-26
ABI3BP: give the matricellular term a real differentia (receptor class)

cmungall:
# ABI3BP: give the matricellular term a real differentia

Takes the 🔵 suggestion from the review of #2262. You were right, and the problem is sharper than a wording fix: removing the unmeasurable negative left the definition with **no** differentia at all. As merged it read

> …acts on cells by engaging cell-surface receptors, thereby modulating their signalling, adhesion, proliferation or differentiation state.

`GO:0008083` growth factor activity satisfies that in full — and, as you noted, plenty of growth factors are matrix-resident (FGFs, HB-EGF, latent TGF-β complexes). So the definition had collapsed to the parent `GO:0048018` plus a localization qualifier, and localization is a weak differentia for a molecular function. **The previous change was correct on its own terms but the intermediate state was a regression**, which is worth stating plainly rather than glossing.

## The fix: receptor class

The material was already in the `justification`, as you pointed out. Pulled into the definition:

> The activity of a matrix-resident secreted protein that acts on cells by engaging a cell adhesion receptor, such as an integrin, thereby altering the activation state of that receptor and modulating cell adhesion, cytoskeletal organisation, proliferation or differentiation. The differentia from the sibling ligand classes is the receptor engaged: a matricellular protein acts through an adhesion receptor rather than through a growth factor receptor kinase or a cytokine receptor, and its output is characteristically restraint of proliferation rather than stimulation of it. Whether a protein with this activity also contributes to the mechanical integrity of the matrix is a separate question, and this term neither asserts nor excludes it.

This is a criterion the siblings genuinely fail, checked against their definitions rather than their labels:

| Sibling | Definition (QuickGO) | Fails because |
|---|---|---|
| `GO:0005125` cytokine activity | "the activity of a **soluble** extracellular gene product…" | matricellular proteins are matrix-resident, not soluble-by-definition |
| `GO:0008083` growth factor activity | "The function that **stimulates** a cell to grow or proliferate" | ABI3BP restrains proliferation; the sign is inverted |
| `GO:0048018` (parent) | any receptor | no receptor-class restriction |

And the positive side is grounded in an existing GO concept rather than invented: `GO:0004895 cell adhesion receptor activity` — *"Cell adhesion receptors include integrins and…"* — so "adhesion receptor" is already a class GO recognises.

The unmeasurable negative is **not** reintroduced, and the structural-agnostic clause is kept, so the term remains consistent with this review's position that a structural role for ABI3BP has never been excluded, only never tested.

The `justification` now records all three candidate differentiae and why two were rejected (unmeasurable negative; matrix residence alone; receptor class as the one that survives), so an editor sees the reasoning rather than just the conclusion. The notes record the three-pass sequence as a pattern worth remembering.

## On the diff: most of it is not a claim change

The `description` block is reflowed to 98 columns as a side effect of an earlier width fix cascading. **That is pure rewrapping** — the edit script asserts the parsed `description` is identical after whitespace normalisation *before* writing, and fails otherwise, so no sentence in it changed. Only two things changed substantively: `proposed_new_terms[0].proposed_definition` and one paragraph of `proposed_new_terms[0].justification`.

No annotation action, term id, label, evidence code, qualifier, reference or `supporting_text` is touched. Still 18 rows = 7 ACCEPT + 3 MARK_AS_OVER_ANNOTATED + 8 NEW.

## Verification

- Edits applied by a script asserting each anchor is present **and unique** before replacing, re-grepping after, plus the description text-identity assertion above; YAML re-parsed to confirm 18 annotations / 10 questions intact.
- `checkquotes.py` — **100 quotes, 0 problems**
- `just validate human ABI3BP` — **✓ Valid, zero warnings**
- `cache/go/terms.csv` — **no change on this branch**; `git diff origin/main HEAD -- cache/go/terms.csv` is empty (the four new terms landed with #2250, and `GO:0004895`/`GO:0008083`/`GO:0005125` appear only in prose, not in a validated slot)
- Branched fresh off `origin/main` after #2262 merged.

### DIFF
--- a/genes/human/ABI3BP/ABI3BP-ai-review.yaml
+++ b/genes/human/ABI3BP/ABI3BP-ai-review.yaml
@@ -9,25 +9,25 @@ description: >-
   ABI3BP (also called TARSH, target of Nesh-SH3) is a large secreted glycoprotein that is
   deposited into the extracellular matrix, where its characterised action is on cells rather than
   on the mechanical fabric of the matrix. The 1068-residue precursor carries an N-terminal signal
-  peptide, two fibronectin type-III domains near either end, a family-specific C-terminal domain shared with FNDC1, and
-  a very large intrinsically disordered central region that is dense in proline-rich SH3-binding
-  motifs and is the site of extensive alternative splicing. It has no catalytic domain. Its
-  characterised receptor is integrin beta-1: matrix-deposited ABI3BP binds integrin beta-1 and
-  holds it in the activated state, which drives paxillin phosphorylation, sequesters Src and
-  ERK at focal adhesions, and thereby suppresses cyclin-D1 induction and S-phase entry. The
-  consequence is a coupled switch in stem and progenitor cells - proliferation is restrained and
-  differentiation is licensed. Mesenchymal stem cells lacking ABI3BP over-proliferate and fail to
-  undergo osteogenic or adipogenic differentiation; cardiac progenitor cells lacking it expand but
-  fail to express cardiomyocyte markers, and hearts lacking it recover poorly from infarction.
-  Consistent with this brake on growth, ABI3BP expression is lost or epigenetically silenced in
-  lung, thyroid and gallbladder tumours, and restoring it suppresses proliferation, motility and
-  epithelial-mesenchymal transition. In the nervous system secreted ABI3BP acts as a diffusible
-  cue that reduces dendritic complexity of olfactory mitral cells during postnatal refinement.
-  The protein is broadly expressed, most highly in lung, and is recovered from the interstitial
-  matrix of aorta, vein, atrial myocardium and cartilage. It is cleaved by thrombin at Arg-337,
-  and its levels are additionally set post-translationally by proteasomal degradation, giving it
-  a half-life of about one hour. Reports of its relationship to cellular senescence point in both
-  directions in different cell types and are not yet reconciled.
+  peptide, two fibronectin type-III domains near either end, a family-specific C-terminal domain
+  shared with FNDC1, and a very large intrinsically disordered central region that is dense in
+  proline-rich SH3-binding motifs and is the site of extensive alternative splicing. It has no
+  catalytic domain. Its characterised receptor is integrin beta-1: matrix-deposited ABI3BP binds
+  integrin beta-1 and holds it in the activated state, which drives paxillin phosphorylation,
+  sequesters Src and ERK at focal adhesions, and thereby suppresses cyclin-D1 induction and
+  S-phase entry. The consequence is a coupled switch in stem and progenitor cells - proliferation
+  is restrained and differentiation is licensed. Mesenchymal stem cells lacking ABI3BP
+  over-proliferate and fail to undergo osteogenic or adipogenic differentiation; cardiac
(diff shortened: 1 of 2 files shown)


cmungall:
Fixed in `9f36a76a2`. You were right, and this one would have sunk the term request rather than merely weakened it.

**The clause was true of ABI3BP and false of the class.** I checked the exemplars I had myself cited two paragraphs earlier: tenascin-C, periostin (αVβ3/αVβ5) and CCN1/CCN2 (αVβ3/α6β1) are characteristically **pro**-proliferative. So "its output is characteristically restraint of proliferation" would have excluded most of the proteins the same justification offers as the reason the term is needed — a definition contradicted by its own motivating set. And you are right that a differentia hedged with "characteristically" is not usable as a differentia at all. Deleted; receptor class alone does the whole exclusion job, so nothing is lost.

I also took your point about placement and removed the sibling comparison from the definition entirely. It is reasoning *about* the ontology, not part of the class definition, `ProposedOntologyTerm` has no `comment` slot, and it already appeared near-verbatim in `justification`. The definition is now just the class and the structural-agnostic clause:

> The activity of a matrix-resident secreted protein that acts on cells by engaging a cell adhesion receptor, such as an integrin, thereby altering the activation state of that receptor and modulating cell adhesion, cytoskeletal organisation, proliferation or differentiation. Whether a protein with this activity also contributes to the mechanical integrity of the extracellular matrix is a separate question, and this term neither asserts nor excludes it.

**Both of your other points taken, and the GO-CAM one verified at source.** I checked `gocams/index.tsv:8672` rather than taking it on trust:

```
UniProtKB:P05556  ITGB1 Hsap  689e7a5d00003515  ...  GO:0004895  cell adhesion receptor activity  GO:0007229  integrin-mediated signaling pathway  GO:0005925  focal adhesion
```

That is a genuinely better argument than the one I had. `GO:0004895` is not merely a plausible class for an adhesion receptor — production GO already assigns it to **the exact receptor ABI3BP engages**, in a model about negative regulation of integrin-mediated adhesion, with `focal adhesion` as the component. The proposed differentia therefore matches existing curation instead of asserting a new typing. Now cited in `justification`.

The reasoner-checkable form is offered too: `GO:0048018 receptor ligand activity and has-target some GO:0004895 cell adhesion receptor activity`, with matrix residence recorded as a non-defining characteristic rather than smuggled into the differentia.

**On the four passes this definition has now taken** — unmeasurable negative → no differentia at all → a differentia that over-claimed at class level → receptor class alone. Each individual change was correct and two of the intermediate states were regressions, which is worth recording rather than quietly converging. The notes now do, with the generalisable form: when tightening a class definition, test it against the *other* members of the class you cited as motivation, not only against the gene in front of you. That is exactly the check I skipped, and it is the same failure shape as the `GO:2000738` inversion earlier in this gene — reasoning from the label and the single case instead of from the definition and the set.

`checkquotes.py` 100 quotes 0 problems; `just validate human ABI3BP` ✓ Valid zero warnings; no `cache/go/terms.csv` change (`git diff origin/main HEAD` empty for that path). No annotation action, term id, label, evidence code, qualifier, reference or `supporting_text` touched anywhere in this PR — still 7 ACCEPT / 3 MARK_AS_OVER_ANNOTATED / 8 NEW.

PR #2266, opened 2026-07-26
Review human AADACL2: an intact catalytic triad, and a PANTHER node placement that is exactly backwards

cmungall:
## AADACL2 (Q6P093) — human arylacetamide deacetylase-like 2

The AADAC family's dark, skin-restricted member. **Seven GOA rows, none experimental** (2 IBA, 5 IEA): three assert an activity at three different levels of generality, and three assert a location — two of which cannot both be true.

This gene is a clean test case for both halves of the campaign's "a domain's name is not an activity" rule, and the two halves came out differently.

### The activity is real machinery, not a fold-derived label

AADACL2's own record annotates a `Ser189-Asp341-His371` triad in a `GDSSG` nucleophile elbow with an `HGG` oxyanion loop and a PROSITE `PS01174` GDXG-lipase-serine match. Because those features carry `ECO:0000250` (by similarity), they were re-tested rather than accepted: align AADACL2 to every characterised protein in its **own** `WITH/FROM` column, then read off what AADACL2 carries at that protein's annotated `ACT_SITE` positions — counting a hit only when the residue matches **and** lands on one of AADACL2's own annotated sites, because at 20% identity a global alignment will happily park a source Asp on some unrelated Asp.

**8 of 15 sources place all three catalytic residues on exactly S189/D341/H371** — the seven closest by identity plus *M. tuberculosis* LipN (26.1%), with every failure at 26.5% identity or below, where the alignment loses register in the C-terminal half. The corroborating set includes human/mouse/rat AADAC (50–51.6%), mouse Nceh1, and three independently characterised mycobacterial carboxylesterases.

So the three vague activity rows are **MODIFY → `GO:0017171` serine hydrolase activity**, whose definition ("a catalytic triad consisting of a serine nucleophile that is activated by a proton relay involving an acidic residue … and a basic residue") is satisfied residue for residue. `GO:0052689` carboxylic ester hydrolase activity, from the family-specific `IPR017157` signature rather than a bare fold, is **ACCEPT**ed.

### The mirror-image error was avoided too — the localisation is genuinely unresolved

The two membrane rows look at first like a textbook paralog-topology artefact: of the **14 reviewed UniProtKB entries carrying `IPR017157`, 12 carry a `TRANSMEM`** — 11 of them annotated explicitly as "Signal-anchor for type II membrane protein", the exception being mouse Aadacl3 with three plain helices — **and AADACL2 is the only one UniProt gives a cleaved signal peptide instead**. Its three `WITH/FROM` protein sources (human AADAC, mouse Aadac, mouse Nceh1) are all type-II anchored.

Measurement did not support removing them:

- UniProt's cleavage site puts **His at −1**, which signal peptidase I essentially does not accept (−1 is A/S/C/G/T in the great majority of sites).
- The hydrophobic core is **more** hydrophobic than AADAC's experimentally validated anchor (peak Kyte-Doolittle over a 19-residue window: **1.94** for AADACL2 1–19 vs **1.81** for AADAC 5–23), and the two N-termini are homologous across the whole segment with the same downstream `TP.PDN.EE.W` motif.
- HPA's secretome vote does put AADACL2 — and only AADACL2 of the five paralogs — in the secreted class, but those are SignalP-family predictors like the one UniProt used, **and the same HPA pipeline fails to call AADAC a membrane protein at all**, so it cannot be used as independent corroboration.
- AADACL2 is `PE 1`, so it was worth asking what the peptides covered: all four observed peptides lie in the catalytic domain, earliest residue **124**. Mass spectrometry establishes expression and does not test the N-terminal fate.

All three location rows are therefore **UNDECIDED**. Recording the contradiction — GOA asserting `GO:0005576` and `GO:0016020` for the same protein, from three pipelines, both readings of one hydrophobic helix — is the deliverable; picking a winner would have been a guess dressed as a finding. `suggested_experiments` carries the protease-protection / conditioned-medium / N-terminal-sequencing design that would settle it.

### The generalisable finding: PAINT's node placement is inverted

Comparing the IBA annotations of all five human family members and the node each came from:

| term | AADAC | NCEH1 | AADACL2 | AADACL3 | AADACL4 |
|---|---|---|---|---|---|
| `GO:0017171` serine hydrolase activity | `PTN002745055` | `PTN002745068` | — | — | — |
| `GO:0016020` membrane | — | `PTN009058713` | `PTN009058713` | `PTN009058713` | `PTN009058713` |
| `GO:0016787` hydrolase activity | — | — | `PTN009058710` | `PTN009058710` | `PTN009058710` |

The well-conserved **mechanism** term sits only at ortholog-specific nodes, so AADACL2/3/4 inherit none of it; what they inherit at the shared family node is the poorly-conserved **localisation**. Their only MF comes from the far deeper `PTN009058710`, whose 16 sources span three Arabidopsis CXE carboxylesterases, Arabidopsis ICME, mouse Afmid and yeast BNA7 (amide hydrolases), *E. coli* Aes, three *M. tuberculosis* esterases and soybean HIDH — a node so broad that bare `hydrolase activity` is all it can license.

The triad is conserved family-wide; the type-II anchor is not. **Moving `GO:0017171` down to `PTN009058713` and dropping `GO:0016020` from it would correct AADACL2, AADACL3 and AADACL4 in one edit.**

### Actions

| term | evidence | action |
|---|---|---|
| `GO:0016787` hydrolase activity | IBA | MODIFY → `GO:0017171` |
| `GO:0016020` membrane (`is_active_in`) | IBA | UNDECIDED |
| `GO:0003824` catalytic activity | IEA (ARBA) | MODIFY → `GO:0017171` |
| `GO:0005576` extracellular region | IEA (SubCell) | UNDECIDED |
| `GO:0016020` membrane (`located_in`) | IEA (ARBA + InterPro) | UNDECIDED |
| `GO:0016787` hydrolase activity | IEA (`IPR013094`, bare fold) | MODIFY → `GO:0017171` |
| `GO:0052689` carboxylic ester hydrolase activity | IEA (`IPR017157`, family) | ACCEPT |

**3 MODIFY, 3 UNDECIDED, 1 ACCEPT.** Plus 1 `core_function` (deliberately with no `locations`), 3 `knowledge_gaps` (WHOLLY_DARK substrate gap, CC_DARK topology gap, MF_DARK curation gap for the node placement), 4 `suggested_questions`, 3 `suggested_experiments`.

### Notes on evidence handling

- **The empty affinage record was not taken on trust.** `gates_passed: True`, `n_discoveries: 0`, `citation_count: 0`. Verified separately: PubMed returns 6 AADACL2 hits, none functional, and the three UniProt `RN` references are all large-scale sequencing. The positive confirmation is PMID:35736449, which states in prose that "*all other members of the AADACL protein family, AADACL2, AADACL3, and AADACL4, have so far been only poorly investigated and no functional roles can be concluded*".
- Two genetic associations exist and neither licenses an annotation: an exome-indel psoriasis screen naming AADACL2 (PMID:31078570, tissue-coherent with skin-restricted expression) and a SIREN stroke GWAS locus explicitly *near* the gene (PMID:38317187).
- **`AADACL2-bioinformatics/`** — reproducible audit (`uv run --no-project --with requests --with biopython python audit_aadacl2_record.py`). Everything fetched at run time from UniProt, InterPro, QuickGO, EBI Proteins and HPA plus this gene's own `AADACL2-goa.tsv`; the script aborts with the regeneration command if that TSV is missing rather than dropping a section. `results.json` was verified to reproduce byte-identically on re-run. The analysis reports a negative result honestly — the hydropathy and cleavage-site measurements undercut the initial hypothesis that the membrane annotation was a simple paralog artefact.
- One UniProt correction to report upstream: the `SIGNAL 1..18` cleavage position places His at −1, which is not a signal-peptidase-I site.
- `cache/go/terms.csv` unchanged (`GO:0017171` was already cached); 0 deletions vs `origin/main`.


cmungall:
Thanks — all five suggestions taken, pushed as 68bed470. Verified each against the tree rather than asserting it.

**1. The stronger nucleophile claim.** You were right that the aggregate buried the best number, and that the data was already in the file. The script now computes it explicitly (`q1_catalytic_machinery.per_target_active_site`) rather than my reading it off by hand:

| AADACL2 position | residue | sources aligning here | with the identical residue |
|---|---|---|---|
| 189 | Ser | **15 / 15** | **14 / 15** |
| 341 | Asp | 8 / 15 | 8 / 15 |
| 371 | His | 8 / 15 | 8 / 15 |

One detail worth adding to your framing: the 15th source, the one without a Ser at 189, is soybean HIDH, whose own annotated nucleophile is Thr164 — and HIDH is the single source in the whole set that is not a hydrolase but a 2-hydroxyisoflavanone **dehydratase**. So the one substitution is the one case where the chemistry differs, which makes it evidence rather than noise. Added as a table in `RESULTS.md` section 1 with that reading, and surfaced in the `GO:0017171` MODIFY reason.

**2. The contradiction between the two "best statement" claims.** Real, and it was mine, not a wording accident — I had two defensible sentences that could not both be true. Reconciled in both places on the substantive ground rather than by softening the language: `GO:0052689` is the most informative *activity* statement on the record, but it is inferred from family membership; `GO:0017171` is the part evidenced on this protein's own residues, **and** the part that survives if AADACL2 turns out to prefer amide bonds — which is not hypothetical, since its closest characterised relative AADAC's name-giving activity is amide hydrolysis of arylacetamides. That asymmetry is now the stated reason `core_functions` uses the mechanism term.

**3. "Mutually exclusive" overstates GO semantics.** Accepted without reservation — you are right that the terms are not disjoint and that the ER lumen is topologically continuous with the extracellular space. Narrowed in all four places (`GO:0005576` reason, `GO:0016020` IEA reason, knowledge gap 2 significance, notes) to what the evidence supports: two incompatible readings of one hydrophobic helix, only one of which can describe the mature protein. GO not forbidding both terms is now said explicitly, so the claim is about this protein rather than about the ontology.

**4. Where the InterPro2GO fix lands.** Added. The `IPR013094` → `GO:0016787` mapping should *not* change — most proteins matching a bare alpha/beta-hydrolase-3 domain warrant nothing more specific, and saying otherwise would break the signature for everything else that carries it. The correction is a protein-level annotation on AADACL2 made from its own residue-level evidence, superseding the signature-level default. Stated in the reason so the MODIFY cannot be read as a request to remap the signature.

**5. `PMID:16641997`.** Fetched, cached and added to `references:` with `relevance: LOW`, so the "all three UniProt `RN` references are large-scale sequencing with no functional content" claim is now checkable inside the YAML instead of only in the notes.

Verification on the pushed tree: `checkquotes` 46/46 including the two new quotes; `just validate human AADACL2` → `✓ Valid` with no warnings; `results.json` re-runs byte-identical after the script change (`diff` clean), so the new table in `RESULTS.md` is not hand-maintained; `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → 0.

Nothing declined.

cmungall:
Correction to my previous comment: PR #2266 was merged at 03:01:37Z, moments before commit `68bed470` (the five follow-ups) landed on the branch, so that delta did **not** reach main with this PR. It is now open as #2272 against current main, with the same content and verification. Apologies for the misleading "pushed as 68bed470" — the commit existed on the branch but the branch was already merged.

PR #2267, opened 2026-07-26
Add the BAR domain's own function to ACAP1, and unlock PMID:25284369's full text

cmungall:
Follow-up to the merged ACAP1 review (#2251). Small and additive: it adds one demonstrated molecular function that GOA lacks, and it fixes a reference flag that was hiding the evidence for the function #2251 already proposed.

### First, a correction to my framing

I was briefed to add `GO:0180020 membrane bending activity`, which is indeed absent from ACAP1's GOA (0 hits in `ACAP1-goa.tsv`; GOA carries only `GO:0005096`, `GO:0005515`, `GO:0005737`, `GO:0005886`, `GO:0010008`, `GO:0016020`, `GO:0030036`, `GO:0055038`). **It was already proposed in #2251**, correctly attributed to the PH domain. So this PR does not add it; it adds the *other* half of the same finding and repairs the evidence beneath both.

### 1. NEW `GO:0042803 protein homodimerization activity`, IDA, PMID:25284369

The BAR–BAR interaction is the **only demonstrated function of ACAP1's BAR domain**, and annotating it is what makes the PH-domain attribution of `GO:0180020` legible rather than surprising: the two rows together encode the paper's whole point — *PH domain bends, BAR domain self-associates*.

I was warned that a `GO:0046982` heterodimerization claim was withdrawn on ACAP3 because AP-MS measures co-complex membership, not stoichiometry. That objection does not apply here, and I want to be explicit about why rather than hand-wave it. Three orthogonal readouts each **resolve the oligomeric state**, not merely detect association:

- 2.2 Å crystal structure: a banana-shaped dimer in one asymmetric unit, interface 4100 Å², largely hydrophobic;
- size exclusion chromatography **with light scattering** (absolute mass, not elution position);
- quantitative gel filtration giving a dissociation constant **below 5 µM**, tighter than the published value for the N-BAR protein endophilin A2.

Crystallography + SLS + a measured Kd say *dimer*. An affinity capture would only have said *interacts*.

**Two limits deliberately respected**, per the instruction to prefer a defensible parent over an overstated child:

- **Dimer, not lattice.** The paper's own wording for the BAR–BAR function is "clustering at the membrane", and it describes tetramers packing helically on the tubule. But the tetramer and the lattice come from fitting the crystal structure into cryoEM maps of ~14 Å (Class I) and ~17 Å (Class II), and *clustering of an unspecified number of subunits is not a dimer*. I annotate the dimer and leave the lattice in prose rather than reaching for an oligomerisation or complex-assembly term the resolution does not support.
- **Fragment, not full length.** Everything was measured on BAR-PH (1–377 in the cryoEM work, 1–364 modelled), so this is a domain-attributed statement about ACAP1, not a demonstration on the intact protein.

Recorded as **non-core** on purpose: homodimerisation is the structural precondition for the curved scaffold, not an output, and it is already described inside the membrane-bending core function. Promoting it would give a prerequisite the same weight as the GAP activity and the cargo-sorting activity.

### 2. PMID:25284369 was wrongly flagged `full_text_unavailable`

`publications/PMID_25284369.md` carries `full_text_available: true`, `pmcid: PMC4198613`, `full_text_extraction_method: xml`. #2251 flagged the reference unavailable and drew its quotes from the abstract. Flag removed; **nine full-text findings added** (3 → 12).

This is not bookkeeping. The abstract states the *conclusion* about the domain division of labour; only the Results give the **assay that makes the annotation an IDA** — liposome binding plus negative-stain EM tubulation of purified protein — and the **negative BAR-domain result in two independent readouts** (liposome sedimentation, and the cryoEM reconstruction of a coated tubule). An IDA supported only by a conclusion sentence is weaker than one supported by the assay, and the assay was sitting unread in the cache.

I swept the other 14 PMIDs in this review for the same defect: **all consistent**, this was the only mismatch.

### 3. The domain attribution, and why it is the easy thing to get wrong

This is the counter-intuitive part, and the reason it deserves emphasis in the row rather than only in the notes:

- ACAP1's **BAR domain can neither bind membrane nor impart curvature** — *"the BAR domain of ACAP1 also showed little affinity to the generated liposomes, regardless of their size"*, and in the cryoEM map *"there is no significant interaction between the BAR domain and the underlying membrane."*
- The bilayer contact and the curvature both come from the **PH domain**, specifically Loop1, whose distal aromatic **Phe-280** inserts into one leaflet: F280A reduces both binding and tubulation, F280E effectively abolishes binding, conservative F280W preserves both.
- The BAR domain instead **binds the BAR domains of neighbouring ACAP1 molecules**.

So "its BAR domain bends membrane" — the phrasing a reader would default to for *any other* BAR protein — is simply false for ACAP1, and "the BAR-PH module tubulates membrane" is right about the fragment and wrong about the mechanism.

`GO:0180020`'s row and core function now carry the assay quotes, the Phe-280 mechanism and both negative results, sourced from the paper rather than from a UniProt feature line.

### 4. QuickGO verification of the "no BAR protein on this term" claim

I was told not to repeat this claim unless I could confirm it. Confirmed — `goId=GO:0180020&goUsage=descendants&taxonId=9606`: **4 annotations, 3 gene products, all IDA.**

| Gene product | Evidence | Reference |
|---|---|---|
| UniProtKB:O43633 CHMP2A | IDA | PMID:36604498 |
| UniProtKB:Q9Y3E7 CHMP3 | IDA | PMID:36604498 |
| UniProtKB:O60313 OPA1 | IDA | PMID:32228866 |
| UniProtKB:O60313 OPA1 | IDA | PMID:37612504 |

CHMP2A/CHMP3 are ESCRT-III subunits; OPA1 is a dynamin-family GTPase. **No BAR-domain protein carries the term**, so ACAP1 would be the first — which is exactly why the PH-domain attribution has to be spelled out on the row instead of letting it read as a routine BAR-protein annotation.

### Gates

- `checkquotes.py` — **125 quotes, 0 problems** (re-run after merging `origin/main`)
- `just validate human ACAP1` — **✓ Valid**, 1 pre-existing warning (no annotation cites the affinage file; it is cited in `references`, deliberately `UNVERIFIED`, and no sentence from it is used as supporting text anywhere in this review)
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` — **empty**; `GO:0042803` was already in the cache, so the file is untouched
- Every claim above re-grepped against the tree with a 23-assertion script; diff is **2 files, +233/−11**, ACAP1 only


cmungall:
Thanks — took all three. Suggestion 1 was a real overstatement on my part, not a wording preference, so it is worth spelling out what changed.

**1. "Inserts into one leaflet" was mine, not the paper's.** You are right that the word appears nowhere in PMID:25284369, and I had asserted it three times as settled fact. Unpacking the source:

- Insertion *into the membrane* is demonstrated — the 5A(276–280) and F280 mutant series act on both binding and tubulation, and the cryoEM map gives *"only one monomer of the ACAP1BAR-PH dimer was observed to insert its Loop1 into the membrane."*
- The **single-leaf** part is a molecular-dynamics-derived model, and the paper hedges it: *"As such, the insertion of F280 into one leaf of the bilayer could reduce membrane surface tension asymmetrically and thereby induce membrane deformation"*. Note also "leaf", not "leaflet".
- Worth adding that Loop1 is introduced even more weakly at first mention — *"which could potentially insert into the membrane"* — so the insertion itself rests on the mutagenesis and the cryoEM density rather than on the paper's structural prediction.

Reworded in all three places (`GO:0180020` row, the membrane-bending core function, `ACAP1-notes.md`) so the demonstrated claim is insertion into the membrane and the asymmetry is attributed to the simulation. Added the two anchoring quotes and Loop1's residue range (276–282, between β1 and β2), which I had left implicit.

**2. IDA vs IPI on `GO:0042803`.** Kept IDA, but the row now says why, because you are right that the pairing looks odd: IPI covers interaction *detection* (two-hybrid, co-purification, co-IP), and detection reports co-complex membership, which cannot support a stoichiometric term at all. Crystallography, static light scattering and a measured Kd assay the oligomeric state of purified protein directly. That is the same distinction that separates this row from the withdrawn ACAP3 heterodimerization claim, so having the evidence code carry it explicitly is an improvement rather than a defence of the status quo. `supporting_entities` retained since the self-reference is still the useful machine-readable statement of *what* it dimerises with.

**3. Row summaries trimmed.** Fair — they had drifted into arguing with a hypothetical reader ("worth stating twice", "the phrasing a reader would default to"). Meta cut from both `review.summary` blocks, substance kept (which domain does what, which readouts, what is deliberately not claimed). The argument now lives only in `ACAP1-notes.md` §9, where the repo convention puts it.

On your runner note: I re-ran both gates locally after merging `origin/main` — `checkquotes.py` reports **127 quotes, 0 problems**, and `just validate human ACAP1` is **✓ Valid** with the one pre-existing affinage warning. The QuickGO table you could not reach was queried at `goId=GO:0180020&goUsage=descendants&goUsageRelationships=is_a,part_of&taxonId=9606`, returning `numberOfHits: 4` over CHMP2A, CHMP3 and OPA1; I agree it is background prose rather than annotation-bearing evidence, which is why no `supporting_text` rests on it.

PR #2268, opened 2026-07-26
Review human ACRBP: GENCODE annotates the splice variant the literature says humans lack

cmungall:
## ACRBP (Q8NEB7) — acrosin-binding protein / sp32 / OY-TES-1

Complete review of all 14 GOA rows (13 distinct annotations — `GO:0005576` ISS from
`UniProtKB:Q29016` appears twice in GOA with different dates and is reviewed once), plus four
proposed new annotations and two GO term requests.

`gates_passed: True` on the affinage record; all 12 of its citations resolve to numeric PMIDs.

### The main finding is an unreconciled discrepancy

Human `GO:0001675 acrosome assembly` is an ISS/IEA transfer out of the mouse *Acrbp* knockout
(PMID:27303034). In that paper the acrosomal-granule defect is rescued by transgenic **ACRBP-V5
alone**, and the same group had reported ACRBP-V5 as an intron-5-retaining variant *not* found in
other mammalian ACRBPs (PMID:23426433). That is the shape of a classic isoform-mismatch
over-annotation: a function transferred from a splice form the target species does not make. The
bioinformatics analysis in this PR was written specifically to demonstrate it, and at the
**sequence level** it found the opposite:

| | human | mouse |
|---|---|---|
| V5-like transcript | **ACRBP-204 / ENST00000536350** | Acrbp-205 / ENSMUST00000112414 |
| protein length | 319 aa (0.587 of canonical) | 316 aa (0.585 of canonical) |
| exon chain | first 5 canonical exons | first 5 canonical exons |
| terminal exon | extended **82 bp** into intron 5 | extended **242 bp** into intron 5 |
| C-terminal pentapeptide | `SLLQL` → **`RYRKF`** at 315-319 | `SLQQL` → **`RYRKL`** at 312-316 |

The mouse change is exactly UniProt `VSP_051965`/`VSP_051966`, i.e. ACRBP-V5. Same intron, same
termination point, same substituted pentapeptide differing only in the last residue — one conserved
splicing event, not a rodent innovation.

**But the expression-level half of the objection survives, and is stronger than PMID:23426433 alone
implies.** PMID:30606959 — same laboratory, three years later — states the human negative outright:
"Porcine, guinea pig, and human spermatogenic cells produce only a single form of Acrbp (termed
Acrbp-W) mRNA, whereas two mRNA forms … are synthesized by pre-mRNA alternative splicing of the
Acrbp gene in mouse". That is an RT-PCR claim about what is transcribed, and it outranks a genome
annotation on exactly that question. Credit to the reviewer for catching it: the sentence is in a
cached full text this review already cited four times, and the first draft missed it and overstated
its conclusion accordingly.

So the honest result is a **discrepancy, not a refutation**: GENCODE annotates `ENST00000536350`,
the primary literature reports the corresponding mRNA as absent from human spermatogenic cells, and
nobody has looked for the protein. That is recorded as a `knowledge_gaps` entry whose `resolution`
is the experiment that would adjudicate it.

`GO:0001675` is nonetheless **ACCEPT**, for a reason that never depended on the isoform argument:
the mouse source is an IMP on a knockout of the **whole gene**, ablating both forms, so what
transferred is a gene-level requirement for normal acrosome formation rather than a V5-specific
activity. UniProt's curator reading of the same full text also credits the mature form with
partially contributing to acrosomal-granule assembly.

### Actions

7 ACCEPT · 4 KEEP_AS_NON_CORE · 1 MODIFY · 1 MARK_AS_OVER_ANNOTATED · 4 NEW

**MODIFY `GO:0002080 acrosomal membrane` → `GO:0043159 acrosomal matrix`.** The human ISS row
transfers pig Q29016's `GO:0002080` IDA, whose method (PMID:15955892) is indirect
immunofluorescence — light microscopy of a sub-micron organelle cannot separate the matrix from the
membrane bounding it. That is a resolution limit, not a curation error, but `located_in` a membrane
is not supportable for this protein: all five reviewed orthologs carry a cleaved signal peptide and
none carries a transmembrane, intramembrane or lipidation feature. `GO:0043159` is `part_of`
`GO:0001669`, is more informative, and is the term the primary literature itself uses ("An acrosomal
matrix protein, ACRBP…"). Rat Q6AY33 and guinea-pig Q60485 carry the same term by the same ISS route
from the same pig row, so the correction applies to four entries.

**MARK_AS_OVER_ANNOTATED `GO:0005634 nucleus`** (HDA, PMID:21630459 sperm-nucleus proteome).
Retained rather than removed — the mass-spectrometric identification is credible — but a
signal-peptide-bearing secretory-pathway protein has no route to the nucleoplasm, and the acrosome
is physically apposed to the anterior face of the sperm nucleus. Consistent with how this PMID is
handled for AAAS, ACADS and ACADM in this repository.

**Four NEW annotations.** Human ACRBP had **no molecular-function annotation of any kind**, despite
being named for what it binds; rat and guinea pig have none either, and only mouse has one
(`GO:0016504`, ISA). Direct QuickGO queries confirm that `GO:0035375`, `GO:0061135`, `GO:0097341`
and `GO:0060046` each return **zero** annotations across all five ACRBP orthologs.

- `GO:0035375 zymogen binding` (MF, ISS) — the documented specificity is the informative part: pig
  sp32 binds the 55/53/49-kDa (pro)acrosin forms and *not* the 43-kDa intermediate or 35-kDa mature
  enzyme. Preferred over `GO:0005515 protein binding` for that reason.
- `GO:0061135 endopeptidase regulator activity` (MF, ISS) — deliberately **sign-neutral**. ACRBP-W
  blocks proacrosin autoactivation in the acrosome, yet purified sp32 *accelerates* autoactivation
  at basic pH in vitro. MGI's choice of the activator direction for mouse captures the in vitro
  result and is contradicted in sign by the in vivo result from the same laboratory 22 years later,
  so the unsigned parent is the strongest claim both datasets support.
- `GO:0097341 zymogen inhibition` (BP, ISS) — an exact existing term ("prevention of zymogen
  activation") for the in vivo role, unused across the whole family. Sign-specific here because the
  in vivo evidence is unambiguous.
- `GO:0060046 regulation of acrosome reaction` (BP, ISS) — the specific step the broad
  `GO:0009566 fertilization` row leaves unnamed. Supported by loss of function in mouse (deficient
  sperm reach the oviduct with normal morphology and motility yet cannot access oocytes) and by
  acute antibody blockade in boar.

### WITH/FROM resolution

Every accession resolved. Pig Q29016 (81.9% identical to human) and mouse Q3V140 (75.2%) are true
1:1 orthologs with their own experimental annotations; `PANTHER:PTN001085565` has one reviewed
member per mammalian genome, so `WRONG_ORTHOLOG_OR_PARALOG` is unavailable as a failure mode
anywhere on this gene; `SL-0007`/`SL-0243` map correctly. The IBA is **not** self-referential — its
with/from cites mouse and pig, not human ACRBP — so `root_cause: NO_FAILURE_CORE`.

### Two GO term requests

- **positive regulation of zymogen activation** — `GO:0097341` has no positive sibling, so the
  acceleration half of ACRBP's behaviour can only be recorded at the generic `GO:0010954`.
- **acrosomal granule assembly** — `GO:0001675` is defined as "the formation of the acrosome from
  the spermatid Golgi" and does not cover condensation of cargo into the granule, which is the
  actual ACRBP-null defect. `GO:0043159` already exists as the matching component.

### Bioinformatics

`genes/human/ACRBP/ACRBP-bioinformatics/` — all figures fetched live from the UniProt and Ensembl
REST APIs, raw responses cached under `data/`, nothing hard-coded, missing input is a hard error.
Verified that a fresh run reproduces `RESULTS.md` byte-identically from both the cache and a
`--refetch`. Every read-out sentence is computed from the data, which is how the two draft claims
that contradicted the tables (rat also has the isoform; human has 8 coding transcripts) were caught
before they reached the review.

### Self-correction made while addressing the review

`PMID:23426433` had been marked `DISPUTED` on the strength of the GENCODE annotation. That was
wrong — a genome annotation does not outrank an expression experiment — and it is now `VERIFIED`,
with the conflict recorded as a knowledge gap rather than as a fault in the reference. No reviewer
asked for this; it followed from the same reasoning.

### Verification

- `checkquotes.py`: 60 quotes checked, 0 problems (including all `file:` quotes, which
  `just validate` skips)
- `just validate human ACRBP`: ✓ Valid, no warnings
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'`: empty (7 rows had been dropped;
  restored from `origin/main` and only the 6 new terms re-inserted, without re-sorting existing rows)
- all 14 GOA rows accounted for, in GOA order, programmatically cross-checked against the TSV


cmungall:
Thank you — this is a good catch and the criticism is correct. The sentence is in the introduction of
`publications/PMID_30606959.md`, a paper I cited four times for the acrosome-reaction mechanism
without reading past the abstract. It states the human negative directly, from the same laboratory,
three years after the knockout paper, and it does outrank a genome annotation on the question of what
is transcribed. My conclusion was overstated in exactly the way you describe.

All three requested fixes are in `f774137e8`, plus one self-correction that followed from the same
reasoning. Each verified against the tree rather than asserted:

**1. The two `GO:0001675` rows, `summary` (lines 172-173 and the ISS row).** Rewritten to separate
the two halves of the objection. The sequence-level half fails; the expression-level half survives
and is independently reaffirmed. The phrase "the isoform-mismatch objection therefore fails and the
term stands" is gone from both rows (`grep`: 0 hits), replaced by an explicit statement that the
honest summary is a discrepancy rather than a refutation. Both rows now carry the quote as
`supported_by`.

The rows also now say why `ACCEPT` is right on grounds that never depended on the isoform argument:
the mouse source is an IMP on a knockout of the **whole gene**, ablating both forms, so a gene-level
requirement transferred — plus UniProt's curator reading of the same full text credits the mature form
with partially contributing to granule assembly. That reason was implicit before; it is now the stated
basis for the action, so the row does not lean on the analysis for more than it can bear.

**2. The `knowledge_gaps` boundary.** It previously read "Not established: that the human transcript
is expressed and translated in spermatids", which as you say inverts the evidential situation. It now
states both sides — the published experimental negative at the mRNA level *and* the conflicting
GENCODE annotation — and says explicitly that "this is not a plain absence of data: there is a
published experimental negative at the mRNA level and a conflicting genome annotation, and they have
never been reconciled". The `gap_statement` now opens by noting the two lines of evidence point in
opposite directions, `provenance` cites the quote, and `resolution` frames the long-read experiment as
*adjudicating* the discrepancy rather than filling a void.

**3. `RESULTS.md`.** Now three limits rather than two, with the quote reproduced in full and the
conclusion stated as "a **discrepancy between GENCODE and the experimental record**, not that human
spermatids make an ACRBP-V5 equivalent". The `## Question` section is scoped up front to the
sequence-level half, and the Part 2b read-out stops at the structural conclusion instead of
generalising to the annotation. Edited through the generator, not the output — a fresh
`uv run python analyze_isoform_transfer.py` reproduces the committed `RESULTS.md` and `results.json`
byte for byte, from both the cache and `--refetch`.

**4. Self-correction you did not ask for.** I had marked `PMID:23426433` `DISPUTED` on the strength of
the GENCODE annotation. That was wrong for the same reason my conclusion was wrong, and it was the
more damaging of the two errors — it impugned a sound paper. It is now `VERIFIED`, with
`review_notes` recording what the earlier draft got wrong and why, and the conflict recorded as a
knowledge gap instead. `PMID:30606959` gains the human-negative statement as a `finding` with
`reference_section_type: INTRODUCTION`, and its `review_notes` note that the sentence was overlooked
while the paper was being cited for other claims.

The PR title and body carried the same overstatement and have been rewritten; the title is now
"GENCODE annotates the splice variant the literature says humans lack", which is what the evidence
actually supports.

No curation action changed: 7 ACCEPT / 4 KEEP_AS_NON_CORE / 1 MODIFY / 1 MARK_AS_OVER_ANNOTATED /
4 NEW, as before.

One incidental fix worth flagging for others: the four new `supporting_text` quotes had to drop their
trailing period. The source sentence ends "...of the Acrbp gene in mouse **[3]**.", so the period is
not contiguous with the quoted span. `just validate` passed with the period included; only the
stricter verbatim check caught it.

Verification: `checkquotes.py` 63 quotes / 0 problems · `just validate human ACRBP` ✓ Valid, no
warnings · `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` empty · all 14 GOA rows
still accounted for in GOA order.

cmungall:
Thanks for the approval. Taking one of the two carry-over suggestions and declining the other.

**Taken — the `GO:0061135` target.** You are right that the row leaned on the reader to infer it. The
term name "endopeptidase regulator activity" naturally reads as "acts on a working protease", which is
the opposite of the mechanism: ACRBP binds the 55-, 53- and 49-kDa (pro)acrosin species and does *not*
bind the 43-kDa intermediate or the 35-kDa mature enzyme, so it never contacts active acrosin. Both
`summary` and `reason` now state that outright, and add the point that the endopeptidase activity being
modulated is the activity of the *product* rather than of the molecule ACRBP holds — which is the
reason `GO:0035375` is paired with this row instead of being subsumed into it. Same commit tightens
the link between the two MF rows so they read as one argument rather than two overlapping claims.

**Declined — `GO:0009566` in `core_functions[0].directly_involved_in`.** I agree it is the broadest of
the three and that `GO:0060046` and `GO:0097341` carry the mechanism. But its GOA row is `ACCEPT`,
which under the schema's definition means "retain as representing the core function of the gene".
Dropping it from `core_functions` while keeping that action would leave the two halves of the document
disagreeing about whether the term is core, and a reader reconciling the row actions against
`core_functions` would find an unexplained gap. The consistent alternatives would be to demote the row
to `KEEP_AS_NON_CORE` — which I do not think is right, since ACRBP's entire reason for existing is
exercised at fertilization and the mouse knockout is a fertility phenotype — or to leave both as they
are. I have left both as they are, with the redundancy accepted as the cost of keeping the row actions
and the synthesis in step.

If you would rather see the row demoted than the parent term retained in `core_functions`, say so and
I will make that change instead; it is a coherent position, just not the one I would pick.

Verification after this commit: `checkquotes.py` 63 quotes / 0 problems · `just validate human ACRBP`
✓ Valid, no warnings · `cache/go/terms.csv` has no deletions against `origin/main` · action counts
unchanged at 7 ACCEPT / 4 KEEP_AS_NON_CORE / 1 MODIFY / 1 MARK_AS_OVER_ANNOTATED / 4 NEW.

PR #2269, opened 2026-07-26
Review human ACRV1: an acrosomal LU-domain protein whose only annotated process is the one it isn't in

cmungall:
Full review of the 10 GOA rows for human **ACRV1** (SP-10, P26436), one review entry per GOA row in GOA order, plus three additive proposals. Final: 2 ACCEPT, 2 MODIFY, 6 MARK_AS_OVER_ANNOTATED, 3 NEW, 0 REMOVE.

## The two IBA rows are less precise than the evidence they were built from

`GO:0005737 cytoplasm` and `GO:0031982 vesicle` (IBA, GO_REF:0000033) share the WITH/FROM triple `MGI:MGI:104590 | PANTHER:PTN008565525 | UniProtKB:P26436`. Resolving it *and* asking what the donor itself carries:

- `MGI:MGI:104590` → **P50289 / ASPX_MOUSE**, mouse Acrv1, the genuine one-to-one orthologue (resolved via UniProt `xref:mgi-104590`; the doubled `MGI:MGI:` form returns HTTP 400).
- QuickGO (`geneProductId=UniProtKB:P50289`, `goId=GO:0005737`/`GO:0031982`, `goUsage=descendants`) returns 5 hits each, of which **two are the donor's own IDA annotations to `GO:0001669` acrosomal vesicle** (PMID:1591350, PMID:16093322), plus an ISO and an IEA to the same term.
- `GO:0001669` is a confirmed descendant of both terms, and human ACRV1 already carries its own IDA to it from HPA.

So the propagation machinery worked and the donor is right, but the term is three levels vaguer than its source and strictly redundant with an existing IDA on the same gene. **MODIFY → `GO:0001669`** on both, with `propagation_review` root_cause `TERM_SCOPING_PROBLEM` / `GRANULARITY_MISMATCH`. `PANTHER:PTN008565525` is reported as an internal tree node (not a protein) and `UniProtKB:P26436` as a valid self-reference, not a circular one.

## `NbExp=3` on the five protein-binding rows is one screen in three flavours

All five `GO:0005515` IPI rows come from PMID:32814053. Querying IntAct directly (`/intact/ws/interaction/findInteractions/P26436`) shows each pair recorded **three times from that single publication** — as `two hybrid array`, `two hybrid pooling` and `validated two hybrid` — both partners `over-expressed`, MI-score 0.56. That is what UniProt's `NbExp=3` is counting; there is no orthogonal assay for any of the five. (IntAct also holds two interactions GOA did not import: TRIM68 anti-tag co-IP and CFTR ubiquitin reconstruction.)

Partners resolved: PPT1 (lysosome/secreted), SPRED1 (plasma membrane/nucleus), "TSC1", CISD2 (ER + mitochondrial outer membrane), PANK2 (cytoplasmic isoform). **Every one is cytosol-facing**, while ACRV1 has a signal peptide and sits in the lumen of a secretory organelle — the Y2H put a signal-peptide-bearing ORF in a yeast nucleus where that peptide is inert. Add the 166-residue disordered Ser/Glu-rich spacer (a classic sticky prey) and the set reads as screen background. They also cohere only as a **neurodegeneration bait panel** (PKAN, CLN1, Wolfram 2, TSC, Legius), not as acrosome biology.

One extra defect worth flagging: the "TSC1" accession **Q86WV8 is an unreviewed 366-aa TrEMBL clone** (submission "Tuberous sclerosis 1", AAH47772, IntAct clone CCSB_14153). Canonical TSC1 is Q92574, 1164 aa. So even the partner identity in that row is a partial ORFeome construct.

`MARK_AS_OVER_ANNOTATED` rather than REMOVE on all five, decided per partner: a Y2H hit is a real observation, `GO:0005515` just conveys nothing and none of these licenses a functional claim.

## `spermatogenesis` has the direction of the relationship backwards

`GO:0007283` (NAS, PMID:21252238). That paper is a study of **TDP-43 repressing the mouse *acrv1* promoter**; `acrv1` is the transcriptional *target* and no property of the ACRV1 protein is assayed. Its defensible content for this gene is expression timing, not process participation. Independently, **CRISPR knockout of mouse *Acrv1* leaves males fertile** with litter sizes indistinguishable from wild type, and the same paper records that anti-SP-10 immunisation of male rodents does not arrest spermiogenesis (PMID:38697008, full text read). `MARK_AS_OVER_ANNOTATED`, with the real role proposed additively rather than substituted, since fertilisation is a different process and not a different level of the same one.

## Three additive proposals, and a resolved apparent contradiction

- **`GO:0007342` fusion of sperm to egg plasma membrane involved in single fertilization (IMP, PMID:10775167).** A mAb against SP-10 residues 135-149 blocks human sperm binding to zona-free hamster oolemma but *not* sperm-zona binding in the hemizona assay, with beta-1 integrins positively excluded as the ligand. IMP because a function-blocking antibody is a specific inhibitor and the inference runs from a perturbation phenotype, not from direct observation. GOA currently has no fertilisation term for ACRV1 at all.
- **`GO:0002079` inner acrosomal membrane** and **`GO:0043159` acrosomal matrix** (IDA), from human immunogold EM and Triton X-114 / thiocyanate partitioning — the sub-acrosomal detail the SubCell keyword cannot express.

The affinage record presents the bovine result (anti-SP-10 reduces sperm-zona **secondary** binding, PMID:8882296) as flatly contradicting the human hemizona result. It does not: the hemizona assay scores *primary* binding by acrosome-intact sperm, secondary binding is post-acrosome-reaction attachment of the inner acrosomal membrane, and SP-10 is inaccessible before the acrosome reaction (PMID:2310816). Both papers agree SP-10 acts *after* the acrosome reaction. Recorded in `reference_review` so a future curator does not mistake it for a citation error.

The fertile mouse null is stated as a caveat travelling **with** the annotation, not suppressed: GO annotates participation rather than requirement, but a curator preferring maximum caution could take the parent `GO:0007338`. That choice is put in `suggested_questions`.

## No molecular function — tested, not assumed

`ACRV1-bioinformatics/` (snapshot-driven: `fetch_data.py` caches UniProt/InterPro/QuickGO into `data/`, `analyze.py` touches no network, so a fresh run reproduces `RESULTS.md` byte-for-byte — verified with `diff`; missing input is a hard error naming the fix command):

1. **The 1990 "unique, no homology" conclusion is superseded.** Residues 188-264 are a canonical Ly-6/uPAR three-finger domain (IPR016054 / PF00021 / CDD `cd23628`) with **10 cysteines — the modal count across the 38 LU domains of the 30 reviewed human LU proteins** — and every cysteine in the protein lies inside it. This matters because that stale claim is why nobody looked for a family-based function, and the affinage record repeats it.
2. **The isoform complexity is not functional complexity.** Mapping all nine `VAR_SEQ` deletions onto the domain: every isoform with a measured abundance keeps the LU domain complete; only isoforms 8 and 11 (in the <1% pool) truncate it by 37 aa. Splicing varies the disordered spacer, not the fold — which is why no `isoform:` field is set on any row.
3. **ACRV1 is one of 11/30 human LU proteins with no GPI anchor**, matching the experimental finding that it is a chaotrope-releasable *peripheral* protein (PMID:1591355) — but the report is explicit that anchoring is *not* the discriminator against transferring a family function, since PATE1/PATE4/SLURP1/SLURP2 are unanchored too. **Reachability is:** 29/30 members with a location annotation are extracellular-accessible (secreted or cell membrane), including all 11 with an acetylcholine-receptor function, and **ACRV1 is the sole exception**. PATE4 is the sharpest test — it shares ACRV1's acrosomal annotation — and it survives, because PATE4 is also annotated as secreted.
4. **ACRV1 is a subfamily singleton in human** (neither PTHR17571 nor cd23628 contains another reviewed human LU protein), so only orthologue transfer is available — no paralogue.
5. The report deliberately records the **complication that undercuts an easy dismissal**: the family's most widely shared experimental MF is nicotinic acetylcholine receptor modulation (7/30), and PATE1/PATE4 — the clade the ACRV1 CDD model is named for — carry `GO:0030548`. Transfer is still wrong, but for stated reasons (surface-exposed members only, not family-wide, ACRV1 unanchored and lumenal), not by assertion.

Also raised but deliberately **not** annotated: PMID:41338461 shows ACRV1 knockdown attenuating ZNF280A/CUX2-driven AKT phosphorylation and glycolysis in ovarian cancer lines. It is the only loss/gain-of-function work on human ACRV1 that exists, but it is a testis-restricted protein in a tissue that never normally expresses it, so it describes ectopic expression, not function. Left as a question for the Consortium.

## Verification

*(Updated after review round 1: `GO:0007342` code changed IDA → IMP, and the family-transfer argument moved from GPI anchoring to a computed topology census. See the reply comment for the full list of seven items addressed.)*

- `just validate human ACRV1` → **✓ Valid**, no warnings.
- All **80** `supporting_text` entries (including the `file:` quotes the repo validator skips) verified as verbatim substrings; all **21** inline quotes in `ACRV1-notes.md` likewise.
- `RESULTS.md` and `results.json` reproduce byte-for-byte on a fresh `analyze.py` run (`diff` clean).
- `git diff origin/main HEAD -- cache/go/terms.csv` → 3 insertions, **0 deletions**.
- `origin/main` merged in before final validation.


cmungall:
Thanks — both blocking items were right, and item 1 turned out to make the analysis materially better rather than just better-worded. All seven points addressed in d7493ce (pushed as 2c48044).

## 1. `RESULTS.md` §5 — you were right, and the fix is a new computation, not a reword

You're correct that §5 leaned on §3 for a discrimination §3 doesn't supply. Rather than just swap the word "anchoring" for "topology" in prose, I made the topology claim computed: `analyze.py` now has a `topology_census()` that classifies every member's UniProt subcellular locations as extracellular-accessible (`Secreted`, `Cell membrane`, `Cell surface`, `Membrane, caveola` — GPI proteins sit in the outer leaflet, so `Cell membrane` counts), rendered as new **§3b**.

Result: **29 of the 30 members with a location annotation are extracellular-accessible, and ACRV1 is the sole exception.** All 11 members carrying any acetylcholine-receptor MF qualify. So the discriminator is reachability, and §5 now says so, opening with "the first is **topology, not GPI anchoring** — PATE1, PATE4, SLURP1 and SLURP2 have no anchor either (section 3), so anchoring cannot be the discriminator."

§3's own "bearing on the review" now carries the concession too, so a reader checking §3 against §5 finds them agreeing rather than in tension.

**The check also surfaced something sharper than the point you raised**, which I'd have missed: **PATE4 is annotated to `Cytoplasmic vesicle, secretory vesicle, acrosome` — ACRV1's own compartment** — as well as carrying `GO:0030548`. That is the strongest available case for transfer, closer than the cd23628 naming argument. It survives, but only because PATE4 is *also* annotated as `Secreted` while ACRV1 has no secreted or cell-membrane assignment at all. §3b and §5 both state this explicitly, and §3b also records that SPACA4 and TEX101 are acrosomal LU proteins. Better argument, and it now rests on a table rather than on prose.

## 2. `IDA` → `IMP` — conceded, the rationale was wrong on the criterion

You're right: native-versus-overexpressed is not the IDA/IMP boundary. GO's IMP scope explicitly covers specific inhibitors, and a function-blocking antibody is one; the inference runs from a perturbation phenotype (block the protein, watch binding fall), not from direct observation. `evidence_type` is now `IMP` and the rationale replaced with the direct-observation criterion.

The two localisation proposals stay `IDA` deliberately — immunogold EM and Triton X-114/thiocyanate phase partitioning *are* direct observations of where the protein is. Thanks for the CRISP1 precedent; good to know the term choice matches repo practice on comparable evidence.

## 3–7. Suggestions, all taken

**3. PAN-GO policy vs curator error.** The `propagation_review` comment on the PANTHER node now records the alternative explanation directly: PTHR17571 has no other reviewed human member, so a heterogeneous clade can't be forcing the broad term, and the remaining explanation may be a restricted PAN-GO term set rather than curator judgement — in which case `TERM_SCOPING_PROBLEM` describes where the annotation sits relative to its evidence, not a mistake.

**4. `GO:0007340`.** Now discussed and explicitly declined, with both relevant sentences quoted. Three reasons: anti-human reagents on bovine sperm; the same abstract reports the antibodies "affected the motility of capacitated spermatozoa, while not affecting the motility of noncapacitated spermatozoa", so a general effect on capacitated cells isn't excluded as the cause; and a protein that only becomes accessible *after* the acrosome reaction is awkward to place as a participant in triggering it. Noted that the sperm-oolemma result avoids the second objection (human sperm, defined-epitope monoclonal, assay scores binding not progression). Both quotes added as `findings` on PMID:8882296 too.

**5. `spermatogenesis` — kept, but the summary now argues for the action it takes.** Fair hit: the text was arguing past `MARK_AS_OVER_ANNOTATED`. The summary now says why I stop short: both objections attack the *evidence*, not the proposition. SP-10 is a spermatid-specific constituent of an organelle assembled during spermiogenesis, so an assembly assay could plausibly earn this term; and a fertile knockout shows dispensability, which isn't non-participation. What's established is that the term rests on nothing at present — which is what `MARK_AS_OVER_ANNOTATED` says — whereas `REMOVE` would assert the stronger, untested claim that ACRV1 takes no part in building the acrosome.

**6. `is_active_in`.** Promoted from aside to recommendation. Both MODIFY `reason` fields now state that the replacement is `located_in GO:0001669`, matching the two existing rows for that term, rather than `is_active_in`, which presupposes a molecular activity ACRV1 hasn't been shown to have.

**7. Fusion label vs binding evidence.** The caveat now travels into `core_functions[1].description`, so a reader arriving at the core function sees it: "The GO term used is labelled for fusion while the measured event is binding: its definition covers both steps as one process, and the evidence reaches only the binding step."

## One thing I found while auditing, not raised in the review

My `description` claimed the Acrv1 null has "normal testis histology and sperm". The detailed figures in PMID:38697008 are shown only for four of the twelve genes, so I went back to check what is claimed for Acrv1 specifically. The abstract does cover it — "we generated knockout mice for 12 genes (Acrv1, ...) ... and examined the appearance and histological morphology of testes, sperm motility, and male fertility" — so the claim holds, but *sperm morphology* was not among the assays. Narrowed to testis appearance, testis histology and sperm motility, added that sentence as a `supporting_text` and a `finding`, and noted that only Gm17266, not Acrv1, showed an abnormality.

## Verification after the changes

- `just validate human ACRV1` → **✓ Valid**, no warnings.
- **80** `supporting_text` entries verified verbatim (was 74; the `file:` quotes into `RESULTS.md` still match after regeneration), plus 21 inline quotes in `ACRV1-notes.md`.
- `RESULTS.md` and `results.json` reproduce byte-for-byte on a fresh `analyze.py` run (`diff` clean) — checked after the §3b addition.
- `git diff origin/main -- cache/go/terms.csv` → **0 deletions**; still the same 3 insertions, since no term ids changed.
- `origin/main` merged again before pushing.
- Action counts unchanged (2 ACCEPT / 2 MODIFY / 6 MARK_AS_OVER_ANNOTATED / 3 NEW); no term id changed. The only annotation-level change is the one evidence code.

`ACRV1-notes.md` records the reasoning for all of the above, including the evidence-code mistake, so the next reviewer of this gene doesn't have to re-derive it.

cmungall:
One more of the same class as item 1, found while auditing my own citations rather than in response to a comment (07b60a0):

The PANK2 two-hybrid row argues that PANK2 is cytosolic while ACRV1 is lumenal, but its `supporting_text` from `RESULTS.md` was the line about ACRV1 lacking a GPI anchor — true, but not support for *that* argument. Swapped for the sentence that is: "its compartment is the lumen of a secretory organelle, topologically separate from the cytosol — which is why its cytosolic yeast-two-hybrid partners cannot meet it in vivo". The GPI line is retained where it belongs, as a `finding` on the `RESULTS.md` reference.

Still ✓ Valid, 80 quotes verified, 0 deletions in `cache/go/terms.csv`, `origin/main` merged. No term id, action, or evidence code changed.

PR #2271, opened 2026-07-26
Review human ACTL7A: an actin fold that kept the pocket but lost the hydrolysis trigger and the filament interface

cmungall:
## ACTL7A (Q9Y615) — an actin fold that kept the pocket but lost the hydrolysis trigger and the filament interface

ACTL7A ("actin-like protein 7A", Arp7A) is the textbook version of this campaign's most repeated
defect: a protein whose GO record carries actin-derived terms because its **name** and **fold** are
actin-like. But the mirror-image error is just as available here — Boëda et al. note that "the
sequence of Arp7A has no major deletions or insertions within its predicted actin fold and has 43%
sequence identity to β-actin, which is higher than that of Arp3". So the fold is *not* obviously
emptied out, and "fold without function" would have been an unsupported assertion.

I resolved this by measuring, not asserting.

### Structure-guided audit (`genes/human/ACTL7A/ACTL7A-bioinformatics/`)

`actin_fold_audit.py` derives its residue sets from experimental structures by neighbour search —
nothing is hard-coded — and maps them through a MAFFT L-INS-i alignment of 22 actins and ARPs, with
the deposited SEQRES sequences included as alignment entries so positions map without hard-coded
numbering. Every mapped position is printed in a gap-free alignment window so the mapping can be
*checked* rather than trusted. A fresh cold run (cache deleted) reproduces `RESULTS.md` and
`results.json` byte-for-byte, and the fail-loud paths (wrong gene symbol, bad accession, empty
contact set) were tested and all raise.

| | ACTL7A | conventional actins | Arp2/Arp3 | Arp1 (forms a filament) | yeast SWI/SNF Arps |
|---|---|---|---|---|---|
| whole chain vs actin | 43.7% | 88–100% | 38–48% | 52.8% | 19–23% |
| nucleotide cleft (19 pos, PDB 2BTF) | **63.2%** | 96.5% | 79.0% | 73.7% | 36.9% |
| protomer interface (79 pos, PDB 8A2S) | **42.3%** | 93.2% | 33.1% | 55.7% | 18.5% |
| ATP-hydrolysis catalytic set (D11/Q137/D154/V159/H161) | **3/5** — `DQEVY` | 5/5 | 5/5 | 5/5 | 1/5 |

Three conclusions, each with a different curation consequence:

1. **The nucleotide cleft is retained.** ACTL7A sits far above the divergent Arps. So no
   "vestigial fold" claim, and no `REMOVE` on that basis.
2. **The ATP-hydrolysis trigger is not.** Actin **His161** — the residue whose side-chain flip on
   filament incorporation triggers hydrolysis (PMID:37009486) — is **tyrosine** in ACTL7A and in
   ACTL9, while **ACTL7B keeps it**. The alignment window is anchored on an identical `PIYEGY`
   motif, so this is not an alignment artefact.
3. **The filament interface is not.** 42.3% puts ACTL7A *below* Arp1, which does form a filament
   in dynactin — quantitatively consistent with the published cell-biological negative: "when
   expressed in HeLa cells, GFP-Arp7A is cytoplasmic and does not assemble into filaments".

And the guard against over-correcting: **the cleft is under disease-relevant constraint.** It covers
only 4.6% of the protein, yet 2 of 5 variants reported in SPGF86 patients land in it (`G246A` ↔
actin Gly182; `G362R` ↔ actin Gly301, a phosphate contact; exact binomial p = 0.019) against 0 of 4
population polymorphisms. A third, `D75A`, maps to actin **Asp11** in the divalent-cation site.

Net: ACTL7A is a **structural/scaffolding** actin-fold protein whose pocket is not decorative.
Nothing licenses a polymerisation term; nothing licenses declaring the pocket dead either. Where
the fold-derived terms survive, they survive as *untested*, not as *demonstrated* — and that is
what the actions say.

### Actions: 30 rows reviewed, 1 new

| action | n |
|---|---|
| ACCEPT | 10 |
| KEEP_AS_NON_CORE | 13 |
| MODIFY | 6 |
| REMOVE | 1 |
| NEW | 1 |
| MARK_AS_OVER_ANNOTATED | 0 |

*(Tally as settled after review. The molecular-function rows moved twice: `GO:0005200` ended at
`REMOVE` and `GO:0005198` at `ACCEPT` as the core MF, once GO's own PAINT record was found to
**negate** `GO:0005200` for the PANTHER node carrying ACTL7A — see `ACTL7A-notes.md` §9. A sixth
`MODIFY` refines the ISS nucleus row to `GO:0001673`, which its own donor already holds by IDA.)*

**`MARK_AS_OVER_ANNOTATED` — `GO:0005200 structural constituent of cytoskeleton`, TAS,
PMID:10373328.** The cited paper is a positional-cloning study: cDNA selection, genomic sequencing,
linkage mapping, Northern blots. No biochemistry, no localisation, no phenotype. The only structural
statement in it is the homology-based label "actin-like" — the activity was read off the name. It
does not stop there: `GO_REF:0000108` uses this MF term as the literal WITH/FROM for the
`GO:0007010` row, so **one 1999 name-based inference generated both an MF and a BP annotation**.
Flagged rather than removed because the *claim* is defensible on post-2011 evidence (the acroplaxome
and perinuclear theca are cytoskeletal structures; GO:0033011 is a GO:0005856 descendant) — what it
needs is re-sourcing, not deletion.

**`MODIFY` `GO:0007010 cytoskeleton organization` → `GO:0030036 actin cytoskeleton organization`.**
Neither human ACTL7A nor mouse Actl7a carried *any* annotation in the actin-process branch (checked
by QuickGO `goUsage=descendants` on both), despite the central result of PMID:36734600 being total
loss of subacrosomal F-actin in the KO — and specifically *local* loss: "The KO spermatids did not
lose F-actin in other cellular regions". `GO:0030036` rather than a polymerisation or nucleation
term because the authors are explicit that the mechanism is unresolved.

**`MODIFY` all three `GO:0001669 acrosomal vesicle` rows → `GO:0033011 perinuclear theca`.** This is
the localisation finding, and it is definitional rather than a challenge to anyone's data.
`GO:0001669` denotes "A structure in the head of a spermatozoon that contains acid hydrolases ...
derived from the lysosome" (exact synonyms: *acrosome*, *acrosomal granule*) — a membrane-bounded
vesicle. ACTL7A is a cytosolic cytoskeletal protein on the **nuclear** side of the inner acrosomal
membrane, so `located_in` that vesicle is a category error however good the immunofluorescence is;
what IF shows is an acrosomal-*region* signal, which at light-microscope resolution necessarily
includes the acroplaxome underneath. The 2011 paper corrected this explicitly: "We have now extended
these earlier studies and shown that Arp7A is actually localized in the subacrosomal layer, which is
also known as the acroplaxome". `MODIFY`, not `REMOVE` — the observation is real, the compartment is
one step away, and both reasons say a curator with figure access should confirm (the two
experimental rows are cached abstract-only).

**`MODIFY` `GO:0005515 protein binding` (TES) → `GO:0030274 LIM domain binding`.** PDB **2XQN**;
residues 28–49 of ACTL7A occupy the apolar pockets of TES LIM2 and LIM3; `A31Y` and `A41Y` each
abolish binding. Note the binding surface is the **disordered N-terminal extension unique to
ACTL7A, not the actin fold** — the one well-characterised biochemical activity of this protein lives
outside the part that gave it its name. The CYLC1 and ACTL9 `protein binding` rows are kept and
marked non-core instead: both partners are real and corroborated by convergent phenotypes, but no GO
MF descendant exists for binding a cylicin or an actin-like protein, so `MODIFY` has no target.

**`NEW` `GO:0007343 egg activation`, IMP.** GOA's most specific process term for ACTL7A was
`fertilization`. But ACTL7A is an **oocyte-activation-deficiency gene**: patient variants deplete and
mislocalise PLCζ, calcium oscillations fail, and the block is bypassed by calcium ionophore, "achieved
a live birth for the patient with ACTL7A variants". `GO:0007343` is a `GO:0009566` descendant, and
there is direct precedent — human **PLCZ1** carries it with an IMP. Additive; the fertilization rows
stay.

### WITH/FROM: all 15 tokens resolved *and* queried for their own evidence

Two results changed my conclusions:

- **The `nucleus` IBA is not the family artefact it looks like.** Two of five sources are yeast
  ARP9 and *Candida* Arp9 — nuclear SWI/SNF-complex subunits, exactly the "PANTHER lumps nuclear
  ARPs with cytoplasmic actins" story I expected to write. But the mouse ortholog carries its **own**
  nucleus EXP/IDA *and* `GO:0001673 male germ cell nucleus` by IDA. The term is right; only
  `is_active_in` overreaches, since no nuclear activity is demonstrated. Kept non-core.
  (`GO:0001673` refinement raised in `suggested_questions` rather than applied, because
  PMID:10373328 reports ACTL7A message "in a wide variety of adult tissues", so a somatic nuclear
  pool is not formally excluded.)
- **The `GO:0005198` IBA has no protein source at all** — both WITH/FROM tokens are internal
  PANTHER tree nodes, and across the entire ACTL7 orthologue set the only non-IBA support for that
  term or any descendant is the gene's own 1999 TAS row. Accepted anyway, because the KO phenotype
  independently supports it; but the row's *route* is recorded as family-level with no protein-level
  evidence.

Also worth a curator's eye: `GO:0001669` / **IMP** / PMID:34727571 uses an IMP code for a
*cellular-component* term; the underlying data are immunofluorescence on patient sperm, which is
IDA-shaped. Recorded in `reference_review.review_notes`.

### Ontology gaps proposed

- **`acroplaxome`** (CC, child of `GO:0033011`). GO's perinuclear-theca definition names both parts —
  "a subacrosomal layer and, continuing caudally beyond the acrosomic system, the postacrosomal
  sheath" — and GO has a term for the caudal part (`GO:0033150 cytoskeletal calyx`) but **none** for
  the subacrosomal layer. That asymmetry is what forces the acrosome/acroplaxome conflation above,
  and the same term is needed for ACTL9, CYLC1, ACTRT2, SPACA1, MYO6 and IZUMO3.
- **`acrosome attachment to nuclear envelope`** (BP, child of `GO:0001675`). `GO:0001675` is defined
  as "The formation of the acrosome from the spermatid Golgi" and cannot express the separately
  mutable **anchoring** step that fails in ACTL7A, CYLC1 and ACTL9 patients alike.

### Validation

`✓ Valid`, quote checker `81 quotes, 0 problems`, `git diff origin/main HEAD -- cache/go/terms.csv |
grep '^-GO:'` empty, branch fast-forwarded onto current `main` before final validation.

One warning is left standing deliberately: *"No annotations reference available deep research
files"*. The affinage report (`gates_passed: True`) was used as a lead-generator only — per campaign
rule, provider sentences are not quoted as `supporting_text` for mechanistic claims. Two of its
citations are excluded outright and the reason is recorded in `ACTL7A-notes.md` §10:
`PMID:bio_10.1101_2025.03.27.645694` is a bioRxiv DOI in a PMID-shaped field, and `PMID:38464253`
is a preprint whose HDAC and INO80/SWI-SNF docking claims are in-silico (they are the reason the
nuclear pool is filed as a knowledge gap rather than dismissed).


cmungall:
Both 🟡 items were right, and the first was a real error rather than a presentation problem. Fixed in 23ec924ea; durable detail is in `ACTL7A-notes.md` §6b and §11 rather than here.

**1. The cytoplasm quote argued the opposite.** Correct, and worse than a mismatch: PMID:12672658 names `t-actin 1/mACTl7b` and `t-actin 2/mACTl7a`, so "T-ACTIN-1 was found in the cytoplasm and T-ACTIN-2 was located in the nucleus" assigns the **nucleus** to ACTL7A and the **cytoplasm** to the paralogue. Attached to an ACTL7A cytoplasm row it was evidence against the term.

Three changes, because the error had a second edge I had missed. The row now cites PMID:21278383 ("GFP-Arp7A is cytoplasmic") and PMID:36734600 ("the nucleus and subacrosomal space and later ... postacrosomal regions"). The alias key is recorded as a `findings` entry on the reference so the trap is documented rather than merely avoided. And the mouse **Actl7b** donor `MGI:MGI:1343053` is downgraded to `SUPPORTS_SOURCE_BUT_NOT_TARGET` — its cytoplasm IDA *is* the paralogue-specific half of that same experiment, so the cytoplasm transfer survives only because mouse Actl7a carries independent cytoplasm IDAs. That is the pointed version of the finding: a paralogue donor whose cited evidence, read with the alias key, supports the *opposite* compartment for the target.

Generalisable lesson: when a paper studies a paralogue pair under aliases, the alias-to-gene key is part of the evidence and has to be quoted alongside the claim.

**2. Flagging the specific MF down while promoting its parent.** Agreed, and the incoherence was mine — the `reason` argued GO:0005200's term was defensible and needed only re-sourcing, while the action said it overshot, and `core_functions[0].description` began "Structural constituent of the acroplaxome". Term correctness and citation quality are separable judgements, so they are now separated:

- **`GO:0005200` → `ACCEPT`, and it is the core molecular function.** The acroplaxome is an F-actin/keratin plate, `GO:0033011` is a `GO:0005856` descendant, the KO loses subacrosomal F-actin, and the structure audit finds a structural rather than polymerising profile. The inadequate 1999 TAS citation and its `GO_REF:0000108` knock-on are argued inside `reason`, where provenance criticism belongs.
- **`GO:0005198` → `KEEP_AS_NON_CORE`**, a true but redundant parent — the same treatment already given the `cytoplasm` and `cytoskeleton` parents.

Nothing now carries `MARK_AS_OVER_ANNOTATED`. On inspection no term on this gene actually overshoots; the defects were always the *citation* and the *inference chain*. That reads better anyway: the 1999 name-derived guess landed on the right term, and the damage was the over-general BP that `GO_REF:0000108` then derived from it.

**Non-blocking items, all taken.** `core_functions[2]` now states why it asserts no molecular function (direct PLCζ binding vs architectural is undetermined — the knowledge gap says so, so a guessed activity should not stand in for it) rather than leaving the slot silently empty. The `GO:0001669` IMP row records that `UNDECIDED` was considered and why `MODIFY` won: the argument is definitional and does not depend on anything in the figures, so deferring would withhold a checkable proposal rather than avoid an unsupported one. And you were right that the `GO:0001673` deferral hedged a protein-localisation claim with an mRNA survey — the Northern data are now explicitly *not* the reason; the reasons are donor heterogeneity on the phylogenetic row and the absence of any somatic-nucleus experiment either way.

**Paralogue cross-check (notes §6b), added at the coordinator's prompt** since ACTL7B is under independent review. The pair is separable on four axes, two of which constrain curation:

| | ACTL7A | ACTL7B |
|---|---|---|
| catalytic set D11/Q137/D154/V159/H161 | `DQEVY` — His161 **lost** | `DQEVH` — His161 **retained** |
| nucleotide cleft / phosphate contacts | 63.2% / 66.7% | 68.4% / 75.0% |
| F-actin protomer interface | 42.3% | 36.4% |
| mouse germ-cell localisation | nucleus; in sperm heads **and** tails | cytoplasm; **not** in sperm |

- The "cannot run actin's hydrolysis cycle" argument **does not carry over to ACTL7B**, which keeps the trigger histidine.
- `GO:0030274 LIM domain binding` **must not be assumed to transfer**: it is carried by ACTL7A's N-terminal extension, and "Arp7B, whose N-terminal extension is distinct from that of Arp7A ... is also strongly enriched in the acroplaxome" — same destination, different adaptor sequence.

**On your "do the donors agree?" rule**, applied back to my own `propagation_review`s: I dropped `GRANULARITY_MISMATCH` from the nucleus IBA row. That donor set spans a yeast SWI/SNF subunit, a *Candida* homologue, a trypanosome protein and two mammalian ACTL7 proteins, so `GO:0005634` is the correct least common ancestor and `GO:0001673` would not fit the fungal and protozoan donors — the specificity question belongs on the ortholog-specific ISS row, where it now sits as a question. I kept `GRANULARITY_MISMATCH` on the rows with homogeneous single donors (cytoplasm IEA/ISS, cytoskeleton IEA, nucleus ISS, and the `GO:0007010` inter-ontology row), where a true-but-uninformative parent is exactly the situation.

**Note on your sandbox limitation:** `just validate-changed` (the exact CI command) passes here — `✓ Valid`, one warning, plus `84/84` supporting_text quotes verbatim via an independent checker, `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` empty, and the committed `RESULTS.md`/`results.json` reproduce byte-for-byte from a cold run with `data/` deleted. Branch is merged up to current `main`.

Thanks for re-deriving the His161 call from the printed alignment windows independently — that is exactly what they were put there for.

cmungall:
Both taken; pushed in 64b88c86e.

**🟡 `root_cause` on the `GO:0005198` row** → now `NO_FAILURE_NON_CORE` with `GRANULARITY_MISMATCH`. You were right that this is worse than cosmetic: `propagation_review` is the machine-readable half of the row, so a stale `NO_FAILURE_CORE` would have had anything aggregating the corpus count this row as core — the exact claim the previous round retracted. I also did what the absence of a validation rule implies and checked it mechanically rather than by eye: an assertion over all 31 rows that `action` and `root_cause` agree, which now reports no mismatches. That check is worth having as a repo-level best-practices rule, since `✓ Valid` demonstrably does not cover it.

**🔵 Two notes paragraphs arguing the superseded decisions** → both rewritten. The "Why no REMOVE anywhere" paragraph no longer calls `MARK_AS_OVER_ANNOTATED` the right instrument for `GO:0005200` (it now points at the resolution paragraph above it, so the section and the table agree), and the nucleus paragraph no longer offers the 1999 Northern survey as the reason for deferring `GO:0001673`. It now says plainly that citing an mRNA tissue distribution to hedge a protein-localisation question was a first-draft error, and gives the real reasons: donor heterogeneity on the phylogenetic row, and the fact that no study has looked for ACTL7A protein in a somatic nucleus. Fair catch — the notes are the durable audit trail, and leaving a retracted argument standing in them is worse than leaving it in a PR comment, which is why I moved that material there in the first place.

cmungall:
Merged current `main` (clean — no conflicts remained by the time I fetched) and took both remaining 🔵, plus a cross-gene check that only became possible once the paralogue merged. Pushed once, in c94cecf1c.

**🔵 `SOURCE_EVIDENCE_WEAK` on `GO:0005198`** — added. You were right that this is the same argument I accepted for `root_cause` a round earlier: the PANTHER-nodes-only provenance was argued in `reason` but invisible to anything reading the machine-readable field, and the file already pairs the two modes on the `GO:0007010` row.

**🔵 §11 round log** — now has round 3 and round 4 entries.

**The ACTL7B cross-check found a real divergence, and it is the most interesting thing in this round.** Your sibling review **REMOVEs** `GO:0005200`; I **ACCEPT** it and make it the core MF. Same term, same 1999 citation, same family. Rather than quietly diverge, both rows now argue it, because the asymmetry is checkable rather than a matter of taste:

- ACTL7A's specific compartment `GO:0033011` **is** a `GO:0005856` descendant, and its integrity demonstrably fails in the knock-out → ACTL7A is a structural constituent of a cytoskeletal structure.
- ACTL7B's specific compartment `GO:0001669` **is not** a `GO:0005856` descendant — a point your own review makes — so there is no annotated cytoskeletal structure for it to be a constituent of, and removal is right there.

I also owe your review an answer on its strongest point, that the clade's `GO:0005198` IBA may encode PAINT having *declined* to specialise to `GO:0005200`. I think that reading is correct as a description of what PAINT did, and it is now recorded on the `GO:0005198` row — but a family-level generalisation cannot see a gene-specific knock-out published years afterwards, which is the whole content of the divergence. That is the general lesson worth carrying: a name-derived annotation is applied per *family* while its correctness is per *gene*, so the pair can legitimately split on it.

Two overstatements of my own that the cross-check corrected, both from making the paralogues look more different than they are: ACTL7B is **not** purely cytoplasmic (your review gives it a low-abundance nuclear pool in early spermatids, so PMID:12672658's split is about the *dominant* pool, not presence vs absence), and "not in sperm" is about **mature** sperm only. §6b is rewritten against your review, and it notes that our two independent analyses reach the same filament conclusion by different routes — your DNase-I-binding-loop replacement, my contact-based interface number of 36.4% for ACTL7B, the lowest of the pair.

One constructive collision: your review *adds* `GO:0001669` to ACTL7B on the strength of it being on the developing acrosome, while mine argues the same term *away* from ACTL7A. Not a contradiction — different gene, different evidence — but "on the developing acrosome" is precisely the description the missing **acroplaxome** term would disambiguate, so the proposed term now has two candidate users in this pair alone.

**One further MODIFY, on the ACRV1 pattern.** `GO:0005634` ISS → `GO:0001673 male germ cell nucleus`: its *sole* donor, mouse Actl7a, already carries the specific term by IDA, so the transfer landed less precise than its own donor. The other three nucleus rows stay `KEEP_AS_NON_CORE` for route-specific reasons — heterogeneous fungal and protozoan donors make `GO:0005634` the correct LCA on the phylogenetic row, the `GO_REF:0000120` WITH/FROM mixes in a generic keyword, and the HDA row is a human experimental call. That produces a second warning ("inconsistent review actions for GO:0005634"), which I am leaving standing with the reasons on the rows, since the routes genuinely differ and collapsing them would be the smoothing you have twice caught me doing in the other direction.

**And the check you suggested earned its keep immediately.** Running `action` vs `root_cause` mechanically over all 31 rows before pushing caught the newly-`MODIFY`ed nucleus row still carrying `NO_FAILURE_NON_CORE` — the same defect class as last round, on a row I had just edited. Fixed to `TERM_SCOPING_PROBLEM` before the push rather than after your review. That is a strong argument for the repo-level rule.

Gates in the order that matters: checkquotes `84/84` → `just validate human ACTL7A` `✓ Valid` (2 warnings, both deliberate) → `terms.csv` diff last, no deletions.

cmungall:
You found the thing that made two of my rounds wrong, and it was in this repository the whole time. Pushed in 0b2f98b88 — and I have gone further than you asked, because the evidence you surfaced warrants it.

**I verified the record independently, and it is stronger than described.** It is not one row but three, and they are one curatorial act:

```
PTN000940351  GO:0005200  F  IBD  false  …|UniProtKB:P61160|…|UniProtKB:P60709|…  20250805
PTN008986528  GO:0005198  F  IBA  false  PANTHER:PTN000940351                     20250805
PTN008986528  GO:0005200  F  IRD  true   PANTHER:PTN000940351                     20250805
```

`GO:0005200` is IBD-supported at the ancestral actin node from conventional actins and Arp2/Arp3, then rejected on descent into the divergent-ARP branches — 8 IRD rows, `PTN008986528` among them. And `GO:0005198` occurs **nowhere else in the family table**: at that node it exists *only* as the substitute for the rejected child. So the IBA I had been treating as a parent left standing for want of anything better is the positive half of a deliberate generalisation. `Q9Y615` is in `PTHR11937-entries.csv`, and you are right about the dates being per-row.

**So I have reversed the ACCEPT, which you explicitly did not ask for.** Once the IRD is a fact rather than a possible reading, my remaining argument does not survive its own scrutiny: *residing* in a `GO:0005856` descendant is not *being a structural constituent of* one, and I was leaning on the ontology asymmetry to do work it cannot do. Worse, my own bioinformatics is the best available argument for GO's position — the route by which real actins earn `GO:0005200` is filament formation, and I measured ACTL7A's inter-protomer interface at 42.3% against 93–100% for conventional actins. I was arguing against my own evidence.

- **`GO:0005200` TAS → `REMOVE`**, on three independent grounds: the cited paper contains no functional data at all (TAS from a gene name, so no curator who read data is overruled — your sibling review's point); GO's pipeline has explicitly negated it for this clade, after the relevant literature; and the structural route is closed. This also brings the paralogue pair into **agreement** rather than divergence, which is the better outcome — a name-derived annotation is applied per family, so when it is wrong it is wrong family-wide, and GO's pipeline had already worked that out.
- **`GO:0005198` IBA → `ACCEPT` and core MF.** Now framed as the level GO chose on purpose, not a fallback.
- Both PANTHER nodes off `NOT_RELEVANT` → `SUPPORTS_TRANSFER`, with the IBA/IRD pair and the IBD seeding recorded on each. The PAINT table is cited with the rows as verbatim `findings`.
- **The round-4 `SOURCE_EVIDENCE_WEAK` you asked for is dropped.** It was right about the file as I then understood it and wrong about the biology: a deliberate generalisation seeded from an IBD over real actins is not weak evidence. Flagging that explicitly since it reverses something you requested.
- `core_functions[0]` reworded from "Structural constituent of the acroplaxome" to "Architectural component", and it now states why the weaker MF is correct.
- Notes §4's table corrected — it recorded both tokens' evidence as `n/a`, exactly as you said; §9 documents all three drafts including the two withdrawn arguments; §6b's divergence paragraph rewritten now that the pair agrees.

**`GO:0007010` → `GO:0030036` stands, and the divergence from your sibling review there is deliberate.** ACTL7B's review removes its `GO:0007010` as purely derivative of the removed MF. Mine keeps a row in that branch because ACTL7A has independent phenotype evidence ACTL7B lacks: the knock-out loses subacrosomal F-actin completely while retaining it everywhere else. Removing the derivative parent and substituting the evidenced child is not the same as keeping the derivative parent.

**The methodological lesson is the one I want on the record.** "PANTHER:PTN… is a tree node, not a protein" is a true statement that invites you to stop looking, and I stopped. The campaign rule is that an unresolved WITH/FROM cannot be dismissed, only deferred — the same must apply to an *unresolvable* one. A PANTHER node cannot be resolved to a protein, but it can be looked up in the family's PAINT table, where it may carry an explicit negation of the very term under discussion. That is now in notes §11 and §4.

Two smaller things: I fixed the campaign quote-checker to mirror the repo validator's `file:` resolution order (`genes/` then project root), because it was silently reporting repo-root references like the PAINT table as MISSING SOURCE rather than checking them — all three PAINT quotes verify verbatim, 89 quotes / 0 problems. And the stale PR-body action table is updated to the current tally: ACCEPT 10, KEEP_AS_NON_CORE 13, MODIFY 6, REMOVE 1, NEW 1.

Gates in order after merging current `main`: checkquotes 89/89 → `just validate human ACTL7A` `✓ Valid` (2 deliberate warnings) → `terms.csv` last, additive only, 4 insertions and no deletions.

PR #2274, opened 2026-07-26
Review human ACTR10 (ARP11), the dynactin pointed-end capping subunit

cmungall:
# ACTR10 (Q9NZ32) — ARP11, the dynactin pointed-end capping subunit

Reviews all **19 GOA rows** one-for-one plus **1 proposed new annotation** (20 entries).
`just validate human ACTR10` → `✓ Valid`, no warnings; 68 `supporting_text` quotes
verified verbatim, including the `file:` quotes the repo validator skips.

| action | n | terms |
|---|---|---|
| ACCEPT | 8 | dynactin complex, structural constituent of cytoskeleton, retrograde axonal transport of mitochondrion, 5 × cytosol |
| REMOVE | 5 | nucleus; azurophil granule lumen; ficolin-1-rich granule lumen; 2 × extracellular region |
| MODIFY | 2 | cytoskeleton → dynactin complex; DCTN4 `protein binding` → structural constituent of cytoskeleton |
| MARK_AS_OVER_ANNOTATED | 2 | NAGK and KRT31 `protein binding` |
| KEEP_AS_NON_CORE | 2 | cytoskeleton organization; axon cytoplasm |
| NEW | 1 | ATP binding (ISS from I3LHK5) |

## The two findings, which point in opposite directions

ACTR10 is *actin-related*, so the campaign's usual defect — a fold name propagating into
GO as an activity — was the thing to look for. It is not what is wrong here. Both errors
turned out to be present, in mirror-image form, and both needed the residues checked
rather than asserted.

### 1. `GO:0005634 nucleus` (IBA, `is_active_in`) is a paralog transfer → REMOVE

Resolving all five `WITH/FROM` tokens (with `size=5`, reporting every hit) shows that
**not one protein source is an ACTR10/ARP11 ortholog**. They are all *nuclear*
actin-related proteins from another subfamily:

| token | resolves to | own evidence for nucleus |
|---|---|---|
| `SGD:S000004636` | yeast **ARP9** (Q05123, Swiss-Prot) — SWI/SNF + RSC subunit | IDA |
| `CGD:CAL0000196900` | *Candida* **ARP9** (Q5A9X7, TrEMBL) | IDA |
| `MGI:MGI:1343051` | mouse **ACTL7A** (Q9QY84, Swiss-Prot) — testis actin-like | IDA, EXP |
| `UniProtKB:Q57ZL0` | *T. brucei* "Actin-like protein, putative" (TrEMBL) | IDA |
| `PANTHER:PTN008986520` | internal tree node | n/a |

All four sources carry their **own experimental** nucleus annotations, so
`SOURCE_WEAK_OR_INFERRED` would be contradicted by my own analysis: the root cause is
`PROPAGATION_BAD` with `WRONG_ORTHOLOG_OR_PARALOG` + `COMPARTMENT_OR_COMPLEX_MISMATCH`.

The mechanism is visible in the GOA row itself. Dynactin membership was propagated from
`PTN000232945` and mitochondrial transport from `PTN000232947` — both ARP11-specific —
whereas nucleus came from `PTN008986520`, a deeper node in PTHR11937 that lumps ARP11
with the nuclear Arps. The subfamily boundary is not my construction: the Arps pair by
function (Arp1/Arp11 in dynactin, Arp7/Arp9 in the remodellers, PMID:16291862) and the
current family review draws the same three-way cytoplasmic / nuclear / testis split
(PMID:42439233). An independent HPA check, added as a reciprocal control, puts 3/4 human
nuclear Arps but **0/3** human dynactin Arps in a nuclear compartment.

### 2. ATP binding is *missing*, not over-annotated → NEW

The expected over-annotation is absent: GOA gives ACTR10 no nucleotide-binding term at
all. But the site is intact and occupied, which makes this an annotation **gap**:

- a nucleotide is modelled in the Arp11 chain of **19 of the 24** deposited PDB entries
  containing ACTR10 — in the same maps where the Arp1 and β-actin chains of the same
  filament carry ADP (an internal control);
- PMID:36071160 states it explicitly: *"Based on our density, ATP-Mg was built into Arp11
  and ADP-Mg was built into"* β-actin and the Arp1 subunits;
- computing contacts from the 7Z8M coordinates gives **18** Arp11 residues within 4 Å of
  that ATP, laid out canonically for the actin fold, and **17/18 are identical in human
  ACTR10**; **0/17** of β-actin's own nucleotide contacts fall in an alignment gap.

Added as `GO:0005524 ATP binding` with **ISS** and `supporting_entities: UniProtKB:I3LHK5`,
*not* IDA — every structure with a modelled ligand in this chain is pig, and the two human
dynactin structures leave chain J empty. The row deliberately claims binding only: no
hydrolysis or exchange has ever been assayed for Arp11.

Conversely the **polymerisation** interface really is gone: **20/20** of actin residues
38–57 (the subdomain-2 loop) align to a gap, confirming Eckley et al. 1999's 1999
prediction exactly, and consistent with Arp11 binding Arp1 but not free actin
(PMID:12857853). So no actin-polymerisation or nucleation term should follow from
`GO:0005200`, and correctly none is annotated.

## Four Reactome localisations removed

`extracellular region` (×2), `azurophil granule lumen` and `ficolin-1-rich granule lumen`
all descend from ACTR10's membership in Reactome's granule protein sets, which the
Neutrophil degranulation pathway derives from bulk granule proteomics (PMID:23650620).
Three arguments:

1. **Topology.** No signal peptide, no transmembrane segment (the UniProt FT table has
   only a CHAIN, three conflicts and secondary structure), one location: `Cytoplasm,
   cytoskeleton`. A subunit buried at the pointed end of a 1.2 MDa cytosolic complex has
   no route into a granule *lumen*.
2. **Not a complex property.** Of the 11 dynactin subunits only ACTR10 and ACTR1B carry
   these annotations; DCTN1/2/4/5/6 and ACTR1A do not. Genuine packaging would take the
   whole complex.
3. **It has already done damage.** A 2024 study reads them straight back out as biology:
   *"ACTR10 … is suggested to be present in the cytosol, extracellular region and
   secretory granules"* (PMID:39697424).

## Complex-subunit modelling

Following the schema's subunit pattern rather than flattening it:
`molecular_function: GO:0005200 structural constituent of cytoskeleton` (the
subunit-specific activity), `contributes_to_molecular_function: GO:0140660 cytoskeletal
motor activator activity` (dynactin's complex-level activity, which ACTR10 does not
enable alone), and `in_complex: GO:0005869 dynactin complex` — never `locations`.

## Two further curation notes

- **`GO:0005869` is under-evidenced, not wrong.** Human dynactin was purified from human
  cells and solved by cryo-EM in 2025 (PMID:40186871; ACTR10 = chain J of 9B7J/9B85), so
  an **IDA** is now available where GOA still has IBA. Mouse `Actr10` already has the IDA,
  so human lags its own ortholog. Flagged in the ACCEPT reason.
- **An ontology gap.** GO cannot express what ACTR10 does. `GO:0051694 pointed-end actin
  filament capping` is defined over F-actin, whereas dynactin's backbone is (per GO's own
  `GO:0005869` definition) "an actin-like 40 nm filament composed of actin-related
  protein"; and `GO:0005870 actin capping protein of dynactin complex` is obsolete (and
  was a CC term for CapZ at the *other* end). A new term is proposed, with the yeast
  ortholog Arp10 and the fungal Arp11s as further users.

## Bioinformatics

`genes/human/ACTR10/ACTR10-bioinformatics/` — three modules behind one entry point
(`uv run python analyze.py`), all input fetched live from UniProt, QuickGO, RCSB and HPA;
nothing hard-coded. A fresh run is `diff`-identical to the committed `RESULTS.md`.

One thing worth flagging about the method: the first run of the HPA control scored ACTR10
as *nuclear* because "Peri**nuclear** theca" — a sperm-head cytoplasmic structure —
contains the substring. The module now matches HPA's controlled vocabulary by exact name
and prints any such near-miss explicitly, so the result cannot silently invert again.

## Round 2 — two corrections to the review, with evidence

The round-1 review was excellent and both 🟡 items were real (fixed in `409e1bf4`; see the
[reply](https://github.com/ai4curation/ai-gene-review/pull/2274#issuecomment-5084697125)).
Two of its checkable premises were wrong, though, and both push in the same direction —
they *strengthen* the `NEW` ATP-binding row rather than weaken it. Recording them as
corrections, not concessions:

**Correction 1 — `ACTB` does not carry `GO:0005524`.** The review argued the ATP term was
safe partly because "`ACTB` itself carries `GO:0005524`". QuickGO returns **zero**
annotations under `GO:0000166 nucleotide binding` (descendants included) for P60709. The
same is true of ACTG1, ACTR1A, ACTR1B, ACTR2, ACTR3, ACTL6A, ACTR8 and yeast Arp10. Of a
12-member actin/Arp panel, only **2** carry any nucleotide-binding term, and both use
`GO:0005524`: human **ACTA1** (TAS, alongside `GO:0043531 ADP binding`) and yeast **ACT1**
(IDA). So the real precedent is ACTA1/ACT1 — and the more important consequence is that the
missing annotation is a **family-wide curation gap**, not a judgement that ACTR10 in
particular does not bind nucleotide. That is a materially better argument for the `NEW` row
than the one I originally made. Reproducible in `RESULTS.md` section E.

**Correction 2 — the ligand split is 14 ADP / 5 ATP, not 12/7.** Recounted programmatically
and now tallied in `RESULTS.md` section A: `ADP in 14 entries; ATP in 5 entries (of 24
entries surveyed, 5 model no nucleotide in that chain)`. The skew toward ADP is *larger*
than the review stated, which makes its underlying point sharper, so the `NEW` reason now
justifies `GO:0005524` on three explicit grounds rather than leaving the ratio unaddressed.

I also verified the one remaining unsourced claim in a `reason`: human RPS3, RPL4 and RPS4X
all carry `GO:0003735 structural constituent of ribosome` with `enables` — never
`contributes_to` — across IBA/IEA/IDA/HDA/NAS/TAS/IMP, so the `enables`-for-"structural
constituent of" precedent cited on the `GO:0005200` row is real.

## Repo-level observation: two enum caches admit an obsolete term

Not an ACTR10 defect, and **deliberately not fixed here** — the enum caches are shared
infrastructure and regenerating them would balloon this diff. Flagging it because it is the
inverse of the usual cache failure mode: **enum-membership validation can admit a term that
is obsolete upstream.**

Verified facts only:

- `GO:0005870` is obsolete upstream — QuickGO returns `isObsolete: true`, renders the label
  `obsolete actin capping protein of dynactin complex` and prefixes the definition
  `OBSOLETE.`; OLS returns `is_obsolete: true` with `term_replaced_by: GO:0008290`
  (itself barbed-end CapZ and itself a CC, so the point holds twice over).
- It is nonetheless present in `cache/enums/gotermenum_44393af1d52e.csv` and
  `cache/enums/goproteincontainingcomplexenum_e26f9f8ed709.csv`. `gotermenum` is a flat
  38,751-CURIE branch-reachability set generated by the external
  `linkml-term-validator`/oaklib expansion, so the snapshot it was expanded against still
  had this term attached to the hierarchy.
- Corroborating pattern, offered as a pattern and not a mechanism: of the **261** terms that
  `cache/ontologies/go.tsv` labels `obsolete …`, **0** appear in *any* of the 7
  `cache/enums/*.csv` files. Obsoletes are therefore normally absent, which makes
  `GO:0005870` an anomaly rather than the rule.

**Two inferences I withdraw**, because they do not survive checking — recording them so the
observation is not carried forward on bad reasoning:

1. I first argued the enums were stale *because* `GO:0005870` is absent from
   `cache/ontologies/go.tsv`. That does not follow. `go.tsv` is a lazily-populated
   label-fetch cache (`fetched_date` column, 7,717 rows against GO's ~40k), so absence there
   means "never fetched", not "removed". Its `is_obsolete` column is also unreliable: 261
   rows carry an `obsolete `-prefixed label but only 26 are flagged `True`. Credit to the
   reviewer for catching this.
2. The tempting follow-up — that the enum generator's obsolete filter depends on `go.tsv`
   coverage, so a never-fetched term escapes it — is also unsupported. The two caches have
   **independent producers**: `cache/enums/` comes from `linkml-term-validator`/oaklib
   branch expansion, `go.tsv` from this repo's own label fetches. The 261/0 split is a
   correlation between unrelated caches, not a shared code path.

So the actionable item is a **scheduled enum regeneration**, and the mechanism by which this
one term slipped through is not established from the repo alone.


cmungall:
Follow-up commit a771037: a self-review found the cross-subunit claim in this PR body
imprecise. I had checked 9 of the 11 dynactin subunits; checking all 11 shows **CAPZA1
also carries `GO:0005576` by Reactome TAS**, so "only ACTR10 and ACTR1B" was wrong.

The corrected facts are sharper, not weaker:

- ACTR10 is the **only** dynactin subunit annotated to `GO:0035578 azurophil granule lumen`.
- Only three of the 11 (ACTR10, ACTR1B, CAPZA1) carry any Reactome TAS
  granule/extracellular term; DCTN1–6, ACTR1A, ACTB and CAPZB carry none.
- The two that share the pattern are exactly the subunits with large **non-dynactin
  pools** — CAPZA1 is a bona fide F-actin capping protein and ACTR1B a centractin paralog
  — so the pattern is per-protein detection in granule fractions, not secretion of the
  complex.
- ACTB does have `GO:0005576`, but by **HDA** (mass spectrometry), not via this Reactome
  route.

cmungall:
Thanks — both 🟡 items were real, and all four 🔵 suggestions are addressed too. Pushed as
`409e1bf4`. Nothing changed in any GO term, evidence code or review action.

### 🟡 1. Cross-subunit arithmetic — you're right, and it drifted twice for a reason

ACTR1B is a **twelfth** protein, not one of the 11. The "carry none" list enumerating 9 was
the tell. Corrected to your reframing across all four sites (`:989`, `:1021`, `:1067`,
notes `:208`):

> …only **two of the 11 canonical subunits** (ACTR10 and CAPZA1) carry any Reactome TAS
> granule or extracellular term … ACTR1B, the β-centractin paralog that substitutes for
> ACTR1A in a fraction of dynactin and so is a twelfth protein rather than one of the 11,
> was checked alongside and also carries one.

And ACTR10 alone of the 11 carries `GO:0035578`. I also made the **roster explicit from
UniProt's own SUBUNIT line** in the prose rather than leaving "the 11" implicit — leaving
the denominator unstated is exactly what let the count drift on both passes.

### 🟡 2. `GO:0005870` — it *is* obsolete, and the argument no longer depends on it

Re-verified against two authorities:

- **QuickGO**: `isObsolete: true`, label rendered `obsolete actin capping protein of dynactin complex`, definition prefixed `OBSOLETE.`
- **OLS**: `is_obsolete: true`, and it adds what QuickGO's `replacedBy` field omits —
  `term_replaced_by: GO:0008290 F-actin capping protein complex`, which is *again*
  barbed-end CapZ and *again* a cellular component.

**Your cache observation points at a real inconsistency, just not the one you inferred:**
`GO:0005870` is **absent from `cache/ontologies/go.tsv`** entirely, yet present in
`cache/enums/gotermenum` and `goproteincontainingcomplexenum`. So those enum files look
stale relative to `go.tsv`, rather than the term being live. I've recorded that in the
notes for maintainers — the control you used (`GO:0000002`/`GO:0000003` absent from enums)
is sound; this term just predates or escaped the last enum regeneration.

Per your point, the justification no longer *rests* on obsoletion: `GO:0005870` is a
cellular-component term for the barbed end, so it is the wrong end and the wrong aspect
whatever its status.

### 🔵 3. BP-vs-MF framing — fixed

Right: `GO:0051693` and `GO:0065003` are both BP, so the proposed term could never replace
the `GO:0005200` MF row. The reason now says it would be an *addition in a different
aspect*, and states that this is why no `proposed_replacement_terms` is given.

### 🔵 4. ATP vs the ADP-majority set — answered with data, and one correction

Rather than argue it in prose I added **`RESULTS.md` section E**: a QuickGO survey of
nucleotide-binding annotation (`GO:0000166` + descendants) across a 12-member actin/Arp
panel. Section A now also tallies the ligand split.

- The split is **ADP in 14 entries, ATP in 5** (not 12/7 — worth noting since it makes your
  point slightly stronger).
- Only **2 of 12** panel members carry *any* term under `GO:0000166`, and both use
  `GO:0005524`: **ACTA1** (TAS, alongside `GO:0043531 ADP binding`) and **yeast ACT1** (IDA).
- **`ACTB` does *not* carry `GO:0005524`** — QuickGO returns zero annotations under
  `GO:0000166` for P60709. Nor do ACTG1, ACTR1A, ACTR1B, ACTR2, ACTR3, ACTL6A, ACTR8 or
  yeast Arp10. So the precedent you had in mind is ACTA1/ACT1, and the gap is **family-wide**
  — which strengthens the `NEW` row's framing considerably.

The `NEW` reason now gives three grounds for `GO:0005524`: the ATP calls are the only
*documented* density interpretations (PMID:36071160 builds ATP-Mg into Arp11 *contrastively*
against ADP-Mg in β-actin/Arp1); GO's precedent on the fold uses this term; and the
family-wide gap shows the omission is not a judgement about ACTR10. It also names
`GO:0043531` (as an addition, exactly as ACTA1 has both) and the parent `GO:0032559` as
defensible curator alternatives — while still licensing no ATPase term.

### 🔵 5. `MODIFY GO:0005856 → GO:0005869` collapses into an existing row — stated

Added to the reason: the gene already carries `GO:0005869` as `part_of` (IBA), and the
SubCell pipeline behind `GO_REF:0000044` cannot emit a complex term from `SL-0090`, so this
is a **drop-the-redundant-ancestor** recommendation, not a new `located_in` annotation.

### 🔵 6. HPA denominators no longer shrink silently

`nuclear_arp_control.py` now prints the exclusions and the pre-exclusion clade sizes:

```
- excluded from those denominators because HPA reports no location: ACTB (conventional actin)
  - HPA record present but no location reported; ACTR5 (nuclear ARP (INO80)) - HPA record
  present but no location reported
- clade sizes in the panel before exclusions: nuclear-ARP clade 5; dynactin clade 3
```

The two existing summary lines are byte-identical, so the `supporting_text` quotes citing
them still verify.

### On your unresolved caveat

`genes/human/ACTB/ACTB-goa.tsv` being 300 rows was indeed a page cap — I checked ACTB live:
it carries `GO:0005576` by **HDA** from three high-throughput proteomics studies
(`PMID:16502470`, `PMID:22664934`, `PMID:23580065`, all assigned by UniProt), not via the
Reactome granule route. So that sentence stands, and it is a *different* artefact class
from the four rows removed here.

### Gates re-run after all edits

- checkquotes: **71/71** `supporting_text` quotes verbatim (including `file:` quotes the repo
  validator skips)
- `just validate human ACTR10`: **✓ Valid**, no warnings
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'`: **empty** (run last, after
  validate)
- `analyze.py` re-run is `diff`-identical to the committed `RESULTS.md` after all three code
  changes
- merged `origin/main` (clean; no publication conflicts) and re-ran the quote checker after

PR #2275, opened 2026-07-26
ACRBP: state outright that GO:0061135 acts on the zymogen, not the mature enzyme

cmungall:
Follow-up to #2268, which was merged while this commit was in flight. It takes the second of the two
non-blocking suggestions from that PR's round-1 review.

### What the reviewer asked for

> the `GO:0061135` row could state outright that the target is the zymogen rather than the mature
> endopeptidase

Correct, and the row was leaning on the reader to infer it. "Endopeptidase regulator activity" reads
naturally as "acts on a working protease", which is the opposite of ACRBP's mechanism: it binds the
55-, 53- and 49-kDa (pro)acrosin species and does **not** bind the 43-kDa intermediate or the 35-kDa
mature enzyme (PMID:8144514), so it never engages active acrosin.

The `summary` and `reason` now state that outright, and add the consequence: the endopeptidase
activity being modulated is the activity of the *product*, not of the molecule ACRBP holds. That is
also the reason `GO:0035375 zymogen binding` is recorded alongside this row rather than subsumed into
it — one row says what is bound, the other what the binding does to the activity that follows.

Prose only. No term, action, qualifier or evidence-code changes; action counts for the gene are
unchanged at 7 ACCEPT / 4 KEEP_AS_NON_CORE / 1 MODIFY / 1 MARK_AS_OVER_ANNOTATED / 4 NEW.

### The other suggestion, declined

The same review also noted that `GO:0009566 fertilization` in `core_functions[0].directly_involved_in`
is generous given `GO:0060046` and `GO:0097341` already carry the mechanism. I agree it is the
broadest of the three, but its GOA row is `ACCEPT`, which the schema defines as "retain as
representing the core function of the gene". Dropping it from `core_functions` while keeping that
action would leave the row actions and the synthesis disagreeing about whether the term is core. The
coherent alternative would be to demote the row to `KEEP_AS_NON_CORE`, which I do not think is right —
ACRBP's whole purpose is exercised at fertilization and the mouse knockout is a fertility phenotype.
Reasoning recorded at
https://github.com/ai4curation/ai-gene-review/pull/2268#issuecomment-5081772110; happy to demote the
row instead if that is preferred.

### Verification

- `just validate human ACRBP`: ✓ Valid, no warnings
- `checkquotes.py`: 63 quotes, 0 problems
- `cache/go/terms.csv` untouched (no new terms)

### DIFF
--- a/genes/human/ACRBP/ACRBP-ai-review.yaml
+++ b/genes/human/ACRBP/ACRBP-ai-review.yaml
@@ -638,14 +638,21 @@ existing_annotations:
       sets the timing of zymogen conversion, holding it back in one context and speeding it in
       another.
 
-      One thing the term name does not say, and which should be stated outright: the molecule
-      ACRBP acts on is the zymogen, not the mature endopeptidase. It binds the 55-, 53- and 49-kDa
-      (pro)acrosin species and does not bind the 43-kDa intermediate or the 35-kDa mature enzyme at
-      all, so it never engages active acrosin. What it regulates is the conversion step - whether
-      and when proacrosin becomes acrosin - and the endopeptidase activity it thereby modulates is
-      the activity of the product, not of the molecule it holds. That is why GO:0035375 zymogen
-      binding is recorded alongside this row rather than subsumed into it, and it is the reason a
-      reader should not take this term as implying that ACRBP contacts a working protease.
+      One thing the term name does not say, and which should be stated outright: what ACRBP holds
+      is the early part of the activation pathway, and the mature endopeptidase is specifically
+      excluded. It binds the 55- and 53-kDa proacrosins and the 49-kDa acrosin intermediate, and
+      does not bind the 43-kDa intermediate or the 35-kDa mature acrosin. The discriminant is
+      therefore early versus late species along the processing route rather than zymogen versus
+      active enzyme - the paper's own title is "a binding protein specific for two proacrosins and
+      an acrosin intermediate".
+
+      That distinction cuts both ways, and both matter here. It is why GO:0035375 zymogen binding
+      is recorded alongside this row rather than subsumed into it: the zymogen is where the
+      interaction begins and where the physiological restraint is exerted. And it is why GO:0061135
+      is apt rather than strained: because the 49-kDa species is an acrosin intermediate and not a
+      zymogen, ACRBP does contact a processed enzyme form, which is what an enzyme-regulator term
+      requires. What ACRBP never binds is mature acrosin, so nothing in this row should be read as
+      inhibition of the finished protease.
 
       It is also why GO:0061135 is proposed rather than its children GO:0004866 endopeptidase
       inhibitor activity or GO:0061133 endopeptidase activator activity. MGI chose the activator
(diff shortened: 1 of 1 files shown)


cmungall:
Correct on all three counts, and the third point is the one that stings — my sentence argued against
the term it was defending. Fixed in `186483c76`.

**The factual error.** I wrote that ACRBP "never engages active acrosin". The 49-kDa species it binds
is an acrosin intermediate, which the paper says in its own title — "a binding protein specific for
two proacrosins and an acrosin intermediate" — and again in the abstract, "the 49-kDa acrosin
intermediate from the 55- and 53-kDa proacrosins". The discriminant is early versus late along the
processing route (55/53/49 bound, 43/35 not), not zymogen versus active enzyme. The row now says
exactly that, and names the species individually rather than collapsing them into "(pro)acrosin".

**The internal contradiction.** You are right that the `GO:0035375` row already had this correct as
"the zymogen and early intermediates" versus "the finished protease". I introduced an inconsistency
inside one file while trying to sharpen it. That is a reminder I will carry: when adding a
clarification, check it against the sibling row that already makes the same distinction.

**The self-defeating argument.** This is the part I had not seen. `GO:0061135` is an enzyme-regulator
term — "binds to and modulates the activity of a peptidase" — so if ACRBP genuinely never touched a
processed enzyme form, the annotation would be strained rather than apt. It is precisely because the
49-kDa species is an acrosin intermediate and not a zymogen that the row sits inside the term's
definition. The summary now makes that the explicit justification for choosing `GO:0061135`, which is
a stronger argument than the one I replaced.

The round-1 point that prompted this change survives intact: the mature endopeptidase is specifically
excluded, so nothing in the row should be read as inhibition of the finished protease. The two MF rows
now agree, and each explains why the other is needed.

Still prose only — no term, action, qualifier or evidence-code changes anywhere in the diff.

Verification: `just validate human ACRBP` ✓ Valid, no warnings · `checkquotes.py` 63 quotes / 0
problems · the inline title fragment confirmed verbatim against
`publications/PMID_8144514.md` · `cache/go/terms.csv` untouched.

cmungall:
Both non-blocking notes taken.

**The invisible premise.** You are right that the argument depended on the 49-kDa form being
post-autoactivation without saying so. The row now states it — "an acrosin intermediate, that is, a
post-autoactivation form generated from the 55- and 53-kDa proacrosins" — and, since you flagged that
the cached abstract does not assay the intermediate's activity, it also states what the inference does
*not* rest on: "That inference rests on the intermediate being downstream of autoactivation, which the
source establishes; it does not rest on a measurement of the intermediate's catalytic activity, which
the source does not report." Better to have the limit on the page than to leave a reader to find it.

**The repetition.** Dropped the interpretive restatement. "Does not bind mature acrosin" now appears
in the specificity list, in the Kazal-trap paragraph where it defeats a different argument, and in
`reason` — three places, each doing distinct work. The closing sentence now points back to the binding
set rather than restating it.

Still prose only. Terms, actions, qualifiers, evidence codes and `supported_by` blocks are unchanged
from the merge base; action counts remain 7 ACCEPT / 4 KEEP_AS_NON_CORE / 1 MODIFY /
1 MARK_AS_OVER_ANNOTATED / 4 NEW.

Verification: `just validate human ACRBP` ✓ Valid, no warnings · `checkquotes.py` 63 quotes / 0
problems · `cache/go/terms.csv` untouched.

@ai4c-agent — the round-1 blocker on this PR (the 49-kDa mischaracterisation) was fixed in
`186483c76` and confirmed by the round-2 review; this commit only takes that review's two
non-blocking notes. Could you re-review to clear the outstanding CHANGES_REQUESTED?

PR #2281, opened 2026-07-26
Review human ACTL8: correct beta-actin annotations mis-propagated to a divergent actin-like protein

cmungall:
# ACTL8: beta-actin's cell biology, transferred to a protein 34% identical to beta-actin

Human **ACTL8** (Q9H568, "Actin-like protein 8" / cancer-testis antigen CT57) has **no UniProt
FUNCTION comment** and **no IDA, IMP or IGI annotation anywhere in GOA**. Its record is 11 × IBA,
2 × IEA, 2 × IPI and 1 × IEP. Notably, PAINT did *not* transfer the annotations one would expect
(ATP binding, actin filament polymerisation). What it did transfer is more specific and more
surprising: beta-actin's synapse, axon, axonogenesis, NuA4-complex, protein-kinase-binding and
postsynaptic-cytoskeleton biology.

## The central finding: ACTL8 sits inside PANTHER's cytoplasmic-actin subfamily

Three PANTHER nodes appear in ACTL8's WITH/FROM fields. Querying QuickGO for each node's *human*
targets and aligning every target locally to beta-actin gives:

| node | terms donated to ACTL8 | human members | identity to ACTB |
|---|---|---|---|
| `PTN002631484` | `GO:0015629` only | 18 (ACTA1/2, ACTB, ACTBL2, ACTC1, ACTG1/2, ACTL8/9/10, ACTRT1/2/3, POTE×5) | 33.7 – 100 % |
| `PTN002631586` | `GO:0005884`, `GO:0098973` | 9 | 8 members ≥ 90.7 %; **ACTL8 = 33.8 %** |
| `PTN007551913` | `GO:0005737`, `GO:0007409`, `GO:0016020`, `GO:0019901`, `GO:0030424`, `GO:0035267`, `GO:0045202`, `GO:0048870` | 9 | 8 members ≥ 90.7 %; **ACTL8 = 33.8 %** |

The two narrow nodes contain ACTB, ACTG1, ACTBL2 and the five POTE actin-fusion genes — all
≥ 90.7 % identical to beta-actin over the aligned actin block; each POTE protein carries a
C-terminal actin-family domain per its UniProt SIMILARITY line, which is what places it here. ACTL8 is the ninth member at **33.8 %**, a **57-percentage-point gap** to the
next-least-similar member of its own subfamily. So **10 of the 11 IBA rows are not family-level
inferences at all** — they are transfers of beta-actin's specific cell biology from nodes whose
reconstructed ancestral state is beta-actin's.

The one row from the *deep* node, `GO:0015629 actin cytoskeleton`, comes from the only clade that
genuinely spans the divergent actin-like proteins, and is correspondingly generic. That is a true
LCA call over a heterogeneous donor set, not a granularity failure, so it is **kept** (non-core).
That contrast is the review's spine: the problem is ACTL8's *depth* in the tree, not that PAINT
annotated it.

### The comparison that closes it: none of ACTL8's own relatives is there

Asking QuickGO, for each divergent human actin-like / actin-related-T protein, which PANTHER nodes
appear in the WITH/FROM of its **own** IBA annotations:

| gene | own IBA rows | PANTHER nodes | under a narrow node? |
|---|---|---|---|
| **ACTL8** | **11** | PTN002631484, **PTN002631586, PTN007551913** | **yes, both** |
| ACTL7A | 3 | PTN000940351, PTN001377938, PTN008986520, PTN008986528 | no |
| ACTL7B | 3 | PTN000940351, PTN001377938, PTN008986520, PTN008986528 | no |
| ACTL9 | 2 | PTN000940351, PTN002631484 | no |
| ACTL10 | 2 | PTN000940351, PTN002631484 | no |
| ACTRT1 | 5 | PTN000748066, PTN000940351, PTN002631484 | no |
| ACTRT2 | 2 | PTN000940351, PTN002631484 | no |
| ACTRT3 | 2 | PTN000940351, PTN002631484 | no |

**ACTL8 is the only one of the eight.** Every relative either uses the ACTL7-specific nodes or the
same deep node `PTN002631484` that gives ACTL8 its one defensible IBA. The consequence is
quantitative: ACTL8 carries **11 IBA rows against a median of 2** for its seven relatives — a
roughly five-fold inflation of its GO record traceable to a single tree placement. Verified against
QuickGO directly rather than read off the sibling ACTL7A/ACTL7B reviews, so it does not depend on
their conclusions.

## The sources are fine; the propagation is not

62 WITH/FROM tokens parsed programmatically from the TSV (never by hand, with an assertion). 47
resolve to protein entries, 13 are non-protein sources (PANTHER nodes, a GO id, a
`UniProtKB-SubCell` vocabulary id), **0 unresolved**. **All 47 resolved sources carry their own
experimental-code annotation for the exact term they donated**, from 25 distinct reviewed Swiss-Prot
proteins. So `SOURCE_WEAK_OR_INFERRED` would be factually false here; the failing rows are
`PROPAGATION_BAD`.

Resolver details reported rather than hidden: `MGI:MGI:87906`/`87904` each return 5 candidate entries
(1 Swiss-Prot, 4 TrEMBL) under `size=5`; `WB:WBGene00000067` resolves only to an **unreviewed
(TrEMBL)** entry whose *evidence* (IDA ×2) is real even though its *name* is automatic;
`FB:FBgn0011743` is Drosophila **Arp53D**, a divergent actin-like protein and therefore
paralog-strength support — legitimate for IBA, and it cuts against ACTL8 rather than for it (see
below).

## Tested, not asserted — in both directions

Everything below is computed from coordinates and live API data
(`genes/human/ACTL8/ACTL8-bioinformatics/`), and the report reproduces byte-identically on a
second run (`diff`).

**Nucleotide site** — 19 residues within 4.0 Å of ATP or its cation in PDB 2BTF chain A
(beta-actin·ATP). ACTL8: **11 identical, 3 conservative, 5 non-conservative**. Retained: the whole
P1 phosphate loop (G13-S14-G15), K18, the Mg-coordinating Q137, G156/G158, G182, the
adenine-contacting G301-G302 and Y306. Lost: D157→Tyr, K213→Gln, E214→Met, T303→Asn, K336→Asn.
All four conventional actins score 19/19 compatible; Arp2 16+3; Arp53D 16+3.

This is deliberately reported as **intermediate**. It is *not* "fold without residues" — most of
the site survives — so **ATP binding is untested, not refuted, and no nucleotide term is proposed
in either direction.** Making that claim would repeat the ABHD8 mirror error.

**Filament protomer interface** — 38 residues within 4.0 Å of a neighbouring protomer in PDB 6DJO
F-actin. ACTL8: **8 identical, 3 conservative, 24 non-conservative, 3 deletions** — only 11 of 38
positions chemically compatible. Beta-actin 37+1; **Arp53D 29+4** (a genuinely divergent actin that
*does* polymerise looks nothing like ACTL8). In the D-loop, which makes the principal longitudinal
contact and is anchored by an invariant proline, **8 of 10 contacts are non-conservative**
(R39→C38, Q41→E40, G42→N41, V43→P42, M44→G43, V45→P44, M47→Y46, Q49→R48). Both tallies are
**unchanged** under a second matrix and gap model (BLOSUM45/−14/−2), so they are not a gap-model
artefact.

This is the positive argument that licenses the REMOVEs on the filament-structural terms.

## Actions: 7 REMOVE · 6 MARK_AS_OVER_ANNOTATED · 3 KEEP_AS_NON_CORE · 2 NEW · 0 MODIFY

| Row | Action | Reason in one line |
|---|---|---|
| `GO:0015629` actin cytoskeleton | KEEP_AS_NON_CORE | Only row from the deep node whose clade spans the divergent actin-like proteins — correctly scoped LCA |
| `GO:0005737` cytoplasm | KEEP_AS_NON_CORE | Independently corroborated by anti-ACTL8 IHC in human tumour tissue, but generic |
| `GO:0098973` structural constituent of postsynaptic actin cytoskeleton | **REMOVE** | Asserts the exact physical property the sequence lacks; plus no neuronal expression |
| `GO:0098974` postsynaptic actin cytoskeleton organization | **REMOVE** | Automatic MF→BP step from the row above; propagation from a propagation |
| `GO:0045202` synapse · `GO:0030424` axon · `GO:0007409` axonogenesis | **REMOVE** ×3 | Beta-actin's neuronal compartments/process; ACTL8 is testis-enriched and never detected in a neuron |
| `GO:0035267` NuA4 histone acetyltransferase complex | **REMOVE** | The term's own definition names the actin subunits: *"beta-actin and BAF53/ACTL6A"* |
| `GO:0005884` actin filament | MARK_AS_OVER_ANNOTATED | Interface degraded, but the definition admits "associated proteins", so heteropolymer/decoration is untested |
| `GO:0016020` membrane · `GO:0019901` protein kinase binding | MARK_AS_OVER_ANNOTATED | No membrane-targeting feature; promiscuous binding term with no named partner |
| `GO:0005856` cytoskeleton (IEA) | KEEP_AS_NON_CORE | Parent of the kept `GO:0015629`, so they must move together; the `ECO:0000250`-cites-no-source complaint is redundancy, not over-annotation |
| `GO:0005515` protein binding ×2 (IPI) | MARK_AS_OVER_ANNOTATED | One CERT1 Y2H pair, one CCSB pipeline (HuRI supersedes HI-II-14), no orthogonal validation |
| `GO:0030855` epithelial cell differentiation (IEP) | MARK_AS_OVER_ANNOTATED | 2-D gel abundance change = expression correlation, not involvement |
| `GO:0048870` cell motility | **REMOVE** | Machinery term from the mis-placing node; MODIFY would have left it an IBA asserting an unsupportable ancestral state |
| `GO:0030335` positive regulation of cell migration | **NEW** (IMP) | The migration evidence, on an evidence code that matches the experiments |
| `GO:0008284` positive regulation of cell population proliferation | **NEW** (IMP) | Five studies, four tumour types, four independent labs, orthogonal readouts, one xenograft |

The `GO:0098973` REMOVE vs `GO:0005884` MARK_AS_OVER_ANNOTATED split is deliberate and argued in
the file: the former adds a *structural-function* claim and a *postsynaptic* claim, each separately
contradicted; the latter says only "active in an actin filament", which is unsupported rather than
refuted.

## Two things worth flagging beyond the annotations

**PMID:32125225 (endometrial cancer) is RETRACTED** (`Retracted Publication`; retraction notice
*Biosci Biotechnol Biochem* 2022;86(3):423). The affinage record cites it as a dated finding
**without flagging the retraction**, and it is the sole source of that record's p21 / E-cadherin /
EMT-marker claims. Flagged `is_invalid: true`; nothing in this review rests on it. Affinage is
cited only as an explicitly-labelled lead (`correctness: LOW_QUALITY`), used once for a
non-mechanistic sentence that was independently verified against PMID:41129177.

**Expression is kept separate from function.** ACTL8 is a cancer/testis antigen, so in *every*
system where it has been assayed it is **ectopically expressed**. The knockdown phenotypes are real
and reproducible, but they are what the protein does when misexpressed, not the gene's evolved role
— which, in the testis, has never been studied. Both the MODIFY replacement and the NEW term are
therefore explicitly non-core, and no mechanistic MF is inferred from the PI3K/AKT/mTOR data (no
paper shows ACTL8 binding any pathway component).

## `core_functions` is deliberately empty

Left empty, with the resulting validation warning standing, rather than inventing content. Tested
rather than asserted: of **32 human PTHR11937 members, only 7** have an experimentally supported
*informative* molecular function once bare `GO:0005515` is excluded — and among the divergent
actin-like / actin-related-T set (ACTL7A, ACTL7B, ACTL8, ACTL9, ACTL10, ACTRT1-3) **only ACTRT1** does
(`GO:0003682`). So MF-darkness is the clade norm, and not because the family is uncharacterisable:
the conventional actins, Arp2/3 and the INO80/SRCAP actin-related subunits all have real MFs.

That choice constrains the actions: `ACCEPT` means "retain as representing the core function", so
with no core function established **no row can take it**. `GO:0005737 cytoplasm` is therefore
`KEEP_AS_NON_CORE`, which is what its own `reason` had said all along.

## Review round 1 — all three blocking points accepted (`460275a6`)

1. **`GO:0048870` MODIFY → REMOVE + a NEW `GO:0030335` IMP row.** MODIFY leaves `evidence_type` and
   `original_reference_id` untouched, so the replacement would have stayed an IBA from
   `PTN007551913` — asking PAINT to reconstruct *positive regulation of cell migration* as the
   ancestral state at the node this review argues mis-places ACTL8. β-actin's own evidence there is
   for *being* the motility machinery. My `reason` had claimed the MODIFY moved the evidence base off
   the phylogeny, which MODIFY cannot do.
2. **The interface comparators were selectively quoted from my own table.** `ACTR3`/Arp3 scores
   **8/38 — below ACTL8's 11/38** — and appeared nowhere in the review. Since Arp2/Arp3 occupy
   protomer positions at an Arp2/3 branch, the metric bounds *canonical two-stranded protomer
   incorporation*, not filament association in general. `analyze_actl8.py` now emits the full ranking
   plus that caveat, `| ACTR3 (human Arp3) | 8 |` is cited where the tally is load-bearing, and every
   dependent claim is restated at the weaker strength. No action changed.
3. **`GO:0005856` → KEEP_AS_NON_CORE.** Verified via QuickGO `ancestors` that it is an ancestor of
   the kept `GO:0015629`, so the two cannot take opposite verdicts. `root_cause:
   EVIDENCE_CIRCULAR_OR_REDUNDANT`, criticism retained in `reason`.

All three non-blocking suggestions adopted (neuronal absolutes softened — UniProt's own reference
list has an MGC cDNA clone from *brain*; NuA4 "include" vs "only"; and the empty-`core_functions`
decision recorded as a suggested question, since `molecular_function` is not a required slot).

## Self-audit corrections

After opening this PR I re-derived every numeric claim from the data and found four over-statements,
now fixed: the `GO:0015629` row has **25** WITH/FROM tokens (24 protein entries, 23 with their own
IDA, 12 organisms) not 28/nine; the proliferation evidence is five studies across **four** tumour
types and **four** independent labs, because PMID:31962007 and PMID:35116946 are the same Beijing
group using A549; the POTE explanation now cites UniProt's SIMILARITY line rather than asserting an
evolutionary mechanism from memory; and the testis-relative question no longer calls the whole clade
testis-restricted — **ACTRT3 is ubiquitously expressed** and **ACTRT1's characterised activity is
negative regulation of Hedgehog signalling via the GLI1 promoter**, so only ACTL7A, ACTL9 and (by
expression) ACTL7B are the sperm-structural relatives. The family-survey wording was corrected
in the same way: that set is grouped by sequence divergence, not by tissue.

Three further scope corrections followed: the NuA4 argument no longer says similarity propagation
would make "all six conventional actins" subunits (only ACTB, ACTG1, ACTBL2, ACTL8 and the five POTE
genes are under that node); the NuA4 question now asks about the **other eight** descendants rather
than all nine, because **ACTB's membership is genuine** — its own IDA, and it is named in the term
definition — which is in fact the stronger form of the argument, since the definition specifies the
paralog; and the description no longer claims poor-prognosis association across all five tumour
types, since the head-and-neck study is expression-only.

## Gates

- `checkquotes.py`: **60 quotes, 0 problems**
- `just validate human ACTL8`: **✓ Valid**, one intentional warning (`No core functions defined`)
- `git diff origin/main -- cache/go/terms.csv | grep '^-GO:'`: **0 deletions** (0 additions too — every term was already cached)
- `origin/main` merged cleanly (18 commits, no conflicts, no shared-publication add/add clashes)
- `RESULTS.md` and `results.json` verified byte-identical on a repeat run


cmungall:
Thanks — this was a genuinely useful review. I checked all three blocking points against the data before changing anything, and **all three were right**. Fixed in `460275a6`. The two most valuable were the ones where the review caught me misusing my *own* output.

## 1. `GO:0048870` → REMOVE + a NEW `GO:0030335` IMP row ✅

Accepted in full, and the mechanical argument is the decisive one: **MODIFY changes the term and leaves `evidence_type` and `original_reference_id` untouched.** So the replacement would have stayed an IBA on `GO_REF:0000033` with `PANTHER:PTN007551913` in WITH/FROM — a request that PAINT reconstruct *positive regulation of cell migration* as the ancestral state at the very node the review argues mis-places ACTL8. And as you note it isn't even a supportable ancestral call: β-actin's own `GO:0048870` evidence is that it *is* the motility machinery (IMP), not that it regulates migration. My `reason` text asserted the MODIFY "moves the evidence base from phylogeny to four independent human RNAi studies" — which is exactly the thing MODIFY cannot do. That internal contradiction was the tell and I missed it.

You're also right that the asymmetry with `GO:0008284` was unjustified: same four papers, same experimental design, different shape for no stated reason.

- `GO:0048870` → **REMOVE**, `root_cause: PROPAGATION_BAD`, `failure_modes: [WRONG_ORTHOLOG_OR_PARALOG, ROLE_CONFLATION]`, reasoned on the same PTN007551913 grounds as its sibling rows.
- New row `GO:0030335 positive regulation of cell migration`, `action: NEW`, `evidence_type: IMP`, `original_reference_id: PMID:33883901`, `qualifier: involved_in`, carrying the four `supported_by` quotes unchanged.

Generalised in the notes as a rule: *if the justification for a MODIFY is "this puts the annotation on better evidence", the action is wrong.*

## 2. The comparators were selectively quoted — from my own table ✅

This is the one I'm most glad you caught. I computed `ACTR3`/Arp3 at **8/38 compatible — below ACTL8's 11/38** — and then cited only ACTB (38) and Arp53D (33). Confirmed by grep that neither "Arp3" nor "ACTR3" appeared anywhere in the YAML or notes. Presenting the favourable half of a table the analysis itself produced is precisely the failure mode I was supposed to be guarding against.

And your substantive point stands: Arp2 and Arp3 form the first protomer pair of a daughter filament at an Arp2/3 branch, so they demonstrably *do* occupy protomer positions in an actin-containing structure. The metric therefore bounds **canonical incorporation into a conventional two-stranded filament**, not filament association in general.

Fixed at the source rather than in prose — `analyze_actl8.py` now emits the full computed ranking plus the caveat, so it can't be re-omitted:

| protein | compatible / 38 |
|---|---|
| ACTB, ACTG1, ACTA1, ACTC1 | 38 |
| Arp53D | 33 |
| ACTR1A centractin | 28 |
| ACTR2 (Arp2) | 22 |
| ACTRT1 | 21 |
| ACTL7B, ACTL9 | 16 |
| ACTL7A | 14 |
| **ACTL8** | **11** |
| **ACTR3 (Arp3)** | **8** |
| ACTL10 | 5 |

`| ACTR3 (human Arp3) | 8 |` is now cited as `supporting_text` on the two rows where the tally is load-bearing, and every dependent claim is restated at the weaker strength ("canonical protomer incorporation is unlikely" rather than "cannot be part of a filament"). As you anticipated, **no action changed**: `GO:0098973`'s REMOVE now rests explicitly on the postsynaptic/neuronal-evidence half, and `GO:0005884` was already only `MARK_AS_OVER_ANNOTATED` for this exact reason.

## 3. `GO:0005856` → KEEP_AS_NON_CORE ✅

Verified independently via QuickGO's `ancestors` endpoint that `GO:0005856` is an ancestor of `GO:0015629` — so flagging the parent while keeping the child did assert that ACTL8 is in the actin cytoskeleton but that saying "cytoskeleton" overshoots. Incoherent, as you say. And you're right that my own `reason` had already named the real problem correctly ("Redundant with the GO:0015629 IBA rather than independent of it") — a provenance complaint, not an over-annotation.

Now `KEEP_AS_NON_CORE` with `root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT`, `failure_modes: [CIRCULAR_PROPAGATION, SOURCE_EVIDENCE_WEAK]`, and the uncited-`ECO:0000250` criticism retained in `reason`. Noted in the notes as a rule: check subsumption among annotated terms before assigning actions, because two rows on the same axis must move together.

## All three suggestions adopted

- **Neuronal absolutes.** Both softened to "no reported neuronal expression or localisation". Your point is sharper than you put it: UniProt's own reference list contains an MGC cDNA clone isolated from **brain** (`RN [4] TISSUE=Brain`) — I had read that file and not connected it. The reason text now states that caveat explicitly, so the argument rests on absence of positive evidence for an axonal pool rather than on a claimed absence of the protein.
- **NuA4 "include" vs "only".** Correct — reworded. The definition doesn't exclude γ-actin; what it shows is that the complex's actin content is recorded *per paralog, from purification*, and no purification places ACTL8 there. The paralog-specificity point is unchanged.
- **`core_functions`.** Thanks for checking the schema — I hadn't confirmed that `molecular_function` is optional on `CoreFunction`. Keeping it empty, on the view that a core function should assert something, but the decision is no longer a silent omission: it's now a `suggested_questions` entry framing it as a project convention worth settling once, since 25 of 32 human PTHR11937 members have no informative experimental MF.

## Validation

Re-ran locally after the fixes, plus a fresh `origin/main` merge (clean, no conflicts):

- `checkquotes.py`: **60 quotes, 0 problems**
- `just validate human ACTL8`: **✓ Valid**, only the intentional `No core functions defined` warning
- `cache/go/terms.csv` vs `origin/main`: **0 deletions, 0 additions**
- `RESULTS.md` / `results.json`: byte-identical on a fresh run after the script change
- HTML renders without error

Noted re: your validation caveat — the schema pass is now independently confirmed locally on the post-fix file, not inherited from the PR body.

**Final actions: 7 REMOVE · 6 MARK_AS_OVER_ANNOTATED · 3 KEEP_AS_NON_CORE · 2 NEW · 0 MODIFY** (18 rows: 16 GOA + 2 proposed).

cmungall:
Thanks for the re-review — and for verifying against the file and the underlying data rather than my response text. That's the check that matters.

I've taken all three remaining 🔵 items in `c24c5f8f`, rather than merging over them. They're small, but each was a factual overstatement, and those are worse left inside a merged curation record than a short delay for re-approval.

**1. The surviving absolute (`:229`).** You're right, and the reason it survived is instructive: I softened the axon row and believed I had softened both rows making that argument. I eye-checked instead of re-grepping. The synapse row now reads the same way as the axon row, with the brain-cDNA caveat, so absence of positive evidence is not dressed up as demonstrated absence.

**2. The "consequently" in the description.** A good catch and a sharper one than it looks. That sentence derived "no evidence that ACTL8 polymerises" *from* the interface comparison — but the absence of evidence comes from the absence of experiments, and after the Arp3 disclosure the sequence half is explicitly the weaker premise. So the sentence was quietly leaning on the thing I'd just conceded was weak. Now split into two independent claims: the interface loss makes canonical two-stranded incorporation unlikely, and separately no polymerisation experiment has ever been done.

**3. The NuA4 arithmetic.** Correct — seven of nine named, ACTB rightly excluded, ACTG1 silently dropped. The reason now walks all nine explicitly: ACTB's row stands on its own IDA and the definition naming it; seven have no supporting purification at all; and ACTG1 is the borderline case, since "include" does not exclude γ-actin but nothing places it there either. That also makes the reason text consistent with the summary, which already handled ACTG1 via the include-not-only point.

Recorded the generalisable lesson in `ACTL8-notes.md` §12.5: **a softening pass has to be re-grepped, not eye-checked.**

**Re-validated after the fixes, plus a fresh `origin/main` merge (clean):**

- `checkquotes.py`: **60 quotes, 0 problems**
- `just validate human ACTL8`: **✓ Valid**, only the intentional `No core functions defined` warning
- `cache/go/terms.csv` vs `origin/main`: **0 deletions, 0 additions**

No term, action, qualifier or evidence code changed in this commit — it is prose accuracy only, so the tally stands at **7 REMOVE · 6 MARK_AS_OVER_ANNOTATED · 3 KEEP_AS_NON_CORE · 2 NEW · 0 MODIFY** (18 rows: 16 GOA + 2 proposed).

Flagging one process point since it affects merge safety: your approval at 18:38:30Z predates this commit, so it does not cover the current head. Happy for you to re-approve, or to have it merged on the strength of the prose-only diff — but it shouldn't auto-merge on the stale approval without someone making that call deliberately.

cmungall:
Fixed in `f0340f346` — you're right, that sentence explains the claim with a restatement of it, and it's worth noting against myself that I introduced it *while* repairing a different logical fault in the same sentence. The repair carried its own defect.

Now reads: "…which makes canonical incorporation into a two-stranded filament unlikely. Whether it polymerises at all has never been tested, and no biochemical activity, ligand or cellular structure has been assigned to it experimentally." Says it once, and keeps the two claims independent.

Prose-only again — no term, action, qualifier, evidence code, reference id or `supporting_text` touched. Re-validated: `checkquotes.py` 60/60 clean, `just validate human ACTL8` ✓ Valid with only the intentional core-functions warning, `cache/go/terms.csv` still identical to `origin/main` (0 deletions, 0 additions), and `origin/main` is already merged with nothing new to pull.

Flagging the same merge-safety point as last time: this commit postdates your 18:44:41Z approval, so that approval no longer covers the head.

PR #2282, opened 2026-07-26
Review human ACTR1A (centractin/Arp1): restore the lost structural and nucleotide MFs, and an inverted PAINT call vs ACTR10

cmungall:
# ACTR1A (alpha-centractin / Arp1) — full GO annotation review

All 43 GOA rows reviewed one-to-one, plus 4 proposed new annotations, 2 core
functions, 1 proposed new GO term, 7 questions and 5 experiments.
`just validate human ACTR1A` → **✓ Valid (1 warning, deliberately left standing —
see below)**. 106 `supporting_text` quotes verified verbatim.

**Actions:** ACCEPT 5 · KEEP_AS_NON_CORE 33 · MODIFY 2 · MARK_AS_OVER_ANNOTATED 3 · NEW 4.

## Headline finding: ACTR1A has lost both of its defensible molecular functions, and one of them went to the wrong paralog

Current GOA gives human ACTR1A exactly **two** MF rows: a fission-yeast-derived
`GO:0106006 cytoskeletal protein-membrane anchor activity` IBA, and a bare
`GO:0005515 protein binding` IPI. The UniProt entry fetched the same day still
cross-references two MF terms that QuickGO returns **0 hits** for on P61163:

```
DR   GO; GO:0005524; F:ATP binding; IEA:UniProtKB-KW.
DR   GO; GO:0005200; F:structural constituent of cytoskeleton; IBA:GO_Central.
DR   PAN-GO; P61163; 2 GO annotations based on evolutionary models.
```

Two different causes:

1. **`GO:0005524 ATP binding`** reached ACTR1A only via the UniProt-keyword
   pipeline `GO_REF:0000043`, which GOA has retired — a QuickGO query for
   `reference=GO_REF:0000043&taxonId=9606` now returns **0 hits for the entire
   human proteome**. The collateral damage is family-wide: **human ACTB has also
   lost ATP binding**, while keeping `GO:0016887 ATP hydrolysis activity` (IDA).
   So the actin family currently has no nucleotide-binding MF in GOA.
2. **`GO:0005200 structural constituent of cytoskeleton`** — the exact analogue of
   "structural constituent of ribosome", carried by ACTB itself and by budding-yeast
   ARP1 **by IDA** (PMID:9658168) — is absent from ACTR1A but **present on
   ACTR10/Arp11 by IBA with the `enables` qualifier**. Within one family, PAINT has
   given the structural-constituent function to the single-copy subunit whose job is
   to *cap and terminate* the Arp1 filament, and withheld it from the subunit present
   in **eight copies** that actually *forms* it. That inversion is the most useful
   thing this review found and is filed as a question for PAINT.

## The fold-name trap, settled with coordinates rather than argument

The campaign's default scepticism about actin-like proteins does **not** apply to
Arp1 — it applies to ACTR10. Rather than assert this, `ACTR1A-bioinformatics/`
computes it (`analyze_arp1_actin_fold.py`; every position derived from coordinates
or a run-time alignment, nothing hard-coded, `cache/` gitignored, and a fresh run
reproduces the committed `RESULTS.md` byte-for-byte):

| | global % id to ACTB | actin nucleotide-site residues retained |
|---|---|---|
| ACTR1A | 53.1 | **19/25 (76%)** |
| ACTR1B | 54.4 | 20/25 (80%) |
| ACTR10 | 28.0 | **10/25 (40%)** — loses actin's D157/G158 phosphate pair |

And in the **human** dynactin structure (PDB 9B85): **ADP is modelled in 8/8
ACTR1A protomers** (β-actin chain H has AMPPNP), with **20 of 22** nucleotide-
contacting positions coinciding with β-actin's own ATP contacts. The largest
ACTR1A–ACTR1A interface uses ACTR1A residues **42–52 and 65–68**, mapping onto
β-actin 38–48 / 61–64 — actin's subdomain-2 DNase-I-binding loop, i.e. Arp1
polymerises through actin's *own* longitudinal surface. ACTR10 appears once and
touches only the two terminal protofilament chains; CAPZA1/CAPZB cap the other end.

**But the transfer is bounded.** I propose `GO:0051258 protein polymerization` and
explicitly argue **against** `GO:0030041 actin filament polymerization`: the polymer
is an Arp1 mini-filament (8 Arp1 + 1 β-actin) and the monomers added are Arp1, not
actin. No actin-filament *dynamics*, treadmilling, severing or depolymerisation term
should be applied — the assembled filament is doubly capped and of exactly defined
length, and even free Arp1 polymers "never attained the length of conventional actin
filaments" (PMID:10074429).

## New annotations proposed

| term | qualifier | evidence | why |
|---|---|---|---|
| `GO:0005200` structural constituent of cytoskeleton | enables | IDA (PMID:40186871) | restores the family-standard MF, now on *human* structural evidence |
| `GO:0043531` ADP binding | enables | IDA (PMID:40186871) | the ligand actually observed, in the conserved actin cleft, in all 8 protomers |
| `GO:0045504` dynein heavy chain binding | contributes_to | ISS (PMID:25814576, `with` F2Z5G5) | dynein tails dock on **inter-protomer clefts**, hence contributes_to not enables |
| `GO:0051258` protein polymerization | involved_in | IDA (PMID:40186871) | generic term chosen deliberately over the actin-specific one |

I propose `GO:0043531 ADP binding` rather than restoring `GO:0005524 ATP binding`
because Arp1 ATP binding is still only a prediction — PMID:10074429 says so in as
many words ("on the basis of conserved sequence features, is predicted to bind ATP")
— and no structure has resolved an ATP-bound Arp1. Note the ADP is in the deposited
coordinates of 9B85/9B7J but is not discussed in the paper's text.

## MARK_AS_OVER_ANNOTATED, with the REMOVE/over-annotation line drawn explicitly

**`GO:0106006` (IBA, contributes_to).** Three arguments against transfer:
(i) every ACTR1A contact resolved in 9B85 is intra-complex (Arp1 neighbours,
β-actin, ACTR10, CAPZA1/CAPZB, p50 shoulder) plus the dynein tail — no membrane or
membrane protein; (ii) in human mitosis the cortical link is Gαi–LGN–NuMA and the
dependency runs the *other way* — "we found that LGN was required for the cortical
localization of dynein-dynactin" (PMID:22327364), so the anchor sits upstream of
dynactin; (iii) the sole interpretable donor is one **EXP** annotation on *S. pombe*
arp1 where Num1/Mcp5 is the anchor and dynactin cooperates with it. That source is
**sound**, which is exactly why `root_cause: PROPAGATION_BAD` (not
`SOURCE_WEAK_OR_INFERRED`) and `failure_modes: [ROLE_CONFLATION,
COMPARTMENT_OR_COMPLEX_MISMATCH]`. I stop short of REMOVE because `contributes_to`
only asserts the *complex* has the activity, and in fission yeast dynactin genuinely
is required for cortical dynein anchoring: the mammalian case is **untested, not
refuted**.

**`GO:0070062` extracellular exosome (2× HDA)** — cytosolic subunit of a 1 MDa
complex with no free pool and no secretory features; the peptides are real, the
location is not.

## A demonstrably wrong source annotation that is actively spreading

Resolving the `GO:0030473` WITH/FROM set turned up a genuine error at source.
`UniProtKB:F2Z5G5` (pig ACTR1A) carries **`GO:0030989 dynein-driven meiotic
oscillatory nuclear movement`** with evidence **IDA**, assigned by **ComplexPortal**,
citing **PMID:36071160** — Chaaban & Carter's *cryo-EM structure* of dynein–dynactin
on microtubules. A structure cannot be a direct assay of a meiotic movement process,
and the term is defined around "an astral microtubule array emanating from the
spindle pole body", which pigs do not have. The same row was applied to pig DCTN1,
DCTN2, DCTN4, DCTN5, CAPZA1 and CAPZB, and **RGD has propagated it to rat Dctn3,
Dctn5, Dctn6, Actr10 and Capza1 by ISO**. It matters here because F2Z5G5's *only*
experimental annotation anywhere under the `GO:0030473` branch is that bad row — so
the human IBA's only mammalian support is an artefact, leaving fungal-only donors
(A. nidulans nudK germ-tube nuclear migration; S. pombe meiotic horsetail
oscillation) for a process that is real in mammals but untested for this gene. Hence
`KEEP_AS_NON_CORE`, not REMOVE: `NO_FAILURE_NON_CORE` + `[CONTEXT_OR_TISSUE_MISMATCH,
SOURCE_MISCITATION]`, and PMID:36071160 flagged `correctness: MISCITED` with the
explanation in `review_notes`.

## WITH/FROM resolution — every token, plus each source's own evidence

`source_entities` are built **from the GOA WITH/FROM column by script**, with an
assertion that counts match. All 7 distinct tokens resolved; each source was then
queried for **its own** evidence for the propagated term (QuickGO `goUsage=descendants`):

- `PomBase:SPBC1347.12` → **O94630** `ARP1_SCHPO` (Swiss-Prot): own IMP/EXP/IPI
- `SGD:S000001171` → **P38696** `ARP1_YEAST` (Swiss-Prot): own IDA + IPI, **and its own `GO:0005200` IDA**
- `UniProtKB:F2Z5G5` → pig ACTR1A (Swiss-Prot): own IPI for `GO:0005869`; the `GO:0030989` row is bad (above)
- `UniProtKB:Q5BBX7` → *A. nidulans* ANIA_01953/nudK, **unreviewed TrEMBL, submitted name "Uncharacterized protein"** — its *name* is not cited as evidence, only its real IMP
- `WB:WBGene00013168` → **two** TrEMBL accessions, **Q9NA98 and U4PR70**, both `arp-1`, neither with a recommended name. The `size=1` trap: a single-hit query would have hidden the ambiguity. Q9NA98 carries its own IDA; U4PR70 has none
- `PANTHER:PTN000233666`, `PANTHER:PTN007551901` → tree nodes, **not gene products** — reported as unresolvable-in-principle, not merely unresolved

So for `GO:0005869`, **4 of 5** resolvable donors carry their own experimental
evidence. "The donors only carry the same family-level inference" would have been
factually false, so that framing is not used anywhere.

## MODIFY ×2 (both from the cited paper's own content)

- `GO:0005875 microtubule associated complex` (TAS) → **`GO:0005869 dynactin complex`**.
  QuickGO's ancestor closure for GO:0005869 contains GO:0005875, and *both* rows cite
  PMID:7696711 — a term and its parent from one evidence line.
- `GO:0016192 vesicle-mediated transport` (TAS) → **`GO:0047496 vesicle transport along
  microtubule`**. PMID:1528266 draws precisely this contrast ("unlike conventional
  actins which work in conjunction with myosin motors, actin-RPV may be involved in
  cytoplasmic movements via a microtubule-based system"). Directionality is *not*
  upgraded, because the paper does not measure it for ACTR1A.

## Proposed new GO term: `dynactin complex assembly`

GO can express dynactin as a structure (`GO:0005869`) and as a binding target
(`GO:0034452`) but has **no assembly process** — a QuickGO ontology search for
"dynactin" returns only those two plus two obsolete terms. Yet assembly is the only
process in which Arp1's polymerisation activity is deployed, and it is mechanistically
distinctive: the shoulder and its extended p50 regions "recruit eight Arp1s and
stabilize their polymerization", CapZαβ caps the barbed end and Arp11 the pointed
end, while free Arp1 polymers by contrast vary in length. GO already has the exact
sibling for the partner complex, `GO:0070286 axonemal dynein complex assembly`.
Proposed parent `GO:0065003`.

## Paralog distinctions stated rather than assumed

**vs ACTR1B** (reviewed in parallel): 91% identical, both genuine dynactin subunits,
but incorporated at **~15:1 in favour of ACTR1A** despite *equal* mRNA distribution
(PMID:7696711) — so the difference is incorporation, not transcription — and neither
has a free pool. ACTR1B retains 20/25 nucleotide-site residues, so fold integrity is
*not* the discriminator. All eight filament positions in 9B85 are assigned to ACTR1A
(the depositors' assignment, not an independent 91%-identity discrimination — stated
as such). Nothing here says ACTR1B is functionally *different*; that is filed as a
question.

**vs ACTR10** (reviewed in parallel): 1 copy, 10/25 nucleotide-site residues, no
nucleotide modelled, contacts only the two terminal chains, and its function is to
*stop* elongation — "Only Arp11 directly caps the pointed end". So nucleotide-binding
and polymerisation annotations belong on ACTR1A, and `GO:0005200` belongs on ACTR1A at
least as much as on ACTR10.

## Deliberately not annotated (with reasons)

- **TLR2 signalling** (PMID:31221720): one unreplicated study, abstract-only, and
  dynactin-dependent receptor trafficking could produce the phenotype indirectly.
- **SETD3 methylation** (PMID:41142317): in vitro only, no in-cell site, and the
  authors hedge ("may be one of them"). A candidate PTM, not a function.
- **Dre1**: binds the *pointed end*, not the Arp1 filament — PMID:40186871 is cited
  for the human structure only.

## Two deliberate choices a reviewer will notice

1. **The one remaining validation warning is left standing on purpose.** "No
   annotations reference available deep research files" is true: the affinage record
   (`gates_passed: True`, 2 real numeric PMIDs, no bioRxiv-DOI-in-PMID-field entries)
   covers only the TLR2 and SETD3 leads, both declined above, and its own mechanism
   profile lists no molecular activity, no localization and no pathway. Citing it to
   silence the warning would mean quoting provider text for a claim I am not making.
   Its *silence* on dynactin architecture is likewise **not** treated as evidence that
   such literature is absent — the structural papers were retrieved from the UniProt
   reference list and PubMed. The judgement is recorded in the reference's
   `reference_review` instead.
2. **`GO:0005829 cytosol` is deliberately absent from `core_functions.locations`.**
   It is correct, but it arrives as **27 identical Reactome reaction-level rows**, all
   reviewed `KEEP_AS_NON_CORE` — several for centriole/ciliogenesis reactions with no
   ACTR1A-specific evidence, where ACTR1A appears only because Reactome lists dynactin
   among the participants. Promoting it would have required marking all 27 as core,
   contradicting the reason given on those rows. The functionally meaningful locations
   carried into `core_functions` are the centrosome and the mitotic cell cortex, both
   `ACCEPT`ed.

## Provenance / reproducibility

- Review YAML **generated from `ACTR1A-goa.tsv`** by a builder script asserting: every
  GOA row has a hand-written review; no review left unused; exactly 43 GOA rows and 27
  cytosol rows; `source_entities` counts match GOA WITH/FROM; every `supporting_text`
  is a verbatim whitespace-normalised substring of its source.
- 25 inline `[PMID:… "quote"]` provenance strings in `ACTR1A-notes.md` verified verbatim.
- `cache/go/terms.csv`: no deletions (`git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → empty), checked **after** the final validate run.
- 16 publications newly cached; 8 with full text.


cmungall:
Thanks — the blocking item was a real defect and I've fixed it, along with all five non-blocking suggestions. Pushed as e558140b.

## 🟡 Blocking: `PMID:36071160` `correctness` → `VERIFIED`

You're right, and the `review_notes` contradicting the enum in the same block is exactly the tell. `correctness` describes *this* review's use of the reference; this review cites PMID:36071160 affirmatively (a `findings` entry, and `supporting_text` for the `GO:0045504` NEW row), so `MISCITED` was both wrong and machine-misleading. Now `VERIFIED`, with the ComplexPortal problem kept in `review_notes` and an explicit pointer that the defect is recorded where it belongs — `source_status: SOURCE_BAD` + `failure_modes: [SOURCE_MISCITATION]` on the `GO:0030473` `propagation_review`, and escalated in `suggested_questions`. Verified there is now no `MISCITED` anywhere in the file.

## 🔵 1 — `GO:0051258`: why a structure is an IDA for a process

Kept PMID:40186871 as the primary reference and said why, rather than leaving it implicit: what the structure resolves is the **product** of the process in human material — eight protomers in the polymerised state, joined through the canonical longitudinal interface, at defined stoichiometry. PMID:10074429 is the kinetic demonstration but was performed on purified native Arp1 **whose species the abstract does not state**, so promoting it to primary would have silently converted an unverified-species assay into a human IDA. The reason now also names the honest alternative you suggested: ISS on PMID:10074429 with a `with` entity, *once the source species is confirmed from its full text* (which I cannot see). I'd rather leave that as a stated option than guess the species.

## 🔵 2 — `GO:0043531`: name the deposition

Added `supporting_entities: [PDB:9B85, PDB:9B7J]` (same pattern as `GO:0045504`), and the reason now says outright: "The evidence is the deposition, not the prose … named in supporting_entities so a downstream curator does not search the text of PMID:40186871 for it and fail to find it."

## 🔵 3 — the UniProt/GOA divergence runs both ways

This was the sharpest catch. I computed the set difference rather than restating the claim:

```
in UniProt DR GO only:  GO:0005200  GO:0005524
in GOA only:            GO:0005515  GO:0005815  GO:0005856  GO:0030473  GO:0106006
```

So the block is a **divergent snapshot**, not a record of one-way loss, and "UniProt still lists it" alone is consistent with staleness. The `GO:0005200` reason, the UniProt file's `reference_review`, and the notes now all say this explicitly and relocate the weight onto the QuickGO queries (0 human hits for `GO_REF:0000043`; 0 hits for `GO:0000166` with `goUsage=descendants` on ACTB/ACTR1A/ACTR1B/ACTR10). The finding survives the correction — it just now rests on the evidence that actually carries it.

## 🔵 4 — `RESULTS.md` now names the bovine reference

Rather than hard-code "bovine", the script reads `_entity_src_gen.pdbx_gene_src_scientific_name` from the mmCIF and computes the entity's identity to human ACTB at run time. The line now reads:

> Reference: PDB **2BTF** chain `A` (BETA-ACTIN, source organism *Bos taurus*, 99.7% identical to human ACTB over the modelled region) … The reference is not human, so contacts are transferred through human ACTB rather than read off directly.

## 🔵 5 — the interface claim no longer rests on one chain pair

You were right that "the eight protomers are joined through 42-52 and 65-68" was a generalisation from E–G alone, with the subdomain-3/4 positions dropped. Instead of trimming the prose I extended the computation: the script now splits the ACTR1A–ACTR1A interfaces at the largest gap in their contact counts, which separates **six** large intra-protofilament interfaces (≥230 atom contacts: A–C, B–D, C–E, D–F, E–G, G–I) from the smaller lateral ones, and reports the consensus present in **all six**:

> ACTR1A **44–49, 51, 52, 65–68, 205, 213, 243–246** — in β-actin numbering, subdomain-2 members 40–45, 47, 48, 61–64 (the DNase-I-binding loop, which carries the polymerisation argument) and subdomain-3/4 members 200, 208, 242–245 (partner-face contacts, reported for completeness).

Full per-pair residue lists are tabulated in `RESULTS.md` with nothing omitted, and the YAML reason now says the consensus "is a property of the filament, not of one hand-picked chain pair". So the claim is stronger than before, not just better hedged.

## Gates after the changes

- `just validate human ACTR1A` → **✓ Valid (1 warning)** — the deliberate deep-research warning, unchanged and explained in the PR body.
- Quote checker: **109 quotes, 0 problems** (3 new ones added); notes: 26 inline provenance strings, 0 problems.
- Fresh `uv run python analyze_arp1_actin_fold.py` reproduces the committed `RESULTS.md`/`results.json` exactly.
- `git diff origin/main HEAD -- cache/go/terms.csv | grep '^-GO:'` → empty, checked **after** the final validate.
- Merged current `origin/main` (no conflicts) before pushing.

One process note worth recording: I applied these as multi-site edits through the builder script with a per-anchor assertion, and the check-against-the-tree pass caught that one edit (the ADP "evidence is the deposition" wording) had been silently lost when an earlier patch script aborted before writing. It is re-applied and verified present in the tree, not just in my intent.

Agreed that the ACTR10 PAINT-inversion and the ComplexPortal `GO:0030989`/rat-ISO chain both deserve escalation to GO_Central independently of this PR.

PR #2284, opened 2026-07-26
Review human ACTR1B: beta-centractin is documented in dynactin, but its MF term describes the wrong activity

cmungall:
Full GO annotation review of human **ACTR1B** (beta-centractin / Arp1B), the second dynactin Arp1 paralog. All **21 GOA rows reviewed one-to-one in GOA order** (the seeded stub collapses the two `PMID:12857853` rows into one; they are two different experiments and are split back out here), plus **2 proposed NEW** annotations.

**ACTR1A was reviewed independently and in parallel.** That is what makes the arguments below possible, and several of them are only available from the pair.

### Documented, not assumed — and specifically *not* by paralogy

The founding paper raised beta-specific antibodies *because* the existing reagents did not work on beta:

> "As antibodies previously raised against alpha-centractin reacted only poorly with beta-centractin, new antibodies were produced and combined with two-dimensional gel electrophoresis to discriminate the two isoforms." (PMID:7696711)

and then found both isoforms confined to the complex, at a fixed ratio:

> "Both isoforms were found predominantly in the cytosolic fraction as a part of a previously identified 20S complex ... with no evidence for a free pool of either isoform." / "The isoforms were found in a constant ratio of approximately 15:1 (alpha:beta) in the dynactin complex." (PMID:7696711)

IntAct (146 binary records for P42025) adds the full roster — DCTN1-6, CAPZA1/2, CAPZB, ACTR10 — **and ACTR1A itself**, so the two paralogs occupy the *same* filament rather than forming separate alpha-only and beta-only dynactins. Over 8 protomers at 15:1 that is roughly one beta per second particle: a genuinely sub-stoichiometric subunit. `GO:0005869` **ACCEPT** on both the IBA and the TAS row.

**Curation recommendation:** the TAS code understates its own evidence. `PMID:7696711` is not a review — it is the primary paper that made the isoform-specific reagents and ran the 2D gels. It should be **IDA**, on ACTR1A's row from the same paper too.

### The molecular-function row is the real defect — and it is a *node* problem, not a paralog problem

`contributes_to GO:0106006 cytoskeletal protein-membrane anchor activity`, IBA, dated 2026-04-16. Three findings:

1. **The node is narrow and single-sourced.** Asking QuickGO which human proteins hold `GO:0106006` by IBA returns 5: HIP1 and HIP1R (from `PTN000045135`), DCTN2 (from `PTN000394123`), and ACTR1A + ACTR1B from **`PTN007551901`** — a different node from the one carrying the other two centractin IBAs (`PTN000233666`). The entire experimental basis is one *S. pombe* donor.
2. **The donor shows a process requirement, not a bridging activity**, and puts the membrane binding elsewhere: "Cortical factor Num1 ... bound to dynein independently of dynactin", while "Arp1, Mug5 and Jnm1 were involved in the regulation of shrinkage and bundling of microtubules" (PMID:25736293) — in fission-yeast meiotic prophase.
3. **The term's own differentia fails.** It requires bridging to "a plasma membrane lipid or membrane-associated protein, in order to maintain the localization of the cytoskeleton at a specific **cortical membrane** location". Arp1 *does* have a membrane-skeleton partner — "Arp1 binds directly to the Golgi-associated betaIII spectrin isoform" (PMID:11461920) — but that recruits dynactin "to **intracellular** membranes", not to a cortical location.

**MODIFY -> `GO:0140378` protein complex scaffold activity**, whose definition ("a structural molecule activity of a protein-containing complex component that serves to hold the complex together") is verbatim what UniProt records for the fungal Arp1 donors. Two notes in the review: the qualifier should become `enables` (scaffold activity is the subunit's own activity), and `GO:0005200` is an acceptable alternative since *S. cerevisiae* ARP1 holds it by its own IDA. Because the term arrives at both paralogs from one node, **fixing `PTN007551901` fixes ACTR1A and ACTR1B in one edit** — and would restore a structural MF that both currently lack entirely. Note UniProt's DR block for P42025 still lists `GO:0005200 ... IBA:GO_Central`, absent from current GOA for both paralogs.

Deliberately **not** proposed: `spectrin binding`. That work is alpha-centractin-directed, so extending it to beta would be the very paralog transfer this review is checking for.

### Four REMOVEs decided by the paralog pair

The Reactome neutrophil-degranulation rows — `GO:0005576` x2, `GO:0034774` secretory granule lumen, `GO:1904813` ficolin-1-rich granule lumen — are on **ACTR1B but not ACTR1A**. ACTR1B has no signal peptide and no TM (the P42025 feature table is CHAIN/MOD_RES/VARIANT only) and no free pool outside dynactin, so it cannot be topologically inside a granule lumen. The decisive step is the pair: ACTR1A is in the *same complex at fifteen times the abundance*, so if beta were granule-lumen cargo alpha necessarily would be too. The asymmetry can only be granule-fraction detection stochasticity. Same argument retires the `GO:0016020` membrane HDA (also beta-only), where the paper concedes the point itself: "The remaining species were largely involved in cellular processes and molecular functions that could be predicted to be transiently associated with membranes." (PMID:19946888)

### `protein binding`, resolved per partner

| Reference | Partner | IntAct method | Call |
|---|---|---|---|
| PMID:12857853 | DCTN1 | **density sedimentation** | KEEP_AS_NON_CORE — co-complex, not contact |
| PMID:12857853 | mouse Arp11 | **pull down** | KEEP_AS_NON_CORE — real direct pointed-end contact |
| PMID:26638075 | DCTN1 | BioID | KEEP_AS_NON_CORE |
| PMID:33961781 | DCTN1 | AP-MS | KEEP_AS_NON_CORE |
| PMID:32814053 | HTT | Y2H x3 | **MARK_AS_OVER_ANNOTATED** |

The DCTN1 rows carry a caution: p150 does not touch the filament directly — "The unique architecture of the shoulder securely houses the p150 subunit and positions the four identical p50 subunits in different conformations to bind dynactin's filament" (PMID:33734450) — and DCTN1 is not even modelled in the human dynactin structure. And on HTT: UniProt reports `NbExp=3`, which reads like triplicate support, but **all three IntAct records come from PMID:32814053 alone** as three Y2H formats in one screen.

## Residue analysis: the fold argument, done on the right structure

`genes/human/ACTR1B/ACTR1B-bioinformatics/` derives the contact residues **from human dynactin itself** (PDB **9B85**, the 2025 Dre1-purified cryo-EM structure) rather than reciting them for conventional actin. Reruns reproduce the committed report byte-for-byte; the PDB-to-UniProt offset is asserted at runtime.

The structure is itself a finding: all **8** Arp1 protomers are modelled as **ACTR1A**, each with bound **ADP**, and **PDBe maps 0 entries to P42025** — ACTR1B is in no deposited structure. At 3-5 A local filament resolution two 90%-identical paralogs could not have been distinguished anyway, so that absence is not evidence against incorporation.

| | nucleotide site (18) | Arp1-Arp1 interface (37) |
|---|---|---|
| **ACTR1B** | **17/18** | **32/37** |
| pig ACTR1A (donor) | 17/18 | 37/37 |
| pombe arp1 (donor) | 13/18 | 20/37 |
| beta-actin | 11/18 | 16/37 |
| ACTR10/Arp11 | 6/18 | 10/37 |

All 6 divergent positions are conservative, none is a gap, and all 6 are **fixed in mouse Actr1b**. The nucleotide-site change is the most informative: **K215 -> R**, and pombe, yeast, worm, *A. nidulans* **and beta-actin all carry Arg there** — ACTR1B holds the consensus residue and ACTR1A is the outlier. Beta's ATP site is intact and arguably more canonical than alpha's.

Consistent with the brief's caution about the fold: Arp1 does polymerise ("Arp1 has a vanishingly low critical concentration for polymerization", PMID:12857853) but its filament is a fixed-length ruler set by other subunits, so *dynamics* terms would not transfer. None was present, so none needed removing.

One asymmetry recorded as a hypothesis, not a conclusion: the interface is **37/37 invariant** between human and pig alpha at 90.3% overall identity, yet ACTR1B differs at **5/37** at 90.4% — strict purifying selection along the alpha lineage, background-rate divergence between the paralogs. Consistent with relaxed constraint on a protomer that never tiles the whole filament. In `suggested_experiments`.

## Proposed NEW

- **`GO:0005524` ATP binding** (ISS) — lost to the `GO_REF:0000043` keyword-pipeline retirement, not refuted. UniProt still lists it; every Arp1 protomer in 9B85 has ADP bound; 17/18 site residues retained.
- **`GO:0047496` vesicle transport along microtubule** (IC, with `GO:0005869`) — ACTR1B's only BP in GOA is the non-core `GO:0030473`, so nothing records the transport function dynactin exists to perform. Inferred from *experimentally established* complex membership, not from the paralog; UniProt's own FUNCTION line for P42025 says "Component of a multi-subunit complex involved in microtubule based vesicle motility".

## Other findings reported upstream

- **A donor annotation looks definitionally impossible.** Pig ACTR1A (F2Z5G5), a WITH/FROM donor on both human centractins' `GO:0030473` IBA, carries `GO:0030989` **dynein-driven meiotic oscillatory nuclear movement** by IDA from ComplexPortal, sourced to PMID:36071160 — a reconstituted dynein-dynactin-BICDR1 motility/cryo-EM paper. `GO:0030989` is defined by oscillation led from a **spindle pole body** during meiosis I, which pig cells do not have, and the paper assays neither meiosis nor nuclei. In `suggested_questions`.
- **`GO:0140660` cytoskeletal motor activator activity has zero human annotations**, though dynactin is the textbook activator of cytoplasmic dynein. Used as `contributes_to_molecular_function` in `core_functions` and flagged as a complex-wide gap.
- **The paralogs are *not* distinguishable by tissue**, contrary to a common shorthand: "alpha-centractin and beta-centractin mRNAs are equally distributed in all populations of mRNA examined, whereas the expression of gamma-centractin appears to be tissue specific" (PMID:7696711). They *are* distinguishable by abundance (15:1) and by CRISPR-screen hit rate (ACTR1B 197/1155 screens vs ACTR1A 557/1172). No disease association: ACTR1B was screened with ACTR1A and DCTN1-6 in 136 IPN patients with no significant variants (PMID:26662454).

## Verification

- `checkquotes.py`: **36 quotes, 0 problems**; `file:` quotes separately confirmed to lie on one physical line (11/11).
- `just validate human ACTR1B`: **✓ Valid**, zero warnings; `just validate-terms`: **✅ passed**.
- Merged `origin/main` before final validation; no conflicts.
- `source_entities` built **from GOA column 11 by construction** and re-verified afterwards against the WITH/FROM field of the matching row: 8 propagation reviews, 0 mismatches.
- `cache/go/terms.csv`: restored from `origin/main` and **appended** only `GO:0140378` and `GO:0140660` (never re-sorted). `git diff origin/main -- cache/go/terms.csv | grep '^-GO:'` is **empty**.

**Actions:** 8 KEEP_AS_NON_CORE · 5 ACCEPT · 4 REMOVE · 3 MARK_AS_OVER_ANNOTATED · 2 NEW · 1 MODIFY.

One expected non-warning: the `core_functions.molecular_function` term `GO:0140378` reaches `existing_annotations` as a `proposed_replacement_terms` value on the `GO:0106006` MODIFY row rather than as a separate `NEW` row, since adding both would double-count the same recommendation.

---

**Follow-up commits after the initial push** (all pushed before any review):
- Provenance attached to the CRISPR-screen contrast in `knowledge_gaps` (the ACTR1B BioGRID-ORCS line from the UniProt record).
- A note in the `GO:0047496` NEW row explaining the term choice: dynactin's cargo is heterogeneous, so `GO:0030705` cytoskeleton-dependent intracellular transport would be the true LCA; the specific child was chosen because it is what the cited UniProt FUNCTION line actually states. Happy to generalise if curators prefer the LCA framing.
- The bioinformatics analysis was re-verified as reproducible **from a cold cache** (`cache/` deleted, everything refetched): both `RESULTS.md` and `results.json` reproduce byte-for-byte.


cmungall:
Thanks — this was a genuinely useful review, and one of the suggestions turned up evidence I had missed rather than just a presentation issue. All five items addressed in `c67fcd5d7`.

## 🟡 #1 (blocking) — `description` contradiction: fixed

You are right, and the error was mine rather than a wording slip: I compressed "equally *distributed* in all populations of mRNA examined" into "expressed at comparable levels", which is a different claim and directly contradicts the 15:1 ratio one sentence earlier. Now reads:

> …at a constant ratio of roughly fifteen alpha to one beta. **Both paralogs are ubiquitously expressed without tissue restriction — it is a third, gamma-centractin mRNA species that is tissue-specific — but beta is the minor of the two by protein abundance.**

## 🔵 #2 — PMID:40186871 *does* assay Arp1b, and I missed it

This is the most valuable item in the review. You are right, and the reason for the miss is instructive: **the paper spells it `Arp1b`, lower-case b, and my grep pattern was `Arp1[ab]` for the first sweep but `Arp1B` for the context extraction** — so the second pass returned nothing and I concluded the paper was ACTR1A-only. Seven occurrences, and two are load-bearing:

> "These Dre1 variants were tested in co-transfection assays for binding to multiple dynactin subunits (p150glued, p50-dynamitin [DCTN2], Arp1b) as well as to dynein"

> "Eluates were immunoblotted with anti-p150glued, anti-p50-dynamitin, anti-dynein 74 kDa, **anti-Arp1b**, anti-FHL2, anti-Strep, and anti-GAPDH antibodies."

That is a **2025, human, paralog-specific recovery of beta-centractin from affinity-purified dynactin**, independent of AP-MS — and it closes the loop with exactly the trick the 1994 paper used, a reagent that cannot be reporting alpha. Both quotes are now on the `GO:0005869` TAS row's `supported_by`, and the row's summary says so.

There is a small irony worth recording, and I've put it in the notes: these authors used **anti-Arp1b** as their representative Arp1-filament readout while depositing a structure that models all eight protomers as alpha. That is the cleanest illustration in this whole review of why "the structure models ACTR1A" is not evidence about paralog occupancy.

The `reference_review` for PMID:40186871 now lists three uses and says explicitly that the third was added after review and why it was missed, rather than quietly absorbing the correction.

## 🔵 #3 — the ADP point: adopted, and it changes the term

You flagged that the ISS rests on an ADP-bound structure while the term is ATP binding. I checked this independently rather than just softening the wording, and the honest conclusion is stronger than a caveat — **the term should be `GO:0043531 ADP binding`, not `GO:0005524 ATP binding`.** The paper that purified native Arp1 and characterised its polymerisation says outright:

> "is predicted to bind ATP and possibly polymerize" (PMID:10074429)

So ATP binding has never been *shown* for Arp1; it is a fold-based prediction, which is exactly what the retired keyword pipeline was propagating. ADP, by contrast, is modelled in all eight Arp1 protomers of 9B85 (entity 10 = ADENOSINE-5'-DIPHOSPHATE; the single conventional-actin protomer, chain H, has AMP-PNP instead). The row is now `GO:0043531`, and:

- `supporting_entities` is `UniProtKB:P61163` alone — your point about `PDB:9B85` in an ISS with/from is right, and the PDB entry is cited as a reference instead;
- `GO:0000166` nucleotide binding is named in the `reason` as the fallback if a curator objects even to ADP;
- PMID:10074429 is added as a reference with a `reference_review`. It also bounds the polymerisation claim in the other direction — Arp1 filaments annealed but never reached conventional actin-filament length — which is why no filament-dynamics term is proposed.

## 🔵 #4 — `GO:0140660` aspirational status: now stated

Added to `core_functions.description`: the term is the semantically correct complex-level activity for a dynactin subunit but has **zero** human annotations, so it is recorded there and raised as a complex-wide gap in `suggested_questions` rather than proposed as a `NEW` row for this one gene, and the asymmetry with the structural MF is intentional.

## 🔵 #5 — QuickGO queries: now reproducible from the repo

New section `## 2b` in `ACTR1B-notes.md` records the three load-bearing queries verbatim with the run date (2026-07-26): the two whole-gene annotation pulls behind the paralog comparison (totals 21 and 43, with the term-set difference spelled out both ways), the `GO:0106006` + `evidenceCode=ECO:0000318` query behind "five human proteins, `PTN007551901` reaches only the two centractins", and the per-donor `goUsage=descendants` pattern plus the `GO:0140660` → zero-hits query.

---

## One change that goes beyond your review, flagged rather than slipped in

You endorsed `GO:0140378` as the `GO:0106006` replacement. I have made **`GO:0005200` structural constituent of cytoskeleton the first-listed replacement, with `GO:0140378` second.** Both remain proposed; only the ordering changed. Two reasons, and I'd rather you push back than not see it:

1. **The term sits on a shared node.** `PTN007551901` delivers `GO:0106006` to ACTR1A and ACTR1B and to nothing else. One node cannot carry a different structural MF per gene, so a recommendation that names `GO:0140378` for ACTR1B while the independently-reviewed ACTR1A (#2282) names `GO:0005200` is not actionable — PAINT would receive two competing edits for one node. The convergence itself is corroboration, since neither review saw the other before reaching it.
2. **Family precedent runs entirely to `GO:0005200`.** Conventional beta-actin carries it; *S. cerevisiae* ARP1 holds it by its own **IDA** (PMID:9658168); **ACTR10/Arp11 — the capping subunit of this very filament — already holds it by IBA with the `enables` qualifier**; and it is what this PANTHER family previously propagated to these two genes, as UniProt's stale DR block still shows. `GO:0140378` has the tighter definition and no precedent here.

`GO:0005200` is also not a category error for a complex-internal filament: GO's own ancestor list places `GO:0005869` under `GO:0015629` actin cytoskeleton.

## Verification after the changes

- `checkquotes.py`: **39 quotes, 0 problems** (up from 36 — the two new Arp1b quotes and the PMID:10074429 quote). Notes file separately checked: **25 quotes, 0 problems**.
- `just validate human ACTR1B` → **✓ Valid**, zero warnings. `just validate-changed …` (CI's exact recipe) → **✓ All changed files passed**.
- `git merge origin/main` → already up to date, no conflicts.
- `git diff origin/main -- cache/go/terms.csv cache/enums/ | grep '^-GO:'` → **empty**; additions still only `GO:0140378` and `GO:0140660`. `GO:0005200` and `GO:0043531` were already cached, so no new rows were needed for the term switch.

On your note that you couldn't run `just validate` in the sandbox — understood, and thanks for saying so rather than ticking the box. I've run both the plain and the `validate-changed` forms after every edit.

cmungall:
Thanks for the approval — and I've taken the three remaining items rather than leaving them, because #1 was a claim-strength problem I introduced in round 2 and I'd rather the record be right than merely approved. Pushed as `5de5017c3`. **This deliberately resets the approval gate**: the APPROVED review predates this commit, so it does not cover the current head and auto-merge should wait for a re-review rather than take the previously-approved tree.

**#1 — you were right, and the fix strengthens rather than weakens the claim.** Both quotes I added in round 2 are Methods sentences ("were tested in co-transfection assays for binding to…", "Eluates were immunoblotted with…"), so "beta-centractin is recovered" was an inference from them. But the paper's *results* text closes the gap, and I had not read far enough:

> "A single substitution in CR1, in which phenylalanine 195 was changed to proline or to alanine (F195P or F195A) and which is predicted to disrupt the strand fold, **decreased binding of endogenous dynactin subunits** and dynein to the transfected Dre1 variants"

plus "a conservative replacement of phenylalanine with tyrosine (F195Y) did not affect Dre1 binding to dynactin". Binding can only be *decreased* from a baseline, so the endogenous subunits assayed do bind wild-type Dre1 — and Arp1b is one of the three, detected with a paralog-specific antibody. The summary now states the chain in that order and says explicitly that the per-subunit bands are in Figure 2D, which the cached text does not render, so the claim sits at exactly that strength and no further. The results quote is added to `supported_by`.

**#2 — ADP specificity is now caveated as you suggested.** Added: 9B85 is a single post-hydrolysis state of a purified complex, so it evidences that the site is *occupied by a nucleotide* far more strongly than it evidences ADP-over-ATP specificity; an actin-family protein would be expected to bind both. The honest reading is "ADP is what has been seen, ATP is what has been predicted", which is why `GO:0000166` remains named as the fallback.

**#3 — the apparent mismatch is intentional and now says so.** The ATP-binding UniProt DR line is cited as evidence that *a nucleotide-binding MF was present and has been lost from GOA*, not as evidence of which nucleotide is bound. That is now stated in the `reason` so a curator doesn't read it as a copy-paste slip.

Thanks also for engaging with the two unrequested changes on the merits rather than as scope creep — the `PTN007551901`-cannot-carry-two-terms argument was the one I was least sure would land.

## Verification after round 3

- `checkquotes.py`: **40 quotes, 0 problems**. Notes file: **26 quotes, 0 problems**.
- `just validate human ACTR1B` → **✓ Valid**, zero warnings.
- Merged `origin/main` (which had moved) before validating; no conflicts.
- `git diff origin/main -- cache/go/terms.csv cache/enums/ | grep '^-GO:'` → **empty**; additions unchanged at `GO:0140378` and `GO:0140660`.
- Action tally unchanged: 8 KEEP_AS_NON_CORE · 5 ACCEPT · 4 REMOVE · 3 MARK_AS_OVER_ANNOTATED · 2 NEW · 1 MODIFY.

No recommendation changed in this round. This is everything I intend to push.

cmungall:
Fixed in `5ad2a3d0d`. You were right that the two passages read inconsistently, and it is worth naming the failure: this is the "fixed in N places, landed in N−1" pattern — I corrected the `GO:0005869` summary in round 3 and did not carry the same correction into the `reference_review` making the same claim.

So rather than patch the flagged line, I swept **the whole gene folder** for any surviving Methods-only framing: five phrase patterns (`paralog-specific recovery`, `is recovered`, `recovery of beta-centractin`, `recovered as part of`, `recovered from affinity-purified`) across every string field of the YAML by walking the parsed document, plus the notes file. **Exactly one instance survived — the one you flagged.** The notes file was already correct. The `reference_review` now reads:

> …immunoblotting Strep affinity-purification eluates for Arp1b. **Those are Methods statements; the result is carried by** the accompanying report that fold-disrupting Dre1 substitutions 'decreased binding of endogenous dynactin subunits' while a conservative F195Y replacement did not, both presupposing a baseline, in the same Figure 2D panel. So the paper does establish a 2025, human, paralog-resolved recovery of endogenous beta-centractin on Dre1, independent of AP-MS — but the per-subunit bands are in Figure 2D, which the cached text does not render, so the claim is made at that strength and no further.

Your point about the AP-MS "all 11 subunits" sentence is well taken and I'm glad you flagged that it is *not* interchangeable — it would have re-imported the very method the argument is trying to be independent of. That is a sharper statement of the reason than I had in my head when I chose the co-transfection result.

**No recommendation, action, term or evidence code changed in this commit.**

## Verification

- `checkquotes.py`: **40 quotes, 0 problems** (unchanged — no quote was added or altered).
- `just validate human ACTR1B` → **✓ Valid**, zero warnings.
- `git merge origin/main` → already up to date.
- `git diff origin/main -- cache/go/terms.csv cache/enums/ | grep '^-GO:'` → **empty**.
- Action tally unchanged: 8 KEEP_AS_NON_CORE · 5 ACCEPT · 4 REMOVE · 3 MARK_AS_OVER_ANNOTATED · 2 NEW · 1 MODIFY.

This is genuinely everything I intend to push.

PR #2286, opened 2026-07-26
Harmonise the AADAC-family PTN009058710 hydrolase row across AADACL2, AADACL3 and AADACL4

cmungall:
## The defect

**AADACL2**, **AADACL3** and **AADACL4** each carry one `GO:0016787 hydrolase activity` IBA row
(`GO_REF:0000033`) transferred from PANTHER node `PTN009058710`, with the **same 17 `WITH/FROM`
tokens** in all three GOA records — measured from the committed TSVs *and* independently from
QuickGO per accession, cross-checked. One row, one node, one donor set, and three reviews reached
three different verdicts:

| gene | PR | verdict as merged |
|---|---|---|
| AADACL2 | #2266 | `MODIFY` → `GO:0017171`, `TERM_SCOPING_PROBLEM` + `GRANULARITY_MISMATCH` |
| AADACL4 | #2263 | `MODIFY` → `GO:0052689`, `TERM_SCOPING_PROBLEM` + `GRANULARITY_MISMATCH` |
| AADACL3 | #2264 | keep `GO:0016787` as the genuine LCA; replace only as **redundant**, `EVIDENCE_CIRCULAR_OR_REDUNDANT` — but `GRANULARITY_MISMATCH` retained on the literal reading |

## Which verdict is correct, and how it was settled

By measurement, in a new shared node-level audit
(`genes/human/AADACL2/AADACL2-bioinformatics/audit_node_PTN009058710.py` →
`NODE_PTN009058710.md`, `node_PTN009058710.json`). All 17 tokens resolve — 16 proteins plus the
tree node itself. For each donor the chemistry is read off **two** sources independent of the
protein's *name* (its own EC numbers / `CATALYTIC ACTIVITY` reactions, and its own curated GO
molecular-function annotations classified by **fetched** ontology ancestry), and the nucleophile
off its own `ACT_SITE` features, with the GDXG elbow pentapeptide reported alongside so a
positionally-inferred call can be checked at the sequence level. Each candidate term is then
tested against every donor:

```
GO:0016787 hydrolase activity:                  TRUE 16, FALSE 0, UNDETERMINED 0
GO:0052689 carboxylic ester hydrolase activity: TRUE 14, FALSE 2, UNDETERMINED 0
GO:0017171 serine hydrolase activity:           TRUE 15, FALSE 1, UNDETERMINED 0
```

**None of the refinements is true of every member, and `GO:0016787` is refuted by nobody.** The
two refutations lie on **different axes**, so neither candidate can be rescued by choosing the
other:

- **bond-type axis** — blocked by the two kynurenine formamidases, mouse Afmid (`MGI:MGI:2448704`
  → `Q8K4H1`) and yeast BNA7 (`SGD:S000002836` → `Q04066`), both EC 3.5.1.9, both carrying
  `GO:0004061 arylformamidase activity` (IMP and IDA respectively). That term sits under
  `GO:0016810` (C–N bonds), a **sibling** of the ester branch, not a descendant.
- **mechanism axis** — blocked by soybean HIDH (`UniProtKB:Q5NUF3`), whose nucleophile-elbow
  residue is **Thr164**, against `GO:0017171`'s definition demanding "a catalytic triad consisting
  of a serine nucleophile".

`GO:0016788` and `GO:0016810` are siblings whose only common ancestor below `GO:0003824` is
`GO:0016787`, so PAINT's term is not merely unrefuted — it is the **exact lowest common ancestor**
of its donor set.

**So AADACL3's reading is the correct one.** The row is correct, correctly scoped and validly
transferred; there is **no granularity and no term-scoping defect**. What is true of it is that it
is subsumed by the `InterPro:IPR017157`-derived `GO:0052689` each of the three genes independently
carries — `EVIDENCE_CIRCULAR_OR_REDUNDANT`, "the target already has stronger direct evidence".

### The Thr164 call, and why the verdict does not rest on it

The whole mechanism verdict turns on one residue, so: UniProt annotates HIDH `ACT_SITE 164` as
**"Proton acceptor"** (`ECO:0000305`), *not* as a nucleophile, so the call is a **fold-position**
inference (T164–D263–H295, elbow apex). Two things carry it instead of a label:

- **the sequence.** Fifteen of sixteen donors read G-x-**S**-x-G at the elbow (`GDSAG` ×12,
  `GQSAG` in ICME, `GHSAG` in Afmid, `GHSVG` in BNA7); HIDH alone reads **`GETSG`**.
- **a sensitivity analysis.** Score HIDH `FALSE`, score it `UNDETERMINED`, or discard every
  positionally-inferred nucleophile in the set — under all three readings `GO:0017171` is not true
  of every donor and `GO:0016787` remains the only term that is.

### Two premises both merged reviews relied on, and both are wrong

1. **HIDH is not "a lyase, not a hydrolase at all".** It is bifunctional and *both* halves are
   experimental: `GO:0033987` dehydratase by **IDA** *and* `GO:0106435 carboxylesterase activity`
   by **IDA**, EC 4.2.1.105 *and* EC 3.1.1.1. So it does not threaten `GO:0016787` and does **not**
   block `GO:0052689` either — the two formamidases do. HIDH refutes only the *serine* term.
2. **Yeast BNA7 does resolve.** `xref:sgd-S000002836` → `Q04066`, `ACT_SITE 110` labelled
   "Nucleophile" and reading as **Ser**. AADACL3's audit reached it through an Alliance record and
   reported the nucleophile unresolved; with BNA7 in, the tally is **15 of 16 serine, one
   threonine**, and *both* formamidases are themselves serine hydrolases.

### Where the mechanism term does belong

`GO:0017171` is not wrong about this family; it is attached to the wrong node. At the **family**
node `PTN009058713`, the three donors PAINT cites — human AADAC, mouse Aadac, mouse Nceh1 — are all
`IPR017157` members, all serine, all ester hydrolases, and **all three hold `GO:0017171` by IDA**.
All three blockers at the deep node lie **outside** `IPR017157`. So the recommendation is a **node
move** (`PTN002745055`/`PTN002745068` → `PTN009058713`), not a term change on the row — stated once,
naming all three genes.

## What changed

All three genes now record the identical treatment for this row: `MODIFY` → `GO:0052689`,
`root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT`, **no** granularity failure mode, and `GO:0017171`
pursued as a node move.

**AADACL2** — replacement `GO:0017171` → **`GO:0052689`** on the `GO:0016787` IBA row, the
`IPR013094` IEA row and the `GO:0003824` ARBA row, so all three general MF rows collapse onto the
one specific term the record has; `root_cause` corrected; `core_functions.molecular_function` moved
to `GO:0052689`, matching the other two (`GO:0017171` remains stated as true of AADACL2 on its own
residues, but pursued as the node move rather than asserted as a second activity);
`supporting_entities` added as all 17 GOA tokens; PAINT `knowledge_gaps` and `GO_REF:0000033`
`reference_review` strengthened with the measured result; new node-move `suggested_questions` entry.

**AADACL4** — replacement term was already right, the classification was not: `root_cause` →
`EVIDENCE_CIRCULAR_OR_REDUNDANT` on both hydrolase rows, `GRANULARITY_MISMATCH` removed, HIDH's
`source_status` corrected to `SUPPORTS_TRANSFER` with the bifunctionality spelled out, `Q9HTI0`
added as the unreviewed donor, `supporting_entities` expanded from **5 of 17** to all 17 (built
from the GOA TSV with an assertion), new node-move `suggested_questions` entry.

**AADACL3** — #2264 merged while this PR was in review, so the third gene is harmonised here
rather than only recommended. Its verdict, actions, terms and `core_functions` are **untouched**;
what changes is the encoding it had kept on the literal reading of the enum plus two donor facts:
`GRANULARITY_MISMATCH` removed from both hydrolase rows (`SOURCE_EVIDENCE_WEAK` kept on the
`IPR013094` row — a separate, still-correct judgment about a fold signature reaching outside the
subfamily); BNA7 recorded as resolved, 15-of-16 superseding its local 13-of-14, with its own
generated `RESULTS.md` left as generated; HIDH moved from `SOURCE_WEAK_OR_INFERRED` to
`SUPPORTS_TRANSFER`.

**New shared artefact** in `AADACL2-bioinformatics/`: `audit_node_PTN009058710.py`,
`NODE_PTN009058710.md`, `node_PTN009058710.json` — one row shared by three genes, audited once and
cited from all three reviews (cross-gene `file:` references are established here: ATP5MC1/2/3,
ANOGA D7r1, PSEPK moaA/moaC). A fresh run reproduces the committed JSON byte-for-byte, and the
prose table, counts, sensitivity lines and family-node numbers are cross-checked against it
programmatically.

**Notes**: the adjudication and all four review rounds are recorded in all three `*-notes.md`
files, with superseded items flagged in place rather than rewritten.

## One open repo question

Removing `GRANULARITY_MISMATCH` goes beyond the enum's literal text ("Parent term is true but
uninformative"), which does describe this row. It is removed because `failure_modes` is documented
as the *biological shape of a propagation issue* and this propagation has none — the parent is
uninformative because the donors are heterogeneous, not because the transfer could have been more
specific. That is the rule this campaign already adopted ("is this term the LCA of its donors?").
The reasoning is now inline on all six rows rather than implicit. **@cmungall** — the concrete ask
is a one-line clarification to `GRANULARITY_MISMATCH`'s description in `gene_review.yaml`
distinguishing the two readings; deliberately not done in a gene PR. If it goes the other way the
fix is one enum entry restored on six rows.

## Gates

- `checkquotes.py` across all three reviews: **133 quotes, 0 problems**
- `just validate human {AADACL2,AADACL3,AADACL4}`: **✓ Valid** (AADACL3 and AADACL4 each carry one
  pre-existing deep-research warning, unchanged from `main`)
- `git diff origin/main -- cache/go/terms.csv | grep '^-GO:'`: **no deletions, no additions**
- `origin/main` merged in; every claim above re-grepped from the tree (48 invariant checks, all pass)


cmungall:
Thanks — sharp review, and the two IMPORTANT items were both real. Round 2 pushed as `122fe08`. I conceded five items and held one, with the reasoning now recorded inline rather than only in the PR prose.

### 1. "Byte-identical across all three" — fixed by *measuring*, not by softening

You were right: `genes_sharing_the_row: ["AADACL2","AADACL4"]` while the prose claimed three, and `read_with_from` returning `{"present": False}` did contradict its own docstring. Rather than weaken the sentence, I made the claim measurable, because it *is* measurable — the row exists in GOA whether or not the folder exists in the tree:

- the audit now fetches the audited row's `WITH/FROM` set **per accession from QuickGO** for `Q6P093`, `Q5VUY0` and `Q5VUY2`;
- where a committed TSV also exists, the two sources are **cross-checked** against each other and disagreement is a hard error;
- a gene covered by *neither* source is a hard error, so no gene can drop out of the equality test silently;
- the missing-TSV path now prints the absence and its reason and records `genes_with_no_tsv_in_this_tree`, instead of quietly returning an absence.

AADACL3 (`Q5VUY0`) comes back with the **same 17 tokens**, so `genes_sharing_the_row` is now `["AADACL2","AADACL3","AADACL4"]` with `measured_from_committed_tsv: ["AADACL2","AADACL4"]` and `measured_from_quickgo` all three. Both `review.reason` texts now say which source covered which gene.

### 3. All three family-node donors carry `GO:0017171` by IDA — conceded, and it does strengthen the recommendation

Verified: `P22760` `IBA,IDA,IEA`; `Q99PG0` `IBA,IDA,IEA,ISO`; `Q8BLF1` `IBA,IDA`. The audit now records this per member as `family_node.mechanism_term_support` with `n_members_with_GO_0017171_IDA: 3`, so the number comes out of the JSON rather than out of prose. Corrected in six places across the two reviews and the audit file.

### 4. The Thr164 call — conceded, and this was the most useful item

Right on the facts: UniProt annotates HIDH `ACT_SITE 164` as **"Proton acceptor"** (`ECO:0000305`), not as a nucleophile, and the script reached it through the positional fallback, unlike Afmid and BNA7. Since the whole mechanism verdict turns on that one residue, I did two things rather than just re-labelling it:

**(a) moved the call onto the sequence.** The audit now reports the GDXG **elbow pentapeptide** for every donor. Fifteen of sixteen read G-x-**S**-x-G — `GDSAG` ×12, `GQSAG` in ICME, `GHSAG` in Afmid, `GHSVG` in BNA7 — while HIDH alone reads **`GETSG`**. The elbow serine is substituted; there is a Ser one position downstream (S165), but no curated Ser–Asp–His triad exists for HIDH, so the term cannot be asserted of it on present evidence. That is checkable without trusting the positional rule *or* the UniProt label.

**(b) measured whether the verdict depends on the rule.** New sensitivity block:

```
only_the_refuting_donor_undetermined:                     TRUE 15, FALSE 0, UNDETERMINED  1, true_of_every_donor=False
every_positionally_inferred_nucleophile_undetermined:     TRUE  2, FALSE 0, UNDETERMINED 14, true_of_every_donor=False
```

Score HIDH `FALSE`, score it `UNDETERMINED`, or throw out every positionally-inferred nucleophile in the set — under all three readings `GO:0017171` is not true of every donor, and `GO:0016787` remains the only term that is. So the verdict is robust to exactly the uncertainty you identified. Both rows' `reason` now states the basis and the robustness.

### 5. and 6. — conceded

- "all fifteen … catalytic **serine** on position 189" is now "catalytic **nucleophile** … and fourteen of the fifteen carrying a serine there", in the row summary and in `core_functions`. You were right that `RESULTS.md` had the qualifier and it was dropped in transit.
- `core_functions` now leads with the evidential case for `GO:0052689` — `IPR017157` is a family-specific signature across almost the whole chain, and every biochemically characterised family member is a carboxylic ester hydrolase — with the paralog consistency mentioned afterwards rather than doing the work. Your point stands that the node audit settles the *row*, not the gene-level synthesis.

### 2. `GRANULARITY_MISMATCH` — held, and I'd rather argue it than quietly comply

Your premise is correct: the enum's literal text does fit this row, and my "presupposes donors that agree" constraint is not in the schema. Two reasons I've kept the removal:

- **The slot is about the propagation, not about the term's height.** `failure_modes` is documented as "**Biological subtype for a propagation or inference issue**", and `PropagationFailureModeEnum` sits under a `propagation_review`. Here the propagation has no defect at all: the parent is uninformative *because the donor set is heterogeneous*, not because the transfer could have delivered something narrower. Read literally, "parent term is true but uninformative" applies to every general term on every record, which would make the mode carry no information.
- **It isn't a new constraint, it's the campaign's existing rule.** The working guidance I'm operating under already says: ask "is this term the LCA of its donors?" — `GRANULARITY_MISMATCH` is only apt when the donors agree and the term still sits above them. That is what produced the LCA test in the first place, so applying it here is consistency, not novelty.

What I've changed in response is to stop leaving it implicit: all four rows now say *why* the mode is absent, so a reader isn't left inferring it from a missing key. And I agree this needs a maintainer ruling rather than a per-PR convention — @cmungall, the concrete ask is a one-line clarification to `GRANULARITY_MISMATCH`'s description in `gene_review.yaml`, distinguishing "the propagation could have been more specific" from "the term is general because the donors are heterogeneous". I've deliberately not touched the schema in a gene PR. If you rule the other way, the fix is one enum entry restored on four rows and #2264 needs no change at all.

### 7. `MODIFY` semantics — noted, no change

Agreed it is "drop as redundant" in substance. Keeping `MODIFY` + `proposed_replacement_terms` because it matches AADACL3 and because the schema has no cleaner encoding for pure redundancy; the `reason` is explicit that the replacement is a statement about the gene's annotation set and not about what the node can license.

### #2264

Unchanged advice, now with one more item it can pick up for free: the QuickGO route also confirms AADACL3's own `WITH/FROM` set, and the BNA7/HIDH corrections are in `NODE_PTN009058710.md` once this merges.

### Gates after round 2

`checkquotes.py` **97 quotes, 0 problems** (two quotes were realigned after the audit prose was rewritten); `just validate` **✓ Valid** for both genes, AADACL4's single warning unchanged from `main`; `cache/go/terms.csv` **no deletions and no additions**; a fresh audit run reproduces the committed JSON, and the prose table, counts, sensitivity lines and family-node numbers are all cross-checked against that JSON programmatically rather than by eye.

cmungall:
All five optional items taken — `70cb197`. None changed a conclusion, and two of them were the kind of thing this PR is nominally about, so thanks for not letting them slide.

**8. Duplicated argument.** Correct — the round-1 sentence ("GRANULARITY_MISMATCH would be the wrong classification here…") survived alongside the round-2 paragraph that supersedes it. Round-1 sentence dropped. I also stopped the third occurrence being a verbatim repeat: the `IPR013094` IEA row now cross-references the argument on the IBA row instead of restating the whole paragraph, which is also what "state the cross-gene reasoning once" asked for.

**9. Fixed in the YAML but not at its source.** The sharpest of the five, because it is exactly the failure mode this PR set out to fix — a corrected claim in one file and the uncorrected original in another. `RESULTS.md` now reads "places its own catalytic **nucleophile** on AADACL2 position 189, and fourteen of them carry a serine there", and the `supporting_text` that quotes it moved in the same change, since fixing either alone would have left a stale quote or a broken one.

**10. "True of every donor at the family node."** Conceded, and this was the one I'd have most wanted caught: the family donor set is the `GO:0016020` row's `WITH/FROM`, the PANTHER tree is never fetched, and so `PTN009058713`'s membership is not enumerated. Reworded to "every donor PAINT **cites** at that node" in the audit prose, both reviews and the finding statement, with `membership_enumerated: false` and a `what_is_measured` note now in the JSON so the distinction survives in the machine-readable output rather than only in prose. The audit file also now says explicitly which half of the blockers-are-outside-the-family argument is measured (the `IPR017157` split, per donor) and which is an absence-from-one-WITH/FROM.

**11. Script infelicities.** Both real:
- the `uncovered` guard could not fire, since `from_quickgo` is built from `PARALOG_ACCESSIONS`, which covers all three genes. It is now an `assert` that documents the invariant rather than a test of an impossible condition — it fires if a gene is ever added to `GENES` without an accession, which is the case it was meant to protect against.
- `family_mechanism` now `die()`s on a family-row token absent from the resolved donor set instead of `continue`ing past it, which is the same discipline `query_for` applies to an unknown database prefix. It also skips the `PANTHER:` node explicitly rather than incidentally.

**12. Quote hygiene.** Both duplicates reduced to the single clean quote; nothing now starts mid-bold.

**On your check of item 4** — the S165 observation ("the S165 alternative would put E at the -2 glycine slot, so 164 is the elbow apex") is a better statement of the register argument than mine and confirms it from a direction I hadn't used. Worth having on the record.

**Gates after round 3:** `checkquotes.py` **97 quotes, 0 problems**; `just validate` **✓ Valid** for both genes with AADACL4's single pre-existing warning unchanged; `cache/go/terms.csv` **no deletions, no additions**; a fresh audit run reproduces the committed JSON, and the donor table, counts, sensitivity lines, family-node numbers and the `membership_enumerated` flag are all re-cross-checked against that JSON programmatically. Diff still confined to `genes/human/AADACL2/` and `genes/human/AADACL4/`.

Nothing outstanding from my side. The one open question is still the schema one for @cmungall — whether `GRANULARITY_MISMATCH`'s description should distinguish "the propagation could have been more specific" from "the term is general because the donors are heterogeneous". Whichever way that goes, the fix is small and mechanical: one enum entry restored on four rows, and #2264 then needs no change at all.

cmungall:
Item 13 conceded, and it is the sharpest catch of the four rounds: I corrected "two of them" in six places and left the seventh in the very file whose stale-quote problem was item 9. `29a853b` also finishes the harmonisation on AADACL3, which merged while this was in review.

### Item 13 — fixed by sweeping, not by patching the reported line

Patching named line numbers is exactly what let this survive two rounds, so the fix is a sweep: one script greps **all** changed files for every `IDA … two of them` variant and every unqualified `every donor at the family node`, applies the correction idempotently, then **re-greps and hard-fails** if any occurrence survives. That found **four more** beyond the one you reported:

- the round-1 sentence in `AADACL2-notes.md` *and* `AADACL4-notes.md` ("held by **IDA** in two of them");
- `NODE_PTN009058710.md`'s own "Not two of three: every donor **at** the family node" — the file making the correction carried the uncorrected phrasing;
- two places in `AADACL4-ai-review.yaml` (`:108` and the `:822` finding statement) that round 3's line-targeted edit missed, which is the same defect you flagged as optional item 10.

So the reported instance was one of five. Logged in all three notes files as the round-4 lesson: when a phrase is corrected, grep the changed file set for the **phrase**, not for the line.

### Items 10 and 11 follow-ups

- Item 10's rewording is now applied everywhere (`AADACL2-ai-review.yaml:673`, `AADACL4:108`, `AADACL4:822`, the notes files and the audit prose), caught by the same sweep rather than by hand.
- The `uncovered` guard is an explicit `if … die()` again — you are right that `python -O` strips assertions, and the guard is the mechanism by which the "no gene drops out of the equality test" promise is kept, so it must not be strippable.
- The round-1 "IDA in two of them" sentences in both notes files: rather than marking them superseded, I corrected them, since they were plain factual errors rather than judgments that changed.

### AADACL3 is now in the tree, so the third gene is harmonised here

#2264 merged mid-review. Its verdict was the correct one and is **untouched** — same actions, same terms, same `core_functions`, verified by diffing against `origin/main`. What changes is the encoding it kept on the literal reading of the enum, plus two donor facts the shared audit settled:

- **`GRANULARITY_MISMATCH` removed** from both hydrolase rows, so all three paralogs now record `EVIDENCE_CIRCULAR_OR_REDUNDANT` with no granularity failure mode. Its position was in tension with itself: it argued `GO:0016787` is the genuine LCA *and* recorded a granularity mismatch. `SOURCE_EVIDENCE_WEAK` is **kept** on its `IPR013094` row — a separate and still-correct judgment about a fold signature reaching outside the subfamily, not a granularity claim, so removing it would have been over-reach.
- **BNA7 recorded as resolved.** Its own audit reached `SGD:S000002836` through an Alliance record, failed, and honestly reported the nucleophile unresolved — hence 13 of 14. The shared audit resolves it directly (`Q04066`, `ACT_SITE 110` labelled nucleophile, reads Ser), so 15 of 16. `AADACL3-bioinformatics/RESULTS.md` is deliberately **left as generated** — 13 of 14 is what its resolver really produced — and the correction is recorded as a supersession citing the shared audit, so nothing claims a number its own script did not compute.
- **HIDH `SOURCE_WEAK_OR_INFERRED` → `SUPPORTS_TRANSFER`**: `GO:0106435 carboxylesterase activity` is IDA, so its esterase activity is slight but not *inferred*.

With all three folders present the audit now measures the 17-token equality from the committed TSVs **and** from QuickGO for every gene. I kept the QuickGO path rather than dropping it as redundant — two independent sources having to agree is a better guarantee than one, and it is what made the claim measurable when a folder was missing.

The PR body and title are updated to cover all three genes.

### Gates

`checkquotes.py` across all three reviews: **133 quotes, 0 problems**. `just validate` **✓ Valid** for AADACL2, AADACL3 and AADACL4 (AADACL3 and AADACL4 each carry one pre-existing deep-research warning, unchanged from `main`). `cache/go/terms.csv`: **no deletions, no additions**. `origin/main` merged in and re-fetched clean. A fresh audit run reproduces the committed JSON, and 48 invariant checks over the three genes — action, replacement term, root cause, absence of the failure mode, `supporting_entities` equality with the GOA `WITH/FROM` column, core MF, shared-audit citation, notes coverage — all pass.

The `GRANULARITY_MISMATCH` schema question is unchanged and still for @cmungall; it now affects six rows across three genes rather than four across two, which if anything makes the one-line enum clarification more worth having.

cmungall:
Both items conceded — `6d07d80`. You are right that round 4 created them: harmonising AADACL3 made every "these three disagree" statement false, and round 4's sweep grepped for the `two of them` phrasings but not for those. That is the second instance of the same failure in this PR, so the guard is widened rather than the two reported lines patched.

**17. AADACL3's `suggested_questions` conflict entry.** Removed. It said the merged AADACL2 review resolves the same row as `MODIFY → GO:0017171` with `TERM_SCOPING_PROBLEM` + `GRANULARITY_MISMATCH`, that "both cannot be right about one row", and that "PR #2266 needs a follow-up to settle it" — which after merge would tell a curator a follow-up is outstanding when the file two directories away already agrees with it. The entry now asks PAINT only what is genuinely still open, where the mechanism term should sit, and keeps the merges-two-existing-rows argument, which is what settled the question in the first place.

**18. AADACL2's schema-request clause.** The justification changes rather than the request. It no longer cites a divergence that does not exist; it now says all three paralogs encode the row on the propagation-shape reading, which the enum text does not itself state, so the convention is carried by argument rather than by the schema. The same clause was live in AADACL4's reason too — the sweep caught that one, I had not noticed it.

**The widened guard.** It greps all three reviews plus the audit prose for `TERM_SCOPING_PROBLEM`, "both cannot be right", "needs a follow-up" and the divergence clause, and hard-fails unless the only survivor sits in text explicitly marked historical. That last exemption is load-bearing: the audit's "Why this audit exists" section legitimately needs to describe the three verdicts that motivated it, so it is now prefixed *"Historical: this section records the state that motivated the audit, not the current state"* and closed with a **Settled outcome** paragraph stating what all three reviews now record. The verbs in it moved to past tense as well.

**Optional items, all three taken.** AADACL3's earlier conflict section carries the inline superseded marker (with a note that the argument inside it is kept because it is what settled the question); its local 13-of-14 serine count now points at the shared audit's 15 of 16 and says why the two differ; and I agree the pending schema ruling now costs six rows rather than four, which argues for making the `gene_review.yaml:2468-2469` clarification soon.

**The generalisable lesson**, now logged in all three notes files: when a change makes a claim false, grep for the **claim**, not for the sentence you remember writing — and when the change is "these two now agree", the claims to hunt are the ones asserting that they do not. Round 3 learned the first half of that and round 5 needed the second.

**Gates:** `checkquotes.py` across all three reviews **133 quotes, 0 problems**; `just validate` **✓ Valid** for AADACL2, AADACL3 and AADACL4 (the two pre-existing deep-research warnings unchanged); `cache/go/terms.csv` **no deletions, no additions**; `origin/main` re-fetched with nothing new; 19 invariant checks across the three genes — action, replacement term, root cause, absence of the failure mode, `supporting_entities` equality with the GOA `WITH/FROM` column, core MF, shared-audit citation, notes currency, and the audit verdict itself — all pass.

cmungall:
Item 22 conceded, and you and the coordinator are both right that the answer is automation rather than more care — `abc09d9`. This was the fifth instance of one defect (9, 13, 17, 18, 22), and this time the false claim was in the journal bullet recording the lesson, which is about as clear a signal as I am going to get.

**Root cause of item 22, since it is instructive.** Round 5's edits were in one script: apply item 17, then item 18, then the AADACL3 13-of-14 pointer, then the notes marker. The item-18 anchor assertion failed — the text was wrapped differently than I had it — and the script aborted *after* writing item 17 and *before* the last two. I then re-ran only the item-18 part, and wrote the journal bullet from my intent rather than from the tree. Partial application plus a from-memory changelog.

### The two edits applied, not the bullet corrected

- **`AADACL3-notes.md`** no longer opens "This review now conflicts with the merged AADACL2 review, and the conflict should be resolved there rather than papered over here", and no longer says a curator gets contradictory advice or asks for a #2266 follow-up. Rewritten in the past tense with the resolution stated first, keeping the merges-two-existing-rows argument, since that is what settled the question.
- **`AADACL3-ai-review.yaml`** now carries the 15-of-16 supersession *in the passage where the 13-of-14 count is stated*, naming why the two differ (Alliance record vs `xref:sgd-S000002836`). Your item 20 was right that a `propagation_review` comment and a reference entry elsewhere in the file is not the same as the pointer being where the count is.

**And the bullet's own wording.** It said "`suggested_questions` **here**" and "the conflict section earlier in **this file**" — true of AADACL3, false in the AADACL2 and AADACL4 copies. Rewritten to name the file in each case, so the same text is true in all three.

**Two further live statements the widened guard found**, neither reported: `AADACL4-notes.md` still recorded "`root_cause: TERM_SCOPING_PROBLEM`, `failure_modes: [GRANULARITY_MISMATCH]`" as the decision taken, and `AADACL3-notes.md`'s earlier reconciliation section predated the harmonisation entirely.

### The guard, committed and self-tested

`genes/human/AADACL2/AADACL2-bioinformatics/check_paralog_agreement.py`, covering the **reviews, the notes files and the audit prose** — nine files:

1. **the agreement invariant** over all six `GO:0016787` rows: `MODIFY` → `GO:0052689`, `EVIDENCE_CIRCULAR_OR_REDUNDANT`, no `GRANULARITY_MISMATCH`, `supporting_entities` equal to each gene's own GOA `WITH/FROM` column, one shared 17-token set across the three, core MF `GO:0052689`, shared audit cited;
2. **stale-claim greps** for every phrasing the harmonisation falsified, plus a rule that a superseded count must carry its pointer *in the same passage*.

Two design points worth stating, because a naive version of this is unusable:

- **Notes files are journals.** A journal recording "X was wrong, now fixed" necessarily contains X, so a blanket grep flags 63 false positives — I tried it. They are scanned paragraph by paragraph, and a stale phrase is allowed only where the paragraph or a marker at the top of its section marks the passage retrospective. An unqualified stale sentence in running prose fails, which is exactly the shape of the AADACL3 section that survived four rounds.
- **The section-level exemption needs a *strong* marker** (superseded / historical / since resolved), not any past tense, so appending a new live claim to an old section is not laundered by its header. There is a self-test mutation for precisely that.

### The self-test paid for itself on its first run

`--self-test` copies the tree to a temp directory, applies one mutation at a time, and requires each to be caught. First run:

```
FAIL superseding pointer removed from a count: NOT caught - this guard does not work
self-test: 10/11 guards demonstrably fire
```

The fault turned out to be the **mutation**, not the guard: it only reworded the lead-in and left "15 of 16" inside the search window, so nothing was actually broken and the guard was right to stay silent. Reading the guard would not have found that — and note that the naive conclusion ("the guard is broken, loosen it") would have made things worse. The mutation now deletes the whole clause and **raises if its target text has moved**, so the self-test cannot quietly start passing for the wrong reason later. It also caught a real guard bug on the way: the pointer window looked only forward, so a supersession stated *before* its count read as absent.

```
11/11 guards demonstrably fire
```

### Gates

`check_paralog_agreement.py`: **no problems across 9 files**. `checkquotes.py` over all three reviews: **133 quotes, 0 problems**. `just validate` **✓ Valid** for AADACL2, AADACL3 and AADACL4 (two pre-existing deep-research warnings). `cache/go/terms.csv`: **no deletions, no additions**. `origin/main` merged in (one new commit) and the guard re-run after the merge. One push, as asked.

Item 2 stays open in `suggested_questions` as the repo-level `GRANULARITY_MISMATCH` question; it is not mine to settle here.

PR #2287, opened 2026-07-26
Remove 80 stale full_text_unavailable flags; add an audit for them

cmungall:
## The defect

A reference marked `full_text_unavailable: true` whose cached publication actually **has** full text is a silent defect. No validator checks the pair, and the flag actively discourages extracting the evidence the annotation needs.

Found while reviewing PAINT genes: ACAP1's review carried the flag on `PMID:25284369`, whose cache reports `full_text_available: true` with a PMCID. Only the Results section contains the assay that makes the annotation an IDA, and the negative BAR-domain result — the abstract has neither.

## Scope

A sweep of **4031 reviews** against **24107 cached publications** found **79 stale flags in 24 reviews across 9 organisms**. The removal pass took **80** (one review carries the same PMID twice).

The signature is sharper than the count: **39 sit on a reference with zero `findings`**, and **all 32 in `human/CDK2` are zero-findings** — consistent with the flag having been applied across a reference list rather than per paper, then suppressing extraction wholesale.

| review | flags |
|---|---|
| `human/CDK2` | 32 (all 0 findings) |
| `DANRE/dph2` | 7 |
| `human/BCL2L13` | 6 |
| `human/DPT` | 5 |
| `human/CALCOCO1` | 4 |
| `human/CHAMP1` | 3 (all 0 findings) |
| 18 others | 1–2 each |

Verified case: `human/CDK2` `PMID:11953320` is flagged unavailable, its cache has `full_text_available: true` and `pmcid: PMC125963`, and the reference carries **0 findings**.

## What is and isn't touched

Only the contradicted direction. The converse — no flag, cache lacks full text — is **not** a defect, since the flag is optional. A cache that never recorded `full_text_available` is never read as "unavailable".

Editing is **line-wise**, not a `yaml.dump` round-trip, so quoting style, block scalars, key order and comments are untouched. The diff is exactly **80 deleted lines across 24 files and zero other changed lines** — a reformat would have produced an unreviewable diff across two dozen unrelated genes.

## Added

- `src/ai_gene_review/tools/audit_fulltext_flags.py` — reports, or fixes with `--fix`, and exits non-zero while flags remain so it can serve as a CI gate.
- `ai-gene-review audit-fulltext-flags` and `just audit-fulltext-flags`.
- **8 tests**, covering what makes bulk running safe: a cache missing the key is never read as unavailable; the key is ignored beyond the frontmatter; an adjacent *legitimately* flagged reference keeps its flag; and removal preserves folded block scalars and quoting byte-for-byte.

## Verification

- `just validate` passes on a sample of edited genes: `human/CDK2`, `human/DPT`, `worm/csr-1`, `DANRE/dph2`.
- `mypy` clean on the new module; 8/8 tests pass.
- The audit now reports **zero** stale flags repo-wide.
- `cli.py` diff is **+53/-0** — I reverted an accidental `ruff format` pass that had rewritten 548 unrelated lines.

## Deliberately not done

This does **not** re-extract the suppressed findings. The 39 zero-findings references are now identifiable as candidates for that work, but re-reading them is a curation task rather than a mechanical one.


cmungall:
Both items were real bugs, and the first one corrects this PR's own accounting.

**🟡 1 — over-removal.** You're right, and the mechanism matters: `find_stale_flags` only inspects reference-level flags, so the post-fix re-check was **structurally blind** to anything the mutator stripped beyond that scope. Deletion is now gated on the flag sitting at the same column as the `id:` key that set the current reference, and `reference_id:` still doesn't reset it (a `SupportingTextInReference` block is keyed by that name — treating it as a new reference would let a nested flag be attributed to it).

`audit()` now also asserts **detected == removed** and fails on a mismatch, so a future scope divergence between detector and mutator is a hard error rather than a silent over-edit. That guard is the general fix; the indentation check alone would only have closed this instance.

**And your correction to my accounting is the part I should own.** I noticed the 79-vs-80 gap, reasoned that a duplicate PMID explained it, and wrote that into the PR description. A duplicate can't produce it — it yields two detections *and* two removals. The extra line was the clpP2 nested flag. I had the discrepancy in front of me and rationalised it instead of investigating.

clpP2 is restored with its nested flag intact. The delta is now **79 detected, 79 removed, 79 deleted lines across 24 files, zero non-flag lines changed**.

**🟡 2 — CLI duplication.** Extracted one shared `audit()` returning an exit code; both the module entry point and the subcommand call it. Worth noting the copy had drifted within an hour of being written — it had already lost `--gene-dir` and changed the failure message, which is a fair argument that I shouldn't have duplicated it even "temporarily". `cli.py` is now **+11/−41**.

**All three suggestions taken:**

- **Frontmatter is parsed as a real `---` block** instead of reading a fixed 4096 bytes. You're right that the test name papered over the truncation: the guard was against full text quoting the key in prose, but it would have silently missed the key in any record whose frontmatter outgrew the window. There's now a test proving the key is found beyond that window while a body occurrence is still ignored. I didn't share `validator.py:817-832` directly — it's inline inside a much larger annotation check rather than an extractable helper, and lifting it out would widen this PR into that function's blast radius. Happy to do it as a follow-up if you'd rather have one implementation.
- **Exit code tested** — non-zero while flags remain, zero after a fix, zero on a clean tree, plus `FileNotFoundError` when `publications/` is absent. Still unwired to CI deliberately: adding it to `just test` would fail the suite for anyone who adds a flag mid-review, which I'd want your call on rather than assume.
- **`reports/suppressed-evidence-candidates.md`** gives the 39 zero-findings references a durable home, split by priority, generated from the **pre-fix tree** so it records exactly what was removed rather than a reconstruction.

Tests 8 → 13. mypy clean, ruff formatted, `just validate MYCTU clpP2` passes, and the audit reports zero repo-wide after merging current `main`.

On your closing note — you couldn't run `uv`/`mypy`/`validate`, so those were taken on trust. For what it's worth, the two things you caught by reading were both things those tools would have passed: the over-removal produced valid YAML that validated cleanly, and the duplicated CLI copy type-checked fine.

PR #2288, opened 2026-07-26
AADACL3: drop the one failure_modes divergence #2286 left behind

cmungall:
Follow-up to #2286, which harmonised the `GO:0016787` treatment across AADACL2/3/4.

All six rows now `MODIFY → GO:0052689` with `root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT` — replaced because `IPR017157` already supplies the specific term per gene, not because the generic term is too coarse.

**One row was left inconsistent.** AADACL3's IEA row alone carried `failure_modes: [SOURCE_EVIDENCE_WEAK]`; the same row in AADACL2 and AADACL4 carries none. The three IEA rows are **identical in GOA** — same `GO_REF:0000002`, same single source `InterPro:IPR013094` — so nothing about AADACL3 justifies a different assessment.

It is also the wrong mode on its own terms. The root cause is redundancy with a more specific InterPro-derived term, which says nothing about source strength, and `IPR013094` (the α/β-hydrolase fold signature) is not a weak source. `SOURCE_EVIDENCE_WEAK` is the value this campaign has repeatedly had to retract once the sources turned out to be sound.

Removed rather than replaced: with the root cause already stating why the row is superseded, no failure mode applies — same as the five sibling rows.

Six rows across three genes now agree, which was the point of #2286.

Diff is **2 deleted lines**. Validation passes, 36 quotes verbatim, no `cache/go/terms.csv` change.

---

> [!NOTE]
> **Low Risk**
> Curatorial YAML and notes only; no ontology, GOA, or executable pipeline changes.
> 
> **Overview**
> Follow-up to **#2286**: the generic `GO:0016787` hydrolase rows were already harmonized on **redundancy** with subfamily-derived `GO:0052689`, but AADACL3's **InterPro `IPR013094` IEA** row still diverged.
> 
> **`AADACL3-ai-review.yaml`** drops `failure_modes: SOURCE_EVIDENCE_WEAK` on that IEA row (AADACL2/AADACL4 carry none) and changes the `InterPro:IPR013094` source entity from `SOURCE_WEAK_OR_INFERRED` to **`CIRCULAR_OR_REDUNDANT`**, consistent with the existing comment that the fold match does not carry the activity conclusion.
> 
> **`AADACL3-notes.md`** adds section **1b** documenting why weak-source labeling was wrong here—the issue is redundancy with `IPR017157`, not source strength—and that all six hydrolase rows and three `IPR013094` entities now agree across the paralog reviews.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1b8d0f83f52211ca8d023024b84e6223cff3fe5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/genes/human/AADACL3/AADACL3-ai-review.yaml
+++ b/genes/human/AADACL3/AADACL3-ai-review.yaml
@@ -413,7 +413,7 @@ existing_annotations:
       source_entities:
       - source_id: InterPro:IPR013094
         source_label: Alpha/beta hydrolase fold-3 (fold-level domain)
-        source_status: SOURCE_WEAK_OR_INFERRED
+        source_status: CIRCULAR_OR_REDUNDANT
         comment: >-
           A fold signature, not an activity. It reaches beyond the subfamily whose chemistry is being
           asserted: it also matches soybean HIDH at 74-298, which matches neither IPR017157 nor
--- a/genes/human/AADACL3/AADACL3-ai-review.yaml
+++ b/genes/human/AADACL3/AADACL3-ai-review.yaml
@@ -410,8 +410,6 @@ existing_annotations:
       supporting_text: '| HIDH_SOYBN (Q5NUF3) | - | - | 74-298 | - |'
     propagation_review:
       root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT
-      failure_modes:
-      - SOURCE_EVIDENCE_WEAK
       source_entities:
       - source_id: InterPro:IPR013094
         source_label: Alpha/beta hydrolase fold-3 (fold-level domain)
--- a/genes/human/AADACL3/AADACL3-notes.md
+++ b/genes/human/AADACL3/AADACL3-notes.md
@@ -481,10 +481,22 @@ arguing that `GO:0016787` is the genuine LCA. Those two are in tension: `failure
 the *biological shape of a propagation issue*, and this propagation has none — the parent is
 uninformative because the donor set is heterogeneous, not because the transfer could have been
 more specific. `root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT` alone now carries the row, matching
-AADACL2 and AADACL4. `SOURCE_EVIDENCE_WEAK` is kept on the `IPR013094` row: that is a separate
-and still-correct judgment about a fold signature reaching outside the subfamily, not a
-granularity claim. (The schema's enum description does not yet distinguish the two readings and
+AADACL2 and AADACL4. (The schema's enum description does not yet distinguish the two readings and
 is worth clarifying; that is a repo question, raised in #2286.)
+
+**1b. Superseded: the `IPR013094` assessment was aligned too.** This section originally kept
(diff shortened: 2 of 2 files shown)


cmungall:
You caught this PR doing exactly what it was written to fix. `source_status: SOURCE_WEAK_OR_INFERRED` sat **one field below** the `failure_modes` I deleted, carrying the same retracted assessment, so "six rows now agree" was false when I wrote it.

Aligned to `CIRCULAR_OR_REDUNDANT`. Your argument is decisive and I have nothing to add to it — the case against the first value applies unchanged to the second: the recorded root cause is redundancy with the `IPR017157`-derived `GO:0052689`, which says nothing about source strength; `IPR013094` is not a weak source; this review's own comment on that entity already concludes the fold match carries nothing (*"the activity conclusion here is carried by the intact Ser-Asp-His triad and by the subfamily signature IPR017157, not by this fold match"*), which is what `CIRCULAR_OR_REDUNDANT` means; and AADACL2/AADACL4 make the identical fold-level argument while recording `CIRCULAR_OR_REDUNDANT`.

**The notes then asserted the old state a third time**, explicitly defending it as *"a separate and still-correct judgment"*. Corrected with a superseded marker rather than a silent rewrite — the original reasoning is what went wrong, so deleting it would hide the thing worth recording.

What I did differently: instead of re-reading the block I had just edited, I grepped **every field** of the review and the notes for both retracted values. The only surviving occurrence is inside the superseded paragraph describing what was there. Then I checked the invariant programmatically across all three genes, reading AADACL2/AADACL4 from `origin/main`:

```
AADACL2  IBA MODIFY -> GO:0052689  root=EVIDENCE_CIRCULAR_OR_REDUNDANT modes=None  IPR013094=[]
AADACL2  IEA MODIFY -> GO:0052689  root=EVIDENCE_CIRCULAR_OR_REDUNDANT modes=None  IPR013094=CIRCULAR_OR_REDUNDANT
AADACL3  IBA MODIFY -> GO:0052689  root=EVIDENCE_CIRCULAR_OR_REDUNDANT modes=None  IPR013094=[]
AADACL3  IEA MODIFY -> GO:0052689  root=EVIDENCE_CIRCULAR_OR_REDUNDANT modes=None  IPR013094=CIRCULAR_OR_REDUNDANT
AADACL4  IBA MODIFY -> GO:0052689  root=EVIDENCE_CIRCULAR_OR_REDUNDANT modes=None  IPR013094=[]
AADACL4  IEA MODIFY -> GO:0052689  root=EVIDENCE_CIRCULAR_OR_REDUNDANT modes=None  IPR013094=CIRCULAR_OR_REDUNDANT
```

Worth stating plainly: this is the fifth or sixth time in this campaign that a correction landed in one field and not its neighbour, and it happened here in a **three-line PR whose entire purpose was removing a divergence**. Care demonstrably does not fix it — #2286 ended by committing `check_paralog_agreement.py` for precisely this invariant, and the honest conclusion is that the guard should be extended to cover `source_status` and run over these three genes, rather than my promising to look harder next time. Flagging that as the follow-up rather than pretending the pattern is closed.

Validation passes; 36 quotes verbatim; no `cache/go/terms.csv` change.

PR #2292, opened 2026-07-26
Guard the IPR013094 source_status — the field the last two PRs both missed

cmungall:
#2286 committed `check_paralog_agreement.py` because five of its six rounds hit the same defect: a claim corrected in one place and left standing in another. **The guard did not cover `source_status` on the source entity, and that field then diverged in exactly the way the guard exists to prevent — twice.**

AADACL3 kept `source_status: SOURCE_WEAK_OR_INFERRED` on `InterPro:IPR013094` where both siblings use `CIRCULAR_OR_REDUNDANT`. #2286 missed it, and so did **#2288 — the three-line PR written specifically to remove this class of divergence** — because the field sits one line below the one being fixed, so re-reading the edited block could not reveal it. A reviewer caught it both times.

## What the check enforces

For every `GO:0016787` row citing that signature:

- `source_status == CIRCULAR_OR_REDUNDANT`, matching the siblings; and
- a **named** rejection of `SOURCE_WEAK_OR_INFERRED` / `SOURCE_EVIDENCE_WEAK`, with the reason in the message — those are the specific values retracted for this row, because the recorded root cause is redundancy (which says nothing about source strength) and the α/β-hydrolase fold-3 signature is not a weak source.

Distinguishing the retracted values from a merely-different one matters: the error mode here is *reinstating the old assessment*, and a bare equality check would report that identically to a typo.

## Verification

- Added to `MUTATIONS`, so `--self-test` proves it fires: **12/12 guards demonstrably fire**.
- The guard **caught the live defect on `origin/main`** before #2288 merged — so it was validated against the real bug, not only a synthetic mutation.
- Baseline passes on the rebased tree (`exit=0`), now that #2288 is in.

Stacked on #2288, which is already merged; this rebases cleanly onto `main` with only the guard change remaining.

---

> [!NOTE]
> **Low Risk**
> Changes only the AADACL paralog validation script and its self-tests; no runtime services or curator YAML content in this diff.
> 
> **Overview**
> Extends **`check_paralog_agreement.py`** so the paralog agreement invariant also covers **`InterPro:IPR013094`** on the shared `GO:0016787` IEA rows—the fold signature that was left divergent across harmonisation and a follow-up fix because **`source_status`** sits adjacent to the fields those edits touched.
> 
> **`check_invariant`** now requires the IPR013094 entity’s **`source_status`** to be **`CIRCULAR_OR_REDUNDANT`**, with explicit rejection of the retracted **`SOURCE_WEAK_OR_INFERRED`** status and **`SOURCE_EVIDENCE_WEAK`** in **`failure_modes`** (each checked in the enum where it is legal). For IEA rows it also fails if that source entity is missing, so the guard cannot be bypassed by deleting the entity.
> 
> Documentation at the top of the script describes this gap. **`--self-test`** gains three mutations (wrong fold **`source_status`**, deleted fold entity, re-added **`SOURCE_EVIDENCE_WEAK`**), bringing the demonstrably firing guard count to **12**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd6290d1e5dd42be09b81675798ca771b1ec94b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/genes/human/AADACL2/AADACL2-bioinformatics/check_paralog_agreement.py
+++ b/genes/human/AADACL2/AADACL2-bioinformatics/check_paralog_agreement.py
@@ -71,8 +71,12 @@ HISTORICAL_MARKER = "Historical: this section records the state that motivated t
 FOLD_SOURCE = "InterPro:IPR013094"
 FOLD_SOURCE_STATUS = "CIRCULAR_OR_REDUNDANT"
 # Values retracted for this row: the root cause recorded on it is redundancy, which says nothing
-# about source strength, and the fold-3 signature is not a weak source.
-RETRACTED_STATUSES = {"SOURCE_WEAK_OR_INFERRED", "SOURCE_EVIDENCE_WEAK"}
+# about source strength, and the fold-3 signature is not a weak source. They live in DIFFERENT
+# enums, and a first version of this guard rejected both in source_status - where SOURCE_EVIDENCE_WEAK
+# is not a permissible value at all, so that half was unreachable and read as coverage while the
+# field it can legally occupy went unguarded.
+RETRACTED_SOURCE_STATUS = "SOURCE_WEAK_OR_INFERRED"   # PropagationSourceStatusEnum
+RETRACTED_FAILURE_MODE = "SOURCE_EVIDENCE_WEAK"       # PropagationFailureModeEnum
 
 # Phrasings that were true before the three reviews were harmonised and are false after it.
 # Each is (regex, why it is stale). Applied to reviews, notes and the audit prose alike.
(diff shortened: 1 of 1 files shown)


cmungall:
Both items were real, and the first is the same shape of hole this PR exists to close — a guard that can be defeated by **removing the thing it guards**.

**🟡 Presence is now asserted.** The loop was continue-on-mismatch, so it passed silently when the `IPR013094` entity was absent entirely: deleted, `source_entities` dropped, or `source_id` relabelled. A `seen_fold` flag now requires it on the IEA row, matching every neighbouring check, and a new mutation **deletes the entity outright** to prove the assertion fires. You were right that it was out of step with its neighbours; I wrote the check for the failure I had just seen rather than the class it belongs to.

**Your enum observation is correct and I verified it against the schema rather than taking it on trust:**

| enum | has `SOURCE_WEAK_OR_INFERRED` | has `SOURCE_EVIDENCE_WEAK` |
|---|---|---|
| `PropagationSourceStatusEnum` | ✅ | ❌ |
| `PropagationFailureModeEnum` | ❌ | ✅ |

So rejecting `SOURCE_EVIDENCE_WEAK` in `source_status` was **unreachable code that read as coverage**, while the field where it *is* legal went unguarded — only `GRANULARITY_MISMATCH` was checked in `failure_modes`. That's worse than the gap it purported to close, because it looked like protection.

The constants are now split by the enum each belongs to — `RETRACTED_SOURCE_STATUS` and `RETRACTED_FAILURE_MODE` — each rejected where it can actually appear, with a third mutation covering the failure-mode case. And `SOURCE_EVIDENCE_WEAK` in `failure_modes` is worth guarding on its own merits: every protein donor on this node carries its own experimental evidence, so "weak sources" is factually unavailable here, which is precisely the retraction this cluster kept having to make.

**🔵 target-drift assertions added** to all three mutations, following `_break_superseded_pointer`'s precedent. A mutation that silently no-ops proves the guard fires when it does not — the same failure mode as the two above, one level up.

**Verification:** self-test **12/12 → 14/14 demonstrably fire**, baseline still passes. On your closing note that you couldn't run `--self-test`: worth recording that both things you caught by reading were invisible to running it — the deleted-entity hole had no mutation to exercise it, and the unreachable branch would never have been reached by any mutation I'd written. The self-test proves the guards I thought of fire; it cannot tell me which guard I failed to write.

PR #2290, opened 2026-07-26
Review human ACTR8 (ARP8): a complex-level phenotype that became five gene-level rows

cmungall:
## ACTR8 (ARP8, hArp8) — INO80's nucleosome-recognition module

`Q9H981`, 624 aa, Swiss-Prot. **37 rows: 14 ACCEPT · 13 KEEP_AS_NON_CORE · 4 MARK_AS_OVER_ANNOTATED · 6 NEW · 0 REMOVE · 0 MODIFY.**

ARP8 is unusually well characterised for a nuclear actin-related protein — a 2.6 Å ATP-bound crystal structure (PDB 4FO0), nanomolar `K_d`s for nucleosomes and the H3–H4 tetramer, a measured ATPase, a demonstrated ssDNA preference, and an INO80-independent mitotic role. GOA reflects almost none of it. So the risk here was mis-scoping rather than over-annotation, and the two defects found are both about **how evidence travelled**, not about wrong biology.

### 1. A complex-level phenotype became five gene-level rows on ACTR8 — and on 14 other genes

Five of the 31 GOA rows (`GO:0000723`, `GO:0006282`, `GO:0045739`, `GO:0045995`, `GO:1904507`) are Ensembl Compara `IEA` transfers from mouse Actr8. Tracing the mouse side:

- all five mouse annotations come from **one** paper, `PMID:23979016`, and all five are `assignedBy: ComplexPortal`;
- QuickGO `reference=PMID:23979016` returns **80 annotations = 16 entities × 5 terms** — `CPX-878` plus **all 15 INO80 subunits** (`Ino80`, `Ino80b/c/d/e`, `Ruvbl1/2`, `Yy1`, `Tfpt`, `Nfrkb`, `Mcrs1`, `Uchl5`, `Actl6a`, `Actr5`, `Actr8`) with byte-identical evidence sets;
- the paper's experiment is an **`Ino80` conditional knockout**. `Actr8` is not named in the abstract and no *Actr8* mouse knockout exists. The phenotype behind `GO:0045995` is literally *"mIno80−/− mouse embryos die early during embryogenesis"*.

ComplexPortal's projection is a documented convention with the complex recoverable from `CPX-878`. The defect is the **second** step: Compara consumes a projected subunit annotation as if it were gene-specific mouse experimental evidence and re-projects it across species, at which point the complex context is gone. Two inferences stacked, presented as one `IEA` from an orthologue.

[...]

PR #2295, opened 2026-07-26
Review human ACTMAP: the actin maturation protease is annotated as a methionine aminopeptidase

cmungall:
## ACTMAP (C19orf54) — the enzyme that acts *on* actin, and the term that calls it a MetAP

ACTMAP is the inverse of the actin-cluster genes reviewed around it: not an actin-related protein,
but the cysteine protease that **acts on** actin. It excises the Nα-acetylated initiator residue of
immature actin post-translationally — the step NAA80 then follows by re-acetylating the newly exposed
Asp2/Glu2. All 6 GOA rows reviewed one-to-one in GOA order; 2 new annotations proposed.

**Actions:** 4 ACCEPT · 2 MODIFY · 1 NEW · 0 REMOVE · 0 UNDECIDED  ·  *(round 2: the second NEW was withdrawn — see below)*

| # | term | evidence | action |
|---|---|---|---|
| 1 | GO:0005737 cytoplasm | IEA (SubCell) | ACCEPT |
| 2 | GO:0004239 initiator methionyl aminopeptidase activity | IEA (Ensembl) | MODIFY → GO:0070005 |
| 3 | GO:0016485 protein processing | IEA (Ensembl) | ACCEPT |
| 4 | GO:0070005 cysteine-type aminopeptidase activity | IDA (FlyBase) | ACCEPT |
| 5 | GO:0004239 initiator methionyl aminopeptidase activity | IDA (UniProt) | MODIFY → GO:0070005 |
| 6 | GO:0016485 protein processing | IDA (UniProt) | ACCEPT |
| 7 | GO:0005522 profilin binding | IPI (proposed) | NEW |

### The headline finding: `GO:0004239` is the wrong term, and it is an ontology gap rather than a curator error

Three grounds, each **computed** by the committed script rather than asserted:

1. **The term means canonical MetAP.** Definition: *"Catalysis of the release of N-terminal initiator
   methionine from peptides"*; **only** EC xref `3.4.11.18`. ACTMAP does not release methionine —
   UniProt's own four Rhea reactions have it releasing `N-acetyl-L-methionine` (or
   `N-acetyl-L-cysteine`), and UniProt deliberately assigns **`EC 3.4.11.-`, not `3.4.11.18`**. The
   2022 paper's central point is that cytoplasmic actin *"defies this rule"*.
2. **The branch requires a free N-terminus that the substrate does not have.** Both `GO:0004239` and
   `GO:0070005` descend from `GO:0008238 exopeptidase activity`, whose definition ends *"…in a
   reaction that **requires a free N-terminal amino group**…"*. ACTMAP's substrate is Nα-acetylated.
   **GO already models the exact analogue correctly:** `GO:0016920 pyroglutamyl-peptidase activity`
   releases a single N-terminal residue whose α-amino group is blocked, and sits under
   `GO:0008242 omega peptidase activity` **+** `GO:0008234 cysteine-type peptidase activity` and
   **not** under `GO:0004177 aminopeptidase activity`. Computed: `under omega peptidase: True`,
   `under cysteine type peptidase: True`, `under aminopeptidase: False`.
3. **The term now conflates two enzyme families.** Censused over five model organisms: 86
   annotations on **74 gene products**, of which **4 are ACTMAP-family** (PTHR28631: human, mouse,
   zebrafish, *Drosophila*) and **70 are not** — `METAP1/1D/2`, yeast `MAP1/MAP2`, `RNPEPL1`, fly
   `MAP1A/MAP1B/und`. One term for both the metal-dependent co-translational MetAPs and the
   cysteine-nucleophile ACTMAP family. **That five-taxon figure understates the family** - it cannot
   see the bovine and *Xenopus* members - so the script also asks each reviewed PTHR28631 member
   directly: **6 of 6 carry the term** (`A6QQD2` bovine ISS, `B0BM95` *Xenopus* ISS, `B0V3H4`
   zebrafish ISS, `J3QPC3` mouse IDA+ISS+ISO+IEA, `Q5BKX5` human IDA+IEA, `Q9VCE8` *Drosophila* ISS).
   Every non-human one is an ISS/ISO/IEA descendant of the human annotation.

**Explicitly not a criticism of the curator.** `GO:0004239` was obsoleted in 2015 and **reinstated
on 2023-03-14** with its present narrow definition (QuickGO: *"This term was reinstated from
obsolete"*); the UniProt IDA is dated **2023-03-31**, 17 days later. The curator took the closest
term that existed. So this is filed as a `proposed_new_terms` request — a cysteine-type *omega*
peptidase term modelled on `GO:0016920` — plus MODIFY to the best term available today (`GO:0070005`,
already on the record with the same IDA support), which retains single-N-terminal-residue release and
the cysteine mechanism while dropping the false EC 3.4.11.18 implication. **One new term would fix
all six reviewed ACTMAP-family gene products, in six species.**

### The activity claim rests on a tested residue, not a domain name

`ACT_SITE 132` in a `Peptidase C39-like` region, with `C132A` catalytically inactive
(`ECO:0000269|PubMed:36173861`) — and independently re-tested four years later by a different lab
with covalent chemistry: butynamide stereoprobes react with WT but **not** C132A ACTMAP, **C119A and
C271A remain reactive** (the specificity control), and the probe destabilises WT but not C132A. So
`GO:0070005` is ACCEPT-ed on mechanism, with the exopeptidase-definition tension recorded as a
`knowledge_gap` rather than used to demote it.

### Both Ensembl IEA rows are reciprocally circular

`WITH/FROM` resolves to `J3QPC3` → `ACTMP_MOUSE`, 361 aa, Swiss-Prot, requested accession returns
itself (not a dead/merged entry). The donor **does** hold its own IDA — but **from the same paper**
(Haahr et al. assayed the human enzyme *and* a mouse knockout in one study), and **three of its four
`GO:0004239` annotations are themselves projections back from human `Q5BKX5`**. So the rows add zero
evidence independent of the human IDA beside them. Kept (the donor is the true 1:1 ortholog), with
`propagation_review` recording `CIRCULAR_PROPAGATION` so they are never counted as a second line of
support.

### The predicted role conflation did not happen in the annotations — it happens in the ontology

The obvious risk for an enzyme acting on actin was that actin's own cell biology would leak onto it
(the ACTL8 shape). **Checked and refuted:** ACTMAP's whole GOA record is 6 rows — cytoplasm, two
protease MF terms, protein processing. No actin-binding, no cytoskeleton, no sarcomere, no
`GO:0005515`. The conflation that *did* occur is at term level (the MetAP-family term above).

Round 1 tried to close the remaining gap — GOA never records that the processed protein *is* actin —
with `GO:0030047 actin modification`, the term **NAA80 carries by IDA from three papers** for the
adjacent pathway step. **That was wrong and has been withdrawn** (see the round-2 section). The
substrate is now recorded with `RO:0002233` has_input extensions to ACTB/ACTG1 on both accepted IDA
rows, which states the same fact with no branch violation and no unwanted ancestry.

### Other checks, including the ones that came back negative

- **No IBA rows at all.** `DR PAN-GO; Q5BKX5; 0 GO annotations based on evolutionary models` —
  despite PTHR28631 being a single-subfamily family (1109 proteins, 845 proteomes, 2999 taxa) with
  one reviewed member per species and IDA-grade human evidence. A PAINT coverage gap, not an error.
- **MGI has never curated the orthologue:** mouse Actmap still carries root `ND` in **all three**
  aspects (`GO_REF:0000015`), so the knockout phenotype from the naming paper is unrepresented in GO.
- **Interaction corpus belongs to a catalytically dead isoform.** All 66 partners in UniProt's
  `CC INTERACTION` block, and **212 of 216** ACTMAP-side IntAct records, are against `Q5BKX5-3`,
  whose `VSP_039884` replaces residues 1–216 and deletes `Cys132`. IntAct's 220 records are 197
  yeast-host Y2H logged under three sub-method names (`two hybrid array` 70 + `prey pooling` 70 +
  `validated two hybrid` 67) — the ACRV1 "NbExp=3 is one screen" pattern. Profilin appears nowhere in
  IntAct. GO has imported none of it, which is right; the profilin annotation rests on the primary
  paper instead, and no part of the Y2H corpus is proposed.
- **The renaming did not split the literature.** PubMed: 3 records for `C19orf54`, 5 for `ACTMAP`,
  the 3 a subset of the 5. The old name survives only in resource metadata (PANTHER family name
  "UPF0692 PROTEIN C19ORF54", `BioMuta; C19orf54`).
- **Sibling cross-check.** Re-run after rebasing onto the `main` that now includes #2290/#2291/#2293:
  no merged review — `ACTR5`, `ACTR8`, `ACTB`, `ACTL7A/7B`, `ACTL8`, `ACTR1A/1B`, `ACTR10` — mentions
  ACTMAP, actin maturation, N-terminal processing or NAA80, and ACTMAP is the only review in the repo
  touching `GO:0004239`; GO:0030047 was likewise untouched before this PR and is untouched again after the round-2 withdrawal.
- **Reference hygiene.** No retractions (`CommentsCorrections` checked for all three PMIDs). The
  affinage record attributes its 2026 finding to "PMID:42159598, PMID:41757055" — the JACS paper and
  **its own bioRxiv preprint** (`UpdateOf`/`UpdateIn`), i.e. one study presented as two sources;
  flagged `LOW_QUALITY` with the preprint kept at `LOW` relevance for provenance only.

### Reproducibility

`genes/human/ACTMAP/ACTMAP-bioinformatics/analyze_actmap_go.py` (standard library only, no deps)
computes every number above from QuickGO and UniProt at run time, builds its source list **from the
GOA `WITH/FROM` field with an assertion** that the counts match token for token, fails loudly on a
missing input file, and fetches `size=2` on identifier lookups with an explicit
`primaryAccession == requested` liveness check. A fresh run was byte-compared against the committed
`RESULTS.md` and `results.json`: both reproduce exactly.

### One of my own claims was wrong, and it is now linted

My first pass wrote "4 gene products in 4 species", read straight off the taxon-restricted census.
The number was right *as scoped* and the sentence built on it was wrong, because the census taxa
exclude two family members. Corrected at all five assertion sites at once - `review.reason`,
`knowledge_gaps[0].resolution`, `suggested_questions[0].question`, the notes, and this PR body - by a
script that asserts each anchor is present before replacing, re-greps afterwards, and verifies that
exactly the intended leaf values changed and no YAML structure did.

Because that is the failure this campaign keeps repeating, the invariant is now committed:
`ACTMAP-bioinformatics/audit_actmap_claims.py` holds the retracted phrasings and the required claims
**with expected occurrence counts**, and `--self-test` breaks the files in a copy to prove every check
can fire (clean baseline silent; retracted phrasing reintroduced, required claim dropped, and guarded
file deleted all reported). Two defects were found by trying to break it rather than by reading it:
my hand-guessed occurrence counts were wrong (2 where the file has 3), and the first version flagged
the notes' own retraction *quote* as a violation.

### Round 2: the reviewer was right, and the decisive fact was not the one it used

`ai4c-reviewer` requested changes on `GO:0030047`: a covalent-modification term proposed for a
proteolytic event, on an *acetylation* precedent that does not transfer — and this file already
demonstrated the side-effect-free alternative, `RO:0002233` has_input extensions.

Checked before conceding, and the check went **against** me, further than the reviewer's own argument:
`GO:0016485 protein processing` is under `GO:0006508 proteolysis` and **not** under
`GO:0036211 protein modification process`, while `GO:0030047` is under `GO:0036211` and **not** under
`GO:0006508`. **The branches are disjoint**, so this is not a question of how loosely "covalent
modification" reads — GO's structure already says a proteolytic process does not live there. (The
term's only child, `GO:0007014 actin ubiquitination`, is additive too.) Same shape as the ABR lesson
that GO keeps the Rac and Rho *regulation* branches disjoint: the structure, not the label, decides.

So: annotation withdrawn, `GO:0030047` removed from `core_functions.directly_involved_in` and from
`cache/go/terms.csv`, has_input extensions added to the accepted `GO:0016485` row (the `GO:0070005`
row already had them), and the `suggested_questions` item rewritten so the **definitional** question
is primary and the placement question secondary. `process_branch_audit()` now computes the branch
flags, NAA80's three IDAs and the fly ortholog's FlyBase ISS rows, so the ancestry and precedent
claims are no longer the only uncomputed ones.

All four non-blocking suggestions taken:

- **The exopeptidase argument does not discriminate the two terms.** Correct — both are under
  `GO:0008238`. The row now says so and attributes that ground to the *term request*, leaving grounds
  1 (wrong released product / wrong EC) and 3 (family conflation) to carry the MODIFY.
- **Row 4's FlyBase-assigner explanation was unsourced.** Split into the verified part (FlyBase holds
  `GO:0070005` and `GO:0016485` ISS on fly `Q9VCE8`, WITH/FROM `UniProtKB:Q5BKX5`, now printed by the
  script) and the causal step, which is labelled a plausible inference rather than a fact.
- **`substrates` listed ACTA1 where the extensions listed only ACTB/ACTG1.** ACTA1 removed; the
  knowledge gap now records that the omission is deliberate.
- **The description stated the alpha-actin activity flatly.** Hedged in both descriptions: established
  for mouse, inferred for human by similarity.

The lint gained the round-2 retractions plus four **structural** invariants on the parsed YAML (no
`GO:0030047` row, none in `directly_involved_in`, has_input on both accepted IDA rows, substrates
limited to the two demonstrated actins), and two new self-test mutations that exercise them — both
report, and the clean baseline stays silent.

### Round 3: a duplicate YAML key, and a number I explained away

The reviewer found a real bug in the round-2 edit: adding provenance to the accepted `GO:0016485` row
produced **two `supported_by:` keys in the same mapping**, and PyYAML keeps the *last* — so the two
`RESULTS.md` entries were deleted before the validator, the renderer, the exporters, `checkquotes.py`
or my own harness could see them. Merged into one four-entry list; verified by the entry count on that
row (2 → 4) and by loading the whole file through a duplicate-key-rejecting loader.

**How it hid is the part worth recording.** The file holds **27** `reference_id: file:` lines while
the parsed check reported **25**, and the notes still carried the round-1 figure of **26**. I had
written "the count moved from 26 to 25 when the `GO:0030047` annotation was withdrawn" — true
arithmetic (that row carried one `file:` quote) and *wrong as an explanation*, because the same round
added two entries that should have made it 27. A coincidence made a rationalisation look like a
reconciliation, on the very number I had just edited. And it is the scope-blindness failure in a new
guise: **every quote gate here walks the parsed document, so none can see a quote parsing removed.**
The gate reporting clean was not weak, it was blind. Post-fix: raw 27 = parsed 27, `checkquotes`
40/40.

Two guards added: `audit_actmap_claims.py` now loads the review through a `_StrictLoader` that
**raises** on a duplicate mapping key, and separately cross-checks **raw against parsed** so a
discarded entry surfaces as an arithmetic mismatch even if the loader were bypassed. A seventh
self-test case reintroduces the exact duplicate key; both fire.

Whether a duplicate-key-rejecting loader belongs in the **shared** validator is deliberately left
open — `src/ai_gene_review/validation/` is shared infrastructure and out of scope for a gene PR — but
nothing in the repo would flag this today, so it is worth raising separately.

Both other suggestions taken: the `render()` disjointness sentence had its boolean interpolated but
its explanatory clause hardcoded, so it is now generated from the per-term flags with a warning
appended if the test ever fails; and `assert_complete_page()` now refuses any QuickGO response where
`len(results) != numberOfHits`, naming the fix, so `census()` and `family_wide_usage()` can no longer
report a total from `numberOfHits` beside a breakdown from a capped page. Verified against a synthetic
truncated response. The stale `26` in the notes is dropped rather than re-stated.

### Round 5: the guard its own self-test could not reach, and a count that conflated two keys

Both approval-round observations were right, and between them they were the last places the tooling
claimed more than it demonstrated.

**The self-test could not reach the raw-vs-parsed check.** Case 7 reintroduces the duplicate key, but
`_StrictLoader` raises and `check_structural()` returns *before* the arithmetic runs — so the one check
whose purpose is to survive a bypass of the loader had no case that fired it, in a file whose stated
standard is "prove every check can actually fire". And the notes recorded a verification "on a
synthetic document" that lived only in a throwaway command — not reproducible from the repo, the same
defect as the coverage sentence fixed a round earlier. Fixed by factoring out `_cross_check(raw, doc)`
and adding **case 8**, which calls it directly on a synthetic (raw, parsed) pair: silent on a matched
pair, reporting on a lossy one. Eight cases, all firing, baseline silent.

**The arithmetic covered only `file:` references**, leaving PMID and GO_REF ids outside it, so a
duplicate key discarding a PMID-only `supported_by` list would have been invisible to the very check
that exists to back the loader up. Now counts every `reference_id` on both sides.

**One measurement correction.** The review cites "46 `reference_id` lines", and a bare
`grep -c reference_id:` does report 46 — my own grep had agreed. But **7 of those are
`original_reference_id:`**, a different key naming the annotation's own GOA reference, carrying no
supporting_text and belonging to no list. The honest total is **39**, and raw 39 now equals parsed 39.
Insisting the numbers reconcile rather than adopting the quoted figure surfaced two real bugs in the
new code: the first regex omitted the `- ` list marker and so matched **nothing**, reporting a spurious
39-entry loss, and the same omission made the synthetic assertion fail. Both were visible only from
running the check on real input.

### Gates

Figures are as of the current head. `checkquotes.py` **40/40** clean; an additional **exact** (not
whitespace-normalised) check on all **27** `file:` quotes, to catch a UniProt quote crossing a
`CC       ` continuation — clean, and **raw 27 = parsed 27**, which is the reconciliation that the
round-3 duplicate-key bug broke. `just validate human ACTMAP` → `✓ Valid`, with one warning left
standing deliberately ("no annotation references the deep research file"): the affinage record's
substance is all traceable to PMIDs that are cited directly, and the standing rule forbids quoting a
provider sentence as `supporting_text` for a mechanistic claim. `cache/go/terms.csv`: **0 deletions**,
exactly **1** addition (`GO:0070005`) appended at EOF without re-sorting, the two pre-existing
duplicates unchanged. Note `just validate` silently **de-duplicated** main's `GO:0001675` and
`GO:0009566`; that unrelated change was reverted. The lint's **7** self-test cases all fire with a
silent clean baseline, and a fresh script run byte-reproduces the committed `RESULTS.md` and
`results.json`.

### Round 4: the three non-blocking items from the approving review

Two were accuracy defects in committed files rather than polish, so they were fixed rather than
deferred. **The truncation guard covered 2 of 6 `/annotation/search` call sites** while the notes
described it generally — an overclaim, and the uncovered `source_evidence()` is exactly what the
`CIRCULAR_PROPAGATION` disposition rests on; all six are now guarded and the notes name the single
deliberate exception (the `limit=1` probe that exists only to read `numberOfHits`, where the guard
would correctly abort). **The raw-vs-parsed check keyed the parsed side on `supporting_text` being
truthy** while the raw side counted every `reference_id: file:` line, so a future bare `file:`
provenance entry — legal in the schema — would have been reported as a discarded entry when nothing
was discarded; a false positive in a guard is worse than no guard, so the count is now keyed on
`reference_id` alone and verified on a synthetic document mixing both forms. The third item was these
Gates figures, now current.

---

> [!NOTE]
> **Low Risk**
> Adds documentation, gene-local Python audit scripts, and ontology cache data only; no changes to shared application runtime or auth paths.
> 
> **Overview**
> Introduces the **human ACTMAP (Q5BKX5)** annotation review package: curated sources (`ACTMAP-goa.tsv`, UniProt dump, Affinage notes), a large `ACTMAP-ai-review.yaml`, and reviewer notes documenting disposition of all six GOA rows.
> 
> **Curation conclusions:** Accept cytoplasm, `GO:0070005`, and protein processing; **modify** both `GO:0004239` (initiator methionyl aminopeptidase) rows to **`GO:0070005`** because ACTMAP releases N-acetyl-methionine, not free Met, and the term conflates METAPs with the ACTMAP family; **propose** `GO:0005522` profilin binding; record **ACTB/ACTG1** `has_input` extensions; withdraw a proposed `GO:0030047` actin modification row after showing proteolysis and modification sit in disjoint GO branches. A **`proposed_new_terms`** entry requests an omega-peptidase-style term for Nα-acetyl N-terminal release.
> 
> **Reproducibility:** `analyze_actmap_go.py` pulls QuickGO/UniProt, audits WITH/FROM circularity, peptidase-branch placement, and a **6/6** PTHR28631 family census; committed `RESULTS.md` / `results.json`. `audit_actmap_claims.py` lints cross-file numeric claims, rejects duplicate YAML keys, and cross-checks raw vs parsed `reference_id` counts.
> 
> **Other diffs:** PANTHER **PTHR28631** entries/metadata; **`cache/go/terms.csv`** gains **`GO:0070005`** (cysteine-type aminopeptidase activity).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b80829b2a6142902602ecf4af52d2ac38d047629. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — the blocking item was right, and checking it before conceding turned up a stronger fact than the argument you used.

**`GO:0030047` withdrawn.** You objected that it is a covalent-modification term proposed for a proteolytic event, on an acetylation precedent that does not transfer. I checked the branch structure, and it settles the matter without needing to interpret "covalent modification" at all:

| term | under `GO:0006508` proteolysis | under `GO:0036211` protein modification process | under `GO:0030036` |
|---|---|---|---|
| `GO:0016485` protein processing | **True** | **False** | False |
| `GO:0030047` actin modification | **False** | **True** | True |
| `GO:0007014` actin ubiquitination | False | True | True |

**The two branches are disjoint.** A proteolytic process is not in the modification branch by GO's own structure, so the term was unavailable to ACTMAP whatever the definition's wording permits — and the term's only child being ubiquitination (additive, like NAA80's acetyl transfer) is consistent with that reading rather than incidental to it. Same shape as the ABR lesson that GO keeps the Rac and Rho *regulation* branches disjoint: the structure decides, not the label.

So I took your first option rather than the widen-the-question one: annotation removed, `GO:0030047` dropped from `core_functions.directly_involved_in` and from `cache/go/terms.csv`, and `RO:0002233` has_input ACTB/ACTG1 added to the accepted `GO:0016485` row — the `GO:0070005` row already had them, so the substrate is now on both accepted IDA rows with no branch violation and no unwanted ancestry. I also kept a version of your second option, because the question is still worth asking for NAA80: the `suggested_questions` item is rewritten so the **definitional** question is primary (does GO intend "covalent modification" to include hydrolytic removal?) with the `GO:0030036` placement question secondary to it.

`process_branch_audit()` now computes the branch flags, NAA80's three IDAs and the fly ortholog's rows, so your point about those being the only uncomputed claims is addressed too.

**All four non-blocking items taken.**

- **The exopeptidase argument does not discriminate the two terms.** Correct, and it was a real defect in the argument: both `GO:0004239` and `GO:0070005` are under `GO:0008238`. The row now states that explicitly and attributes that ground to the *term request*, leaving ground 1 (releases `N-acetyl-L-methionine`, not methionine; `EC 3.4.11.-` not `3.4.11.18`) and ground 3 (the METAP-family conflation) to carry the MODIFY.
- **Row 4's FlyBase-assigner explanation was unsourced.** Split: the verified part is that FlyBase holds `GO:0070005` and `GO:0016485` ISS on fly `Q9VCE8` with WITH/FROM `UniProtKB:Q5BKX5` (now printed by the script); the causal step — that the human IDA was created to serve as that ISS source — is labelled a plausible inference, not a fact.
- **`substrates` listed ACTA1 where the extensions listed only ACTB/ACTG1.** ACTA1 removed; the knowledge gap now records that the omission is deliberate and why.
- **The description stated the alpha-actin activity flatly.** Hedged in both the top-level `description` and the core-function description: established for the mouse enzyme, inferred for human by similarity (`ECO:0000250|UniProtKB:J3QPC3`).

**One correction I found on myself** before your review landed, in the same PR: "4 gene products in 4 species" came straight off the five-taxon census, which cannot see the bovine and *Xenopus* members of PTHR28631. Querying each reviewed member directly gives **6 of 6** carrying `GO:0004239`. Fixed at all five assertion sites, and since this is the failure this campaign keeps repeating, the invariant is now committed rather than hand-checked: `audit_actmap_claims.py` holds the retracted phrasings, the required claims **with expected occurrence counts**, and four structural invariants on the parsed YAML, with `--self-test` breaking the files in a copy to prove each check fires. Writing it found three of its own defects, every one by trying to break it rather than by reading it: my hand-guessed occurrence counts were wrong twice, and the first version flagged the notes' own retraction *quote* as a violation.

Gates re-run after every change: `checkquotes.py` clean; an additional **exact** (not whitespace-normalised) check on all 25 `file:` quotes, which is what catches a UniProt quote crossing a `CC       ` continuation line — clean; `just validate human ACTMAP` → `✓ Valid`; `cache/go/terms.csv` 0 deletions, now exactly **1** addition (`GO:0070005`), duplicates unchanged; and a fresh script run byte-reproduces the committed `RESULTS.md` and `results.json`.

On the one warning left standing — "no annotation references the deep research file" — I'd rather leave it than satisfy it: everything substantive in the affinage record is traceable to PMIDs that are cited directly, and quoting a provider sentence as `supporting_text` is the thing the campaign rule forbids. Happy to reconsider if you'd prefer it cited for something non-mechanistic.

cmungall:
Fixed, and the diagnosis was better than the bug report — you reconstructed the arithmetic (27 raw − 2 dropped = 25) from the outside, which is how I should have read my own numbers.

**The duplicate `supported_by` is merged** into one four-entry list. Verified two ways: the entry count on that row goes 2 → 4, and the whole file now loads through a duplicate-key-rejecting loader without complaint. Post-fix `checkquotes` reports **40** quotes and the exact `file:` check reports **27**, matching the 27 `reference_id: file:` lines in the text.

**The part I want to record, because it is worse than the typo.** The notes said "the count moved from 26 to 25 when the `GO:0030047` annotation was withdrawn". That is arithmetically true — that row carried one `file:` quote — and it is a *wrong explanation*, because the same round also added two entries that should have taken the count to 27. A coincidence made a rationalisation look like a reconciliation, on the very number I had just edited. That is this campaign's own "do not rationalise a numeric discrepancy — investigate it" rule, failed on the number closest to hand.

It is also the scope-blindness failure in a new guise: **every quote gate in this repo walks the parsed document, so none of them can see a quote that parsing removed.** The gate reporting clean was not weak, it was structurally blind. Two guards rather than one, since a single check that a bypass defeats is not much:

1. `audit_actmap_claims.py` loads the review through a `_StrictLoader` that **raises** on a duplicate mapping key, instead of trusting a downstream consumer to notice.
2. It also cross-checks **raw against parsed** — `reference_id: file:` lines in the text versus entries surviving `safe_load` — so a discarded entry shows up as an arithmetic mismatch even if the strict loader were bypassed. A seventh self-test case reintroduces the exact duplicate key; both fire, and the clean baseline stays silent.

On your last point there: I've deliberately **not** touched `src/ai_gene_review/validation/`. A duplicate-key-rejecting loader in the shared validator looks right to me — nothing in the repo would flag this today, and the failure is silent by construction — but that is shared infrastructure and out of scope for a gene PR. Worth its own issue.

**Both 🔵 items taken.**

- **The hardcoded disjointness clause.** Correct, and it was the one place the narrative could have contradicted its own number. The clause is now generated from the per-term flags, so it reads `is under GO:0006508 and is NOT under GO:0036211` from the booleans, with an explicit "the disjointness test did NOT hold on this run" warning appended if the flag ever flips.
- **Pagination.** `assert_complete_page()` now refuses any QuickGO response where `len(results) != numberOfHits` and names the fix, applied in both `census()` (`limit=200`) and `family_wide_usage()` (`limit=100`). Tested by feeding it a synthetic `{numberOfHits: 500, results: [3 items]}`, which aborts. You're right that this is the same class as the error the PR corrected on itself — a denominator set by the query rather than by the biology — which is why it was worth doing even though nothing was truncated at 86 and 6 hits.
- The stale `26` is **dropped** rather than re-set, for the reason §12 already gives about restating counts.

Gates after all of the above: `checkquotes` 40/40 clean; exact `file:`-quote check 27/27 with raw = parsed; `just validate human ACTMAP` → `✓ Valid` (same single deep-research warning, still left standing); `cache/go/terms.csv` 0 deletions, net +1 (`GO:0070005`), the two pre-existing duplicates untouched; the lint's 7 self-test cases all fire with a silent clean baseline; and a fresh script run byte-reproduces the committed `RESULTS.md` and `results.json`.

cmungall:
All three taken. Two of them were accuracy defects in committed files rather than polish, which is why I pushed rather than deferring them past the approval.

**1. The truncation guard covered 2 of 6 call sites.** You're right that the notes described it generally — that is an overclaim in a committed file, and the uncovered `source_evidence()` is precisely what the `CIRCULAR_PROPAGATION` disposition rests on, so it was the worst one to leave bare. All six `/annotation/search` calls now go through `assert_complete_page()`: `source_evidence()`, the NAA80 `GO:0030047` probe, the fly-ortholog query and `curation_state()` join `census()` and `family_wide_usage()`. §15 now states the coverage instead of implying it, and names the single deliberate exception — the `limit=1` probe that exists only to read `numberOfHits`, where the guard would correctly abort. That one carries a comment saying so, so it doesn't read as an oversight.

**2. The raw-vs-parsed check would have raised a false positive.** Also right, and worth more than its label: a guard that cries wolf is worse than no guard, because it trains you to ignore it. `supporting_text` is optional in the schema, so keying the parsed side on it while the raw side counted every `reference_id: file:` line would have reported a legal bare `file:` provenance entry as a discarded one. The count is now keyed on `reference_id` alone (`_count_file_reference_ids`), and I verified it on a synthetic document holding one entry with a quote and one without — it returns 2, where the old form would have returned 1.

**3. Gates figures.** Patched to the current head: 40 quotes, 27 `file:` quotes with raw = parsed, `cache/go/terms.csv` net +1, 7 self-test cases. I also added the sentence that the figures are as of the head, since restating counts in prose is what went stale twice already.

Thanks for the note on `_no_duplicate_keys` being a plain function rather than a generator constructor — that eager-construction property is why a duplicate at any depth raises, and I had not reasoned about it explicitly when writing it. Worth stating in the docstring for whoever copies it next.

Gates re-run at this head: `checkquotes` 40/40; exact `file:` check 27/27 with raw = parsed; `just validate human ACTMAP` → `✓ Valid` (same single deep-research warning, still standing); `cache/go/terms.csv` 0 deletions / net +1 (`GO:0070005`) with the two pre-existing duplicates untouched; lint clean with all 7 self-test cases firing and a silent baseline; fresh script run byte-reproduces `RESULTS.md` and `results.json`.

Nothing outstanding from my side. The one item I have deliberately left alone is a duplicate-key-rejecting loader in `src/ai_gene_review/validation/` — I think it belongs there, but it is shared infrastructure and out of scope for a gene PR, so it wants its own issue.

cmungall:
Both taken, and pushing them turned up a measurement correction to the review.

**The self-test could not reach the cross-check.** You're exactly right, and it is the sharper of the two: case 7 reintroduces the duplicate key, `_StrictLoader` raises, `check_structural()` returns, and the arithmetic below it never runs — so the one check whose whole purpose is to survive a bypass of the loader had no case that fired it, in a file whose stated standard is "prove every check can actually fire". You also spotted that §16's "verified on a synthetic document" lived only in a throwaway command, i.e. not reproducible from the repo — the same defect as the §15 coverage sentence I'd fixed one round earlier, which is not a coincidence I like. Fixed as you suggested: `_cross_check(raw, doc)` is factored out, and **case 8** calls it directly on a synthetic (raw, parsed) pair — silent on a matched pair, reporting on a lossy one. Eight cases now, all firing, baseline silent.

**The arithmetic covered only `file:` ids.** Also right, and the framing is what convinced me: the arithmetic exists to *back up* the loader, so covering a subset of what the loader covers defeats the point. Now counts every `reference_id` on both sides.

**One correction to the premise, offered because it changed my code.** The comment cites 46 `reference_id` lines, and a bare `grep -c reference_id:` does report 46 — my own grep had agreed with yours. But **7 of those are `original_reference_id:`**, a different key: it names the annotation's own GOA reference, carries no `supporting_text`, and belongs to no list. The honest total is **39**, and raw 39 now equals parsed 39.

That mattered practically. Insisting the two numbers reconcile, instead of coding to the quoted 46, surfaced two real bugs in the new code within a minute: my first regex omitted the `- ` list marker and therefore matched **nothing**, reporting a spurious 39-entry loss on a clean file; and the same omission made the synthetic assertion in case 8 fail. Both were visible only from running the check against real input rather than reasoning about it — which is the third time in this PR that a number refusing to add up was the actual bug report.

Gates at this head: `checkquotes` 40/40; exact `file:`-quote check 27/27; raw 39 = parsed 39 on all reference_ids; `just validate human ACTMAP` → `✓ Valid` (same single deep-research warning, still standing deliberately); `cache/go/terms.csv` 0 deletions, net +1 (`GO:0070005`), the two pre-existing duplicates untouched; lint clean with 8 self-test cases firing; fresh script run byte-reproduces `RESULTS.md` and `results.json`.

I'm stopping here. The curation content has been settled since round 2 — MODIFY `GO:0004239` → `GO:0070005` with the `GO:0008242`-parented term request, `GO:0030047` withdrawn on computed branch disjointness, has_input on both accepted IDA rows, `GO:0005522` profilin binding, Ensembl rows kept with `CIRCULAR_PROPAGATION` recorded — and rounds 3-5 have been about the tooling that guards it. The one thing I've deliberately not done is add a duplicate-key-rejecting loader to `src/ai_gene_review/validation/`: I think it belongs there, but it is shared infrastructure and wants its own issue rather than a gene PR.

Issue #2294, opened 2026-07-26
cache/go/terms.csv: duplicate check needs to run on main, not only per-PR (inter-branch collisions are invisible to it)


cmungall:
## What happened

`GO:0031011 Ino80 complex` landed twice in `cache/go/terms.csv` on `main`:

| line | timestamp | added by |
|---|---|---|
| 3456 | `13:43:14` | #2291 (ACTR5), sorted position |
| 9070 | `13:40:43` | #2290 (ACTR8), appended at EOF |

ARP5 and ARP8 are both INO80 subunits, so **both reviews needed the same new term** for `core_functions.in_complex`. Fixed in #2293, but the interesting part is why neither PR caught it.

## Why the existing gate cannot catch this

The gate reviewers and agents run is

```
cut -d, -f1 cache/go/terms.csv | sort | uniq -d
```

Within *either* branch the term appears exactly **once**, so both PRs passed legitimately. The rows landed in **different positions** (one sorted, one at EOF), so git auto-merged both with no conflict. **The duplicate exists only in the merge result** — it is not observable from either branch.

This is distinct from the previously-documented failure mode (#2283, `GO:0043159`), which is *self-inflicted within one branch*: a row is hand-inserted, then a later `just validate` re-appends it at EOF because the appender cannot see the inserted row. That one a per-PR check does catch. This one it cannot, by construction.

## Suggested fix

Run the duplicate check **on `main`**, where the merge result exists — e.g. in the existing build/test workflow:

```bash
dupes=$(cut -d, -f1 cache/go/terms.csv | sort | uniq -d)
# main currently carries two pre-existing duplicates that are out of scope here:
expected=$'GO:0001675\nGO:0009566'
[ "$dupes" = "$expected" ] || { echo "unexpected duplicate curies in cache/go/terms.csv:"; comm -13 <(echo "$expected") <(echo "$dupes"); exit 1; }
```

Two notes:
- `main` already carries **two** pre-existing duplicates, `GO:0001675` and `GO:0009566`, so a bare "must be empty" assertion would fail immediately. Either allow-list them or clean them up first.
- The file is **not globally sorted** (rows are appended at EOF by several branches), so any fix must not re-sort — re-sorting rewrites unrelated lines and manufactures spurious deletions in every subsequent PR diff.

## Why it will recur

[...]

PR #2298, opened 2026-07-26
Review human ACTL10; Swiss-Prot sequence starts mid-actin-fold

cmungall:
## ACTL10 (Q5JWF8) — a dark divergent actin, and a Swiss-Prot sequence that starts mid-fold

ACTL10 has **three GOA rows** — two IBA and one IEA derived from one of them — and **no experimental annotation of any kind, in any aspect**. UniProt has no FUNCTION, no SUBCELLULAR LOCATION and no SUBUNIT line; its only `CC` line is a bare family-membership statement with `ECO:0000305` (curator inference). Pharos calls it `Tdark`. IntAct holds zero interactions.

### Actions

| row | evidence | action |
|---|---|---|
| `GO:0015629` actin cytoskeleton | IBA, node PTN002631484 | **KEEP_AS_NON_CORE** |
| `GO:0005200` structural constituent of cytoskeleton | IBA, node PTN000940351 | **MARK_AS_OVER_ANNOTATED** |
| `GO:0007010` cytoskeleton organization | IEA from `GO:0005200` | **MARK_AS_OVER_ANNOTATED** |

No REMOVE. `core_functions` is deliberately empty, and the resulting validator warning is left standing rather than silenced with invented content.

### Central finding: Q5JWF8 begins in the middle of the actin fold

Human ACTL10 is annotated at **245 aa**; mouse is 346 and several mammals 366–368. Two results:

1. **The length variation is not phylogenetic.** Across 87 mammalian ACTL10 entries, **4 of 4** sister-taxon pairs disagree — *Sapajus apella* 368 vs *Cebus imitator* 245 (sister genera), *Marmota* 346 vs *Sciurus* 245, *Urocitellus* 346 vs *Ictidomys* 245.
2. **The human genome encodes the missing region, in frame.** The MANE transcript is **single-exon** with 555 nt of leader contiguous with the CDS. Translating that leader in the CDS frame — with the frame *proven* by first asserting the CDS translates to the Swiss-Prot sequence — gives **167 uninterrupted codons** running into the annotated initiator, and their translation is unmistakably actin, including **phosphate-binding loop 1**.

Extending the sequence raises the global alignment score against *Sapajus* ACTL10 from **1117 to 1835** at 96.5% identity (mouse 868→1383; elephant 905→1594). And the **Met1 of both *Sapajus* and *Callithrix* ACTL10 aligns to a human CTG codon at chr20:33667129**, not an ATG — loss of the ancestral initiator with the reading frame still intact.

**Scoped deliberately:** this establishes that Q5JWF8 is not the whole of ACTL10's actin homology. It does **not** establish which product the cell makes — a lost initiator with a conserved downstream frame is compatible with a genuinely shortened protein *or* with non-AUG initiation / an unannotated upstream exon. That needs N-terminal proteomics. No length is asserted for the human protein.

### This corrects a merged sibling analysis, and the correction flips the sign

The committed ACTL8 panel already included ACTL10 and ranked it **last of fourteen** at the filament interface (5/38 compatible, below Arp3). My script **asserts that it reproduces ACTL8's committed numbers before drawing any conclusion** — it reproduces all three shared rows exactly (ACTB 37/1/0/0, ACTL8 8/3/24/3, ACTL10 3/2/13/20) — then splits the single `gap` column by cause:

| ACTL10 sequence | id/cons/non-cons/int-gap/**outside** | present | compatible |
|---|---|---|---|
| Q5JWF8, 245 aa as annotated | 3/2/13/0/**20** | 18 | **5/18** |
| extended ORF, 412 aa | 7/4/23/4/**0** | 38 | **11/38** |

**20 of the 38 "interface" positions were never substitutions — they are positions the annotated sequence does not reach.** Repaired, ACTL10 scores **11/38, exactly ACTL8's own 11/38**, with *Sapajus* ACTL10 independently giving 11/38 too. ACTL10 is an ordinary member of the divergent-actin band, not an outlier below Arp3.

The nucleotide site is sharper still: the five positions Q5JWF8 fails to reach are **exactly phosphate-binding loop 1**, and the extended frame retains all five (4 identical + 1 conservative) for **15/19 overall — better than ACTL8 (14/19) and ACTL7A (14/19)**. The extended human ORF and *Sapajus* ACTL10 give identical calls at all 19 positions.

So the "fold without function" reading is **refused by measurement**, which is why no residue-decay argument is used and no nucleotide term is proposed either (residues being present is a possibility, not a measurement).

### Why REMOVE was withheld

`MARK_AS_OVER_ANNOTATED` needs no positive argument; REMOVE does. The interface degradation is real (11/38 against β-actin 38/38 and Arp53D 33/38, a divergent actin that *does* polymerise; stable under a second matrix and gap model). But **this review's own control refuses REMOVE: human Arp3 scores 8/38, below ACTL10, and Arp3 does make actin-like protomer contacts** at an Arp2/3 branch. The metric therefore bounds *canonical two-stranded filament incorporation*, not all structural roles.

### Sibling consistency — all three rows are byte-identical to merged reviews that disagree

`GO:0015629`↔ACTL8 (KEEP_AS_NON_CORE), `GO:0005200`↔ACTR10 (ACCEPT), `GO:0007010`↔ACTL7A (MODIFY) / ACTL7B (REMOVE) / ACTR10 (KEEP_AS_NON_CORE). Checked programmatically. Each divergence is resolved on stated gene-specific grounds:

- **ACTR10's ACCEPT** rests on bona fide dynactin membership — ACTL10 has no complex, location, phenotype or biochemistry.
- **ACTL7A/7B's REMOVE** rests on a **TAS** row citing a paper with no functional data *plus* an explicit PAINT IRD negation at their node — ACTL10's is a live IBA and PAINT has not negated its branch.
- **ACTL8's KEEP_AS_NON_CORE** rests on the node being the correct **LCA** of a heterogeneous clade, which is a property of the node; ACTL10 sits on the *same* node with the *same* 25 tokens, so that one does transfer.

### PAINT finding (one node-level fix would correct five genes)

`GO:0005200` is asserted by **IBD at exactly one node** (PTN000940351) and **negated by IRD at eight descendant nodes** — Arp2, Arp3, Arp5, Arp6, Arp8, ACTL6A/B, ACTR1A/B, ACTL7A/7B — with the parent `GO:0005198` substituted at the ACTL7 node. QuickGO shows the ten human genes still receiving it are the four conventional muscle actins **plus ACTL9, ACTL10, ACTR10, ACTRT1, ACTRT2, ACTRT3**. Only ACTR10 has independent evidence (dynactin). Extending the negation, or dropping to `GO:0005198`, would fix five genes at once.

The merged **ACTR8** review reached the same 8-plus-1 tally independently and asked for the general extension; this PR supplies the concrete target list and credits it rather than restating it. Also filed: Arp2 and Arp3 **seed** the ancestral IBD (each with its own IDA) while the term is IRD-negated at their own clade nodes — not strictly contradictory, but worth confirming as deliberate.

### WITH/FROM resolution

All 36 tokens resolved. `GO:0015629`: 24/24 protein donors carry their own experimental evidence across 12 organisms. `GO:0005200`: **10/10 do**. So `SOURCE_WEAK_OR_INFERRED`/`SOURCE_EVIDENCE_WEAK` would be contradicted by the analysis — root cause is `PROPAGATION_BAD`. Two donors resolve only to TrEMBL entries; reported rather than hidden. Resolver gotcha fixed: WormBase tokens need bare `xref:WBGene…`; the `xref:wormbase-…` form returns zero hits, which reads as "no such source" rather than "wrong query".

### Affinage was empty — and that is not evidence

`gates_passed: True`, `n_discoveries: 0`, `citation_count: 0`. An independent Europe PMC search found **PMID:32742462, which has ACTL10 in its title** (a TCGA survival correlation; no functional data — its authors state the gene is essentially unstudied) and **PMID:35180326** (mouse Actl10 testis-enrichment; states its mutants "have not been linked to germline biology", which is why the `GO:0007010` row is *not* modified to `GO:0030036` as ACTL7A's was). All PMIDs retraction-checked.

### Reproducibility

`analyze_actl10.py` computes everything at run time; a fresh run was verified by `diff` to reproduce both `RESULTS.md` and `results.json` byte-for-byte. The ACTL8 cross-check guard was **break-tested two ways** (perturbed cutoff → aborts; missing expectation row → aborts rather than passing vacuously). `check_source_entities.py` asserts the review's `source_entities` equal the GOA WITH/FROM field in both directions, with a 4-mutation self-test that asserts each mutation target exists first.

Gates: checkquotes 20/20 verbatim, 0 problems · `just validate human ACTL10` ✓ Valid (2 genuine warnings) · terms.csv unchanged (all terms already cached), no deletions, duplicates identical to main's pre-existing `GO:0001675`/`GO:0009566`.

---

## Three late checks (second commit), run against primary sources rather than inherited

**1. Reference-projection check.** ACTL10 has **no PMID-backed GOA rows**, so the "one paper, N entities" pattern has no literal form here. The node-level analogue does apply, and both IBA rows **share the reference `GO_REF:0000033`**: `GO:0015629` projects from PTN002631484 to **18** human genes and `GO:0005200` from PTN000940351 to **10**, each with identical evidence. The two rows are not independent of each other either — **6 of the 10** `GO:0005200` protein donors also sit on the `GO:0015629` row.

New finding from asking what each donor *contributes*: the **4 donors unique** to the `GO:0005200` row — the only ones making it more specific than a generic actin-cytoskeleton call — are yeast **Arp1**, yeast **Arp10**, human **Arp2** and human **Arp3**. Arp2 (`P61160`) and Arp3 (`P61158`) are seeds **at** PTN000233596 and PTN000233796, two of the eight nodes carrying the `GO:0005200` IRD negation; yeast Arp1 is a dynactin filament subunit and the centractin clade node PTN007551901 also carries an IRD. **The row's specificity rests largely on donors from clades GO's own pipeline excludes from the term they donate** — a substantive strengthening of `PROPAGATION_BAD`.

**2. Publisher-Correction check.** Read `CommentsCorrections/RefType` from the efetch XML for both load-bearing PMIDs: **32742462 and 35180326 have no `CommentsCorrections` element at all** (no correction, erratum or retraction). ACTL10's affinage record has `citation_count: 0` and cites nothing, so it cannot share ACTL8's retracted `PMID:32125225`; verified by grep, not assumed.

**3. The ACTL8 census, re-derived from QuickGO rather than read off ACTL8's review.**

| gene | IBA rows | under PTN002631586 / PTN007551913? |
|---|---|---|
| ACTL7A | 3 | no |
| ACTL7B | 3 | no |
| **ACTL8** | **11** | **yes, both** |
| ACTL9 | 2 | no |
| **ACTL10** | **2** | **no** |
| ACTRT1 | 5 | no |
| ACTRT2 | 2 | no |
| ACTRT3 | 2 | no |

ACTL8's finding is **confirmed** (it alone is under either narrow node; 11 rows vs a median of 2), and PTN002631484's 18-member human set matches its claim exactly. It **does not extend to ACTL10**, which sits at the family median and under neither narrow node. Clean negative, recorded rather than dropped: **ACTL10 is correctly placed in the tree — its defect is the mis-annotated protein sequence, not mis-placement.**

**terms.csv / issue #2294:** this branch does not touch `cache/go/terms.csv` at all (all five terms it uses were already cached), so it cannot collide with ACTMAP #2295 / ACTRT3 #2296 / ACTRT2 #2297. Both gate directions were re-run after the final `just validate`: no deletions, duplicates exactly `GO:0001675` / `GO:0009566` as on `main`.

No action changes from these checks.

---

> [!NOTE]
> **Low Risk**
> Documentation and offline analysis only; no runtime services or auth changes. Large YAML and scripts are review artifacts with self-tests, not production code paths.
> 
> **Overview**
> Adds a **complete human ACTL10 (Q5JWF8) annotation review** for a testis-enriched, largely uncharacterized actin-family gene, with **no new core functions** proposed.
> 
> **GO verdicts:** `GO:0015629` actin cytoskeleton → **KEEP_AS_NON_CORE**; `GO:0005200` structural constituent of cytoskeleton → **MARK_AS_OVER_ANNOTATED**; `GO:0007010` cytoskeleton organization (IEA from `GO:0005200`) → **MARK_AS_OVER_ANNOTATED**. Reviews explain divergences from merged sibling genes (ACTL8, ACTR10, ACTL7A/B, ACTRT3) on gene-specific grounds.
> 
> **Central sequence finding:** Swiss-Prot **245 aa starts mid–actin-fold**; the MANE single-exon leader encodes **167 in-frame upstream codons** (including phosphate-binding loop 1) absent from the annotated chain—framed as an annotation/data-quality issue, not proof of the cellular protein length.
> 
> **New bioinformatics:** `analyze_actl10.py` regenerates `RESULTS.md` / `results.json` from UniProt, Ensembl, PDB, QuickGO, and cached PAINT; separates **outside-span vs substitution** at actin nucleotide and filament interfaces (correcting prior ACTL8-panel interpretation); audits IBA WITH/FROM donors and **GO:0005200** IBD vs eight IRD negations. Companion lints **`check_source_entities.py`** (review `source_entities` vs GOA) and **`check_claim_scoping.py`** (P-loop claims must name which sequence).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9286901485f8a12b7bf760160598ddf85ba70bde. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — all three suggestions were real and all three are addressed in `c2f89fa9a`. No action changes.

**1. P-loop-1 claims scoped.** You were right that this cuts the wrong way for the entry GO and UniProt actually publish: Q5JWF8 as annotated is ABSENT at all five of those positions per my own `RESULTS.md`, so an unscoped "ACTL10 conserves phosphate-binding loop 1" is false of it. Four sites said it unscoped — `description`, the `GO:0015629` summary, the `GO:0005200` reason, and one in the notes — and all four now name the sequence.

Since this is one claim asserted at a dozen sites, I made the invariant mechanical rather than a hand-checked list: `check_claim_scoping.py` requires every P-loop-1 mention in the review and notes to sit within 300 characters of a phrase naming which sequence is meant. It inspects 13 and **fails loudly if it finds zero**, so it cannot pass vacuously.

Two bugs in that lint are worth reporting because they are both instructive:

- **My first draft was defeated by YAML line-wrapping.** Folded scalars store `does not reach` as `does not\n      reach`, so literal matching flagged three *correctly scoped* sites. The tempting fix was to widen the regex until they passed — which would have been rationalising a discrepancy rather than investigating it, and would have shipped a lint that emits false positives. The actual fix is to whitespace-normalise before matching.
- **The self-test's own mutations reproduced that same bug.** They asserted on normalised text while substituting on raw text, so the phrases they meant to strip did not match and the mutation silently did nothing. The harness reported FAILURE rather than passing, which is the right behaviour, but the fault was the mutation, not the lint. Fixed by mutating normalised text and asserting `detected == changed`. Two earlier mutations also deleted only one of *two* scoping phrases in a passage, so the claim stayed scoped and the lint was correct to keep passing — a mutation has to remove **all** scope from a claim's window, or add a claim that has none.

**2. `is_active_in` now discussed.** Agreed, and it is a genuine over-assertion rather than a stylistic quibble: `is_active_in` asserts the protein *carries out its molecular function* in the actin cytoskeleton, and no molecular function has been measured for ACTL10 at all, so the qualifier presupposes precisely what is unknown. `located_in` would say what the phylogenetic inference actually supports.

I have **not** changed the term or the action. The qualifier is correct for the conventional-actin donors, so this is a transfer artefact of the same kind as the term's breadth, not a curator error — and a qualifier swap is a GOA-side change rather than a term-level one, which is not what `MODIFY` expresses. It is recorded in the row and raised for PAINT alongside the node recommendation, noting that the two rows differ (`enables` on `GO:0005200`, `is_active_in` on `GO:0015629`) and that both over-assert on this target.

**3. Page-size saturation asserted.** `orthologue_lengths` now reads the total from the **`x-total-results` header** instead of `len(results)`, aborts if a full page comes back, and asserts parsed count equals declared count. Break-tested by setting `page_size = 10`, which aborts with *"returned a full page (10) ... (87 declared)"*. A fresh run reproduces `RESULTS.md` and `results.json` byte-for-byte, so the guard changed no number.

**Gates re-run:** checkquotes 20/20 verbatim · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · `cache/go/terms.csv` still untouched by this branch, no deletions, duplicates only `main`'s pre-existing `GO:0001675`/`GO:0009566`.

Note the previous approval predates this head commit, so it does not cover it.

cmungall:
All three nits fixed in `58b4ef8e9`. One of them was more than cosmetic, so worth flagging:

**The slice bug would have failed exactly when it mattered.** For a claim inside the first 60 characters of a file, `rel - 60` went negative and Python sliced from the *end* of the window — so the FAIL message would have been garbled at the only moment the lint ever prints one. Rewritten with the offset clamped, and I exercised it directly against a synthetic file whose claim sits at offset 0; the message now begins at the claim rather than in the trailing filler. Good catch.

**`r` → `resp`.** You're right that it is harmless today only because `declared` is read before the loop — it is a trap for the next edit, so the response object is now `resp`. A fresh run reproduces `RESULTS.md` and `results.json` byte-for-byte, so no number moved.

**The "12 claims" count.** Rather than change 12 to 13, I removed the number. A figure in prose that has to track a script's output drifts the moment the text changes, which is precisely how the 12 became wrong — I added a claim to the notes in the same round that introduced the sentence. The paragraph now says to run the lint for the current figure and records why it does not quote one. (Same reasoning as building `source_entities` from the GOA field rather than by hand.)

Gates: checkquotes 20/20 · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · `cache/go/terms.csv` untouched by this branch, no deletions, duplicates only `main`'s pre-existing pair · both lints pass with self-tests (3 and 4 mutations caught).

This head commit postdates your approval of `c2f89fa9a`, so it isn't covered by it.

cmungall:
Applied in `e647c7beb`. Only `analyze_actl10.py` changes; the curation surface is untouched and `RESULTS.md` / `results.json` reproduce byte-for-byte, so no number moved.

New `quickgo_all()` helper requires `numberOfHits` to be present, asserts the returned row count **equals** it, and additionally aborts on a full page. Both call sites converted — `quickgo_evidence` and `go0005200_census` — with an assertion that no unguarded QuickGO call remains, so a future third call site can't quietly skip it.

Your measurement of the headroom is what made the priority clear, and I agree with your ranking: **the census is the one that mattered.** The review states that *exactly ten* human genes still receive `GO:0005200` from PTN000940351 and names them, so a short page would silently understate a load-bearing enumeration — and an under-count there would have made the PAINT recommendation look narrower than it is. Whereas truncation in `quickgo_evidence` could only ever produce a false negative on `has_experimental`, and every donor returned true, so the 24/24 and 10/10 WITH/FROM results and the `PROPAGATION_BAD` diagnosis were never at risk from paging. Guarding both anyway is the cheaper choice than remembering which one is safe.

Break-tested by forcing `limit=5` on the census: aborts with *"QuickGO truncated the human GO:0005200 IBA census: 5 of 43 rows returned at limit=5; add pagination before trusting this count"*.

Gates: checkquotes 20/20 · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · `cache/go/terms.csv` untouched, no deletions, duplicates only `main`'s pre-existing pair · both lints pass with self-tests.

I have nothing further queued, so this is my final commit unless review raises something.

cmungall:
Fixed in `a7c83526b`. Only `analyze_actl10.py` changes; `RESULTS.md` and `results.json` reproduce byte-for-byte, so the curation surface is untouched.

**1. You were right, and the failure was the interesting kind.** The assertion existed only in the throwaway script that made the edit, so the commit message stated something true of the *moment* and not of the *artefact* — and it did not close the future-edit case it claimed to close. That is the same shape as the "fixed in N places, landed in N−1" pattern this branch has been guarding against, committed in a commit message instead of code. It is now `assert_quickgo_calls_are_guarded()`, called from `main()`: the endpoint URL must appear literally exactly once, and the single fetch through the constant must sit lexically inside `quickgo_all`.

Worth reporting how it went wrong first: **my textual version matched its own error-message strings** and reported four call sites where there is one — a checker whose pattern matches its own diagnostics. Rewritten over the AST, which sees code only. Break-tested four ways, each aborting correctly: a second literal URL; an extra guarded fetch outside the guard; renaming `quickgo_all` away; and moving the fetch into a helper outside the guard.

**2. The `limit` KeyError** now raises with a message saying that without an explicit limit the page size is the server default and coverage cannot be verified.

**3. On `total == 0` I partly disagree,** and implemented the distinction rather than the blanket rule — flagging it since you asked for consistency with the line-141 convention.

Those two zeros do not mean the same thing. For the **census**, zero rows means the query broke, and it now aborts. For a **per-donor evidence lookup**, a donor carrying no annotation for the term it donated is a *genuine and interesting finding* — it would weaken the propagation and belongs in the report — so aborting would destroy the finding rather than surface it. That is the same reasoning that keeps an ambiguous cross-reference as data rather than as a missing input. So there is an explicit `allow_zero`, default `False` (strict), `True` only at the `quickgo_evidence` call site, with the reason stated at both. Both branches verified against a real zero-hit query: strict aborts, permissive returns `numberOfHits = 0`.

**One more, self-inflicted, since this branch has held itself to "a stated verification is one that actually ran":** my own shell harness lied twice. Piping python into `tail` and chaining with `&&` tests *tail's* exit status, so an aborted run printed "RUN SUCCEEDED" and the idempotency diff that followed was vacuous — it compared two unchanged files because the script had died before writing. Both times the real error was visible one line above. Now capturing the exit code and skipping the diff when the run failed. Recording it because it is exactly the class of thing the reviewer cannot see from the diff.

Gates: checkquotes 20/20 · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · `cache/go/terms.csv` untouched, no deletions, duplicates only `main`'s pre-existing pair · both lints pass with self-tests (3 and 4 mutations).

cmungall:
Both taken, in `28cabdfce`. `RESULTS.md` is unchanged; `results.json` gains only a new always-null field.

**1. You found the escape route already sitting in the file, which is the part that makes this worth fixing rather than noting.** My check required `get_json(QUICKGO_SEARCH, ...)` inside `quickgo_all` — but `SESSION.get(url, params=...)` is right there for the `x-total-results` read, so a future `SESSION.get(QUICKGO_SEARCH, ...)` would have satisfied the check while skipping every coverage assertion. Guarding the constant's **name** instead of one call shape is shorter and closes `get_json`, `SESSION.get`, the keyword-first-arg form and local aliasing in a single rule, as you said.

It now also **asserts presence, not just absence of strays** — a guard defeatable by deleting the thing it guards is worse than none. Break-tested on all four routes, each aborting:

| break | result |
|---|---|
| `SESSION.get(QUICKGO_SEARCH, ...)` | aborts — read outside `quickgo_all` |
| `get_json(url=QUICKGO_SEARCH, ...)` | aborts — same |
| alias to a local, then `get_json(u, ...)` | aborts — same |
| delete the fetch from `quickgo_all` | aborts — "guards nothing" |

**2. The two zeros really were conflated, and it feeds a load-bearing number.** `n_with_own_experimental` is the 24/24 and 10/10 behind `PROPAGATION_BAD`, and "QuickGO does not hold this accession" is the dead-accession trap — it would have read as a property of the donor. `quickgo_evidence` now re-queries once per zero without the `goId` filter and records `accession_absent_from_quickgo`.

It is `null` throughout the committed run, because all 34 own-evidence blocks have `n >= 1` — matching your measurement. So nothing was affected today; the field's value is that the distinction is now *checked* rather than assumed, and a future reviewer can see which of the two a zero was.

**3. The three smaller notes,** all accepted: the coverage test is now `got != total`, so an over-count cannot pass silently and the assertion matches how the earlier commit message described it; both truncation messages now say to raise the limit and re-run, making the deliberate conservative abort actionable rather than just loud; and the guard's reference count is printed (`2 QUICKGO_SEARCH read(s)`), giving it the same non-vacuity signal as `check_claim_scoping`'s "inspected N".

Gates: checkquotes 20/20 · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · `cache/go/terms.csv` untouched, no deletions, duplicates only `main`'s pre-existing pair · both lints pass with self-tests.

Agreed that the shell-harness failure is the most transferable item — a vacuous byte-for-byte check being indistinguishable from a passing one is a good way to put it, and it is the same shape as the `total == 0` conflation above: two different states rendered as one value.

cmungall:
Real bug, correctly diagnosed, fixed in `de66a4f7e`. `RESULTS.md` and `results.json` are byte-identical.

You're right that the three cases collapsed to two, and right about which claim it broke. The probe went through the normal path with `limit=1`, so:

| donor's total `N` | `got` | `total` | outcome |
|---|---|---|---|
| `N > 1` | 1 | `N` | `got != total` → abort |
| `N == 1` | 1 | 1 | `got >= limit` → abort |
| `N == 0` | 0 | 0 | `absent = True` |

So the accession-present-but-lacking-the-term case — the genuine finding the field was added to record, the one that would weaken a propagation — took the run down, and with a message about a phantom API inconsistency rather than about the probe's own `limit=1`. And my commit message said the field's value was that "the distinction is now **checked** rather than assumed" when the code did not check it. That is the same shape as the assertion that lived in a throwaway script two commits ago, so it is worth naming plainly rather than filing as cosmetic.

**`count_only` as you suggested.** I'd add that it is not a loosening: for a caller that reads no rows, completeness of the row set is *inapplicable* rather than merely inconvenient — there is no row set it depends on. The zero check still applies, because the count is precisely what the caller wants. The call stays inside `quickgo_all`, so `assert_quickgo_calls_are_guarded()` passes unchanged at `n=2`.

**Break-tested live, which is what shows the branch is reachable rather than merely written:**

| case | query | result |
|---|---|---|
| A. held, but not for this term | `P60709` × `GO:0004672` | `n=0`, `absent=False` — **previously an abort** |
| B. accession absent entirely | `O15507` × `GO:0004672` | `n=0`, `absent=True` |
| C. real evidence | `P60709` × `GO:0005200` | `n=7`, `absent=None`, probe never fires |

Case B deliberately uses `O15507` — the inactive entry from ACTR10's dead-accession finding — so the trap that motivated the field is the thing the test exercises.

**Both cosmetic notes applied.** (a) The non-vacuity line now says "or the guard itself", matching what is asserted, since one of the two counted reads is in the self-check — you're right that a printed number should say what it counts. (b) The guard now matches `AsyncFunctionDef` and **rejects** a shadowed definition rather than keeping the last: containment is by line range, so two definitions would make the allowed range depend on dict ordering. Neither can occur here, but as you say the guard is the thing other reviews will copy, so it shouldn't carry the trap. Both break-tested (shadowed def and an `async def` reading the constant outside).

Noted for the record on your two remaining observations: the perimeter really is *the URL as one string literal plus the name of the constant*, so a future `QUICKGO_BASE + "/annotation/search"` would evade it — past the point of diminishing returns, but now on the record. And I agree the probe was a third instance of two-states-rendered-as-one-value; that is three in a row, which suggests the pattern is worth looking for deliberately rather than catching case by case.

Gates: checkquotes 20/20 · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · `cache/go/terms.csv` untouched, no deletions, duplicates only `main`'s pre-existing pair · both lints pass with self-tests.

cmungall:
`8a98adef5`. Two things, one of which came from re-running the terms.csv gate against a `main` that had moved again.

## ACTRT3 merged mid-review with the same row and a different action

**#2296** landed titled *"a GO:0005200 that PAINT has already rejected next door"*, carrying the **same** `GO:0005200` IBA row from the same node, resolved **MODIFY → `GO:0005198`** where this review has `MARK_AS_OVER_ANNOTATED`. Two independently reviewed members of the residual six, one row, two actions — the AADACL trio hazard, so it is reconciled in the row rather than left for a reader to notice.

**The diagnosis is now identical:** both give `root_cause: PROPAGATION_BAD` and `FUNCTIONAL_DIVERGENCE`, and I have added `GRANULARITY_MISMATCH` to match ACTRT3, because the term being more specific than what transfers is a property of *the row*, not of the remedy chosen for it.

**Only the action differs, on evidence ACTRT3's own reason names.** It states that `GO:0005198` "is supported for ACTRT3 — it is a component of the perinuclear theca's protein scaffold and contributes to its structural integrity", so generalising moves its row from an unsupported specific claim onto a supported general one. ACTL10 has no counterpart: no complex, no assembly, no location, no phenotype, no partner. Since `GO:0005198` is defined by contribution to the structural integrity of a complex or assembly, substituting it here would trade one unsupported assertion for a slightly less specific unsupported one — and because **neither gene already carries `GO:0005198`**, it would *newly add* a molecular function to a gene that has none. Wrong direction for a Tdark protein.

ACTRT3 performs the same discrimination in the other direction against ACTR10's ACCEPT ("ACTR10 has an ortholog-strength donor in the seed set and ACTRT3 has none"). So this row now carries **three actions across four merged reviews** — ACCEPT (ACTR10), MODIFY (ACTRT3), MARK_AS_OVER_ANNOTATED (ACTL10), REMOVE (ACTL7A/7B, different evidence code) — but unlike the AADACL trio each is pinned to a stated checkable gene-specific fact rather than to a differing judgement about identical evidence. Not uniformity, a reason per gene.

## Your two remaining notes

**(1) `count_only` now returns the count alone.** You were right that documenting "reads no rows" while handing back an unverified `results` array guards the shape rather than the invariant — the same hole the name-based guard closes one level up. The response is now exactly `{"numberOfHits": N}`; verified it has one key, and all three probe branches still behave (`absent` False / True / None). **(2)** The missing-`limit` message no longer asserts a reason that does not hold for a `count_only` caller, where an unset limit costs bandwidth rather than coverage.

## terms.csv

The gate flagged `-GO:0070005` on this pass, and it is not a deletion: `origin/main` advanced to `29b1705cc` and **ACTMAP #2295 added that row**. Verified against the merge base rather than assumed — this branch's own `base..HEAD` delta to `cache/go/terms.csv` is empty, as is worktree-vs-HEAD, and duplicates in the merged result are still only `main`'s pre-existing `GO:0001675`/`GO:0009566`. Second time the moved-target diff has read as a deletion on this branch, which is worth knowing given three siblings were in flight.

`RESULTS.md` and `results.json` remain byte-identical. Gates: checkquotes 20/20 · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · both lints pass with self-tests.

cmungall:
Accepted in full, `928690148`. I verified every premise against the schema and the four merged reviews rather than against the comment, and each one held. Action unchanged; only the coded fields and one sentence move.

## (b) The coded diagnosis contradicted its own prose — and my own argument

This is the item worth dwelling on, because you're right that it is the substantive half. The row carried `source_status: SUPPORTS_SOURCE_BUT_NOT_TARGET`, whose definition is *"Source evidence supports the source annotation…"* — presupposing source evidence, while the comment on that very entity says the source is a GO term carrying none. And `SOURCE_EVIDENCE_WEAK` contradicts **this review's own finding one row above**, that every donor carries its own experimental annotation. I criticised exactly that inconsistency in the `GO:0005200` row and then committed it here.

Verified the sibling convention independently rather than taking it on report:

| review | action | root_cause | modes | source_status |
|---|---|---|---|---|
| ACTR10 | KEEP_AS_NON_CORE | NO_FAILURE_NON_CORE | – | `CIRCULAR_OR_REDUNDANT` |
| ACTRT3 | KEEP_AS_NON_CORE | `EVIDENCE_CIRCULAR_OR_REDUNDANT` | `CIRCULAR_PROPAGATION` | `CIRCULAR_OR_REDUNDANT` |
| ACTL7A | MODIFY | TERM_SCOPING_PROBLEM | GRANULARITY_MISMATCH, SOURCE_EVIDENCE_WEAK | SOURCE_WEAK_OR_INFERRED |
| ACTL7B | REMOVE | SOURCE_WEAK_OR_INFERRED | SOURCE_EVIDENCE_WEAK, FUNCTIONAL_DIVERGENCE | SOURCE_WEAK_OR_INFERRED |

ACTL10 was the only one of four using `SUPPORTS_SOURCE_BUT_NOT_TARGET`. Now `EVIDENCE_CIRCULAR_OR_REDUNDANT` / `CIRCULAR_PROPAGATION` / `CIRCULAR_OR_REDUNDANT`, matching ACTRT3 exactly — and I agree ACTL7A/7B's weak-source values are right *for them*, since their source row is a 1999 TAS with genuinely weak evidence. The distinction is real rather than house style.

**The generalisable part:** the action was correct throughout and only the machine-readable fields were wrong, so nothing in the prose flagged it. **Prose agreeing with itself is no evidence that the codes agree with the prose** — and these fields are the output the taxonomy exists to produce. Diffing coded metadata against sibling reviews on identical rows belongs in the routine, not just actions. Recorded in the notes.

## (a) The fourth sibling on the derived row

Added, with the reason ACTL10 cannot follow it. ACTRT3 holds the term "is true at this level of generality and should not be removed" because it is a perinuclear-theca scaffold component required for acrosome assembly. Truth *at that level of generality* is precisely what is unestablished for ACTL10. So for ACTR10 and ACTRT3 the row is true-but-uninformative; for ACTL10 it is unsupported — a difference `MARK_AS_OVER_ANNOTATED` records and `KEEP_AS_NON_CORE` would hide.

## The nit was a real falsehood, not a phrasing preference

"would newly place a molecular function on a gene that currently has none" — `GO:0005200` *is* an MF row on ACTL10, so as written a curator checking the GOA file would find it false. Now "for which none is supported". Also moved `got` below the `count_only` branch, where it was unused.

## terms.csv, recorded for the campaign as you suggested

The gate misfired twice here — `GO:0031011`, then `GO:0070005` — and **neither was a deletion**; both were rows *added to `main`* by siblings (ACTR5/ACTR8, then ACTMAP #2295) after this branch's base. A tip diff attributes another branch's addition to yours as a removal. The form that doesn't misfire:

```
MB=$(git merge-base origin/main HEAD)
git diff $MB HEAD -- cache/go/terms.csv | grep '^-GO:'                    # did I delete anything
git show origin/main:cache/go/terms.csv | cut -d, -f1 | sort | uniq -d    # will the merge duplicate
```

The sharp edge is that the *documented remedy* would have made it worse: `git checkout origin/main -- cache/go/terms.csv` at that moment would have silently pulled a sibling's row into this PR. Worth flagging given three siblings were in flight.

`RESULTS.md` and `results.json` byte-identical. Gates: checkquotes 20/20 · `just validate human ACTL10` ✓ Valid (same 2 genuine warnings) · terms.csv unchanged by this branch against the merge base; duplicates in the merged result still only `main`'s pre-existing pair · both lints pass with self-tests.

PR #2299, opened 2026-07-26
ACTRT3 follow-up: query each reference by scope, and one mis-statable claim corrected

cmungall:
Follow-up to #2296, which **squash-merged with `reviewDecision: ""` about four minutes before
this delta was ready** — so the review never ran on the PR at all and these corrections never
reached `main`. Same shape as the ABI3BP mis-merge; the content of #2296 landed intact, verified
by diffing the merge commit against my branch point.

The work here is the **reference-projection check** that came out of the ACTR8 review (merged
after ACTRT3 was launched, so it is not in the procedure ACTRT3 followed): query QuickGO **by
reference** rather than by gene, and ask how many entities each supporting publication annotates.
A reference that annotates many entities with identical evidence is a projection, not N
independent findings. New reproducible audit: `ACTRT3-bioinformatics/reference_scope.py`
(+ `reference_scope.json`, byte-identical on a fresh run).

| PMID | total | entities | terms | reading |
|---|---|---|---|---|
| 35793634 | 35 | 19 | 3 | 12 theca proteins carry `GO:0033011` IDA; the phenotype row does **not** spread |
| 18692047 | 5 | **2** | 2 | one co-IP, logged by two databases **and** reciprocally on both partners |
| 33961781 | **9514** | 119+ | 1 | BioPlex; one term across all of GOA, all IntAct IPI |
| 41668650, 34869336, 11750065 | 0 | — | — | not curated by GOA at all |

### 1. `GO:0033011` is not a projection — and the discriminator is which row *fails* to spread

The two perinuclear-theca ACCEPTs rest on mouse Actrt3's IDA from **PMID:35793634**, a paper
titled for *CCIN/Calicin*. Querying by reference shows it carries `GO:0033011` by IDA for **12
mouse theca proteins** (Actl9, Actrt1, Actrt2, Actrt3, Capza3, Capzb, Ccin, Cylc1, Fabp9, Gsto2,
H2bl1, Wbp2nl) — exactly the shape that turned out to be a ComplexPortal projection on ACTR8.

It is not one here, and the reason is checkable: all 12 are `assignedBy: UniProt`, and the
paper's **only functional row — `GO:0007286` IMP — is confined to `Ccin` alone**, the one gene
actually knocked out. A projection spreads the phenotype across the set; this curator did not.
So these are 12 per-protein localisation calls from one immunolocalisation study. **The ACCEPTs
stand, now on a stated basis instead of an assumed one.**

Useful side effect: the same query places **ACTRT1, ACTRT2 and ACTL9 in the theca by IDA
independently** of the 2026 co-immunoprecipitations. That turns "the ARP-T clade is a theca
clade" from an inference resting on one paper into a curated fact, and it strengthens the
`suggested_questions` item addressed to the ACTRT1/ACTRT2/ACTRT3/ACTL7A/ACTL9 set.

### 2. A mis-statable claim in my own prose, corrected

The `GO:0005522` row said mouse Actrt3 carries protein binding "by IPI **twice**" from
PMID:18692047 — which reads as two experiments. That reference covers only **2 entities in
total**, so the two rows are one co-immunoprecipitation logged once by UniProt and once by
IntAct, recorded **reciprocally on Actrt3 and on Pfn3**: one experiment recorded four ways. This
is the ACRV1 `NbExp=3` pattern appearing in my own review after I had applied it to the PDCL3
row. The proposed term never depended on a replicate count — it rests on the co-IP, the
reciprocal *Pfn3*-knockout destabilisation, and the retained profilin surface — so **the term
stands and only the sentence changed.**

### 3. `GO:0005515` quantified

PMID:33961781 accounts for **9,514** `GO:0005515` annotations across GOA, all IntAct IPI, and it
is the only term that publication contributes. That scale is what makes one unreplicated partner
on this gene uninformative rather than merely thin, and it sharpens the contrast with ACTR1B,
which kept its row from the same publication on a partner set coherent with dynactin.

### Two negative checks, reported because a negative result is still a result

- **`CommentsCorrections`/`RefType` read directly on all six cited PMIDs: empty.** No erratum, no
  Publisher Correction, no retraction. Checked on each cited article's own record rather than by
  searching for correction-type publications, because ACTR8's follow-up showed a Publisher
  Correction is not findable that way.
- **Every cross-gene fact this review leans on re-verified against the merged sibling YAML**
  rather than taken from a summary: ACTL7A `GO:0005198` IBA ACCEPT + `GO:0005200` TAS REMOVE +
  core MF `GO:0005198`; ACTL7B the same plus `GO:0015629` TAS REMOVE; ACTR10 `GO:0005200` IBA
  **ACCEPT** with core MF `GO:0005200`; ACTL8 `GO:0015629` IBA **KEEP_AS_NON_CORE**; ACTR1B
  `GO:0005515` from PMID:33961781 **KEEP_AS_NON_CORE**. All five as relied on.

### The new checker needed its own fix

Counting bare `GO:0005515` as a functional claim made *every* interaction paper look like a
projection, and mis-flagged PMID:35793634 as "possible projection" until that term was excluded
from the functional set — found by reading the output against what was already known, not by
trusting it. `audit_claims.py` now also value-checks the three reference-scope figures and
retracts the "twice" phrasing, and it hard-fails if PMID:35793634's `GO:0007286` row ever stops
being confined to one entity, since the whole projection argument depends on that.

No annotation action changes. `cache/go/terms.csv`: **0-line diff**, duplicates unchanged at the
two long-standing `GO:0001675`/`GO:0009566`.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to ACTRT3 review artifacts, notes, and local audit scripts; they do not alter application runtime behavior or GO term files.
> 
> **Overview**
> Introduces **`reference_scope.py`** and committed **`reference_scope.json`**, which query QuickGO **by PMID** (not by gene) to count annotation rows vs distinct entities, flag truncation, probe ComplexPortal, and run a location-vs-function “projection” heuristic (excluding bare `GO:0005515`).
> 
> The **ACTRT3 review and notes** are updated to use that audit: perinuclear-theca **`GO:0033011` ACCEPTs** are justified as twelve per-protein IDA calls (not a complex projection); **BioPlex `PMID:33961781`** is described with filtered-query totals (9,514 annotations, IntAct + ComplexPortal tail); and the profilin **`PMID:18692047`** “IPI twice” wording is corrected to one co-IP logged four ways. A new **`reference_scope.json`** findings block and **`supported_by`** links tie prose to the JSON.
> 
> **`audit_claims.py`** now strips markdown in `norm()`, checks reference-scope figures and structural flags (`rows_complete`, projection tail, `GO:0007286` = 1 entity), and bans the retracted phrase **“IPI twice”**. **No GO annotation actions change** in this delta.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f57f09b71ca188018ec4c53bddcd3d73c14d67d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — the IMPORTANT issue is real, I've confirmed it by measurement, and it was worse than a presentation problem. Fixed in 78965b129.

## 🟡 The sampling defect: confirmed, and it was hiding something

You were exactly right about the mechanism. `reference_scope.py` fetched one `limit=200` page, reported per-term/per-code/per-database breakdowns *from that page*, and took `total_annotations` from `numberOfHits`. Harmless for five of six references (the page covers them); not for PMID:33961781, where 200 of 9514 rows is a 2% sample.

I ran your suggested remedy (1). Both sample-derived claims were **false**:

| claim | measured |
|---|---|
| "it is the only term that publication contributes" | **false** — `GO:0005515` exact = **9508**, not 9514; aspect totals MF 9509 / CC 5 / BP 0 |
| "all assigned by IntAct" | **false** — **IntAct 9509 + ComplexPortal 5** |

And the five rows the sample hid are the interesting part: `GO:0005813 centrosome`, IPI, **assigned by ComplexPortal**, projected onto the ted-tubulin complex and its four subunits. So the reference this review cites to illustrate what is *not* a projection **does carry a projection tail** — invisible in a 200-row sample of a 9,514-row screen. None of it touches ACTRT3 and no action changes, but the prose was asserting coverage it did not have.

Your remedy (2) is implemented and generalised: every count is now its own filtered count query; page-derived fields are renamed `*_seen`; `truncated`, `unaccounted_annotations` and `term_list_provably_complete` are emitted; and **ComplexPortal is probed unconditionally**, since a projected tail can sit entirely outside the sample — which is precisely how this one escaped. `numberOfHits` and page-derived numbers can no longer sit side by side unlabelled.

## Three things that fell out of fixing it properly

1. **The verdict function then inverted.** With bare `GO:0005515` excluded from the functional set, the newly-visible ComplexPortal rows presented as "multi-entity localisation with no functional claim" and were reported as **NOT a projection** — backwards for rows a projecting database assigned by design. The exclusion you validated was correct, but it was load-bearing in a second place I hadn't considered. Fixed by naming projecting-database rows *before* the heuristic runs.
2. **Per-code totals do not partition the total.** QuickGO's `evidenceCode` filter is not exact-only: on PMID:35793634 the per-code counts sum to **57 against a total of 35**. They are kept for orientation, used for **no** claim, and documented as such — so `all IPI` is not asserted anywhere either.
3. **`term_list_provably_complete` is the precondition for the projection test**, not a nicety. PMID:35793634 is complete at 35 of 35 (as you independently verified), so the twelve-versus-one comparison the two `GO:0033011` ACCEPTs rest on is sound — now stated as a checkable property rather than as the absence of a surprise.

## 🔵 All three taken

- **The `GO:0005522` over-correction.** You're right that the co-IP is still a strand. Now: "the case for the term therefore does not rest on an apparent replicate count. All three strands below still stand; the first is one co-immunoprecipitation rather than two."
- **Hardcoded `sc[]` indexing.** Replaced with a helper that names the missing PMID *and* the keys present, and says what to do about it.
- **Non-self-identifying quotes.** `'"distinct_entities": 2'` etc. are gone; the `supporting_text` values are now `summary_line` strings that carry their own PMID, e.g. `PMID:35793634: 35 annotations total, 35 rows examined, 3 terms seen, 0 unaccounted; ...`. A new `summary_line` field exists for exactly this.

## Guards, each break-tested rather than read

`audit_claims.py` gains six patterns and four data-side guards. Verified by deliberately breaking each:

- corrupt `9508` in prose → `WRONG VALUE (scope, BioPlex exact protein-binding count)`
- corrupt the per-database split → `WRONG VALUE (scope, BioPlex per-database split)`
- make the theca phenotype row spread → "GO:0007286 row no longer covers exactly one entity; the 'not a projection' argument depends on it"
- remove the BioPlex projection tail → "the review states PMID:33961781 carries a ComplexPortal projection tail, but reference_scope.json no longer finds one"
- delete a cited PMID block → the named `RuntimeError`

Also guarded: `term_list_provably_complete` going false, and the assigning databases ceasing to sum to the total.

## Note on the merge

For the record, since it affects the trail rather than this PR: **#2296 squash-merged with `reviewDecision: ""`** about four minutes before this delta was ready, so no review ran on it at all. Its content landed intact (verified by diffing the merge commit against my branch point), which is why this is a follow-up branch rather than a force-push.

`reference_scope.json` reproduces byte-identically on a fresh run; `cache/go/terms.csv` diff remains 0 lines with duplicates unchanged at `GO:0001675`/`GO:0009566`. No annotation action changed in either round.

cmungall:
Confirmed and fixed in 9b4224a02. You're right, and the committed file disproves the old name by itself — which is the part I should have noticed.

## 🟡 `true_entities_per_term` held annotation counts

`count(reference, goId, goUsage=exact)` returns `numberOfHits`, which counts **annotation rows** and never collapses per gene product. Your disproof reproduced exactly: PMID:18692047's `GO:0005515` read **4** against **2** entities (Q8BXF8, Q9DAD6, each logged once by UniProt and once by IntAct). The double-logging this PR is about was mislabelled as an entity count *inside the JSON that explains it*.

Your reading of why the biology survives is right, and I want to be explicit that it was an asymmetry plus luck rather than care:

- **`GO:0007286` = 1 is rigorous either way** — entities ≤ annotations, so one annotation implies exactly one entity. "Confined to Ccin alone" holds regardless of which quantity the field held. That is the half carrying the weight.
- **Only the 12 needed distinct entities.** It is 12 either way (no theca protein is double-logged there), but `audit_claims.py` was confirming the sentence against a field that *could* have read 12 while the truth was 11. A guard that is right by luck of the data is not a guard.

Your remedy implemented as specified:

| change | effect |
|---|---|
| `true_entities_per_term` → `true_annotations_per_term` | name matches semantics |
| new `entities_per_term` from distinct `geneProductId` | **only** when `term_list_provably_complete` |
| new `projection_test_basis` | records which quantity `verdict()` used |
| `audit_claims.py` repointed | the two prose claims now check entity counts |

The gating matters: PMID:33961781 is truncated, so it gets **no** `entities_per_term` at all and falls back with the fallback recorded — a distinct count from a sampled page is a lower bound wearing a total's clothes. PMID:18692047's basis line now reads `GO:0005515=2`, not 4. `unaccounted_annotations` deliberately stays on annotation counts, since it is `total − sum(annotations)` and is sound precisely because those are rows — as you noted.

## Two guard gaps this exposed, both the round-2 blindness in miniature

Found by break-testing, not by reading:

1. **The retraction string was too long.** Banning the whole sentence let an emphasised or reworded variant past — which is exactly how the phrase survived in notes §5 while §14b declared it corrected. Now banned as the two words that carry the claim, and `norm()` strips markdown emphasis and backticks. Verified: `**IPI twice**` and `` `IPI twice` `` both fail now; neither did before.
2. **A guard that runs after the thing it guards is not a guard.** The availability check for `entities_per_term` sat *below* the patterns indexing into it, so its message was unreachable behind a `KeyError` — which is what your 🟡 would have produced in the field. Moved above, and entity counts now go through an accessor that raises a named error naming the field, the reason and the fix. Verified by deleting the field: the message appears, the `KeyError` does not.

## 🔵 All four taken

- **The surviving §5 phrasing** is corrected, and the prose now states the distinction explicitly: "4 protein-binding annotations spread over only 2 entities … a hit count never collapses per gene product, and conflating the two is exactly how an apparent replicate appears."
- **`'"GO:0007286": 1,'`** is gone. Both occurrences are now `projection_test_line` values carrying their own PMID, e.g. `PMID:35793634 projection test on entities: GO:0005515=12, GO:0007286=1, GO:0033011=12; max functional 1, max localisation 12` — which also makes the basis visible in the quote itself.
- **Long lines**: every line over 150 characters rewrapped; the file's longest is now under that. Two were machine-generated `source_label` values for the *Dictyostelium* multi-gene entry, refolded as block scalars.
- **Trailing whitespace** removed from both scripts.

`reference_scope.json`, `results.json` and `RESULTS.md` all reproduce byte-identically on fresh runs. `cache/go/terms.csv` diff remains 0 lines, duplicates unchanged at `GO:0001675`/`GO:0009566`. No annotation action has changed in any of the three rounds — both `GO:0033011` ACCEPTs stand as argued, and `GO:0005522` still rests on the co-IP, the reciprocal *Pfn3*-knockout destabilisation and the retained profilin surface.

Noted on your runner constraints: the reproducibility and break-test claims are mine and unverified by you. The break-test outputs are quoted verbatim above and each is a one-line command against a restored backup, so they should be cheap to spot-check if you want them.

cmungall:
Confirmed, fixed in 5012417a9. This one is the sharpest of the four and I'd have shipped it: the gate read as if it enforced the docstring's reason, and it did not.

## 🟡 Term completeness ≠ row completeness

`entities_per_term` is collected from distinct `geneProductId` values in the **returned rows**, so the property it needs is that every **row** was seen — `not truncated`, already computed and strictly stronger. `unaccounted_annotations == 0` proves only that every annotation's **term** was queried. The docstring named the right reason while the condition underneath enforced something else.

**Your near-miss reproduces exactly.** PMID:33961781: 9508 + 5 = 9513 against 9514, so a single stray `GO:0005515` descendant the sampled page never showed is the *only* reason its term list stayed open. I checked what would have happened without it:

```
unaccounted would be 0, page still 200 of 9514
-> entities_per_term[GO:0005515] ≈ 119   (distinct_entities_seen)
-> entities_per_term_available: true
-> projection_test_basis: entities_per_term
-> projection_test_reliable: true
```

119 is the number round 1 flagged as materially understated — reappearing labelled complete and used as the projection basis, with `audit_claims.py` inheriting the same precondition and therefore **confirming** it. Round-1's defect returning under a label asserting it had been fixed.

Fixed as you specified:

- `rows_complete = not truncated and total > 0` is the gate for `entities_per_term`
- `term_list_provably_complete` retained as a separate, separately-reported property
- `projection_test_reliable = rows_complete`
- the audit asserts **both**, so a regression in either is visible rather than silently substituted

Break-tested: forcing term closure while rows stay truncated now yields the named unsupported-claim messages *and* the `rows_complete` assertion. Code-level, `complete = rows_complete` no longer references `unaccounted` at all, so term closure cannot enable entity counting by any path.

## Same root, smaller: the two fields the rename missed

`max_entities_on_a_functional_term` / `max_entities_on_a_location_term` held **annotation** counts in the fallback branch under names saying entities — today's `max_entities_on_a_location_term: 5` for PMID:33961781, exactly as you said. Renamed to `max_functional_spread` / `max_location_spread` with a `spread_units` field carrying `entities` or `annotations`, and `verdict()` now prints the units instead of hardcoding the word "entities". The `projection_test_line` also carries them, so a quote of it discloses its own basis.

## 🔵 All three taken

- **Docstring line 43** no longer refers to the removed field name; the naming-discipline paragraph now states the row-vs-term distinction explicitly, since that is the discipline it exists to establish.
- **`entity_count()` raised where the file accumulates.** You're right that this broke the script's own contract of printing every failure in one run. It now appends to `problems_early` and returns `None`, and the dependent patterns are added only when the count is available. Verified: two independent unavailabilities produce **three** messages, where the raising version produced one and aborted.
- **The broken wrap** at notes:450-452 is repaired.

## The through-line, since it is the same error four times at four depths

| round | the mismatch |
|---|---|
| 1 | a **total** reported beside a **200-row sample** |
| 2 | an **annotation count** named an **entity count** |
| 3 | two fields the rename **did not reach** |
| 4 | a **row-derived** count gated on a **term-derived** property |

**A number's name must match how it was obtained.** Every one of these passed schema validation, quote validation, and the numeric audit — because each constituent value was real and each quote verbatim. All four were found by asking what a specific number was actually counting. That is now written up in `ACTRT3-notes.md` §14b as the transferable lesson, and it is worth more than the ACTRT3 review itself.

No prose figure moved in this round and no annotation action has changed in any of the four: both `GO:0033011` ACCEPTs rest on `GO:0007286` = 1, which is rigorous under either quantity since entities ≤ annotations, and `GO:0005522` rests on the co-IP, the reciprocal *Pfn3*-knockout destabilisation and the retained profilin surface.

`reference_scope.json` reproduces byte-identically; `origin/main` merged in and re-validated green; `cache/go/terms.csv` diff 0 lines with duplicates unchanged at `GO:0001675`/`GO:0009566`; no line over 150 characters and no trailing whitespace in any of the four scripts.

cmungall:
Confirmed, fixed in f57f09b71. You're right that it costs nothing that the JSON was correct beside it — `verdict()` is what `main()` prints, and the printed sentence is where the judgement gets formed.

## 🟡 Both hardcoded nouns fixed

- **The general fallback** (`functional rows on {f} entities, location rows on {l}`) — now `{f} {units}` / `{l} {units}`.
- **The localisation branch** — was units-correct on `l` and hardcoded on `f` (`confined to {f} entity`). Now both carry `{units}`. This is the branch the whole theca argument prints through, which is what makes it the worse of the two.

`verdict()`'s docstring now states the invariant explicitly: no printed count is rendered with a hardcoded noun, because the maxima are entity counts only when the basis is entities.

## The structural fix, since this is the second unreachable-branch defect

You noted the reachability shape matches round 3. That's the part I acted on rather than just patching the symptom: **coverage is no longer left to the dataset.** New `selftest()` runs on every invocation, *before* any report is written, driving `verdict()` with synthetic blocks through the fallback, localisation and same-count branches on an annotations basis, and asserting that **no printed count carries a unit the block did not declare** — plus the mirror, that an entities basis says entities.

Verified by reintroducing your exact regression:

```
$ sed -i 's/{f} {units}, location rows on {l} {units}/{f} entities, location rows on {l}/' reference_scope.py
$ uv run python reference_scope.py
  File ".../reference_scope.py", line 341, in selftest
AssertionError: selftest: verdict() said 'entities' for a block declaring spread_units='annotations' ...
```

The run aborts before emitting a report, rather than printing a misnamed sentence. Two rounds of the same class was the argument for making the branches reachable by construction.

## 🔵 All five taken

- **`proj_line`** no longer recomputes the units inline and then repeats them in a suffix; it uses `spread_units` once.
- **Direct indexing in `audit_claims.py`** — you were right that `entity_count()` used `.get()` for a reason the rest of the file didn't follow. Every JSON field now goes through a `flag()` helper that reports a renamed or dropped field **by name**, and the per-database/per-term figures use `.get()`. Verified by renaming `rows_complete`: `PMID:35793634 has no 'rows_complete' field; reference_scope.py's output shape changed and … can no longer be checked` — no `KeyError`.
- **yaml:448-449** no longer justifies the precondition by naming term-list closure. It now says every row was returned, 35 of 35, and states *why* row completeness is the property that matters: the per-term entity counts are collected from the returned rows, so a term list can close while the rows are still a sample.
- **The 19-entities figure** is now value-checked against `distinct_entities_seen` and cited in its own `supporting_text` (the refreshed `projection_test_line`). Verified: changing 19 to 20 fails.
- **notes:454** rewrapped. The remaining >100-character lines are markdown table rows.

## Five rounds, one error

| round | the mismatch |
|---|---|
| 1 | a **total** reported beside a **200-row sample** |
| 2 | an **annotation count** named an **entity count** |
| 3 | two fields the rename **did not reach** |
| 4 | a **row-derived** count gated on a **term-derived** property |
| 5 | the hardcoded noun in the two branches **a reader sees** |

**A number's name must match how it was obtained** — and, from rounds 3 and 5 specifically, **a fix verified only against reachable data is half verified.** Both are now written up in `ACTRT3-notes.md` §14b, which is the part of this PR most likely to be useful to another gene.

No number in the review moved in this round, and no annotation action has changed in any of the five. Both `GO:0033011` ACCEPTs rest on `GO:0007286` = 1, rigorous under either quantity since entities ≤ annotations; `GO:0005522` rests on the co-IP, the reciprocal *Pfn3*-knockout destabilisation and the retained profilin surface.

`reference_scope.json` reproduces byte-identically; `origin/main` merged and re-validated green; `cache/go/terms.csv` 0-line diff with duplicates unchanged; no line over 150 characters and no trailing whitespace in any of the four scripts.

PR #2301, opened 2026-07-26
Review human ACTA1: the canonical actin, where three pipelines substitute a generic one

cmungall:
## ACTA1 (P68133) — skeletal muscle α-actin

This gene inverts the campaign's usual problem. Every actin-family gene reviewed before it (ACTL7A/7B, ACTL8, ACTL10, ACTR1A/1B, ACTR5, ACTR8, ACTR10, ACTRT3) was a **divergent** member where the risk was inheriting canonical actin biology it does not have. ACTA1 **is** the canonical actin, so the phylogenetic transfers argued *down* elsewhere had to be argued *up* here.

**50 GOA rows + 1 NEW: 15 ACCEPT · 16 MARK_AS_OVER_ANNOTATED · 13 KEEP_AS_NON_CORE · 5 REMOVE · 1 MODIFY · 1 NEW**

### PAINT is working on this gene, and that is a result

`GO:0005200` is IBD-asserted at exactly **one** node in PTHR11937 — `PTN000940351`, the conventional-actin node — and **IRD-negated at eight descendant nodes**. ACTL7A, ACTL7B and ACTRT3 each had to act on the negated side. ACTA1 sits on the accepting side and is the recipient the assertion was made for: all ten donors resolve, all ten carry their own experimental evidence, and the AADACL4 heterogeneous-donor caveat was checked and does not apply. Worth recording because this family's PAINT handling has otherwise only been reported when it failed.

### The actual defect: three pipelines each substitute a generic actin for ACTA1

Not disease pleiotropy — GOA has **no** nemaline-myopathy phenotype rows at all. It is isoform substitution, from three independent directions.

**1. AgBase ISS → 5 REMOVE.** `GO:0010628`, `GO:0030027`, `GO:0030175`, `GO:0044297`, `GO:0090131` all have WITH/FROM `UniProtKB:P08023` = **chicken ACTA2, aortic *smooth muscle* actin**, and all five trace to one paper (PMID:10633868), an antisense knockdown of endothelial-mesenchymal transformation in chick cardiogenesis. Three things are wrong:

- **Wrong paralog.** Chicken ACTA2 is the ortholog of human *ACTA2*. Chicken has its own ACTA1 (`P68139`) — the correct ortholog existed in the same species and was not used.
- **Wrong context.** `GO:0044297` is *defined* by the presence of axons, dendrites, cilia or flagella. A multinucleated muscle fibre has none.
- **ISS cannot discriminate.** ACTA1 is **97.9%** identical to ACTA2 and **98.9%** to ACTC1 (measured), so the similarity criterion is satisfied trivially.

The donor is **not** weak — P08023 holds all five terms by its own IDA/IMP — so `root_cause: PROPAGATION_BAD`, not `SOURCE_WEAK_OR_INFERRED`.

**Scope, with a positive control.** QuickGO shows the block on all four human α-actins **and mouse Acta1** — but **not** on ACTB/ACTG1. It was keyed on α-actin-ness rather than actin-ness, which is what makes this a demonstrable paralog transfer rather than a judgement call. Retracting it from ACTA1, ACTC1 and mouse Acta1 removes 15 annotations resting on one chick experiment.

**2. Reactome.** Eight `GO:0005829` rows are one per-reaction export, not eight findings. Three are **E-cadherin adherens-junction** reactions; ACTA1 is in them because Reactome's participant is a generic polymer, `F-actin (all)`, whose reference entities are **all six human actins**. A first draft split the verdicts on these rows and the repo validator was right to object: what they assert in GO is only *cytosol*, equally bland everywhere. All eight are `KEEP_AS_NON_CORE`, and the defect — which is in pathway membership, upstream of the GO term — is filed as a Reactome-side correction.

**3. PAN-GO, running outward from ACTA1.** `GO:0001725` stress fiber is IBD at `PTN000233075` seeded from ACTA1 *itself*, and its sole basis is an EGFP-actin transfection into C2C12 myoblasts (PMID:15198992) whose abstract never mentions stress fibers. All four of mouse Acta1's `GO:0001725` rows carry WITH/FROM `P68133` (ISS, IEA, ISO, IBA), so **one ectopic-expression readout now supports five annotations across two species and four pipelines**. Demoted, not removed — full text unavailable. The three sarcomeric terms on the same node are correct and accepted.

### Measurement, and what it declines to show

`ACTA1-bioinformatics/actin_peptide_specificity.py`: of ACTA1's 63 tryptic peptides in the 7–30 aa window, only **9 (14.3%)** occur in no other human actin, and those collapse to **3 independent regions** (reporting 9 would overstate it — most are nested missed-cleavage variants). **29** of the 54 shared peptides are shared with ACTB/ACTG1.

This **does not refute** the five extracellular HDA rows (prostatic-secretion, parotid and trabecular-meshwork exosomes, plasma microparticles, tears). Attribution *is* achievable, so they are **untested, not refuted** → `MARK_AS_OVER_ANNOTATED`, with a decisive targeted-proteomics experiment specified against the three named regions. Deliberately the same discrimination ACTL8 drew between its refuted filament-interface result and its untested ATP-site result.

### Other findings

- **MODIFY `GO:0006936` → `GO:0003009`** skeletal muscle contraction. α-skeletal actin is the *exclusive* sarcomeric isoform of skeletal muscle from infancy (PMID:16288873) and ACTA2/ACTG2/ACTC1 carry the other branches, so specialising loses nothing and stops ACTA1 standing in for muscle actin in general.
- **NEW `GO:0051371` muscle alpha-actinin binding** — absent from GOA though UniProt maps it to two ACTA1 regions (112–125, 360–372) and the affinity was measured on actin isolated from human patient muscle, where K336E reduced it ten-fold (PMID:16945537).
- **The interaction record is asymmetric.** All 11 `GO:0005515` rows are screen or incidental output — 7 from **one** two-hybrid bait panel logged under three sub-method labels at MI-score 0.56 (the ACRV1 pattern, confirmed against IntAct directly), 2 from one AP-MS pipeline in two non-muscle lines. Meanwhile the two interactions UniProt records with experimental `ECO:0000269` tags have **no GO row**: TTID/myotilin (PMID:10958653, whose 7 GO annotations are all on MYOT) and USP25 (**PMID:16501887 carries zero annotations in the whole of GO**). Adding those would improve the record more than removing any of the nine weak rows. SYNC/syncoilin is separated out as the one screen partner in the right cell and compartment.
- **Considered and rejected** upgrading the DNase I row to `GO:0060703`: the paper shows affinity *purification*, not inhibition, and DNase I is not a physiological partner of sarcomeric actin — annotating it would be the in-vitro-not-in-vivo error.

### Negative results, recorded rather than skipped

- **Reference projection** (the ACTR8 defect): every functional reference queried by PMID and counted by entity — 1, 1, 3, 3 and 5 annotations, all on ACTA1. No complex-plus-subunits pattern, no phenotype spreading past the gene assayed.
- **Retraction / erratum / expression of concern:** 28 cited PMIDs read from `CommentsCorrections/RefType` on each *cited* article's record — not a pubtype search, which cannot see a Publisher Correction. **Zero flagged.**
- **IBA landing above its donor** (ACRV1): none — rows 3–6 land at their donor's own terms.
- **LCA caveat** (AADACL4) on `GO:0005200`: checked, inapplicable — the donors agree.
- **Partner accessions** (ACRV1 TrEMBL/ORFeome): all ten reviewed canonical; `Q6ZQX7-4` is a declared LIAT1 isoform, not a partial clone.
- **WITH/FROM:** all 26 rows resolve, zero unresolved tokens, every model-organism mapping verified against the resolved entry's own cross-references.

### Reproducibility and drift guards

Four committed scripts. `build_source_entities.py` is the **only** thing that writes `source_entities`, and it also verifies the hand-restored `supporting_entities` — the `fetch-gene` stub collapsed seven `PMID:32814053` rows into one, all seven were restored, and since they share a (term, evidence, reference) key the test is a multiset comparison: **11 interaction rows match the GOA one-to-one**. `audit_claims.py` pins every number appearing on more than one prose surface, deriving expected values from the JSON rather than from literals, plus a list of retracted phrasings. Both have self-tests that break the checks and require them to fire.

### Two of my own errors, documented in the notes

1. **A reference title written from memory**, blocked by the pre-write hook — on PMID:16288873, a *load-bearing* reference. `audit_claims.py` now treats the invented string as retracted.
2. **My dead-accession guard did not work, and its own self-test proved it.** `O15507` passed a "primaryAccession must match" check, because an inactive UniProt entry returns **its own accession** and nothing else. The authoritative signal is `entryType: "Inactive"` (`inactiveReason: MERGED → P56159`). A fifth case requires a live accession to still pass, so the guard discriminates rather than refusing everything.

Also fixed a silent zero: QuickGO **rejects every non-UniProtKB gene-product id** in this file's WITH/FROM fields (HTTP 400), and UniProt does not cross-reference WormBase by `WBGene` id, so four *C. elegans* donors first reported "no evidence" when they had never been asked. Corrected count is **24/24**, not 20/24 — and it is the number that makes the `GO:0015629` ACCEPT solid.

Gates: `checkquotes` 51/51 clean · `just validate` ✓ Valid, no warnings · `terms.csv` no deletions, only the 2 pre-existing duplicates, exactly 3 rows added.

---

> [!NOTE]
> **Low Risk**
> Curated annotation review, analysis scripts, and GO term cache additions only; no runtime application or auth/data-path changes.
> 
> **Overview**
> Adds a **complete human ACTA1 (P68133) GO annotation review** (~50 GOA rows adjudicated plus one proposed annotation) and the **reproducible checks** behind it.
> 
> The review **accepts** core sarcomeric biology (thin filament, structural constituent, ATP hydrolysis, myosin binding) and flags **propagation failures**: five **REMOVE** ISS rows from chicken smooth-muscle **ACTA2** (mesenchyme/lamellipodium terms on skeletal actin), many **MARK_AS_OVER_ANNOTATED** interaction and extracellular MS rows, stress-fiber annotations from C2C12 transfection, and **MODIFY** generic muscle contraction → **skeletal muscle contraction**. It proposes **NEW** `GO:0051371` muscle alpha-actinin binding from patient-muscle data.
> 
> **`ACTA1-bioinformatics/`** adds scripts and committed JSON for WITH/FROM donor resolution, actin isoform tryptic peptide specificity (HDA isoform attribution), PTHR11937 nucleotide-term family counts, automated `source_entities` sync to the GOA TSV, and **`audit_claims.py`** to keep review/notes/RESULTS numbers aligned. **`cache/go/terms.csv`** gains three related GO term rows used in the review.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ae7b3c4394a4e876330e3e9d4c8418921850409. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — all five taken. Two were real errors; I verified both against UniProt before conceding rather than taking them on trust.

**(1) The peptide analysis was digesting the ORF, not the protein.** Confirmed: `INIT_MET 1 "Removed"`, `CHAIN 2..377` (intermediate form, N-acetylcysteine at residue 2) and `CHAIN 3..377` for the mature protein after ACTMAP cleaves that acetylated cysteine. So the ORF's N-terminal peptide cannot be detected, and my `suggested_experiments` entry named it and gave the region as starting at residue 1.

Recomputed on the mature chain, and I extended the fix a step further than the suggestion: **comparators now contribute both their ORF *and* their own mature-chain digests**, so a peptide is called distinguishing only if no other actin can produce it in either form — otherwise an ACTA1 peptide could look unique merely because a comparator presents it after its own N-terminal processing.

You predicted the 3-region count would be unchanged, and every count is: **63 peptides, 9 distinguishing (14.3%), 54 shared, 29 shared with ACTB/ACTG1, 3 regions.** The script now computes both forms and asserts they agree, so a future divergence is flagged rather than smoothed over. Region 1 is `3-30` `DEDETTALVCDNGSGLVK…`, and the experiment names that peptide. The numbers were never wrong; the experiment was unbuildable — *a sequence analysis is only as biological as the sequence it starts from, and a `CHAIN` feature is where that is decided.*

**(5) The MARTX cross-link was stated as fact.** Confirmed: both `CROSSLNK` features carry `ECO:0000250|UniProtKB:P60709`, and the PTM comment is `ECO:0000305`. The cross-linking is real; the Lys-52/Glu-272 numbering on the skeletal isoform is transferred from β-actin. The description now says so explicitly.

**(2) `GO:0015629` accepted twice while its parent is demoted, and absent from `core_functions`.** Added to `core_functions.locations`, and the parent's row now argues the asymmetry rather than leaving it implicit: `GO:0015629` is defined as the part of the cytoskeleton composed of actin and associated proteins, so for a filament protomer it is a substantive statement, it is IBD-supported at `PTN002631484` from 24 experimentally-annotated donors, and ACTA1 holds it independently by IMP — none of which is true of the bare `cytoskeleton` parent, which would be equally true of a tubulin or a lamin.

**(3) IPI over IDA for the new `GO:0051371`.** I took the second option you offered and stated why no partner accession is available: PMID:16945537 names only "the Z-line protein alpha-actinin" and does not say whether ACTN2 or ACTN3 was used, so there is nothing to put in a with/from field and asserting one would be a guess. The row now records that, explains the absent `supporting_entities`, and spells out the conversion path to IPI for a curator who gets the isoform from the full text.

**(4) `GO:0043531` leaning on cross-review precedent.** Rewritten to lead with the cycle: ATP-bound G-actin is the assembly-competent species that adds to a growing filament; hydrolysis follows incorporation, and the ADP protomer is the aged, depolymerisation-prone form that gives the filament its polarity and that the severing and recycling factors recognise. So both ligands are bound at different points in one cycle and dropping either would misdescribe the mechanism. The paralog precedent is now corroboration, not the argument.

One process note: adding those three phrasings to `audit_claims.py`'s retracted list **immediately fired on my own post-mortem text**, which had quoted two of them in order to explain them. They are now described rather than reproduced — the same rule I had already applied after the hook caught a reference title I had written from memory.

Gates: `checkquotes` 51/51 · `build_source_entities --check` OK · `audit_claims` OK with self-tests passing · `just validate` ✓ Valid, no warnings · peptide JSON reproducible from a fresh run · no duplicate YAML keys · `terms.csv` zero deletions against the **merge base**, only the two pre-existing duplicate curies, net delta of exactly three rows.

cmungall:
Self-audit follow-up — I found and fixed a wrong number of my own, in the `GO:0043531` row you had asked me to strengthen.

While re-checking the claims in that row I noticed one I had never measured: it placed ACTA1 among a supposed **pair** of PTHR11937 members carrying both nucleotide terms. That came from the merged ACTR10 review, whose line is a list about **ATP binding alone** — and I had converted it into a count about *two* terms. The campaign's own rule is to relay a sibling's claim as a claim rather than as a fact, and this broke it in the way that is hardest to catch: the sibling's summary arrived pre-formatted as a quotable line and carried more apparent authority than the review it came from. Nothing flagged it, because it was internally consistent and verbatim-sourced.

Measured across all **533** PTHR11937 protein members (`nucleotide_terms_in_family.py`, new and committed):

| term | family members carrying it |
|---|---|
| `GO:0005524` ATP binding | **31** |
| `GO:0043531` ADP binding | **1** — ACTA1, by TAS |
| both | **1** — ACTA1 alone |

So it was wrong in both directions at once, and **the corrected fact is the stronger one**: ACTA1 is the family's *sole* ADP-binding holder, which is a better reason for the paralog reviews having treated it as the reference point than "one of two" ever was.

Worth flagging the measurement trap, because it is the same shape as one earlier in this review and it produced a confidently wrong answer first. **Querying by GO term alone paginates into a false negative.** `GO:0043531` has ~205,000 annotations across GO and `GO:0005524` ~9.6 million; one unpaged request returns page 1 of 200, and intersecting that page with the family gives an **empty set for both terms** — which reads as "no member of this family carries either term", the exact opposite of the truth. The script therefore keys the query on the family's own accessions, batches them because 533 × 2 per-accession queries do not finish, and **asserts no batch was itself truncated**, so the fix for the pagination trap cannot silently reintroduce it. It also fails if ACTA1 ever stops being the sole ADP holder, since the review now asserts that it is.

`audit_claims.py` gains the wrong phrasing as retracted and pins both measured figures, deriving them from the new JSON rather than from literals. And for the third time this round the lint immediately fired on my own post-mortem text for quoting the retracted phrase while explaining it — on all three prose surfaces this time — so all three now describe it instead of reproducing it.

That makes three errors of my own caught in this review by its own checks (a reference title from memory, a dead-accession guard that passed the dead accession, and this), plus the two you caught. All are written up in `ACTA1-notes.md` rather than only in commit messages.

Gates unchanged and green: `checkquotes` 51/51 · `build_source_entities --check` OK · `audit_claims` OK with self-tests passing · `just validate` ✓ Valid, no warnings · `terms.csv` zero deletions against the merge base, only the two pre-existing duplicates, net delta of exactly three rows.

cmungall:
Both 🟡 confirmed against the repo and fixed, plus all four 🔵. Every fix read back out of the tree and each new guard break-tested in both directions.

## 🟡 1 — the scope qualifier

Confirmed: `fetch_interpro_family_simple.py` builds `PTHR11937-entries.csv` from `/protein/reviewed/entry/…`, and the family's own metadata reports `proteins: 88887`, so **533 is ~0.6%**. You were right that this matters and right about where it had leaked.

The qualifier now travels with the number on **every** surface — review YAML, notes, `RESULTS.md`, the script docstring, its stdout, and the JSON keys (`n_reviewed_with_atp_binding`, `n_reviewed_with_adp_binding`, `subject_is_sole_adp_holder_among_reviewed`, plus new `member_scope` and `fraction_of_family_measured`). `RESULTS.md`'s "sole holder … in the entire family" is corrected to "among reviewed members".

I took your warning about laundering seriously: **the family-wide version is now stated separately and labelled as an argument**, not folded into the measurement — ACTA1's `GO:0043531` is manual TAS, unreviewed entries get only IEA, no IEA pipeline maps the actin fold to ADP binding, so an unreviewed holder is unlikely, *and unlikely is not measured*.

The `audit_claims.py` guard is rescoped too, and deliberately pins the specific phrase `533 reviewed` on ≥3 surfaces rather than a bare "reviewed", which already occurs for unrelated reasons and would have passed vacuously. Both unqualified forms are now retracted phrasings. Break-tested: stripping the qualifier from one surface produces two failures; restoring it passes.

## 🟡 2 — the truncation guard

Fixed as you specified — `numberOfHits` against `len(d['results'])`, no constant. Empirically QuickGO *does* honour `limit=200` here, so the committed numbers were unaffected, but I re-ran rather than assuming: **counts identical at 31 and 1.** The check no longer depends on that behaviour holding.

## 🔵 all four

- **The sole-holder regression aborting the audit.** Correct and a real bug — it raised from inside `required_claims()`, so it would have hidden every other finding, including the self-test baseline. Now a `problems` entry, and I added a second one pinning the member count so prose and measurement cannot drift apart independently. Break-tested both.
- **`goUsage` unset.** Now set to `descendants` over `is_a,part_of`, matching `resolve_withfrom.py`. It left both counts unchanged at 31 and 1 — stated rather than assumed.
- **Backtick imbalance** fixed.
- **Quote-split RETRACTED bypass.** Already closed locally before your review landed: `flatten()` now strips quotation marks and normalises dashes, and it caught the surviving instance on the next run. No occurrence of that phrase survives on any surface.

## Pass 2's three script notes, also in this commit

The set-versus-cardinality note was the most productive thing in either review. Fixing it **immediately exposed a live disagreement the count had masked** — both forms give 9 distinguishing peptides, but the sets differ in 4 members. The right invariant turned out to be narrower than "sets agree", since the N-terminal peptides are *supposed* to differ: counts agree **and every differing peptide lies at the N-terminus**.

Then break-testing that guard found **it did not fire**. I had written the tolerance as `MATURE_START + 1`, so it scaled with the parameter under test — an offset of 50 bought itself a 51-residue window and passed, emitting a confident wrong region. *A guard whose tolerance is set by the thing it guards is worse than no guard.* The tolerance is now a fixed biological constant, and further break-testing showed its honest limit (an offset of 4 is plausible processing and passes), so **`MATURE_START` is no longer typed in at all** — `mature_chain_start()` parses it from the `CHAIN` feature whose note matches the `RecName`, failing unless that is exactly one feature. It returns 3, agreeing with the hand-typed value; the point is that it is derived rather than transcribed.

Your third note was right too: reusing ACTA1's offset for the comparators was wrong, since ACTB is annotated `CHAIN 1..375` **and** `CHAIN 2..375`. Comparators now contribute digests at offsets 0, 1 and 2 — conservative by construction and independent of six `CHAIN` annotations being complete.

One method note for myself, since it nearly bit me while verifying this: my read-back `grep -c '533 reviewed'` returned 0 for `RESULTS.md` while the lint passed. The content was fine — the phrase wraps a line and the lint normalises whitespace. **Verify with the same normalisation the checker uses**, or the verification disagrees with the check for reasons that have nothing to do with the claim.

Headline numbers unchanged throughout: 63 peptides, 9 distinguishing (14.3%), 54 shared, 29 with ACTB/ACTG1, 3 regions at 3-30 / 287-317 / 338-361; 31 reviewed members with ATP binding, 1 with ADP binding. Verdict mix unchanged: 15 ACCEPT / 16 MARK_AS_OVER_ANNOTATED / 13 KEEP_AS_NON_CORE / 5 REMOVE / 1 MODIFY / 1 NEW over 50 GOA rows.

Gates: `checkquotes` 51/51 · `build_source_entities --check` OK · `audit_claims` OK with self-tests · `resolve_withfrom --self-test` OK · `just validate` ✓ Valid, no warnings · both JSONs reproducible from fresh runs · `terms.csv` zero deletions against the merge base, only the two pre-existing duplicates, net delta three rows.

cmungall:
All three 🔵 fixed and pushed as `e3459fd38` — **this is my last push; I have nothing further queued.** Each was break-tested rather than reasoned about.

**1. The surface-counting bug, which defeated this lint's own stated purpose.** Your example is exact. `P08023` occurs 13/2/2, so a minimum of 3 was satisfied seventeen times over from the review YAML alone — I confirmed by deleting all thirteen mentions from the review and watching the old check pass. That is precisely the corrected-at-all-but-one-site failure the module docstring names, so the lint was blind to the one thing it was written for. It now counts **distinct surfaces** and names the missing one; the same mutation now fails with `missing from ['review']`.

Switching semantics immediately exposed two claims that had only ever reached 3 by summing — `F-actin (all)` and the IntAct score `0.56`, both absent from `RESULTS.md`. I lowered their minimums to 2 rather than adding them to `RESULTS.md`: both are curation evidence, not outputs of any script in that directory, and padding the file to satisfy a lint is how a lint starts shaping the prose instead of checking it. The minimums had simply been written under the old semantics.

**2. `confined = bool(diff) and all(...)`.** Correct — a fully-agreeing run, the *best* possible outcome, evaluated False and aborted with a corruption message. `all()` over an empty set is already True and that is the right reading: nothing disagrees, so nothing disagrees off the N-terminus. Checked all three cases; only the empty one changes behaviour.

**3. `88887` as the last transcribed constant.** Also correct, and the asymmetry you describe is the dangerous part: a refreshed member list moves the numerator and fires the audit, while the denominator and the 0.6% go stale in silence. It is now parsed from `PTHR11937-metadata.yaml` by `family_protein_total()`, which fails loudly if the key is absent — break-tested by renaming it.

Nothing biological changed: no GO term, evidence code, qualifier, action or claim. Counts unchanged — 533 reviewed members, 31 with ATP binding, 1 with ADP binding, 63 peptides, 9 distinguishing, 3 regions.

Worth recording that all three of these were in code I had written *as guards*, in a review whose notes already document three earlier self-inflicted guard failures (a tolerance that scaled with the parameter it guarded, an accession check blind to inactive entries, a retracted-phrase matcher evadable by a quotation mark). The pattern across all six is the same: **the guard reports OK, so nothing draws attention to it, and only reading it or deliberately breaking it reveals the hole.** A passing self-test says nothing about the guard you failed to write.

Gates: `checkquotes` 51/51 · `build_source_entities --check` OK · `audit_claims` OK with self-tests · `resolve_withfrom --self-test` OK · `just validate` ✓ Valid, no warnings · `terms.csv` zero deletions against the merge base, only the two pre-existing duplicates, net delta three rows.

cmungall:
All three pass-5 🔵 fixed in `6ae7b3c43`. I had said the previous push was my last; I reopened for one reason, and it is worth stating plainly.

**🔵 1 was not a new demand — it was my own fix applied in N−1 places.** Last round I parsed 88,887 in the script and pinned the numerator's scope in `audit_claims`, then left the denominator and the derived 0.6% typed into three prose surfaces and unpinned. So a refreshed member list would move the JSON and fire the 533 guard while the denominator went stale in silence — the *same* numerator/denominator asymmetry that motivated parsing it in the first place, fixed in one place and not the other. That is this campaign's most repeated failure shape, and "I announced I was done" is the wrong reason to leave it in the tree. Both the total and the percentage are now pinned from the JSON; break-tested by removing the total from the notes, which fails naming the missing surface.

**🔵 2**: `family_protein_total()` took the first `proteins:` line at any nesting while its sibling `mature_chain_start()` required exactly one match — an inconsistency in strictness between two parsers written days apart, which is how one of them quietly becomes authoritative about the wrong number. Now requires exactly one and refuses to guess. Break-tested with a duplicated key.

**🔵 3**: a differing peptide absent from `positions` would have been skipped by `all()` and passed vacuously. You noted it is impossible today because the mature chain is a suffix of the ORF; I agree, and the guard should not rest on that continuing to hold. Unlocated peptides now fail loudly.

Guard code and comments only — no GO term, evidence code, qualifier, action, `core_function`, description or reference changed, and both JSONs are byte-identical to the previous commit.

**This is genuinely my final push and I have nothing queued.** For the record on where this converged: passes 1–2 found two factual errors in the review (the ORF-versus-mature peptide species and a by-similarity residue assignment stated as fact), and everything from pass 3 onward has been guard code, with the biology untouched since pass 1. Across the whole PR, six defects were found in code I had written *as guards* — a tolerance that scaled with the parameter it guarded, an accession check blind to inactive entries, a retracted-phrase matcher evadable by a quotation mark, a lint that summed occurrences instead of counting surfaces, an agreement check that failed on perfect agreement, and this last half-applied fix. Every one of them reported OK while broken. That is the durable lesson from this gene, and it is written into `ACTA1-notes.md` rather than left in commit messages: **a guard that passes tells you nothing until you have broken it on purpose.**

Thank you for five careful passes — the mature-chain catch in particular changed a proposed experiment from unbuildable to actionable.

PR #2304, opened 2026-07-27
Review human ADAMTSL1; only PTHR13723 member with no extracellular matrix term

cmungall:
## ADAMTSL1 (Q8N6G6, punctin-1) — human, PAINT + affinage campaign

GOA record is unusually thin: **4 rows** (1 IEA + 3 TAS), **no IBA**, **no molecular-function
annotation of any evidence code**. All four reviewed, plus one proposed `NEW` row. Deep research:
affinage, `gates_passed: True`, faith 100%, 10 numeric PMIDs, no preprint ids in PMID-shaped
fields.

### Verdicts

| term | evidence | action |
|---|---|---|
| `GO:0030198` extracellular matrix organization | IEA / `GO_REF:0000002` / `InterPro:IPR013273` | **KEEP_AS_NON_CORE** |
| `GO:0005788` endoplasmic reticulum lumen | TAS / `Reactome:R-HSA-5173005` | **KEEP_AS_NON_CORE** |
| `GO:0005788` endoplasmic reticulum lumen | TAS / `Reactome:R-HSA-5173192` | **KEEP_AS_NON_CORE** |
| `GO:0005788` endoplasmic reticulum lumen | TAS / `Reactome:R-HSA-6785565` | **KEEP_AS_NON_CORE** |
| `GO:0031012` extracellular matrix | IDA / `PMID:11805097` | **NEW** |

5 entries against 4 GOA rows; the extra is the `NEW` proposal.

### The domain question, settled from the feature table first

The task lead was that ADAMTSL proteins lack the catalytic domain. Verified rather than assumed:
the FT block over the 1762-aa precursor lists a signal peptide, nine annotated TSP type-1
repeats, four Ig-like C2-type domains and a PLAC domain — **no metalloprotease domain, no
disintegrin-like domain, no prodomain, no zinc-binding site**, no `CATALYTIC ACTIVITY` comment
and no EC number. Corroborated by `PMID:11805097` and by `PMID:28722276` ("comprising an ADAMTS
ancillary domain but lacking the protease domain and thus, enzymatic activity").

### The peptidase error is real, but it sits upstream of GO

`Q8N6G6` carries **`KW-0378 Hydrolase`**, which generates `GO:0016787 hydrolase activity
IEA:UniProtKB-KW` in the entry's own GO cross-reference list — **contradicted inside the same
entry** by its CAUTION comment. Of the six human ADAMTS-like proteins only ADAMTSL1 and THSD4
carry the keyword; ADAMTSL2/3/4/5 carry the identical CAUTION and no MF keyword, and none of the
six has a `CATALYTIC ACTIVITY` comment.

**It never reached GOA** (keyword-derived annotations were withdrawn for cellular organisms), so
it is filed as a UniProt correction in `suggested_questions` rather than as an
`existing_annotations` row — reviewing rows GOA does not carry would break the row-count
reconciliation.

Two related observations, kept separate because they have different standing:

- **PANTHER's family GO-slim** returns `GO:0004222` + `PC00153 metalloprotease` for Q8N6G6 — but
  the identical block comes back for *every* member queried, including ADAMTS5 (a real protease)
  and ADAMTSL2 (which carries an explicit negated `GO:0004222` in GOA). It is a family summary
  and it is not negation-aware; it is **not** PANTHER individually calling ADAMTSL1 a protease.
- **InterPro2GO got this right**: `IPR013273` is the only one of ADAMTSL1's fourteen InterPro
  entries with any GO mapping, and it maps to `GO:0030198` alone despite spanning the catalytic
  ADAMTS proteases.

And PAINT already knows: node `PTN002673039` carries an **IKR** loss of `GO:0004222` and an
**IRD** loss of `GO:0006508`. In GOA that judgement surfaces as exactly one annotation —
ADAMTSL2's negated `GO:0004222` — and on no other ADAMTS-like member.

### Headline finding: a PAINT propagation gap that lands on the one gene with the best evidence

Scripted census (`ADAMTSL1-bioinformatics/check_family_propagation.py`, live QuickGO + UniProt +
cached PANTHER files):

- `GO:0031012 extracellular matrix` is an **IBD at node `PTN000347317`** (16 gene sources).
- **24 of the 26** human members of PTHR13723 receive it by IBA from that node.
- PAPLN does not, but holds the term from its own evidence — expected PAINT behaviour.
- **ADAMTSL1 has no `GO:0031012` annotation of any kind, from any evidence code. It is the only
  human member of the family in that position.**
- The gap is **species-specific, not subfamily-specific**: mouse Adamtsl1 (`Q8BLI0`), same
  PANTHER subfamily SF157, *does* receive the IBA from the same node, plus three HDA rows.

Meanwhile the human evidence is the strongest in the family: UniProt cites three `ECO:0000269`
references for the location, `PMID:11805097` shows punctate deposition into the cell substratum
(the observation the protein is named for), and `PMID:24281761` recovered **endogenous, untagged**
ADAMTSL1 from human fibroblast secretomes as a direct MMP10 substrate. Hence the `NEW` row rather
than leaving it to PAINT. The tagged/heterologous caveat on the 2002 assay is stated in the row's
reason, which is why the secretome result is cited alongside it.

### Isoforms

The brief's premise was inverted, so stating it plainly: **punctin-1 is the SHORT splice form**
(525 aa, `Q8N6G6-1`); the canonical displayed sequence is the long isoform 3 (1762 aa). UniProt
applies "Punctin-1" as an entry-level AltName, which is the source of the confusion.

This matters because **essentially every ADAMTSL1 experiment used isoform 1** — the 2002
purification, all glycosylation site mapping, all secretion mutants, and the Trp42Arg disease
variant assay, the last explicitly because "A plasmid construct for full-length ADAMTSL1 is
presently unavailable." `isoform: Q8N6G6-1` is therefore set on the `NEW` row to record *what was
tested*, per CLAUDE.md, not to assert isoform restriction. Isoforms 5 and 6 delete residues
1–1299 including the signal peptide and so would not be secreted at all; recorded in
`functional_isoforms`.

### Checks run, including the negatives

- **Retractions / errata / expressions of concern**: all twelve PMIDs checked via each article's
  own `CommentsCorrections` record (a pubtype search misses Publisher Corrections). **None
  flagged.**
- **IntAct**: 16 records reduce to **7 distinct partners**, each counted twice because BioPlex 2.0
  and 3.0 are the same anti-tag co-IP pipeline; all MI-score 0.35 in over-expressing HEK293T,
  several partners cytosol- or peroxisome-facing for a signal-peptide protein. GOA carries **zero**
  `GO:0005515` rows and none of this should be imported. B3GLCT is an expected enzyme–substrate
  encounter; FBN2 is kept as a hypothesis only.
- **Reference projection test**: each Reactome reference annotates ~9 gene products with the same
  three terms — set-membership projection, not independent findings.
- **Sibling-paralog consistency**: the merged ADAMTSL4 review resolves the *identical* three ER
  rows as ACCEPT while its reason calls them "secretory pathway transit, not a functional
  localization". Same substantive judgement, different action. Declared here and raised in
  `suggested_questions` rather than diverging silently.
- **Downward-MODIFY (ACRV1 pattern)**: no IBA rows exist, so no propagation lands above its donor.
  Negative by absence of the row class.
- **Marker vs function**: `PMID:35115729` / `PMID:39880678` use *Adamtsl1* only as a Schwann-cell
  cluster marker, and the second ablates the **cell population**, not the gene. Deliberately not
  annotated.
- **Worm orthologue**: `PMID:22014523` itself says "The biological role of MADD-4 orthologs,
  including ADAMTSL1 and 3 in mammals, is unknown." Not carried into any human annotation.

### A number that refused to add up

I wrote "fifteen experimental sources" for the `GO:0030198` IBD from the GOA WITH/FROM field, then
counted the PAINT seed list and got **14**. The gap is exactly the `PANTHER:PTN000347317` token
that GOA appends to WITH/FROM. Corrected in both the review and the notes, and the seed
composition is now **computed** by the script (section 4 of RESULTS.md) instead of counted by eye.
Likewise "three human family members" became the derived "22 human members, three of them in the
ADAMTS-like branch".

### Reproducibility of the analysis

`check_family_propagation.py` hardcodes no results, aborts if a QuickGO response is paginated,
and asserts each claim RESULTS.md makes (ADAMTSL1 still has zero `GO:0031012` rows; mouse
Adamtsl1 still has the IBA; ADAMTSL1 still carries KW-0378 with no CATALYTIC ACTIVITY and the
CAUTION intact; the IKR row is still in the cached PAINT table). Two consecutive runs reproduce
RESULTS.md byte for byte, and the guards were exercised by deliberately corrupting each cached
input — each mutation asserts its target is present first, so it cannot be a silent no-op.

### What changed after review

Recorded here so this description does not outlive the file it describes.

- **`GO:0030198` moved `ACCEPT` -> `KEEP_AS_NON_CORE`.** The first draft rested the acceptance
  partly on the annotation being "well-constructed and unrefuted", which is absence of
  contradiction rather than support. The reason now rests on three positive but strictly
  family-level grounds: the InterPro2GO mapping restraint, a non-catalytic route available to
  the architecture, and PAINT's independent IBD at `PTN000347317` — shown to sit *above*
  ADAMTSL1 via subfamily SF157 rather than merely somewhere in the family. `core_functions`
  accordingly drops `directly_involved_in` and keeps only `locations: GO:0031012`, the one
  claim the gene's own data establishes.

- **A measured negative was found and folded in — `PMID:22242013`.** Surfaced by the concurrent
  ADAMTSL3 review, not by me: the paper is titled for fibrillin-1, so no ADAMTSL1-keyed search
  reaches it, and it is absent from ADAMTSL1's GOA and from the affinage record. It holds the
  only molecular-function experiment ever performed on human ADAMTSL1, and it is negative twice:
  recombinant human ADAMTSL-1 did not bind the N-terminal half of fibrillin-1 in a panel where
  ADAMTSL-2, ADAMTSL-3 and papilin did, and did not bind the non-catalytic region of ADAMTS-10.
  Both partners that place the rest of the family in the fibrillin-1 pathway are excluded for
  this one. Ground two above is qualified accordingly; the verdict stays `KEEP_AS_NON_CORE`
  because `GO:0030198` is much broader than fibrillin-1 binding, so excluding one mechanism is
  not refuting the process. Bounded by the isoform caveat: the methods say "full length" without
  a length, and the Apte lab stated in 2017 that no 1762-residue construct existed. The proposed
  binding experiment was rewritten — as first drafted it would have repeated a published negative.

- **A wrong causal story was removed from the census.** The report had explained PAPLN's missing
  IBA as PAINT declining to overlay an existing direct annotation. PAPLN's only `GO:0031012` rows
  are IEA and TAS, neither experimental, so that cannot be the reason. The script now *tests* the
  condition and renders whichever branch the data selects. Converges independently with #2305.

- `PMID:24634412` and `PMID:39880678` added as `references` with `reference_review` recording why
  neither is annotated, and the `GO:0030198` census figure is computed rather than hand-counted.

**Cross-review note.** I was given the premise that ADAMTSL1 holds the `GO:0030198` IBA while
ADAMTSL5 does not, which would justify the two reviews' differing verdicts on the identical row.
That premise is false, and #2305's own table agrees: the IBA reaches only ADAMTSL2, ADAMTSL4 and
THSD4. The two genes are in identical evidentiary positions, so the divergence is filed once in
`suggested_questions` as a family-level line-drawing question rather than dressed up as a per-gene
distinction, and the script asserts the parity so it cannot go stale.

### Gates

- `checkquotes.py`: **42 quotes, 0 problems**; notes file: 18 inline quotes, 0 problems.
- `just validate human ADAMTSL1`: **✓ Valid**, no warnings, run after merging `origin/main`.
- Strict duplicate-key YAML load: clean. Raw vs parsed `reference_id` counts reconcile (23 raw
  grep lines − 5 `original_reference_id` = 18 = parsed).
- `cache/go/terms.csv`: untouched by this branch; merge-base diff shows no deletions; in-file
  duplicates are exactly the two pre-existing on main (`GO:0001675`, `GO:0009566`).
- No rendered HTML committed.

### For the campaign brief

- **The pre-write validation hook resolves `file:` references against `$CLAUDE_PROJECT_DIR`, not
  against the agent's own worktree.** Any review in a fresh worktree that cites
  `file:human/<GENE>/...` for a *new* gene folder is hard-blocked by the hook even though
  `just validate` passes in the worktree. Route the write through Bash and run the real validator.
- **`GO:0005615 extracellular space` was obsoleted on 2026-03-06**, `replaced_by GO:0005576
  extracellular region`. Worth knowing before proposing a secreted-protein location term.
- **`fetch-gene` truncates `alternative_products[].sequence_note`** when the UniProt
  `ALTERNATIVE PRODUCTS` list wraps across a `CC` continuation: isoform 5 was seeded as
  `VSP_039322, VSP_039329, VSP_039330,` with `VSP_039331` dropped. Corrected by hand here.

---

> [!NOTE]
> **Low Risk**
> Curatorial YAML, reference caches, and an offline analysis script only; no changes to runtime services, auth, or shared production code paths.
> 
> **Overview**
> Adds a **complete human ADAMTSL1** (Q8N6G6) annotation review package: `ADAMTSL1-ai-review.yaml`, curator notes, cached GOA/UniProt/Affinage inputs, and PMID stubs for the cited literature.
> 
> The review documents that **ADAMTSL1 is the only human PTHR13723 member with no `GO:0031012` extracellular matrix** annotation despite direct human ECM evidence, and **proposes a new IDA** `located_in GO:0031012` (PMID:11805097, isoform Q8N6G6-1). It **reconciles the four thin GOA rows**: InterPro `GO:0030198` and three Reactome ER-lumen TAS rows are **KEEP_AS_NON_CORE**; core function is limited to ECM localization. It also records a **fibrillin-1 SPR negative** (PMID:22242013) qualifying family-level matrix-organization inference, and flags UniProt **Hydrolase** keyword vs non-catalytic domain structure.
> 
> **`check_family_propagation.py`** plus `RESULTS.md` / `results.json` recompute PANTHER/QuickGO/UniProt census with guards so propagation-gap claims cannot go stale silently.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d6826e4d674d9ad9a39e498287b3964fa4ff855. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Both 🟡 items fixed, all three 🔵 taken, and one factual correction to a cross-review premise I was handed. Single push: `a96d7ec25`.

**1. `core_functions` asserted `directly_involved_in` with nothing supporting it — you're right, and the fix went further than patching the provenance.**

The deeper problem was the one you named in passing: `"well-constructed and unrefuted"` is absence of contradiction, and I had used it as the stated basis for `ACCEPT`. So rather than only wiring provenance into the existing structure, I changed what the review claims:

- `GO:0030198` is now **`KEEP_AS_NON_CORE`**, not `ACCEPT`. `ACCEPT` asserts the term as the gene's core function, and the evidence does not reach that.
- The reason now rests on three *positive* family-level grounds — the InterPro2GO mapping restraint, the demonstrably available non-catalytic route via ADAMTSL2/ADAMTSL4/THSD4, and PAINT's independent IBD at `PTN000347317` — and says explicitly that all three are family-level and that is the limit of the case.
- `PTHR13723 PTN000347317 GO:0030198 P IBD false` is attached as `supported_by` on the row.
- `core_functions` **drops `directly_involved_in`** and keeps only `locations: GO:0031012`, the one claim the gene's own data establishes. Its description says why the BP term is deliberately not repeated there.

That is your option 2 plus the provenance from option 1, and it removes the prose/structure mismatch in both directions rather than only one.

**2. `PMID:24634412` and `PMID:39880678` added to `references` with `reference_review`.**

Not annotated, and the reasoning is now in the YAML rather than only the notes. For the chondrosarcoma paper, three grounds: the cached record is abstract-only and **does not state the direction of the effect**, so neither `GO:0008284` nor `GO:0008285` can be chosen without guessing; the assay is a cancer cell line with no normal-physiology counterpart; and a proliferation phenotype narrows neither recorded gap, since it identifies no molecular function and says nothing about matrix organisation. Your framing prompted a useful precision fix: the review's claim is now stated as *no molecular function and no matrix-organisation evidence*, not *no functional observation of any kind*.

**3. 🔵 The "22" is now computed, and fixing that caught a real error of mine.**

`ECM_ORG_TERM` added and the census does a second pass over `GO:0030198` (RESULTS.md §1). While doing it I found my PAPLN sentence was wrong. I had written that PAPLN lacks the IBA but "carries the term from its own evidence, which is the expected PAINT behaviour". **PAPLN's only `GO:0031012` rows are IEA and TAS — neither experimental — so redundancy suppression cannot account for the missing IBA.** The script now derives that (it tests the non-IBA rows against the experimental code set) instead of asserting a causal story. This puts me in agreement with #2305, which reads PAPLN as a second coverage gap and sharpens it further by noting three PAPLN orthologs are themselves seeds at the node.

**4. 🔵 ADAMTS spacer sourced.** Correct that it is not an FT feature. It comes from `DR   Pfam; PF05986; ADAMTS_spacer1; 1.` and `DR   InterPro; IPR010294; ADAMTS_spacer1.`, now a sourced finding on the UniProt reference that says explicitly why it is absent from the feature-table enumeration.

**5. 🔵 `NEW` row — partially taken, and a reason for not taking the rest.** The row `summary` now names the endogenous secretome evidence, so it is visible at row level. I did **not** add a second row, because the two references support different terms: a fibroblast secretome is conditioned medium, which supports `GO:0005576 extracellular region`, whereas `GO:0031012` requires deposited matrix — which is what the punctate substratum result shows and the secretome does not. Making the HDA row primary would attach secretome evidence to a term it does not specifically demonstrate, and adding `GO:0005576` alongside would annotate a `part_of` parent redundantly. So `PMID:11805097`/IDA stays primary with the tagged-heterologous caveat stated, and `PMID:24281761` corroborates. Happy to be overruled if you read the secretome fractionation as matrix-specific.

---

**Correction to a cross-review premise, since it bears on a verdict.**

I was told the ADAMTSL5 review's `MARK_AS_OVER_ANNOTATED` and my verdict could both stand because ADAMTSL1 has the `GO:0030198` IBA and ADAMTSL5 does not. **That is false**, and #2305's own table agrees — ADAMTSL1 is `–` in its `GO:0030198` column. My census independently gives the same answer: the IBA from `PTN000347317` reaches only ADAMTSL2, ADAMTSL4 and THSD4 within the ADAMTS-like branch. **ADAMTSL1 and ADAMTSL5 hold the InterPro IEA and nothing else — identical evidentiary positions** — and if anything ADAMTSL5 has *more* gene-specific matrix evidence (its own IDA to `GO:0031012`, plus microfibril and heparin binding).

So the divergence cannot be justified per gene. It is about where the family draws the `KEEP_AS_NON_CORE` / `MARK_AS_OVER_ANNOTATED` line, and it is filed once in `suggested_questions` to be settled for the family rather than gene by gene. The script now **asserts the parity** (`ADAMTSL1.ecm_org_iba == ADAMTSL5.ecm_org_iba == 0`) so the comparison cannot silently go stale.

Where #2305 and I agree, for the record: it flagged "ADAMTSL1 has no IBA at all; if that review reports IBA support, that is a discrepancy worth resolving" — no discrepancy, my review reports the *absence* and never claims IBA support. And its principle that absence of an IBA at an incoherently-propagating node is a coverage gap rather than a curatorial judgement is exactly how my headline uses it.

**On the 🔵 `isoform` slot doc drift** — agreed, and thank you for checking it rather than flagging it as an error. CLAUDE.md ("indicates what was tested") and the schema slot description ("only when the annotation is isoform-specific") do contradict each other. Reported for the campaign brief; nothing changed here.

**Gates re-run after the merge of `origin/main`:** `checkquotes.py` 46 quotes / 0 problems; notes file 18 inline quotes / 0 problems; `just validate human ADAMTSL1` ✓ Valid with no warnings; strict duplicate-key load clean with raw and parsed `reference_id` counts both 20; `cache/go/terms.csv` untouched, no deletions against the merge base, in-file duplicates exactly the two pre-existing on main. RESULTS.md reproduces byte-for-byte across consecutive runs, and I re-broke the cached PAINT table to confirm the IKR guard still aborts.

cmungall:
Two commits since the approval, both of which I judged worth invalidating it for. `9f0d354d2` takes your residual 🔵; `d863f7c92` folds in a paper that changes part of the argument.

**🔵 residual — taken. You were right that it was the load-bearing step.**

"That node covers ADAMTSL1" is now stated with its derivation: mouse Adamtsl1 is in PANTHER subfamily SF157, the same subfamily as human ADAMTSL1, and it receives the `GO:0031012` IBA **from `PTN000347317`**, so the node is an ancestor of SF157 and therefore of ADAMTSL1. The reason now says explicitly that this is what distinguishes "PAINT reached the term at a node above this gene" from "PAINT reached it somewhere in the family". Your related note about the 4-of-26 vs 2-of-26 asymmetry is recorded in the gene notes, with the point that this review does not lean on the `GO:0030198` propagation pattern as a coverage gap and only claims node ancestry.

**A measured negative I had missed — and I did not find it, PR #2305 did.**

`PMID:22242013` is titled *"Microenvironmental regulation by fibrillin-1"*. It is absent from ADAMTSL1's GOA, absent from the affinage record, and unreachable by any ADAMTSL1-keyed search. It contains the only molecular-function experiment ever performed on human ADAMTSL1, and it is negative twice, both verified verbatim against the cached full text:

> ADAMTSL-2, -3, and papilin polypeptides interacted with the N-terminal half of fibrillin-1, while ADAMTSL-1 did not.

> However, neither ADAMTSL-2 nor -1 bound to ADAMTS-10, indicating that ADAMTS enzymes may partner only with specific ADAMTSL proteins.

So **both** partners that place ADAMTSL2, ADAMTSL3 and THSD4 in the fibrillin-1 pathway have been tested for ADAMTSL1 and both were negative. That is a measured negative rather than an absence, which makes it far stronger than the silence the rest of this review works against.

What it changed:

- **Ground two of the `GO:0030198` reason is qualified.** I had written that a non-catalytic route to the term is "demonstrably available to this architecture". True in principle, but the *specific* route the relatives use is excluded for this protein, and the reason now says so. The verdict stays `KEEP_AS_NON_CORE` rather than moving to `MARK_AS_OVER_ANNOTATED`: `GO:0030198` is much broader than fibrillin-1 binding, and excluding one mechanism is not refuting the process. It does remove the strongest mechanistic analogy, which is a further reason the term cannot be core.
- **The MF `knowledge_gap` is now bounded on both sides** — what is established, *and* what has been measured and excluded — which is what `boundary` is for and what I had left one-sided.
- **The proposed binding experiment was rewritten.** As I first drafted it, it proposed SPR of ADAMTSL1 against fibrillin-1 — which would have repeated a published negative. It is now an unbiased partner search, with fibrillin-1 and ADAMTS-10 retained only as controls and fibrillin-2 / LTBPs / heparin as the untested candidates.
- **Bounded, not blanket.** The methods call the reagent "Recombinant full length ADAMTSL-1" but give no length, and the Apte lab stated in 2017 that no construct for the 1762-residue form existed. Both RefSeqs existed in 2012 (`NP_443098` = 525 aa isoform 1, `NP_001035362` = 1762 aa isoform 3), so the paper does not settle which was assayed; the balance favours punctin-1. Recorded in `reference_review` as firm for what was tested and probably untested for the long form — the same isoform caveat that runs through the whole review.

This is the campaign's "a paper titled for something else holds your gene's answer" lesson in a form I had not guarded against: not a paralog-titled paper, a **partner**-titled one. My retraction/erratum sweep, my IntAct check and my WITH/FROM resolution would none of them have surfaced it. The thing that did was a sibling review of a different gene in the same family.

**Also worth flagging for the campaign, since it nearly caused real damage:** `/tmp` is shared between concurrently running agents. My `/tmp/pr_body.md` was overwritten by the ADAMTSL3 agent's PR body, and I ran `gh pr edit` against it before noticing. It failed only because `gh pr edit` is currently broken on this repo — it exits 1 with a `Projects (classic)` GraphQL deprecation error regardless of content — so the sibling's body never landed on this PR. Two corrections on my side: scratch files now go in a worktree-private directory, and I stopped filtering that error out of the output, which is what let me read a stale body and believe an edit had applied. The body is now updated via `gh api -X PATCH`, which does not touch the deprecated field.

**Gates re-run after merging `origin/main`:** `checkquotes.py` 48 quotes / 0 problems; notes 20 inline quotes / 0 problems; `just validate human ADAMTSL1` ✓ Valid, no warnings; strict duplicate-key load clean with raw and parsed `reference_id` both 20 and no YAML anchors/aliases; `cache/go/terms.csv` untouched, no deletions against the merge base, duplicates exactly the two pre-existing on main; no HTML committed.

cmungall:
All three precision points taken; `ff4b8efab`. No term, action or evidence code changed.

**1. The ADAMTS-10 negative does not discriminate — you're right, and I was double-counting.** ADAMTSL-2 is equally negative against ADAMTS-10 and still reaches `GO:0030198`, ADAMTS-10 binding was shown only for ADAMTSL-3, and THSD4 was never tested against it. The reason now says so explicitly and states that the fibrillin-1 negative alone is load-bearing. The conclusion is unchanged because it survives on that one result.

**2. The parity sentence — narrowed, and the direction carried over.** This is the sharper observation of the three: the claim was true when I wrote it and the paper I added falsified it. Now "the same **IBA** position", which still holds and is still assertion-guarded, plus the wider difference stated with its sign — ADAMTSL5 has its own IDA to `GO:0031012` plus microfibril and heparin binding, while ADAMTSL1 has a measured negative against the very route by which the family reaches `GO:0030198`. **The better-supported gene is the one taking the harsher action**, so the family-wide question is no longer a tie to be broken but an argument with a direction, and both the reason and the `suggested_questions` entry now say that. I have kept `KEEP_AS_NON_CORE` — one excluded mechanism does not refute a process much broader than it — but the honest position is that the evidence moved toward ADAMTSL5's reading, and the review now says so rather than leaving the reader to notice.

**3. "The LTBPs remain untested" — the same error the commit was written to fix.** LTBP-1 and LTBP-4 were coupled to CM5 chips in those very experiments and the LTBP results are in a supplementary table absent from the cached record. Changed to "not reported in that study's main text", with the caveat spelled out, and the proposed experiment now says to check the supplement before assaying them. Asserting a second candidate untested on the paper that had just taught me a candidate can look untested and not be would have been the identical mistake twice in one commit.

**4.** Line 607 rewrapped; the file's longest line is now 95 characters. The re-wrap was verified not to change content by parsing before and after and comparing every string value under whitespace normalisation, rather than by eye — the first attempt at that rewrap silently swallowed nested mapping keys into a folded scalar, and the equality check is what caught it.

**Gates re-run after merging `origin/main`:** `checkquotes.py` 48 quotes / 0 problems; notes 20 inline quotes / 0 problems; `just validate human ADAMTSL1` ✓ Valid with no warnings; strict duplicate-key load clean, raw and parsed `reference_id` both 20, no YAML anchors or aliases; `cache/go/terms.csv` untouched with no deletions against the merge base and only the two pre-existing duplicates; no HTML committed.

cmungall:
Both residuals fixed; `3f71f2d4a`. Neither was cosmetic — one lost information — so they were worth a round.

**1. `knowledge_gaps.resolution`.** You're right, and the failure is worth naming: my LTBP substitution ran through the middle of the fibrillin-2 clause and left it without a predicate, so **fibrillin-2's untested status silently dropped out of that field** while the parallel `suggested_experiments` text kept it. Both are now stated separately: fibrillin-2 is not reported as tested anywhere and FBN2 is in the affinity-purification hits, so it is the one family-shaped hypothesis still standing; the LTBPs are separately flagged as "not in the main text" with the supplementary-table caveat.

**2. The question's closing sentence.** Also right — it was a leftover from the pre-narrowing draft and, as written, denied the difference the same entry had just enumerated. Replaced with the form that is actually true and more useful: the difference between the genes is real but **points the wrong way to explain the divergence**, since it would justify the harsher action on ADAMTSL1 rather than the milder one, so it cannot be what separates the two verdicts.

Both are the same failure mode — string replacement across wrapped YAML producing new garbage — and both survived a diff read because the damaged text looked plausible on the line it sat on. What caught them was your reading the *parsed* sentences rather than the hunks. I have adopted that as the check: this repair was verified by parsing the two fields and printing them end to end, plus an assertion that only `knowledge_gaps` and `suggested_questions` differ under whitespace normalisation.

**Gates:** `checkquotes.py` 48 quotes / 0 problems; `just validate human ADAMTSL1` ✓ Valid, no warnings, after merging `origin/main`; longest line 95 characters; `cache/go/terms.csv` untouched with only the two pre-existing duplicates; no HTML committed.

Nothing further from me — this is my last push unless something new comes up in review.

cmungall:
Both taken; `c17de8bb7`. You were right that these are the same class one level down, and the first was a stale factual claim rather than a presentation issue.

**1. The nested BP gap boundary was left behind by the fold-in.** It still carried the unqualified form of ground two, and still said "Not established: any ADAMTSL1 loss-of-function, **binding** or assembly result bearing on matrix organisation" — contradicting its own row's reason, which now calls exactly such a binding result load-bearing. It now records the fibrillin-1 exclusion with the isoform caveat, and the "not established" clause is narrowed to loss-of-function and assembly results plus the absence of any *positive* binding result. That is the third recurrence of parallel-field staleness in this PR (the `full_text_unavailable`-shaped problem: I updated the field I was thinking about and not its twin), and all three were caught by you rather than by any gate.

**2. Traceability.** `PMID:22242013` is now in the row's `supported_by` and in both `provenance` lists, and the two fields that said "that study" name it, so the `resolution` field reads correctly standing alone.

**A gap in `checkquotes.py`, found by a number that would not add up.** I added three quotes and the checker's total went from 48 to **49**. Investigating instead of accepting it: the checker walks `supported_by` and `findings` only, so **`provenance` lists are invisible to it**. For this file 17 `supported_by` + 32 `findings` = 49, exactly what it reports, and **6 `provenance` quotes are unchecked**. Both counts now reconcile independently, so there is no unexplained residue. I verified all 6 with a walker that matches on the *shape* of an entry — any dict carrying `reference_id` + `supporting_text` — rather than on its parent key; all pass. Recorded in the gene notes, because any review using `knowledge_gaps` provenance and relying on `checkquotes.py` alone is not checking those quotes. It is the same detector/mutator scope-divergence shape as the two residuals above.

**Gates:** `checkquotes.py` 49 / 0 problems, plus the shape-based walker 23 / 0 covering provenance; notes 20 inline quotes / 0; `just validate human ADAMTSL1` ✓ Valid, no warnings, after merging `origin/main`; longest line 95; `cache/go/terms.csv` untouched, only the two pre-existing duplicates; no HTML.

cmungall:
All three taken; `9f53c1190`. Two were over-claims of mine and the third let me state a negative at its real strength. I also answered your gate question by experiment, and it corrects something I asserted.

**1. The boundary was arguing for an action the row did not take.** Correct, and it is the sharper of the two: I attached "tested and excluded" to *the non-catalytic matrix-organising role* when the experiment excludes only *the fibrillin-1 route*, which is the broader claim the same row's reason explicitly refuses. Rescoped, and the boundary now ends that clause by saying why it matters — "that excludes the route, not the process, which is why the row is kept rather than marked over-annotated" — so the field argues for the action actually taken instead of against it.

**2. "No positive binding result of any kind" — too broad, agreed.** The same file treats the FBN2 affinity-purification hit as the one family-shaped hypothesis still standing. Now "no validated binding partner", with the FBN2 caveat stated inline so the two fields agree.

**3. The negative is wider than I wrote, and I have widened it.** *"Binding to the C-terminal half of fibrillin-1 was negative for all ADAMTSL proteins tested"* — so the ADAMTSL-1 negative covers fibrillin-1 entire, not just the N-terminal half. Added as a third quoted finding on the reference and reflected in the boundary. You were right that this is what makes "the strongest mechanistic analogy is removed" as strong as the reason claims; understating it was safe but weaker than the evidence.

**Your gate question, answered by breaking it — and it corrects me.** I reported the provenance gap in a way that implied a project-wide risk. That was wrong. `linkml-reference-validator`, the gate that actually runs in CI via `just validate`, **does** walk `knowledge_gaps` provenance. Injecting `THIS SENTENCE APPEARS IN NO PUBLICATION ANYWHERE ZZQQXX.` as a provenance `supporting_text` flips `just validate` from `✓ Valid` to:

```
✗ Invalid
  ✗ [ERROR] Reference validation failed: Text part not found as substring:
    'THIS SENTENCE APPEARS IN NO PUBLICATION ANYWHERE ZZQQXX.'
```

The probe asserted its anchor was present before mutating, and the tree was restored to a clean `git diff` after. So it is a **local-script gap in `checkquotes.py` only**, not a project-gate gap, and the notes now say that rather than the broader thing I first wrote. Your point about the shape-based walker is also right and I have not claimed otherwise: keying on `reference_id` + `supporting_text` in one dict misses `findings`, which inherit their id from the parent references entry — the two walkers are complementary, and it is `just validate` that covers both.

**Gates:** `checkquotes.py` 50 / 0; shape-walker 23 / 0 covering provenance; notes 20 inline quotes / 0; `just validate human ADAMTSL1` ✓ Valid, no warnings, after merging `origin/main`; longest line 95; `cache/go/terms.csv` untouched with only the two pre-existing duplicates; no HTML.

cmungall:
Both taken, plus a fifth instance of the same defect that this review had not found; `f98b93cf2`.

**1. The `findings` statement — you're right, and it was wrong on the paper as well as stale.** Rewritten so the ADAMTS-10 result is corroborating rather than discriminating, and with your correction folded in: ADAMTS-10 binding was shown only for ADAMTSL-3, ADAMTSL-2 is equally negative against it, and ADAMTSL-6/THSD4 was never tested against it — so ADAMTS-10 is not what places ADAMTSL-2 or -6 in the fibrillin-1 pathway, which my sentence had implied.

**2. `(data not shown)` — the C-terminal negative now carries that caveat**, so it sits at a visibly lower evidentiary standard than the N-terminal result, which has the sensorgrams and is the load-bearing one. Good catch on the inconsistency with how I had treated Table S2. I also harmonised the row reason with the boundary while I was there, carrying the same caveat, so the two no longer state the negative at different scopes.

**3. A fifth instance you did not flag, found by changing method.** After four rounds of the same defect I stopped fixing the field in front of me and swept **every string value in the parsed document** against a list of every phrasing I had retracted during this PR. It found one more: the **top-level MF gap boundary** still carried the double-count framing, in a field that had already been rewritten twice without anyone noticing. Now fixed, and the sweep reports **0 hits across all seven** retracted phrasings:

```
two partners that place | both been tested for ADAMTSL1 and both were negative |
exactly the same evidentiary position | well-constructed and unrefuted |
LTBPs remain untested | no positive binding result of any kind |
divergence is not attributable to a difference between the genes
```

Your count was right and worth stating plainly: this is the ACBD3 failure — one claim asserted at N sites with no generation relationship between them — reproduced five times in a review whose author had read that lesson. The instances were reason vs `core_functions`; MF boundary vs BP boundary; the question tail vs its own opening; `resolution` vs `suggested_experiments`; and `findings` vs reason. The transferable part is not "be careful", it is that **the unit of correction has to be the document, not the field**, and that a retraction list swept over all prose is cheap enough to run every round. Written up in the gene notes. The notes themselves legitimately still contain three of those strings, because explaining a retraction means quoting it — so the sweep belongs on the review YAML, not on prose about it.

**CI:** `Build and test` has since gone green on `9f53c1190`, and this delta changes prose only.

**Gates:** `checkquotes.py` 50 / 0; shape-walker covering provenance 23 / 0; retracted-phrasing sweep 0 / 7; `just validate human ADAMTSL1` ✓ Valid, no warnings, after merging `origin/main`; longest line 95; `cache/go/terms.csv` untouched with only the two pre-existing duplicates; no HTML.

cmungall:
Both correct; `9fe0a5635`.

**1. The sensorgram claim was an over-claim and I have narrowed it to what the figure contains.** Verified against the legend: Figure 5a shows traces for the **binders only** — full-length ADAMTSL-2 (320-0 nM) and the C-terminal end of ADAMTSL-3 (80-0 nM) against rF90, plus the rF90WMΔ control — so the paper contains **no displayed ADAMTSL-1 trace for either half**. The file now states exactly the distinction that survives: the C-terminal result is explicitly flagged data not shown, whereas the N-terminal ADAMTSL-1 negative is reported as an outcome of a panel that is displayed. Read literally my sentence attributed figure data to ADAMTSL-1 that does not exist, which is precisely the error class this review has been correcting all round.

**2. The sixth instance — and you identified why my sweep could not have caught it.** That is the most useful thing in this review round. A retraction list looks for **old** phrasing left standing; instance six is **new** phrasing propagated to some siblings and not others, which is the mirror image and structurally invisible to it. So the fix is guarded differently: the edit now **enumerates the sibling set** for the C-terminal claim — `references.findings[1]`, `existing_annotations[0].review.reason`, and `existing_annotations[0].review.knowledge_gaps[0].boundary` — and asserts that **both the claim and its caveat appear in all three** before the write is accepted. It fired usefully while I was writing it, since the boundary was indeed missing the caveat.

Two directions, two checks: **sweep for what should be gone, assert co-occurrence for what should be everywhere.** Neither alone is sufficient, and I had only the first. Written up in the notes alongside the five-instance history, because the pair is the transferable part rather than either check on its own.

**Gates:** `checkquotes.py` 50 / 0; shape-walker covering provenance 23 / 0; retracted-phrasing sweep 0 hits / 7 patterns; sibling co-occurrence assertion passes for all 3 fields; `just validate human ADAMTSL1` ✓ Valid, no warnings, after merging `origin/main`; longest line 95; `cache/go/terms.csv` untouched with only the two pre-existing duplicates; no HTML.

cmungall:
Both taken; `9fe0a5635` → `23e822e42`. The general point was the valuable one and I have acted on it rather than just patching the instance.

**The co-occurrence check failed, and for exactly the reason you gave.** My sibling list had three members; the widened claim was load-bearing in a fourth. The top-level MF boundary concluded "what is excluded is fibrillin-1 binding" while stating the evidence as the N-terminal half only — so it used the widened scope without carrying what licenses it. That is not the mild pattern of the previous round; it is the one field that draws the conclusion without the premise. Fixed, with the C-terminal result and its data-not-shown caveat now stated where the conclusion is drawn.

**And I have stopped hand-listing siblings.** You are right that enumerating them is the same judgement that produced the six instances, so a list-based control inherits the failure it is meant to catch. The check now **derives** the set: collect every string field in the parsed document, select those that draw the widened conclusion (`excluded is fibrillin-1 binding` / `covers fibrillin-1 entire`), and require each to state the basis (`C-terminal half`). Run against the pre-fix file it flagged `.knowledge_gaps[0].boundary` immediately; it now passes on three fields — the row reason, the row's BP gap boundary, and the MF gap boundary. Two guards keep it from passing vacuously: it asserts the selector matched something at all, and it keys on **the conclusion** rather than on field paths, so the same claim copied into a new field is caught without anyone updating a list.

The generalisable form, which is now in the notes next to the six-instance history: **select the fields by what they assert, not by where they live.** Sweep for what should be gone, derive-and-assert for what should be everywhere.

**Cosmetic point taken too** — the stacked dash-clauses are split, so the sentence no longer briefly asserts "a panel that is displayed" before the next clause withdraws ADAMTSL-1 from it.

**CI:** `Build and test` went green on `9fe0a5635` after your review ran; this delta is prose only.

**Gates:** `checkquotes.py` 50 / 0; shape-walker covering provenance 23 / 0; retracted-phrasing sweep 0 hits / 7 patterns; derived conclusion-basis invariant passes on 3 of 3 selected fields, selector non-empty; `just validate human ADAMTSL1` ✓ Valid, no warnings, after merging `origin/main`; longest line 95; `cache/go/terms.csv` untouched with only the two pre-existing duplicates; no HTML.

cmungall:
Taken; `9d6826e4d`. Your framing of the residual is more useful than the fix, so I have recorded it as the finding rather than quietly patching.

**The selector was phrase-literal and you are right that this is the same problem one level up.** It missed the reference `findings` statement, which draws the conclusion as *"covers fibrillin-1 in its entirety, not only the N-terminal half"*. As you say, nothing is wrong in the file today — that field carries the basis and the caveat and passes on merit — but the check would have gone all-green with the caveat dropped from it, because the field asserting the conclusion in different words was never in the selected set. Broadened to four patterns; the set went 3 → 4 and all four pass, including `references[12].findings[1].statement`.

**The regression across controls is now a table in the notes**, because it is the transferable output of this review:

| control | caught | residual gap |
|---|---|---|
| fix the field in front of me | 1 at a time | 5 parallel fields left stale |
| retraction sweep over all prose | the 5th instance | cannot see *new* phrasing propagated unevenly |
| hand-listed sibling co-occurrence | 3 of 4 siblings | the list is the same judgement that caused the defect |
| conclusion-derived selector | the 4th sibling | phrase-literal, misses paraphrase |
| broadened phrase set | the paraphrase | still phrase-literal in principle |

Each control caught one more than its predecessor and left a smaller version of the same gap. **No text-matching control closes it**, because what is being checked is whether two sentences *mean* the same thing, and I have written that down rather than implying the latest one is sufficient. The other half of the honest summary: a reviewer reading the parsed fields end to end caught what every mechanical control missed, on each of the last five rounds.

**Cosmetic point taken** — the forward-pointing "below" is gone; the clause now names what it licenses instead of where the reader should look.

**CI:** `Build and test` went green on `23e822e42` after your review ran; this delta is prose only.

**Gates:** `checkquotes.py` 50 / 0; shape-walker covering provenance 23 / 0; retraction sweep 0 hits / 7 patterns; conclusion-basis invariant 4 / 4 with a non-empty-selector guard; `just validate human ADAMTSL1` ✓ Valid, no warnings, after merging `origin/main`; longest line 95; `cache/go/terms.csv` untouched with only the two pre-existing duplicates; no HTML.

I am treating the review as converged on substance — the last four rounds have not moved a term, action, evidence code or quote — so unless something material comes up, this is my final push.

PR #2305, opened 2026-07-27
Review human ADAMTSL5; PAINT node propagates incoherently across its own family

cmungall:
Review of human **ADAMTSL5** (Q6ZMM2) for the PAINT + affinage campaign. All **12** GOA rows reviewed — the `fetch-gene` stub seeded only 9 (it collapsed the three `GO:0005515` partner rows into one and the two `GO:0031012` IDA rows into one), so rows were restored to give every partner and every assigner its own verdict.

**Verdicts: 7 ACCEPT · 4 MARK_AS_OVER_ANNOTATED · 1 MODIFY**

## The domain hypothesis: established, but the predicted error did not occur

The brief flagged "ADAMTSL proteins lack the catalytic metalloprotease domain" as a lead to establish. It holds, on three independent lines: UniProt's `CAUTION`, the primary paper (*"In contrast to ADAMTS proteases, ADAMTSLs lack a catalytic domain and thus have no proteolytic activity."*), and a sequence scan finding no `HExxHxxGxxHD` zinc-binding signature and in fact **no `HExxH` substring anywhere** in the 481 residues, with no peptidase signature among its InterPro matches.

**But the predicted peptidase-annotation error is absent from this gene.** There is no `GO:0004222`, no `GO:0008237`, no `GO:0006508` row. Reported as a *hypothesis not confirmed*, not as a finding. PAINT in fact models the family's loss of catalysis correctly — the catalytic terms reached all four catalytic ADAMTS controls and none of the seven ADAMTS-like members.

## `GO:0031012` IBA — ACCEPT, on measurement rather than deference

All **16 of 16** protein donors in the WITH/FROM field carry their own **IDA/HDA** for the term or a descendant, so `SOURCE_WEAK_OR_INFERRED` would be contradicted by the data. The GOA field matches the cached PAINT seed list exactly (asserted in code). The donor set is **heterogeneous** — ECM 14, basement membrane 4, interstitial matrix 1, microfibril 1 — so `GO:0031012` **is** the LCA and no specificity upgrade applies. Checked and negative: the IBA does not land above its donors (14 of 16 hold this exact term), so no downward MODIFY.

`source_entities` are generated **from the GOA WITH/FROM field** by a committed script with a both-directions assertion, never by hand.

## `GO:0030198` IEA — MARK_AS_OVER_ANNOTATED

`InterPro:IPR013273` is named, literally, *"ADAMTS/ADAMTS-like"* (26,580 proteins) and carries exactly one GO mapping — this term. It bundles the catalytic proteases, for which ECM organization follows from ECM proteolysis, with the non-catalytic members. And the **only direct test on ADAMTSL5 was negative**: *"did not identify a consistent difference"* in microfibril density; fibronectin binding also unsupported.

**Not REMOVE, deliberately** — that negative is a *"data not shown"* result from one exogenous-protein assay, so it shows no role has been *demonstrated* rather than refuting one.

A third argument was available and **deliberately not used**: ADAMTSL5 lacks the `GO:0030198` **IBA** that ADAMTSL2/ADAMTSL4/THSD4 got from the same node. Because that node's propagation is incoherent family-wide, the absence is more likely a coverage gap than a curatorial judgement, and leaning on it would be rationalising a number I could not explain.

## `GO:0071953` elastic fiber TAS → MODIFY to `GO:0001527` microfibril

The TAS source is a **review** (`full_text_available: false`) used for **62 distinct entities**, 41 receiving this term, whose abstract never names ADAMTSL5. The same pass assigned the more specific `GO:0001527` to 15 other proteins including THSD4 — so the specific term was available and not chosen. Meanwhile **UniProt already carries `GO:0001527; C:microfibril; IDA:UniProtKB` and GOA lacks it entirely.** `GO:0001527` is `part_of` `GO:0071953`, so the parent survives by closure. The merged ADAMTSL4 review independently proposed the same term.

## Three `GO:0005515` rows — MARK_AS_OVER_ANNOTATED, decided per partner

One HuRI screen logged as **three sub-methods of the same experiment** (`two hybrid array` + `two hybrid prey pooling approach` + `validated two hybrid`, MI-score 0.56) — so UniProt's `NbExp=3` is one screen counted three ways, replicating the ACRV1 finding on a second gene.

| partner | IntAct records | distinct partners | localisation |
|---|---|---|---|
| CYSRT1 | 1670 | **517** | cornified envelope |
| KRTAP5-9 | 842 | **213** | intracellular hair-keratin matrix |
| FHL5 | 316 | **108** | nucleus |
| **ADAMTSL5** | 22 | **12** | secreted / ECM |

All three are topologically inaccessible to a signal-peptide secreted protein. All three resolve to reviewed canonical Swiss-Prot entries at canonical lengths — no ORFeome substitution (negative, recorded).

I differ here from the merged **ADAMTSL4** review, which used `REMOVE` on four comparable Y2H rows. Per campaign convention an unreplicated screen hit is `MARK_AS_OVER_ANNOTATED`; flagged as a cross-family inconsistency rather than diverging silently.

## Main finding for PAINT: one node, incoherent propagation

Node `PTN000347317` carries four IBD terms; their propagation across the eight human PTHR13723 members is incoherent:

| gene | GO:0031012 | GO:0030198 | GO:0004222 |
|---|---|---|---|
| ADAMTSL1 | – | – | – |
| ADAMTSL2 | IBA | IBA | **NOT-IBA** |
| ADAMTSL3 | IBA | – | – |
| ADAMTSL4 | IBA | IBA | – |
| **ADAMTSL5** | **IBA** | – | – |
| THSD4 | IBA | IBA | – |
| PAPLN | – | – | – |
| *ADAMTS1/9/10/17* | IBA | IBA | IBA |

`GO:0031012` reached 5 of 7 and `GO:0030198` only 3 of 7. The sharpest case is **human PAPLN, which received no IBA at all although three of its own orthologs are seeds** for `GO:0031012` at that very node — fly `Ppn`, mouse `Papln`, worm `mig-6/ppn-1`. Human ADAMTSL1 likewise has no IBA at all.

This is a **propagation-coverage** defect, distinct from the "right term, wrong node" (AADACL) and "mis-placed member" (ACTL8) defects. Filed once in `suggested_questions` naming all affected genes.

Also filed: an **ontology issue** (`GO:0001527 microfibril part_of GO:0071953 elastic fiber` contradicts its own definition — *"occurring independently or along with elastin"* — and the elastin-free ciliary zonule), and a **UniProt/GOA discrepancy** over the missing `GO:0001527` IDA.

## Cross-check with the concurrent ADAMTSL1 / ADAMTSL3 reviews

Derived independently from QuickGO, not read off their branches. **ADAMTSL3** carries a `GO:0031012` IBA *identical* to mine (same node, same 17 tokens) and shares the same `GO:0071953` TAS and InterPro `GO:0030198` IEA — all three of my non-ACCEPT verdicts have a direct counterpart there. **ADAMTSL1** has only 4 annotations and **no IBA at all**; if that review reports IBA support, that is a discrepancy worth resolving.

## Checks that came back negative (recorded so they are not re-run blind)

- **Retraction / erratum / expression of concern** on all 9 PMIDs, via publication types **and** each record's own `CommentsCorrections` (a Publisher Correction is invisible to a pubtype query): none flagged.
- **Reference projection test**: `PMID:23010571` annotates only **2** entities — gene-specific curation, not a projection.
- **Partner accession integrity**: all reviewed, canonical, correct length.
- **IBA-above-donor precision defect**: absent.

## Reproducibility

Four committed scripts under `ADAMTSL5-bioinformatics/`, all failing loudly on missing input. `audit_adamtsl5_claims.py` has a **5-case self-test that deliberately breaks each guard**; all five fire. Two of its own defects were found by running it rather than reading it: a raw-vs-parsed count guard whose scope differed from the mutator's (it counted `source_id` document-wide and flagged 18 vs 17, catching the *other* propagation_review's entity — a guard bug, not a data bug), and three hand-guessed claim counts that were all wrong and are now derived from inspected occurrences.

`just validate` passes on the branch **merged into main**; `checkquotes.py` reports 0 problems over 32 quotes; `cache/go/terms.csv` adds exactly one row with no deletions in either direction and only the two pre-existing duplicates.

---

> [!NOTE]
> **Low Risk**
> Annotation and documentation only (YAML, TSV, analysis scripts, one GO terms cache line); no runtime application or auth/data-path changes. Main review risk is curatorial judgment on GO actions, not code regressions.
> 
> **Overview**
> Adds the **human ADAMTSL5** PAINT + affinage campaign deliverable: curated GOA snapshot, UniProt extract, notes, Affinage context, and **`ADAMTSL5-ai-review.yaml`** reviewing all **12** GOA rows (restored from a 9-row stub so each `GO:0005515` partner and duplicate `GO:0031012` IDA assigner has its own verdict). Outcomes are **7 ACCEPT**, **4 MARK_AS_OVER_ANNOTATED**, **1 MODIFY**, plus a **NEW** proposal for **`GO:0001527` microfibril** (IDA from PMID:23010571; present in UniProt but absent from GOA).
> 
> Substantive annotation changes argued in the review include **MODIFY** of bulk review TAS **`GO:0071953` elastic fiber** → evidence-backed **`GO:0001527`**, **MARK_AS_OVER_ANNOTATED** on InterPro **`GO:0030198`** (non-catalytic ADAMTSL5; negative microfibril assay) and three HuRI Y2H **`GO:0005515`** partners, and **ACCEPT** of the experimental core (microfibril/heparin binding, ECM localization) and a well-measured **`GO:0031012` IBA** (16/16 donors with own IDA/HDA). Family-level **`PTN000347317`** propagation incoherence (e.g. human PAPLN/ADAMTSL1 with no IBA) is documented in `suggested_questions`.
> 
> **`ADAMTSL5-bioinformatics/`** adds scripts to resolve WITH/FROM donors, census IBA propagation across PTHR13723, inject `source_entities` from GOA, and **`audit_adamtsl5_claims.py`** (GOA coverage, NEW-term checks, quote guards) with committed JSON/RESULTS outputs. **`cache/go/terms.csv`** gains one row: **`GO:0050436` microfibril binding**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c2b49e1a112d22221f4c717be3e63c18169b39b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — all three checked. Two accepted, one declined with evidence.

**1. `NEW` row for `GO:0001527` — accepted.** You're right that `MODIFY` alone re-emits the specific term carrying `PMID:23962539`'s provenance, i.e. the best term resting on the weakest source. Added a 13th entry: `GO:0001527 microfibril` / `IDA` / `PMID:23010571` / `located_in`, quoting *"the exogenous protein co-localized with fibrillin-1 in microfibrils"* plus the UniProt DR line. I kept the `MODIFY` as well, since re-terming that TAS row is still the right action *for that row* — the two are complementary rather than redundant: one fixes an existing weak row, the other supplies the annotation GOA lacks entirely.

`existing_annotations` is now **13 = 12 GOA rows + 1 NEW**. Rather than leave that as an unexplained count change, I tightened the audit script's coverage check to reconcile it explicitly — non-NEW count must equal the GOA row count, *and* the total must equal GOA + NEW — so the surplus can't silently drift later.

**2. `core_functions[0].locations` — accepted.** Added `GO:0001527` alongside `GO:0031012`. You're right that listing only the coarser term contradicted the entry's own description ("decorates assembled fibrillin microfibrils").

**3. PANTHER node `source_status` — declining, with numbers.** I checked the repo convention across **277** PANTHER-node `source_entities` in merged reviews:

| value | uses |
|---|---|
| `SUPPORTS_SOURCE_BUT_NOT_TARGET` | 149 |
| `SUPPORTS_TRANSFER` | 57 |
| *(none)* | 26 |
| `UNRESOLVED` | 25 |
| `SOURCE_WEAK_OR_INFERRED` | 12 |
| `NOT_RELEVANT` | 8 |
| **`CIRCULAR_OR_REDUNDANT`** | **0** |

`CIRCULAR_OR_REDUNDANT` has no precedent for a tree node and would be semantically wrong here — the node is not the target and is not redundant evidence; that value is carrying the self-referential `UniProtKB:Q6ZMM2` token precisely because *that* one is the gene itself. `NOT_RELEVANT` (3% of uses) also reads wrong: the node is the single most relevant entity on the row, being the propagation locus that licenses the transfer. And `SUPPORTS_SOURCE_BUT_NOT_TARGET`, the plurality value, is used where a propagation is being *challenged* — this row is an `ACCEPT`, so it would misreport the verdict.

I take the underlying point that a node carries no evidence of its own, but the schema defines `source_status` as the "mechanical status of this source with respect to the target annotation", and for a sound node on an accepted transfer that status is that it supports it. Leaving `SUPPORTS_TRANSFER`, which is also the plurality convention for exactly this case.

**On your validation note** — fair, and worth recording: the pre-write hook and the review runner both resolve `file:` reference paths against `$CLAUDE_PROJECT_DIR`, which for an agent working in a sibling `git worktree` is *not* the directory containing the files. I ran `just validate human ADAMTSL5` in the worktree itself, on the branch merged into `main`, and it passes; CI is the gate as you say.

Gates after these changes: `just validate` ✓, `checkquotes.py` 34/34, audit 0 problems with its 5-case self-test all firing, and `cache/go/terms.csv` still adds exactly one row (`GO:0050436`) with no deletions in either direction — `GO:0001527` was already present.

cmungall:
Round 3 — accepted, you were right.

The `reconciliation` guard was a tautology: `new_rows` and `reviewed` partition `anns`, so `len(anns) == n_goa + len(new_rows)` just restates `len(reviewed) == n_goa` and can never fire independently. That is the "unreachable check that reads as coverage" failure — worse than no guard, because it looks like protection. It was also, as you noted, the only guard without a self-test case, which is exactly how it survived.

Replaced it with the invariant the narrative actually rests on: **every `NEW` proposal must name a term GOA does not already carry.** That converts "GOA is missing `GO:0001527`" from an assertion in prose into something the script checks. Added a sixth self-test case that retargets the `NEW` row at `GO:0031012` — a term that *is* a GOA row — and confirms the guard fires:

```
[PASS] duplicate key            [PASS] retracted phrasing
[PASS] broken file quote        [PASS] claim count drift
[PASS] pending action           [PASS] NEW duplicates a GOA term
```

No change to the review YAML this round — `just validate` still passes, 34/34 quotes verbatim, `cache/go/terms.csv` still adds exactly one row.

Thanks also for withdrawing the `source_status` point explicitly rather than leaving it ambiguous; the `does this entity carry evidence` vs `status relative to the target annotation` distinction is worth having on the record for the next PAINT gene.

cmungall:
Round 4 — docstring fixed. You're right that letting item 3 read as if it covered `NEW` rows is the same conflation that produced the tautology, so coverage and NEW-proposal validity are now separate numbered items with a note on why they are distinct. Docstring only; audit still reports 0 problems and all six self-test guards fire. No curation content touched.

cmungall:
Round 5 carry-over closed. All seven guards are now bannered and numbered in execution order (duplicate-key, row coverage, NEW-proposal validity, PENDING, `file:` quotes, required-claim counts, retracted phrasings), and the docstring lists the same seven in the same order — verified programmatically that the two sequences are identical, not just eyeballed. Comments only, no logic changed; audit reports 0 problems and all six self-test guards fire.

Agreed on your framing that this matters mainly because the audit is likely to be copied forward, which is exactly why I did not leave it.

On the merge commit: thanks for diffing against **both** parents rather than only main — the branch-side arithmetic (9072 + 4 = 9076) is the check that would catch a one-sided resolution, and it is a better instrument than the main-side check I had run first. I independently asserted the same union six ways before pushing (every main curie retained, every branch addition retained, nothing invented, exact set union, derived row-count arithmetic, and `duplicates_after − duplicates_on_main` empty rather than `duplicates_after` empty, so the two pre-existing `GO:0001675`/`GO:0009566` duplicates are not mistaken for new ones). Both directions agree.

Your note about `GO:0050436` landing mid-file rather than at EOF is right and worth having on the record — `terms.csv` is an unsorted append-log whose neighbouring rows are already out of both id and timestamp order, so the position carries no meaning and re-sorting it would produce spurious deletions against other branches.

No open items from my side.

PR #2303, opened 2026-07-27
Review human ACTG2; GO:0005884 actin filament never reaches the smooth muscle actins

cmungall:
## ACTG2 (P63267) — smooth muscle gamma-enteric actin

ACTG2 is the commonest cause of visceral myopathy and MMIHS5. Its GOA record has **27 rows and not one low-throughput experimental annotation**: 3 IBA, 5 IEA, 8 ISS, 3 bulk-proteomics HDA, 8 Reactome TAS.

Two opposite errors were available. Twelve actin-family reviews merged before this one argue *down* family-level actin inferences, because their subjects are 33–50 % identical to β-actin and have never been shown to polymerise — but ACTG2 is 98.7 % identical to α-skeletal actin and **38/38 identical at the filament protomer interface**, so those terms are true here. The other error is reading the disease as the function; the mechanism is explicitly dominant-negative, not loss of function, and the two are kept apart throughout.

### Headline finding: `GO:0005884 actin filament` never reaches the smooth muscle actins

Measured from the cached PAINT table plus QuickGO, not asserted. Inside PTHR11937, `GO:0005884` is asserted at exactly **three** nodes, all on 2019-03-01:

| node | IBD seed(s) | human genes reached |
|---|---|---|
| `PTN000233075` | ACTA1 | ACTA1 only |
| `PTN000748220` | ACTC1 | ACTC1 only |
| `PTN002631586` | ACTG1, mouse Actg1 | ACTB, ACTBL2, ACTG1, ACTL8, POTEE/F/I/J/KP |

Two singleton nodes plus one cytoplasmic-clade node. **No node covers the smooth muscle actins**, so QuickGO returns **zero** annotations to `GO:0005884` or any descendant for either ACTG2 or ACTA2, while every other conventional human actin has one. `PTN000748220` likewise confines `GO:0017022` myosin binding, `GO:0033275` actin-myosin filament sliding and `GO:0007015` actin filament organization to ACTC1 alone.

[...]

PR #2307, opened 2026-07-27
Review human ACTA2; actin filament and myosin binding sit on the wrong PANTHER nodes

cmungall:
Full one-to-one review of all 50 GOA rows for human **ACTA2** (P62736, smooth muscle
alpha-actin, aortic), plus 3 `NEW` proposals.

**Actions:** 6 ACCEPT · 1 MODIFY · 23 KEEP_AS_NON_CORE · 19 MARK_AS_OVER_ANNOTATED ·
1 REMOVE · 3 NEW — 53 total. `existing_annotations` (53) − `NEW` (3) = 50 = GOA rows, asserted in code.

## The headline: three terms are attached to nodes that cannot reach the smooth muscle actins

`GO:0005884 actin filament` reaches human genes from three PTHR11937 nodes and from no node
ancestral to all conventional actins:

| node | human genes reached | terms it carries |
|---|---|---|
| `PTN000233075` | ACTA1 only | GO:0001725, GO:0005865, GO:0005884, GO:0030240 |
| `PTN000748220` | ACTC1 only | GO:0005884, **GO:0007015, GO:0017022, GO:0033275**, GO:0030017, GO:0060047 |
| `PTN002631586` | ACTB, ACTBL2, ACTG1, **ACTL8**, POTEE/F/I/J/KP | GO:0005884, GO:0098973 |

So **ACTA2 and ACTG2 are the only two of the six conventional human actins with no `actin
filament` annotation of any kind** — verified with `goUsage=descendants`, not only `exact` —
while ACTL8 has one. Measured on the same 38 inter-protomer contact residues of PDB 6DJO chain C
that the merged ACTL8 and ACTL10 analyses used:

| protein | ident | cons | non-cons | compatible / 38 | holds GO:0005884? |
|---|---|---|---|---|---|
| ACTA1 | 38 | 0 | 0 | 38 | yes |
| **ACTA2** | **38** | **0** | **0** | **38** | **no** |
| ACTG2 | 38 | 0 | 0 | 38 | no |
| ACTC1 | 38 | 0 | 0 | 38 | yes |
| ACTB / ACTG1 | 37 | 1 | 0 | 38 | yes |
| ACTL8 | 8 | 3 | 24 | 11 | **yes** |

`PTN000748220` confines `GO:0017022 myosin binding` to ACTC1 as well. Of 102 human `GO:0017022`
annotations only four are on actins (three ACTC1, one ACTA1 TAS) — so **the gene whose every
disease phenotype is a failure of the actin-myosin interaction has no myosin binding term**,
despite `PMID:26153420` running in vitro motility assays with smooth muscle myosin on
baculovirus-expressed human smooth muscle alpha-actin. Proposed as `NEW`, `IDA` not `IPI`,
because the cached text does not state the myosin preparation's species and I will not assert a
`with/from` I cannot verify.

## Asked in the other direction, the same node structure explains the one REMOVE

The only PTHR11937 node whose entire human reach is **exactly ACTA2 + ACTG2** is `PTN004322804`,
and the only term it gives them is `GO:0005576 extracellular region` — from a single rat *Acta2*
seed whose own evidence for that exact term is itself an `IBA`. PAINT has a node for precisely the
two smooth muscle actins and uses it to place them outside the cell, while the terms unambiguously
true of them sit on two single-gene nodes.

I first wrote `MARK_AS_OVER_ANNOTATED` here. Two checks changed it to `REMOVE`: the extracellular
detection is already carried by this gene's own two HDA rows under `located_in`, and **all six**
conventional actins carry those same HDA rows, so extracellular detection is a family-wide
proteomics signal and only these two got an `is_active_in` claim. Removing the IBA deletes the
unsupported functional assertion and keeps every observation.

## `GO:0005200` verified rather than inherited

Asserted once at `PTN000940351`, IRD-negated at eight descendant nodes, **none containing ACTA2**.
Ten human genes still receive it: ACTA1, ACTA2, ACTC1, ACTG2 (the four *muscle* actins), ACTR10,
ACTL9, ACTL10, ACTRT1, ACTRT2, ACTRT3. Correcting a premise I was given: **ACTB and ACTG1 receive
no IBA at all** — ACTB by TAS, ACTG1 by IC — because ACTB is itself one of the ten IBD seeds.
Eight of the ten seeds carry their own experimental evidence for the exact term.

A tree inconsistency falls out: **human ACTR2 and ACTR3 are IBD seeds justifying the term, and
`PTN000233596` / `PTN000233796`, the clades they seed, are two of the eight nodes where the same
term is IRD-negated.** Filed as a PAINT question, not as a defect in any single row.

## Disease variants: a single-chain contact set gave the wrong answer

Scoring the 19 `FT VARIANT` positions against chain C's contact set returned **zero** pathogenic
variants on the interface. That is an artefact — 6DJO has four protomers, so chain C has no i+2
neighbour and actin protomer contacts are not symmetric. Measured over every chain, **four AAT6
substitutions (Y145C, R292G, T326N, T353N) make direct ≤4 Å inter-protomer contacts**, R185Q and
R212Q sit ~4.2 Å from the nucleotide, and **R179 is 4.42 Å across the strand** — just outside a
4.0 Å cutoff, so `PMID:26637293`'s inter-strand model is corroborated, not refuted. The paper's
own named partners K193 and T196 fall inside the independently computed contact set and are used
in code as a literature-derived control on it. The only variant *in* the ≤4 Å cross-strand set is
**T196S, a non-pathogenic polymorphism**.

## Other findings

- **`GO:0006936` → `GO:0006939` smooth muscle contraction (MODIFY).** 30 human annotations carry
  GO:0006939 including MYH11 and MYLK, but neither smooth muscle actin — GO describes smooth
  muscle contraction with its myosin and its kinase but without its actin.
- **`GO:0005524 ATP binding` (NEW).** UniProt keeps the ATP-binding keyword and GOA keeps ATP
  hydrolysis activity, but `GO_REF:0000043` now returns **zero** human annotations, so the keyword
  route is gone family-wide. The cleft is intact: 18 identical + 1 conservative of 19 contacts.
- **Donor-level projections.** `PMID:24204762` (one rat siRNA study) is the sole source of **five**
  Compara rows; `PMID:10633868` (one chick antisense study) is the sole source of **all five**
  AgBase ISS rows; the two `GO:0019901` rows are one mouse IPI (partner LRRK2) transferred twice.
- **Reference projection.** `PMID:18468998` gives `GO:0032991` to 36 entities; `PMID:23580065`
  gives `GO:0005576` to 95, all HDA; `PMID:16548883` gives `GO:0009615` to 20, all IEP. For
  BioPlex (3731 and 9514 annotations) the entity count is reported **unavailable** rather than
  sampled from one page.
- **`PMID:17994018` carries a 2008 corrigendum** whose PubMed `CommentsCorrections` id is **null**
  — invisible to both a pubtype search and a PMID follow. Crossref resolves it
  (`10.1038/ng0208-255c`, `update-to 10.1038/ng.2007.6`). Nothing here rests on it alone.
- Nine Reactome `cytosol` rows are per-reaction exports of one assertion; reasoned once.

## Relationship to #2303 (ACTG2), reviewed in parallel

Derived independently and **agreeing**: the three-node `GO:0005884` structure, zero reach for both
smooth muscle actins, the `PTN004322804` reverse-reach result, `REMOVE` on the `GO:0005576` IBA,
and `GO:0005200` being correctly retained. Two method notes: this script uses
`"Swiss-Prot" in entryType`, so it is not affected by the `"reviewed" in "unreviewed"` substring
bug; and it reproduces ACTL8's committed interface tallies on **all nine** shared panel members —
because it inherited ACTL8's hand-defined conservative groups rather than re-deriving them as
`BLOSUM62 > 0`. The definition, not the numbers, is what needs harmonising.

## Reproducibility

`ACTA2-bioinformatics/analyze_acta2.py` regenerates `RESULTS.md` and `results.json` in full.
Guards that abort rather than degrade: single point of `GO:0005200` assertion; reproduction of
ACTL8's committed tallies before any number is printed; ACTA2's own sequence length against the
panel median (the ACTL10 truncation artefact — ACTL10 is flagged at 245 aa and its 20 unreached
interface positions reported as `outside_span`, never as substitutions); three numbering controls
including the paper's own "R179 (R177 in α1-actin)"; QuickGO page-coverage assertions with real
pagination; strict duplicate-key YAML loading with raw-vs-parsed `reference_id` reconciliation
(32 = 32); and coverage arithmetic (53 − 3 = 50).

Gates: `checkquotes.py` 29 quotes / 0 problems; all 11 `file:` quotes grep-verified separately
since the validator skips them; `just validate human ACTA2` ✓ Valid; `cache/go/terms.csv`
unmodified in both directions, in-file duplicates only the two pre-existing curies.

---

> [!NOTE]
> **Low Risk**
> Documentation and curation metadata only—no runtime services, auth, or data pipelines change; impact is limited to gene annotation artifacts in the repo.
> 
> **Overview**
> Adds a **complete human ACTA2 (P62736) GO curation package**: a ~2k-line `ACTA2-ai-review.yaml` that adjudicates all 50 GOA rows plus three proposed annotations, and reproducible bioinformatics output (`analyze_acta2.py`, `RESULTS.md`).
> 
> The review **removes** the extracellular-region `is_active_in` IBA (circular PAINT seed; HDA `located_in` already captures detection), **replaces** generic muscle contraction with **smooth muscle contraction**, and **proposes** `actin filament`, **myosin binding**, and **ATP binding** where PAINT node placement left the smooth-muscle actins uncovered despite perfect filament-interface scores. Most other rows are accepted, downgraded to non-core, or flagged over-annotated (BioPlex `protein binding`, basement membrane/juxtaglomerular transfers, ERK and antiviral IEP projections, etc.). The YAML also documents reference quality, propagation failures, core functions, knowledge gaps, and PAINT/Reactome questions for curators.
> 
> The analysis script regenerates PAINT node census, nucleotide/filament residue tallies, multi-chain variant proximity, WITH/FROM donor resolution, and reference-projection checks, with guards tied to the parallel ACTG2 review.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b54259af873d4d2d14ec18acd9e63145cea2827. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Pushed one fix round (d57a09d). Four corrections, one of them a real defect I found in my own first commit while probing what a reviewer would probe.

**1. The ATP-binding ISS donor was wrong.** I proposed `GO:0005524` as ISS from `UniProtKB:P68137` (Sus scrofa ACTA1) because that is the donor of this gene's `GO:0016887` ATP hydrolysis row. QuickGO shows P68137 holds **no** `GO:0005524` annotation and nothing anywhere under `GO:0000166 nucleotide binding` — so the ISS would have transferred an annotation the donor does not have. Donor changed to **human ACTC1 (`P68032`), which holds `GO:0005524` by IDA** (`PMID:16611632`): same species, direct experimental evidence, a stronger transfer than the cross-species one. The `GO:0016887` row's use of P68137 is unaffected — that donor does carry `EXP` for hydrolysis. This is the WITH/FROM discipline applied to an *outgoing* proposal rather than an incoming row, and I had not been applying it there.

**2–4. Extracellular-region provenance, sharpened** after a cross-check against #2303 (ACTG2), which describes the same node differently:

- The **single-seed** claim for `PTN004322804` is now cited to where it is actually readable — the cached PAINT export's IBD row, `PTHR11937  PTN004322804  GO:0005576  C  IBD  false  RGD:621676` — and not to the GOA WITH/FROM field, which names the node and the rat gene without distinguishing donor from co-recipient.
- The `REMOVE` reason now keeps **two chains apart**. Rat *Acta2*'s only experimental annotation under `GO:0005576` is a `GO:0005604` IDA, but that is a *different term* and it reaches human ACTA2 by **Ensembl Compara**, not through PAINT. Whether the PAINT curator generalised it upward into the IBD is a reconstruction neither source records. The `REMOVE` rests on the circular exact-term evidence (the seed's own `GO:0005576` annotation is an IBA from this same node) plus the fact that the two HDA rows already carry the observation under `located_in` — not on that reconstruction.
- `GO:0005604` is now recorded as **ACTA2-specific: ACTG2 does not carry it**, so basement membrane is not smooth-muscle-actin-wide the way the extracellular HDA rows are.

**Coverage arithmetic hardened.** The audit now asserts the GOA TSV has no byte-identical duplicate lines, so `53 − 3 = 50` is against *distinct* rows rather than a raw line count that happens to match. Measured: 50 raw, 50 distinct, 50 distinct ignoring the DATE column.

**Two method points relevant to the parallel ACTG2 review:**
- This script tests `"Swiss-Prot" in entryType`, not `"reviewed" in entryType`, so it is not affected by the substring bug where "reviewed" matches "unreviewed (TrEMBL)". Verified by grep over the source; the two TrEMBL donors it reports are correctly labelled.
- It reproduces ACTL8's committed interface tallies on **all nine** shared panel members, not 4 of 14, because it inherited ACTL8's hand-defined conservative amino-acid groups via the ACTL10 analysis rather than re-deriving them as `BLOSUM62 > 0`. Same-named metric, different measurement, and the disagreement falls on exactly the divergent genes. The *definition* is what needs harmonising once, not the numbers repeatedly.

Gates re-run after the change: `checkquotes.py` 29/0; all 11 `file:` quotes grep-verified separately (the validator skips them); `just validate human ACTA2` ✓ Valid; `cache/go/terms.csv` unmodified in both directions with only the two pre-existing duplicate curies.

cmungall:
Thanks — all four items addressed, three by change and one by argument. Pushed as 38052f8 (single push).

**1. Action tally — you are right, and I can name the cause rather than just conceding it.** The file has always been 19 `MARK_AS_OVER_ANNOTATED`; my PR body said 20 and summed to 54. The off-by-one is not arithmetic, it is a stale edit: `GO:0005576` (the `PTN004322804` IBA) started life as `MARK_AS_OVER_ANNOTATED` and I moved it to `REMOVE` late. I incremented `REMOVE` to 1 and forgot to decrement the other. Verified independently two ways — parsed count and `grep -c '^    action: MARK_AS_OVER_ANNOTATED'` both give 19, and 6+1+23+19+1+3 = 53 = `existing_annotations`. PR description corrected.

**2. `GO_REF:0000024` is worded for ortholog transfer, and these donors are paralogues — conceded, and worth more than a footnote.** Both `NEW` reasons now say so explicitly: the `GO:0005884` donors (ACTA1, ACTB, ACTC1, ACTG1) and the `GO:0005524` donor (ACTC1) are human paralogues, ISS on paralogy is permitted but the reference text does not describe it, and a submitting curator may prefer a different reference — **or an IBA once the tree is fixed**, which is the point I would rather the reader take away. The ISS row is a stopgap for a node-placement defect; if PAINT asserts `GO:0005884` at the conventional-actin ancestor, the right annotation is an IBA and this row should never be submitted at all. That is now said in the reason rather than only in `suggested_questions`.

**3. `PDB:` accessions in `supporting_entities` — conceded and removed.** You are right that a structure is not a gene product and does not belong in a with/from. Both `PDB:6DJO` and `PDB:2BTF` are gone from `supporting_entities`; the removal is lossless because both reasons already name the PDB entries in prose and the quantitative argument is carried by the `file:` reference into `RESULTS.md`. Each reason now states why the accession is absent, so a future editor does not "restore" it.

**4. `core_functions` and `GO:0005524` — I am not making this change, and here is why.** `molecular_function` is single-valued in the schema, so the ATP core function can carry hydrolysis or binding but not both, and `GO:0016887` is the right one to keep on two grounds: it is the term GOA actually holds, and it is the more informative of the pair, since hydrolysing ATP entails binding it while the converse does not. Promoting a term I am *proposing* over a term the record already carries would also invert the usual relationship between `core_functions` and `existing_annotations`. The proposal is surfaced where proposals belong — as a `NEW` row — and the core function's description argues the binding explicitly, so a reader is not left unaware of it. If the schema gains a multivalued `molecular_function`, both belong there.

**On the highlighted item.** Agreed that the ACTR2/ACTR3 seed-versus-IRD contradiction is the most actionable thing here for PAINT, and I deliberately left it as a question rather than acting on it: both donors genuinely carry their own IDA for `GO:0005200`, so neither the assertion nor the negations are wrong *as individual annotations*. The defect only exists at the level of the tree, which means it cannot be fixed from inside any one gene's record — and a per-gene review that tried would be asserting something none of its evidence supports.

Gates re-run on this head: `checkquotes.py` 29/0; 11 `file:` quotes grep-verified separately; `just validate human ACTA2` ✓ Valid; action tally 19/53 confirmed by parse and by grep; `cache/go/terms.csv` unchanged in both directions with only the two pre-existing duplicate curies.

cmungall:
One further push (9b54259) — withdrawing a hedge of mine that was wrong, not adding anything.

The `GO:0005576` `REMOVE` reason said the single-seed fact was readable from the cached PAINT export and *"not from the GOA WITH/FROM field, which names the node and the rat gene without distinguishing donor from co-recipient."* **That disclaimer is false.** The WITH/FROM on this row is two tokens — `PANTHER:PTN004322804|RGD:621676` — one tree node and exactly **one** gene product, and QuickGO's raw `withFrom` carries the same two entries with their source databases named:

```json
[{"connectedXrefs":[{"db":"PANTHER","id":"PTN004322804"}]},
 {"connectedXrefs":[{"db":"RGD","id":"621676"}]}]
```

So the node's single seed is readable directly from the annotation, and the PAINT export corroborates it rather than being the only source. The id is easy to misread only because it arrives in RGD's namespace rather than as a UniProt accession — which is precisely the kind of qualifying field that must not be dropped before reasoning from the value.

The reason now states **two facts separately** instead of merging them, because merging them is what makes this row look better-founded than it is:

1. **Who seeded the node** — from the GOA WITH/FROM, corroborated by the cached PAINT IBD row. This is provenance.
2. **The circularity** — that same seed is *also a co-recipient*: rat *Acta2* carries the identical `GO:0005576` IBA from this node, so its own evidence for the exact term is an IBA, and no experimental annotation to `GO:0005576` exists anywhere in the chain. **This is the substantive basis for the `REMOVE`**, and it is a different claim from (1).

The action is unchanged, and the two supports that make removal lossless are unchanged: both HDA detections stay under `located_in`, and all six conventional actins carry them, so the extracellular signal is a family-wide proteomics observation and only these two genes got an `is_active_in` assertion.

Gates re-run on this head: `checkquotes.py` 29/0; 11 `file:` quotes grep-verified separately; `just validate human ACTA2` ✓ Valid; tally 6/1/23/19/1/3 = 53 confirmed by parse and grep; `cache/go/terms.csv` unchanged in both directions, in-file duplicates only the two pre-existing curies.

Nothing further from me on this PR.

PR #2310, opened 2026-07-27
Review human ADCK1 (Q86TW2) GO annotations

cmungall:
Review of human **ADCK1** (Q86TW2), PAINT campaign, deep research provider **affinage** (`gates_passed: True`, 5 citations, all numeric PMIDs, no bioRxiv ids in PMID-shaped fields).

## GOA / stub reconciliation

| | count |
|---|---|
| `ADCK1-goa.tsv` lines | 9 (8 data rows + header) |
| distinct data rows | 8 |
| `fetch-gene` stub entries | 8 |
| reviewed `existing_annotations` | 8 |
| plus `action: NEW` proposals | 1 |

**Exact match, no collapse.** The gene has zero `GO:0005515` rows, and its three `GO:0005739` rows differ in evidence code and reference, so the seeding key `(GO ID, evidence_type, reference, negated, qualifier)` separates them all. QuickGO independently returns `numberOfHits: 8` for `UniProtKB:Q86TW2`. A committed check derives `supporting_entities` and `propagation_review.source_entities` **from the GOA WITH/FROM column and asserts they match** — no hand-maintained source list.

## Verdicts

| # | term | ev | action |
|---|---|---|---|
| 1 | GO:0005743 mitochondrial inner membrane | IBA | ACCEPT |
| 2 | GO:0007005 mitochondrion organization | IBA | ACCEPT |
| 3 | GO:0055088 lipid homeostasis | IBA | KEEP_AS_NON_CORE |
| 4 | GO:0005739 mitochondrion | IEA | ACCEPT |
| 5 | GO:0005739 mitochondrion | HTP | ACCEPT |
| 6 | GO:0005739 mitochondrion | IDA | ACCEPT |
| 7 | GO:0010637 neg. reg. of mitochondrial fusion | IMP | ACCEPT |
| 8 | GO:1903852 pos. reg. of cristae formation | IMP | ACCEPT |
| 9 | GO:0010954 pos. reg. of protein processing | IMP | **NEW** |

No REMOVE and no MODIFY. Propagation classification for the three IBA rows: `NO_FAILURE_CORE` on rows 1-2, `NO_FAILURE_NON_CORE` on row 3.

## The fold-propagation hypothesis: not confirmed in GOA, confirmed one layer up

[...]

PR #2311, opened 2026-07-27
Review human ADCK2 GO annotations (UbiB/aarF atypical kinase family)

cmungall:
### GOA / stub reconciliation

| count | value |
|---|---|
| `ADCK2-goa.tsv` rows (minus header) | 8 |
| distinct rows | 8 |
| `- term:` entries in the review | 8 |

Exact reconciliation, no collapsed rows to restore. QuickGO independently returns
`numberOfHits = 8` with `len(results) = 8` (unpaginated), so the local TSV is the complete
annotation set.

### Verdicts

| # | term | ev | ref | action |
|---|---|---|---|---|
| 1 | GO:0005739 mitochondrion | IBA | GO_REF:0000033 | **ACCEPT** |
| 2 | GO:0005739 mitochondrion | IEA | GO_REF:0000044 | ACCEPT |
| 3 | GO:0016020 membrane | IEA | GO_REF:0000044 | **KEEP_AS_NON_CORE** |
| 4 | GO:0005515 protein binding | IPI | PMID:27499296 | **MARK_AS_OVER_ANNOTATED** |
| 5 | GO:0005739 mitochondrion | HTP | PMID:34800366 | ACCEPT |
| 6 | GO:0005739 mitochondrion | IDA | PMID:31480808 | ACCEPT |
| 7 | GO:0005739 mitochondrion | IDA | PMID:33988507 | ACCEPT |
| 8 | GO:0010795 regulation of ubiquinone biosynthetic process | IDA | PMID:31480808 | **ACCEPT (core)** |

### The kinase hypothesis: not confirmed in GOA, confirmed one layer up

ADCK2 is a UbiB/aarF atypical kinase, so `GO:0004672`/`GO:0006468` were prime candidates for
fold-name-propagated-into-activity error. **They are not there.** ADCK2 carries no
`GO:0004672`, `GO:0004674`, `GO:0006468` or even `GO:0005524` row. Like ADAMTSL5, the
predicted defect simply is not in the annotation set, and that is reported as a
non-confirmation rather than stretched into a finding.

It *is* present in UniProt, which asserts protein-serine/threonine kinase activity in three
machine-readable fields (`EC=2.7.11.-` in the RecName, the `Serine/threonine-protein kinase`
keyword, and the keyword-derived `DR GO; GO:0004674` line) while its own FUNCTION comment
says it is not known whether the protein has protein kinase activity or what residue class it
would phosphorylate. Because GOA no longer carries keyword-derived rows, this defect is
**invisible from the GO annotation set** and can only be fixed upstream. Filed in
`suggested_questions`.

### What COQ8A/COQ8B were actually shown to do (checked, with one correction)

The brief's premise holds but needed a caveat. PMID:25498144 and PMID:27499294 establish that
neither COQ8A nor yeast Coq8p catalyses canonical protein kinase activity, with the KxGQ motif
sterically occluding the peptide site — and PMID:27499294 explicitly generalises uPKL
functionality to the whole family, which is why it bears on ADCK2 at all.

But the negative has since been **partly contested**: PMID:38425362 (2024) reports COQ8B
phosphorylating COQ3 in vitro, so **COQ8B now holds both a `NOT|enables` and an `enables` IDA
for GO:0004672**, and COQ8A holds a `NOT|enables` IDA alongside an `enables` ISS propagated
*from* COQ8B. Both contradictions are live in QuickGO. Raised as a second question rather than
papered over, because an unqualified family-level negative is no longer supportable.

### Own analysis: `ADCK2-bioinformatics/` (reproducible, with controls)

Clustal Omega MSA of ADCK2 + COQ8A/COQ8B/Cqd1/Cqd2/ADCK1/ADCK5/*E. coli* UbiB + PKA as
negative control, projecting COQ8A's own UniProt-annotated positions. Residue identity and "lands on a
site the target itself annotates" are reported separately, never collapsed.

Register is judged by the control: 6/9 columns place PKA inside one of *its own* annotated
sites; PKA reproduces **4/4** canonical catalytic residues and **0/2** KxGQ positions, so KxGQ
is UbiB-diagnostic rather than an artefact of aligning any kinase. The one column the strict
test cannot confirm (`DFG_D`) is reported as *unconfirmable rather than wrong*: PKA lands on
D185, its own DFG aspartate, identical to COQ8A's D507 — P17612 simply annotates no feature
there. ADCK2's reading is separately corroborated by local context and a raw-sequence scan.

| feature | COQ8A | ADCK2 |
|---|---|---|
| KxGQ motif | K276…Q279 | **K147…Q150** — retained |
| A-rich loop | `AAAS` 337–340 | **`GSGC` 207–210** — not Ala-rich |
| beta3 Lys / cat Asp / cat Asn / DFG Asp | K358 / D488 / N493 / D507 | K311 / D445 / N450 / D493 — all retained |

Two conclusions pulling opposite ways, both stated: ADCK2 **is** a UbiB uPKL by the criterion
the family-wide prediction rests on, and it is **not** a pseudokinase — so the
fold-retained-residues-lost argument that would earn a REMOVE was **not available**. Had a
`GO:0004672` row existed, the honest verdict would have been MARK_AS_OVER_ANNOTATED on the
occlusion argument.

The A-rich loop divergence (Gly where COQ8A has the suppressor alanines, and A339G is exactly
the COQ8A mutation that de-represses autophosphorylation) is deliberately **not** read as
"ADCK2 kept phosphotransfer": that activity is *cis* autophosphorylation, shown dispensable in
vivo. Filed as a hypothesis in `suggested_experiments` with the discriminating mutant.

### Reciprocal-orthologue test (added after cross-checking PR #2310)

PR #2310 reports ADCK1 carrying **A164** where this review reports ADCK2 carrying **G209** —
the same alignment column. Adding yeast Cqd2 (the ADCK1 orthologue) tests the pairing
directly:

> Glycine at the COQ8A-A339-equivalent position appears in **exactly 2 of 8** UbiB proteins,
> **ADCK2 and Cqd1**. All six others — including **Cqd2 and ADCK1** — carry the suppressor
> alanine.

So a single residue reproduces the Cqd1↔ADCK2 / Cqd2↔ADCK1 pairing from sequence alone,
independently of PANTHER and of the genetics. Recorded as corroboration of the **orthology**
only; it is explicitly *not* used to claim the branches differ functionally, and no nucleotide
term is proposed here — same reasoning as #2310, which withholds one for ADCK1 because the
site is untested rather than refuted.

Side effect worth recording for the campaign: **adding one more taxon improved the alignment
register.** The control's DFG column moved from a spurious Ser to its true D185, taking the
control from 3/4 to 4/4. A sparse alignment was mis-seating a column, and more taxa fixed it
where more parameters would not have.

### Agreement with the concurrent ADCK1 review (#2310)

Independently derived, and they agree on every cross-cutting point: the fold-propagation
hypothesis is not confirmed in GOA but is present in UniProt's EC/keyword layer;
`GO_REF:0000043` keyword annotations are withdrawn (verified here: **0** human annotations);
catalytic residues intact in both, so neither is a pseudokinase; the Cqd1/Cqd2 orthology
split; and `just validate` silently collapsing main's two pre-existing duplicate curies in
`cache/go/terms.csv`. No disagreement to resolve.

### PAINT: no defect, in either direction

`PTN000059786` propagates **49 annotations over 49 distinct gene products** (annotation count
and entity count verified equal, unpaginated) to **exactly one term**, `GO:0005739`. Its reach
is the ADCK2/Cqd1 orthologue group — yeast YPL109C is under it, yeast Mcp2/Cqd2 is not. The
reciprocal question returns the same node and the same true term.

More strongly, PAINT's three-way split **matches the published orthology exactly**:

| human | PANTHER family | IBA node | yeast counterpart |
|---|---|---|---|
| ADCK2 | PTHR45890 | PTN000059786 | YPL109C = **Cqd1** |
| ADCK1 | PTHR43173 | PTN005148758 | YLR253W = **Cqd2**/Mcp2 |
| COQ8A/COQ8B | PTHR43851 | PTN000059692 | Coq8 |

PMID:34362905 pairs Cqd1↔ADCK2 and Cqd2↔ADCK1/5 from the biology, and Cqd1/Cqd2 have
**reciprocal** effects on CoQ distribution. **Relevant to the concurrent ADCK1 PR:** the two
genes are in different PANTHER families with different donors and are, if anything,
antagonistic — nothing should be shared between them. ADCK1 also carries no MF row in GOA, so
a kinase-propagation finding there would be a disagreement worth resolving.

### GO:0010795 is right, and right at the right level

Tested rather than assumed. The paper's 3H-mevalonate labelling found **total** cellular CoQ
and cholesterol synthesis *unchanged* with only the mitochondrial fraction depleted — so ADCK2
governs precursor delivery, not a pathway step, and `GO:0006744` (which COQ8A/COQ8B correctly
hold) would be wrong. The curator's choice is confirmed, not upgraded.

Kept separate from phenotype: the beta-oxidation defect is downstream of the CoQ deficit and
CoQ10-rescuable, so **no fatty-acid GO term is proposed** from the knockout phenotype.

### GO:0005515: `NbExp=4` is one experiment counted four times

Third instance of this pattern in the campaign (after ACRV1 and ADAMTSL5). All four IntAct
evidences are one publication, one method (`anti tag coip`), ADCK2 bait / SLC25A5 prey,
MI-score **0.35**, and **3 of 4 are spoke-expanded**. SLC25A5/ANT2 carries **385** IntAct
evidences of its own; ADCK2 has 83 partners over 7 publications. Partner resolved to the
reviewed canonical entry with matching length (no ORFeome substitution). The topology
objection that carried ACRV1 is **absent here and reported as a negative** — both proteins face
the matrix side, so the interaction is possible; possibility is not support.

### Ontology gap

GO has **no coenzyme Q transport or distribution term**. The complete descendant set of
`GO:0006743` is nine terms (biosynthesis, catabolism, seven enzyme activities), checked against
the ontology rather than by keyword search. Yet that is precisely the characterised activity of
Cqd1/Cqd2. `proposed_new_terms` proposes one under `GO:0032365`, scoped so the immediate
holders are the yeast genes and **ADCK2 is a candidate, not an asserted holder**.

### Checks run that came back NEGATIVE (recorded, not skipped)

- **Retraction / erratum / expression-of-concern:** 16 PMIDs checked via `CommentsCorrections`
  and publication types. All clear.
- **Complex-projection test:** PMID:31480808 annotates only **2** entities (human + mouse
  ADCK2), so it is gene-specific, not a projection. For PMID:34800366 and PMID:27499296 the
  QuickGO result is paginated, so entity counts are **unavailable** and the test is recorded as
  unreliable rather than substituted with the annotation total.
- **IBA-less-precise-than-donor (ACRV1):** not warranted. Donors disagree in specificity (mouse
  Adck2 at mitochondrion, Cqd1 additionally at inner membrane by IDA), so `GO:0005739` is the
  correct LCA; per the AADACL4 rule `GRANULARITY_MISMATCH` needs the donors to agree.
- **Self-referential IBA:** `UniProtKB:Q7Z695` is in its own WITH/FROM — valid, `NO_FAILURE_CORE`.

### Provider record

affinage, `gates_passed: True`, 5 numeric PMIDs, no `PMID:bio_*` ids. Marked `LOW_QUALITY`
because its framing conflates CoQ biosynthesis with CoQ delivery. Used only to widen the
search; **no provider sentence is quoted as evidence for any mechanistic claim.**

### Verification

- `checkquotes.py`: 54 quotes, **0 problems**
- `just validate human ADCK2`: **✓ Valid**, one warning left standing on purpose (no annotation
  decision rests on the provider record, by design)
- `cache/go/terms.csv`: merge-base diff shows **no deletions**; net diff vs main's current file
  is **exactly one appended line** (`GO:0010795`); in-file duplicates are only the 2
  pre-existing (`GO:0001675`, `GO:0009566`)
- `origin/main` merged in before final validation, so the merge result is what was validated
- `ubib_motif_analysis.py` reproduces `RESULTS.md` and `results.json` byte-for-byte
- `audit_adck2_claims.py --self-test`: 7/7 guards fire; the audit found two real defects in
  its own first draft (a guard that flagged the *correct* negated statement, and mutations that
  landed in comments the detector could not see)

---

> [!NOTE]
> **Low Risk**
> Documentation, cached ontology metadata, and offline analysis scripts only; no runtime application or auth/data-path changes.
> 
> **Overview**
> Adds a **human ADCK2 (Q7Z695)** gene review package: `ADCK2-ai-review.yaml` walks all eight GOA rows with verdicts (core role **GO:0010795** accepted; **GO:0005515** marked over-annotated; membrane IEA kept non-core), literature synthesis, PAINT/IBA checks, core function text, a proposed GO term for **intracellular coenzyme Q distribution**, and curator-facing questions (e.g. UniProt kinase EC/keywords vs evidence).
> 
> Introduces **`ADCK2-bioinformatics/`**: Clustal-based UbiB/uPKL motif projection (`ubib_motif_analysis.py` → `RESULTS.md` / `results.json`), QuickGO sweeps arguing no existing CoQ transport term (`coq_transport_term_check.py`), and **`audit_adck2_claims.py`** to enforce review/GOA alignment, motif numbers, and retracted phrasing (with `--self-test`).
> 
> **`cache/go/terms.csv`** gains one line for **GO:0010795** (regulation of ubiquinone biosynthetic process).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17da3ca874e5840ce767ef86b298600d96d8fbba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
All six points accepted and applied in c8d5f7f. Thanks — the ontology-gap one was a real hole in the evidence, not a presentation issue.

**Previous suggestion 2 (open): the GO:0006743 enumeration could not have found what it was looking for.** You're right, and it's worse than under-powered: a transport term would never be classified under a metabolic process, so enumerating that branch was *guaranteed* to return nothing regardless of the truth. An absence is only evidence when the query could have returned a hit.

Redone in `ADCK2-bioinformatics/coq_transport_term_check.py`, two complementary sweeps that each could have:

- **label sweep** — all **80** GO terms whose label mentions ubiquinone / coenzyme Q / quinone. Seven have transport-flavoured labels; all seven are *electron transport* terms, where the electron is the cargo and the quinone the acceptor.
- **branch sweep** — the **2584 distinct** descendants of `GO:0006869`, `GO:0010876`, `GO:0032365`, `GO:0120009`, `GO:0006810`. Zero ubiquinone-specific children.

Sweep 1 catches a term filed in the wrong branch, sweep 2 one whose label avoids the word; the script states that the absence is evidence only insofar as the two together are exhaustive. Worth flagging that writing it reproduced this campaign's own anchoring trap — the first run reported 8 hits purely because `"transport" in name` matches "electron transport". Every exclusion is now printed with its reason rather than disappearing into the verdict.

**Suggestion 2 (independence).** Conceded without reservation — PANTHER's subfamily assignment is sequence-derived, so a residue split is not independent of it. Reworded everywhere to "a second sequence-based line agreeing with the genetics, not a third independent line".

**Suggestion 1 (which A-rich column).** Correct, and the report now computes the contrast rather than asserting the scope: `Arich_A3` splits 2/6 exactly along the pairing, while `Arich_A1` carries Gly in **four** proteins — ADCK2, Cqd1, **Cqd2 and ADCK1** — cutting across it. Stated explicitly that the claim is about the A339-equivalent position, not the A-rich loop as a whole.

**Suggestion 3 (identity figures).** Added, with the number that makes the point: the control moved **17.1% → 11.9%** on adding one sequence. Recorded as a property of the MSA rather than of the sequence pairs, with an instruction to recompute rather than compare across alignments.

**Suggestion 4 (negation guard).** A fair hit, and the third defect that guard has had. The window now closes at punctuation *and* at coordinating conjunctions, so a negator in a neighbouring clause no longer suppresses. Two decoy must-fire cases added — one with `no ... , and <claim>`, one with the negator across a full stop — alongside the existing must-not-fire cases. Self-test now 11/11.

**Previous suggestion 3, remaining half (GO:0005524).** Also accepted, and it improves the ask. The question now scopes itself explicitly to the protein-kinase claims and says why it excludes the nucleotide keywords: the alignment finds the beta3 lysine, catalytic Asp and Asn, and the DFG-equivalent Asp all intact, so nucleotide binding is the best-supported part of the keyword set. The nuance I added rather than simply conceding — the open question there is **ligand identity, not existence**, because COQ8A binds ADP in preference to ATP and carries `GO:0043531` rather than `GO:0005524`, and ADCK2 sits on the opposite side of the very A339-equivalent residue that shifts that selectivity. So `GO:0005524` should be re-examined on which nucleotide, not withdrawn as unfounded.

No annotation verdict changed in this commit; `existing_annotations` differs only inside `suggested_questions[0]` and `[2]` and the `proposed_new_terms` justification. Gates re-run after merging `origin/main`: checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicate curies, `RESULTS.md`/`results.json` reproduce byte-for-byte, audit and self-test pass.

cmungall:
All five applied in be541bd. Suggestion 1 was a genuine hole in the evidence, not a polish item — thanks.

**(1) `quinol` missing from `KEYWORDS`.** Confirmed and fixed. `"quinone" in "ubiquinol"` is `False`, so both sweeps really were blind to the reduced form, and that is the one failure mode the two-sweep design is supposed to preclude. Ubiquinol-named GO terms exist — 6 in this repo's own `cache/go/terms.csv`. Adding `quinol` grew the label sweep **80 → 102** and surfaced four terms that had been invisible to both sweeps (`GO:0006122`, `GO:0008121`, `GO:0009486`, `GO:0009496`); all four adjudicate out as electron transport or MF, so the gap conclusion is unchanged, but it now rests on a query that could have contradicted it.

The widening then produced the mirror error immediately: **`"quinol"` *is* a substring of `"quinolinic"`**, and `GO:1903222 quinolinic acid transmembrane transport` cleared every other filter as a genuine BP transport term. Quinolinic acid is a tryptophan-pathway NAD precursor with no relation to CoQ. Excluded explicitly, and only when the label names no real CoQ species, so a term mentioning both would still be caught. That is three substring traps in one small script — `"transport"` matching *electron* transport, `"quinone"` failing to match *ubiquinol*, and `"quinol"` matching *quinolinic* — one too loose, one too narrow, one too loose again. Written up in the notes as the clearest example of the anchoring rule I've hit.

**(2) Negation window.** You're right, and I introduced it: I narrowed the budget from 40 to 30 in the same edit that added the clause-boundary lookahead, and the boundary check had already made the budget redundant. `"no experiment has ever shown that "` is 32 characters, so it would have false-positived. Budget is now 60; the lookahead still closes at punctuation and at `and`/`but`/`while`/`whereas`/`however`, so the two decoy cases still fire. That exact sentence is now a must-not-fire case. Self-test 12/12 (4 must-not-fire, 8 must-fire).

**(3) Hardcoded interpretation.** Fair — same defect I'd already fixed once elsewhere. The `Arich_A1` sentence is now gated on the computed set with three branches (splits the same way / cuts across, naming the computed crossers / is a subset), so nothing about Cqd2 or ADCK1 is asserted by hand.

**(4) Docstring and baked-in history.** Both corrected. The sweep-1 docstring now says the filter is on labels (noting QuickGO's search itself also indexes synonyms), and the MSA caveat states the principle plus the current membership size rather than a historical before/after that would go stale on the next membership change.

**(5) Process commentary in `justification`.** Removed — you're right that it's the notes' job, and it was already there in full. The justification now states only what was swept and what was found.

Gates after merging `origin/main`: checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates, `RESULTS.md`/`results.json` reproduce byte-for-byte, audit and 12/12 self-test pass. No annotation verdict moved; the YAML delta is confined to `proposed_new_terms.justification`.

cmungall:
Fixed in f55ec32. The IMPORTANT finding is right, and it's the campaign's signature defect committed by me in the very commit that congratulated itself on auditability: **the widening from 80 to 102 terms is what made "every transport-flavoured one is an electron-transport term" false**, and I wrote that sentence in the same commit. The script gained a second exclusion category; the paragraph summarising the script did not. And the YAML never named `GO:1903222` at all, so the one surface a curator would read was the one with no trace of it.

Both surfaces now state **both** categories and name the term. Conclusion unchanged — quinolinic acid is a tryptophan-pathway NAD precursor, the exclusion is correct, and there is still no CoQ transport term.

Taking your selector point seriously, I added a guard keyed on the **stable identifiers** (`GO:1903222` and `QUINOLINE_FALSE_FRIEND`) rather than on any wording: if the sweep script contains the false-friend branch, both prose surfaces must mention the GO id it excludes. It fires on a mutation that renames the id away. Self-test 13/13.

**(1) The false-friend reason could misreport.** Correct, and the case you named is real: `pyrroloquinoline quinone biosynthetic process` contains `quinolin` *and* matches on its own `quinone`, so my `quinolin in label and no REAL_COQ` test would have excluded it while claiming it matched only via `quinolin`. Replaced with an exact test — delete the false-friend substring, re-test `KEYWORDS`, and whatever still matches did not come from `quinolin`. Unit-checked on four labels:

| label | matched only via quinolin? |
|---|---|
| `quinolinic acid transmembrane transport` | True |
| `pyrroloquinoline quinone biosynthetic process` | **False** |
| `ubiquinol transport` | False |
| `quinoline metabolic process` | True |

**(2) Stale console headers.** Both derive from `KEYWORDS` now.

**(3) The `{0,60}` budget.** Agreed, and I took the "drop it" option rather than the "match the window" one. With the clause-boundary lookahead in place the budget can only *add* false positives — it cannot make the guard usefully stricter — and it carried its own cliff at 61 in-clause characters. There is now exactly one length limit, the preceding window, instead of two that can disagree. Both decoys and all three must-not-fire cases still behave.

**(4) Doubled crossers / (5) orphaned "exactly that".** Both fixed; the antecedent is restored by an explicit "So no GO term describes the movement or partitioning of coenzyme Q itself."

Since your sandbox couldn't execute: re-run locally on the merge result — sweep exits 0 with the two exclusion categories printed per candidate, `ubib_motif_analysis.py` reproduces `RESULTS.md`/`results.json` byte-for-byte across consecutive runs, audit clean, self-test 13/13, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicate curies. `existing_annotations` is byte-identical to `be541bd`; the YAML delta is the justification paragraph only.

cmungall:
All three closed in d2e606d. None was cosmetic.

**(1) Stale rule in the notes.** You're right that this is a substantive divergence, not wording: the notes still said "excluded only when the label names no real CoQ species", which is the rule the previous commit *deleted*, and the two rules genuinely disagree on `pyrroloquinoline quinone biosynthetic process`. The notes now describe delete-and-retest and keep the old rule only as an explicitly labelled contrast, naming the discriminating label so the difference is visible rather than implied.

Worth stating plainly, since it bears on how much the new guard is worth: **the drift guard could not have caught this.** It checks that `GO:1903222` is *disclosed*, not that the *mechanism* is described correctly. Keying on a stable identifier fixes the "prose gets reworded" failure; it does nothing about "prose describes a rule the code no longer implements". Those are different defects and only the first is now mechanised.

**(2) Half-exercised loop.** Correct, and this is the same shape as the campaign's "a detector and a mutator must agree on scope" lesson — I wrote a two-surface loop and a one-surface test, so the notes branch was asserted coverage, never demonstrated coverage. `run()` now takes a `notes_text` override and a notes-surface mutation drives that branch. Self-test 14/14.

**(3) Unguarded helper.** Also correct — a table in a commit message guards nothing against the next edit. `matched_only_via_quinoline` is now pinned by a check that runs on **every** audit, importing the sweep module and asserting five labels, including a `ubiquinol-quinoline hybrid transport` case for the both-named scoping you verified by hand.

And I tested the guard by breaking it rather than by reading it, per the campaign rule. Reverting the helper to the old `REAL_COQ` rule gives:

```
1 problem(s):
  - false-friend classifier: 'pyrroloquinoline quinone biosynthetic process' -> True, expected False
```

i.e. it fails on precisely the discriminating label rather than passing quietly. Restored, clean.

Gates re-run locally on the merge result (your sandbox can't execute these): motif report reproduces byte-for-byte, sweep exits 0, audit clean, self-test 14/14, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates. `existing_annotations` unchanged since `be541bd`; this commit touches only the notes and the audit script.

cmungall:
Both fixed in dedab2d. Suggestion 2 was a real bug, and its mechanism is worth naming.

**How the gating got inverted.** Last round I applied the edit with a string replacement whose search text was `    check_raw_vs_parsed(problems)` — 4 spaces — against a line indented with **8**. It matched as a substring starting at column 4, so the replacement landed *inside* the existing indentation: the helper inherited the 8-space (gated) position and `check_raw_vs_parsed` came out at 4 spaces, ungated. Exactly the campaign's "string replacement across indented text creates new garbage" failure, and it produced the precise inversion you describe.

**Consequences, as you diagnosed them.** `check_false_friend_helper` depends on neither argument, so gating it meant `run(base)` — the self-test's abort check — never exercised it, and the only path that reached it was the notes mutation, where a problem it appends is indistinguishable from one the disclosure guard appends. A broken helper would therefore have made that mutation report **"caught" for the wrong reason**. Ungated now.

**`check_raw_vs_parsed`: fixed at the root rather than by moving the gate back.** You're right that it was the check that genuinely didn't apply to a mutated document, because it re-read `REVIEW` from disk and ignored `review_text`. Restoring its gate would have preserved the underlying trap — one check inspecting a different document from the rest of `run()`. It now takes the document text instead. Its arithmetic (raw `- reference_id:` lines vs parsed `supported_by` entries) is self-consistent on any input, so with the text passed in it needs no gate at all, and there is no longer any check in `run()` reading a different document than the others.

**Verified by breaking it, not by reading it.** With the sweep helper reverted to the superseded `REAL_COQ` rule, the ungated version now does this:

```
SELF-TEST ABORTED: the unmutated document already fails
  - false-friend classifier: 'pyrroloquinoline quinone biosynthetic process' -> True, expected False
```

i.e. it fails loudly on the unmutated document naming the discriminating label — the behaviour that was previously unreachable. Restored: audit clean, self-test 14/14.

Gates on the merge result: motif report reproduces byte-for-byte, sweep exits 0, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates. This commit touches `audit_adck2_claims.py` only — `ADCK2-ai-review.yaml`, the notes, both analysis scripts and both generated artefacts are byte-identical to `d2e606d`, so no verdict, term or number moved.

cmungall:
Both closed in adb21f35. I verified the reachability argument rather than conceding it, and it holds exactly as you stated.

**The rationale was unreachable.** Injecting a duplicate mapping key:

```
run() on duplicate-key doc -> 1 problem(s)
    YAML: duplicate key 'product_type' at line 4
reached raw/parsed check? False
```

`ConstructorError` is a `YAMLError` subclass, `run()` returns early, and `check_raw_vs_parsed` is never called. StrictLoader owns that case outright. Both the docstring **and the failure message** claimed otherwise — the message would have told a future reader to go looking for a duplicate key that cannot be the cause. Both now state the residual the count comparison actually catches: a `supported_by` entry that has lost its `reference_id` key, or a `reference_id` line inside a block scalar.

**And the coverage gap, with your why/whether distinction honoured.** It was untestable while it re-read `REVIEW` from disk — no in-memory mutation could reach it — and the previous commit made it testable by passing the document text, then didn't add the test. Added now: renaming one `reference_id` key drops a raw line while the parsed entry survives, giving 23 vs 24.

I also checked it is caught for the *right* reason, since that ambiguity is what you flagged two rounds ago:

```
problems: 1
  - raw/parsed: 23 '- reference_id:' lines in the file but 24 parsed supported_by entries...
attributed to raw/parsed? True
```

One guard, correct arithmetic, no credit taken by another check. Self-test 15/15.

**Both carried trivia closed too.** The prose join is now a proper list join (unit-checked at 0/1/2/3 items: `none` / `A` / `A and B` / `A, B and C`) — worth doing precisely because it was unreachable at the current membership and so would have degraded silently the next time a sequence was added, which is how the identity figures moved last time. And the sweep module is cached per process instead of being exec'd on each of ~15 `run()` calls.

Gates on the merge result: motif report reproduces byte-for-byte, sweep exits 0, audit clean, self-test 15/15, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates. This commit touches the two analysis scripts only; `ADCK2-ai-review.yaml` and `ADCK2-notes.md` are unchanged since `d2e606d`, so no verdict, term or reported number moved.

That closes everything raised across all six rounds. Nothing outstanding from my side.

cmungall:
All three closed in bd5815d, and the first was the sharpest finding of the round.

**(1) The gate was one conditional behind its own fix.** You're right, and this is the same hardcoded-interpretation class I "fixed" two commits ago — I moved the *names* into a computed set and left the *relationship* hardcoded in the prose. `elif crossers:` never tested that `expected_gly ⊆ a1_gly`, yet the sentence "the same ADCK2/CQD1 pair plus X" asserts exactly that, so a membership like `{CQD2, ADCK1}` would have had the report claim the pair carries Gly at A1 when neither does. Now gated on `crossers and expected_gly <= set(a1_gly)`, with a separate branch that just describes the set as it is. The lesson I'm taking: gating on *some* computed value is not the same as gating on the condition the sentence asserts.

**(2) Bare join twelve lines below the helper that exists to prevent it.** Fixed — and yes, "fixed in N places, landed in N−1" landing inside the commit that introduced the fix. The remaining `', '.join` calls are comma-separated enumerations inside parentheses and bullet lists, where a plain join is the right form; only prose sentences use `_english_list`.

**(3) Header contradicting the docstring.** Corrected: the header now says StrictLoader closes the duplicate-key case alone and points at `check_raw_vs_parsed`'s docstring for what the count actually earns its place on.

**Also took the low-value note, and it paid for itself.** Pinning `_english_list` at 0/1/2/3 in the tree exposed **two defects in the module loader I added last commit**: it never registered the module in `sys.modules`, which `@dataclass` needs to resolve `__module__` (opaque `AttributeError` otherwise), and it *raised* on import failure — aborting every later check, which is the rule this file's own header states. Import failure is now a reported problem. Both verified by breaking them:

```
- prose join: ['A', 'B', 'C'] -> 'A and B and C', expected 'A, B and C'
- could not import ubib_motif_analysis.py: RuntimeError: boom
```

So the "much lower value" test found more than the check it was pinning. Worth recording: the cost of writing a cheap guard is often paid back by what writing it reveals about the code it touches, not by the guard.

Gates on the merge result: motif report reproduces byte-for-byte, sweep exits 0, audit clean, self-test 15/15, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates. Scope is the two analysis scripts; `ADCK2-ai-review.yaml`, `ADCK2-notes.md`, `RESULTS.md` and `results.json` are unchanged since `adb21f3`, so no verdict, term or reported number moved.

cmungall:
All three plus the note closed in 82d7bb9.

**(1) First third-party dependency, and the failure mode you traced.** Correct on every step: `ubib_motif_analysis.py` imports `requests` at module level (the sweep is stdlib-only), so in a pyyaml-but-no-requests environment the helper appended a problem, `run(base)` came back non-empty, and `--self-test` aborted before any mutation — a missing wheel presenting as a claim defect, and the harness going from degraded to unrunnable. That's strictly worse than not having the check.

The loader now classifies: a `ModuleNotFoundError` naming something that is neither the module under load nor a sibling `.py` is environmental, and the caller prints a loud SKIP saying the check did **not** run, rather than appending a problem. Everything else is still a reported problem. Both paths tested by breaking them:

```
A: import requests -> import definitely_not_installed_xyz
  SKIPPED unit pin on ubib_motif_analysis.py: ModuleNotFoundError: No module named
  'definitely_not_installed_xyz' (environment lacks a dependency; this is not a claim
  defect, but the check did NOT run)
audit: all invariants hold        exit 0

B: appended `raise RuntimeError("boom")`
1 problem(s):
  - could not import ubib_motif_analysis.py: RuntimeError: boom
```

Skip-with-announcement rather than silent skip, since "a null result from a check is a finding" cuts both ways — a check that didn't run must say so.

**(2) "Neither of these" above three calls.** Fixed. Third instance of that shape in this PR, each time in the commit that was tidying something else.

**(3) "Gly is carried by none".** Restored to "no protein" — a wording regression I introduced when `_english_list` replaced the bare join, which is a small instance of the thing that check exists to prevent.

**On the note.** Fair: I did observe that exact traceback (`dataclasses.py:749`, `sys.modules.get(cls.__module__).__dict__` on `None`), so the causal claim was demonstrated in-session but only asserted in the tree. The docstring now says registration is documented practice generally and that `@dataclass` was the *observed symptom that surfaced it*, not necessarily the only consumer. As you say, the catch-and-report makes it moot either way.

Gates: motif report reproduces byte-for-byte, sweep exits 0, audit clean, self-test 15/15, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates. Scope is the two analysis scripts; `ADCK2-ai-review.yaml`, `ADCK2-notes.md`, `RESULTS.md` and `results.json` unchanged since `adb21f3` — no verdict, term or reported number has moved in four commits.

cmungall:
All three closed in 9ef5e39.

**(1) Classification against a reference set.** This was the substantive one — without it, an undeclared or misspelled import in the analysis code got *excused* as a missing wheel, which is the same "degraded reads as fine" failure inverted. The classifier now parses this directory's `pyproject.toml` and treats only a **declared** dependency going missing as environmental. Verified both directions:

```
declared deps parsed: ['Bio', 'requests', 'yaml']
requests                -> True    yaml -> True    Bio -> True
definitely_not_declared -> False
ubib_motif_analysis     -> False   (sibling script, so a genuine intra-repo break)
```

and end to end, injecting `import yaml_absent_stub` into the motif script now reports rather than skips:

```
1 problem(s):
  - could not import ubib_motif_analysis.py: ModuleNotFoundError: No module named 'yaml_absent_stub'
```

**(2) Once per path.** Fixed — and your prediction was exact. My first draft of the reporter pin reset the announce flag on every `run()`, so the self-test printed **sixteen** identical SKIPPED lines interleaved with the mutation log. I saw the shape you described before I'd finished fixing it.

**(3) Reporter pinned.** Both classes plus the classifier, seeded through `_MODULE_CACHE` with no monkeypatching, as you suggested. Writing it caught a defect in itself: the first version leaked a SKIPPED notice naming a synthetic file that does not exist, on every audit run. Capturing stdout fixed that *and* made the announce-once behaviour assertable rather than assumed — two calls must emit exactly one notice. That check would have caught defect (2) had it existed first.

Self-test 15/15 with no synthetic noise; audit clean.

On your closing observation: agreed, and it generalises past this branch. The cheap fix at every step of this thread would have closed the finding while quietly shrinking what the harness actually checks — swallow the import error, restore the gate, keep the summary sentence and reword it. Each time the more expensive fix removed a class rather than an instance, and in three cases (the loader's `sys.modules` bug, the leaked synthetic notice, the sixteen lines) writing the guard is what exposed the next defect. The guards have been worth more for what they surfaced than for what they now assert.

Gates: motif report reproduces byte-for-byte, sweep exits 0, audit clean, self-test 15/15, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates. Scope is `audit_adck2_claims.py` alone; the review YAML, notes, `RESULTS.md` and `results.json` are unchanged since `adb21f3` — no verdict, term or reported number has moved in five commits.

cmungall:
All three plus both notes closed in 1107cfb.

**(1) Submodule vs absent wheel.** Correct, and the case is real: `import requests.sessionz` raises with `name = "requests.sessionz"`, whose top-level component is declared *and installed*, so the old test excused a typo in the sibling script as a missing wheel. Now gated additionally on `importlib.util.find_spec(top) is None`.

Two things worth reporting about the fix. **My own reporter pin caught the semantic change on the first run** — it was asserting that `requests` classifies environmental, which is precisely the behaviour you were asking me to remove. And the absent-wheel branch is now unreachable while every declared dependency is installed, so the pin scopes `find_spec` for the duration and restores it in a `finally`. That is the one place I've used patching, and only because the branch is otherwise untestable in a healthy environment; the alternative was leaving it asserted-not-observed, which is the shape you've flagged twice.

The pin now covers all four: undeclared → report, bad submodule of an installed dep → report, declared-and-absent → skip, sibling name → report.

**(2) Mandatory dist map.** Taken. Unmapped distributions are collected and reported rather than guessed:

```
dependencies = ["requests>=1", "some-new-dep>=1.0"]
names   : ['requests']
unmapped: ['some-new-dep']   -> reported, not silently guessed
```

Guessing `dist.replace("-", "_")` is what would have dropped the next `biopython`-shaped dependency out of the declared set without a sound.

**(3) The comment drifted again**, one commit after being corrected for exactly that — "three below" / "two unit pins" against four calls and three pins. I've rewritten it **without any count at all**, because naming a number is what made it drift twice; it now describes the property (these inspect modules, not the document) rather than the tally.

**Both notes.** `re.split` takes `maxsplit` by keyword now. The dependencies regex is anchored to a line-initial key, verified on a synthetic pyproject where `optional-dependencies` appears first: it correctly returns `{'requests'}` rather than the decoy.

On the bias you named: I checked every degenerate path and they all fail toward reporting — no pyproject, unparseable pyproject, unmapped dist, non-`ModuleNotFoundError`, sibling-script name, and installed package all report rather than excuse. The only path that skips is a declared dependency that is genuinely absent.

Gates: motif report reproduces byte-for-byte, sweep exits 0, audit clean, self-test 15/15, checkquotes 54/54, `just validate` ✓ Valid, `terms.csv` +1/−0 with only the two pre-existing duplicates. Scope is `audit_adck2_claims.py` alone; the review YAML, notes, `RESULTS.md` and `results.json` are unchanged since `adb21f3`, so no verdict, term or reported number has moved in six commits.

cmungall:
Fixed in d95560f — and reproduced before fixing, since you traced it by reading rather than running.

**The finding is exactly right, including the part that matters most: the classifier is correct in both environments and the *expectation* is what was environment-dependent.** With no `requests` wheel, `import requests.sessionz` genuinely *is* a missing-wheel event, so the pin would have appended "requests is present" on a machine where it is not — and because `check_import_failure_reporter` is ungated and `run(base)` gates `self_test`, `--self-test` would abort before a single mutation. Degraded-reads-as-defect, reintroduced by the pin rather than by the classifier, one round after I closed the same property.

Reproduced by reconstructing the pre-fix assertion and running both versions under a scoped `find_spec`:

| | requests present | requests ABSENT |
|---|---|---|
| pre-fix | silent | **FIRES** — "a bad SUBMODULE was excused as a missing wheel; requests is present" |
| post-fix | silent | silent |

Fixed symmetrically, as you suggested: the neighbouring absent-wheel assertion already scopes `find_spec` because it is unreachable in a *healthy* environment; the submodule case needed the mirror because it is unreachable in a *degraded* one.

Worth reporting a wrinkle, because it is the same class of bug: **my first reproduction attempt was itself wrong.** I put the scratch copies in `/tmp`, so `HERE` resolved to `/tmp`, no `pyproject.toml` was found, the declared set came back empty, and a *different* assertion failed in all four cells — which would have read as "the fix doesn't work". Re-run in-place, the table above is clean. A harness that lies to you about its own environment is precisely what this thread has been about.

**Suggestion 2 (unguarded `find_spec`).** Correct and I'd missed the interaction: it is called from inside `_load_local_module`'s `except` handler, whose documented contract is to return rather than raise. `find_spec` can raise — `ValueError` for a module with no `__spec__`, `ImportError` while importing a parent package. That call is now total, resolving any failure to **report** the import failure rather than excuse it, which is the classifier's safe direction. Not error-hiding: the outcome is the strictest one available.

**Suggestion 1 (stale docstring)** and **3 (what the map check pins)** both taken: the docstring no longer claims "no monkeypatching", explains which two assertions scope `find_spec` and why, and records that the mandatory-map assertion pins that every declared distribution *has* an entry, not that the entry is *right* — since checking correctness would mean importing each name and reintroduce the very coupling.

---

**On calibration, and I think this is the right place to say it.** This is now the sixth consecutive round in which nothing in a deliverable moved: `ADCK2-ai-review.yaml`, `ADCK2-notes.md`, `RESULTS.md` and `results.json` are byte-identical since `adb21f3`. Every finding has been genuine and this one would have produced a false claim-defect on any machine lacking `requests`, so it was worth fixing. But the loop has moved from the review to the harness that checks the review, and the harness is now considerably more elaborate than the eight annotations it guards.

So: if the next round's findings do not touch a term, a verdict, a number or a statement in the review or the notes, I'd propose calling it done rather than continuing. I'll of course fix anything that does.

PR #2312, opened 2026-07-27
Review human ADCK5: the UbiB kinase label is a UniProt keyword artefact, not a GOA one

cmungall:
### Row reconciliation

`ADCK5-goa.tsv` has **4** data rows; `existing_annotations` has **4** entries. They
reconcile exactly — no collapse. The two `GO:0005515` rows share a partner (NOTCH2NLA) but
differ by reference, and the seeder keys on reference, so both survived. `cache/go/terms.csv`
is **untouched**: all four GO ids already exist in it, so this PR adds no rows and deletes
none (checked against the merge base, and against main's current file for duplicates — only
the two pre-existing `GO:0001675`/`GO:0009566` dups remain).

### The hypothesis under test, and how it actually resolved

The brief predicted that `GO:0004672 protein kinase activity` / `GO:0006468 protein
phosphorylation` on ADCK5 would be fold-name-propagated-into-activity errors, since COQ8A and
COQ8B are known for what they turned out *not* to be.

**The mechanism is real and demonstrable. The predicted GOA error is not there.** ADCK5's
whole GOA record is four rows, with no molecular-function term except `protein binding`.
QuickGO returns **0 hits** for `GO:0004674` on Q3MIX3 with `goUsage=descendants`. There is
nothing to REMOVE or MODIFY on kinase grounds — reported as a **non-confirmation**, in the
ADAMTSL5 sense.

The unsupported claim does exist, one layer up, in **UniProt**:

```
DE   RecName: ... EC=2.7.11.-;
KW   Serine/threonine-protein kinase;
DR   GO; GO:0004674; F:protein serine/threonine kinase activity; IEA:UniProtKB-KW.
```

…in the same entry that states *"The function of this protein is not yet clear. It is not
known if it has protein kinase activity…"*. GOA no longer imports keyword-derived (SPKW)
annotations, so the GO record is clean while the UniProt record is not. Filed as a **UniProt
correction request** in `suggested_questions`, not as a GO action.

### Verdicts

| # | term | evidence | action |
|---|---|---|---|
| 1 | `GO:0016020` membrane | IEA `GO_REF:0000044` | **ACCEPT** |
| 2 | `GO:0005515` protein binding (NOTCH2NLA) | IPI `PMID:25416956` | **MARK_AS_OVER_ANNOTATED** |
| 3 | `GO:0005515` protein binding (NOTCH2NLA) | IPI `PMID:31515488` | **MARK_AS_OVER_ANNOTATED** |
| 4 | `GO:0005739` mitochondrion | HTP `PMID:34800366` | **ACCEPT** |

No IBA rows exist on this gene, so no `propagation_review` was required.

**Row 1** is true but prediction-derived: SubCell `SL-0162` ← `Membrane {ECO:0000305}` ← a
*predicted* `TRANSMEM 50..67` (`ECO:0000255`). Not refined to `GO:0031966`, because that would
mean joining an experimental mitochondrion call to a predicted membrane call — a composite
inference this review declines to make. The real defect is upstream: ADCK1 and ADCK2 get
`SL-0173 Mitochondrion` from the same pipeline on the *same* MitoCoP evidence; ADCK5 does not.

**Rows 2–3** are the third instance in this campaign of the `NbExp` trap (after ACRV1 and
ADAMTSL5). UniProt records `NbExp=4`, but IntAct shows `PMID:25416956` logging the pair
**three times under three sub-method labels of one Y2H screen** (`two hybrid array` +
`two hybrid prey pooling approach` + `validated two hybrid`), plus one further `two hybrid
array` from the same CCSB lineage — method replication, not orthogonal validation. MI 0.67
throughout; no orthogonal assay in any of ADCK5's 54 IntAct records. NOTCH2NLA is
`Secreted`/`Cytoplasm` and human-specific; ADCK5 is mitochondrial. Not REMOVE — these are
experimental IPI rows whose primary data I have not read.

### Bioinformatics (`genes/human/ADCK5/ADCK5-bioinformatics/`)

MAFFT alignment of the five human UbiB proteins + yeast Coq8p + *E. coli* UbiB + **PKA Cα as
a negative control**. Every published reference residue is asserted against the downloaded
sequence before any column is read.

| column | ref | UbiB | PKA | **ADCK5** | discriminates? |
|---|---|---|---|---|---|
| KxGQ lysine | COQ8A:276 | K | *gap* | **K147** | yes |
| A-rich loop | COQ8A:339 ≡ PKA:53 | A | G | **A209** | yes |
| catalytic-loop Asp | COQ8A:488 | D | D | D360 | no |
| DFG Asp | COQ8A:507 | D | D | D382 | no |
| β3 Lys / αC Glu / cat-loop Asn | PKA:73/92/172 | K/E/N | K/E/N | K228/E281/N365 | no |

**ADCK5 retains both features Stefely et al. identified as inhibiting protein-kinase
activity**, and retains the full catalytic core — so it is neither a canonical kinase nor a
dead pseudokinase. The two independently anchored A-rich-loop columns resolve to the *same*
alignment column (391); the script asserts that rather than assuming it.

Family census (complete, non-truncated queries): **ADCK5 is the only human UbiB gene with
zero IBA annotations** (UniProt agrees: `PAN-GO; Q3MIX3; 0 GO annotations based on
evolutionary models`). On the EC/keyword split, `family_census.json` is the authority and it
is finer-grained than an earlier draft of this paragraph said: **ADCK1 also carries
`EC 2.7.-.-`**, so the pattern is that the two members whose activity was actually assayed —
COQ8A and COQ8B — lost *both* the specific EC and the `Serine/threonine-protein kinase`
keyword, ADCK1 lost the EC but kept the keyword, and **ADCK5 and ADCK2 kept both**. ADCK5
retaining `EC 2.7.11.-` and the keyword while its entry says the function is unknown is the
finding; `RESULTS.md` and the review YAML state it correctly.

**Correction (round 6):** an earlier version of this review argued the UniProt `SL-0173` vs
`SL-0162` asymmetry was "in UniProt's SUBCELLULAR LOCATION line, not in the evidence", on the
grounds that all three genes carry the same MitoCoP HTP row. **That was wrong** and became
checkable once ADCK1's review merged: ADCK1 and ADCK2 each carry
`SUBCELLULAR LOCATION: Mitochondrion {ECO:0000269|PubMed:33988507}`, an experimental
annotation ADCK5 lacks, and MitoCoP is not cited in either entry. The UniProt correction
request survives — the paper states *"ADCK5 and OBSCN were absent from the library"*, so
ADCK5 is **untested rather than tested-and-negative** — but it is now posed on ADCK5's own
evidence rather than on a parity that does not hold.

PAINT: `PTHR43173` has exactly one annotated node, `PTN005148758`, seeded by a single yeast
protein (MCP2, itself SF19). ADCK1 (SF19) inherits its three terms; ADCK5 (SF28) inherits
none. That is a **reach/coverage gap, not a mis-propagation** — the inverse of this
campaign's usual finding — and it is raised as a question rather than acted on, because one
yeast seed is thin support for a family-wide extension.

### Avoiding the mirror error

"UbiB proteins are not protein kinases" would be **too strong**. COQ8A/Coq8p lack *generic
in-trans* activity, but `PMID:38425362` shows *"COQ3, but not COQ6, is phosphorylated by
COQ8B at multiple sites"* — a real protein substrate. (Caveat recorded: that study used
ancestral-sequence-reconstructed COQ8 proteins.) So the correct reading is that *generic*
Ser/Thr kinase activity is unsupported and architecturally disfavoured, while a *specific,
substrate-restricted* activity remains open. GOA's current silence on MF is the right state.

### Checks that came back negative (recorded so they are not re-run blindly)

- **Retractions/errata:** none on any of the 12 PMIDs consulted — `CommentsCorrections/RefType`
  on each cited record, **plus** Crossref `relation`/`update-to` for the six load-bearing DOIs
  (the null-PMID corrigendum hole).
- **Partner identity:** Q7Z3S9 is a reviewed, canonical, full-length Swiss-Prot entry — no
  TrEMBL/partial-ORFeome substitution as on ACRV1.
- **Complex-to-subunit projection:** `PMID:34800366` is a proteome-wide localisation census
  with no functional/phenotype term, so the ACTR8 failure mode cannot apply.

### Two warnings deliberately left standing

`No core functions defined` — ADCK5's molecular activity is genuinely undetermined (UniProt
says so; Pharos **Tdark**), and this review found no measurement to replace that. Inventing a
core function to silence the warning would be exactly the failure CLAUDE.md warns about.
`No annotations reference the deep research file` — affinage's content (SOX9, senescence, JQ1)
underpins none of the four GOA rows, so citing it in an annotation would be artificial. Its
central claim is also **overstated**: it reports "ADCK5 phosphorylates SOX9 at serine 181" as
fact, where `PMID:32277958`'s abstract says ADCK5 *"might regulate"* PTTG1 by phosphorylating
SOX9 and reports no kinase assay. Flagged `LOW_QUALITY` in `reference_review`.
(`gates_passed: True`, `faith_pct: 50.0`, all citations numeric PMIDs.)

### Reproducibility

Deleting the JSON/FASTA outputs and re-running reproduces all four byte-for-byte. Both
analysis scripts and the claims audit ship `--self-test` break-tests, exercised in the
direction each guard exists to catch **and** in the happy direction. Two real defects were
caught that way and are documented in `RESULTS.md`: the anchor guard caught PKA residue
numbers being off by one (classical Knighton vs UniProt numbering), and the claims audit's
first residue check **failed on perfect agreement**, reporting K147 and K228 as conflicting
when both are correct.

---

> [!NOTE]
> **Low Risk**
> New gene-folder documentation and offline Python audits only; no changes to shared application code or GO term seeds beyond this gene’s review artifacts.
> 
> **Overview**
> Adds a **complete curated review** for human **ADCK5 (Q3MIX3)** as a UbiB atypical kinase with **no assayed activity**, arguing the “serine/threonine kinase” label is a **UniProt keyword/EC artefact** rather than something to fix in GOA (QuickGO has no kinase MF rows for this accession).
> 
> The new **`ADCK5-ai-review.yaml`** reconciles four GOA rows: **ACCEPT** prediction-derived membrane and MitoCoP mitochondrion; **MARK_AS_OVER_ANNOTATED** for duplicate **NOTCH2NLA** `GO:0005515` IPI hits (one Y2H screen triple-logged plus a second CCSB Y2H, no orthogonal assay). It retracts earlier **ADCK1/ADCK2 vs ADCK5 SubCell parity** framing in favor of **PubMed:33988507** (ADCK5 absent from the kinome library) and files **UniProt/PAINT/ontology** questions rather than proposing new GO terms for ADCK5.
> 
> Under **`ADCK5-bioinformatics/`**, the PR adds stdlib analysis scripts (**motif alignment**, **five-gene GO/EC census**, **IntAct partner localisation**), committed **`results.json` / `family_census.json` / `partner_localisation.json`** and alignments, plus **`audit_adck5_claims.py`** with **`--self-test`** and frozen fixtures so YAML, notes, and RESULTS cannot drift from computed residue calls, census numbers, withdrawn phrasings, or cross-gene database claims.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca69aa2b5f020862612dc19a44d2ff2d40d8cba6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — this was a genuinely useful review, and I've taken all five points. One push, no interim comments.

### 1 (blocking) — `knowledge_gaps` populated

You're right that the finding belonged in a structured slot, not prose. Two entries added, both grounded in quotes already verified in this PR:

- **`BIOLOGY` / `WHOLLY_DARK` / `OPEN`** — the undetermined catalytic activity, substrate and process. `boundary` separates what is established (mitochondrial localisation; the complete catalytic core K228/E281/D360/N365/D382 behind both UbiB inhibitory features K147/A209) from what is not (any measured activity, nucleotide preference, substrate, process, submitochondrial compartment or topology). It explicitly notes that the fold argument **constrains but does not close the question in either direction** — COQ8A lacks canonical in-trans activity, yet COQ8B phosphorylates COQ3.
- **`ONTOLOGY` / `MF_DARK` / `OPEN`** — as you suggested, mirroring suggested_question 4. Its `boundary` records the cost on the *characterised* paralogs too: COQ8A's demonstrated ATP hydrolysis is carried as generic `GO:0016301` alongside `NOT|GO:0004672`. No `proposed_terms`, deliberately — for ADCK5 there is no measured activity for a term to record.

No `core_function` was invented.

### 2 — `review.reason` on all four rows

Was 0/4, now 4/4.

### 3 — the membrane row's anchor didn't speak to membranes

Agreed, and this was the one substantive evidence defect. Re-anchored to `PMID:25498144` *"they contain predicted single-pass transmembrane (TM) domains that could limit solubility"* (the sentence is about ADCK1–5) plus `PMID:27499294` *"endogenous COQ8A is partially buried in the inner mitochondrial membrane with its C-terminus facing the matrix"*. Row 3 now has a `supported_by` too.

While re-checking these I caught one of my own: a `file:` quote I'd added used a hyphen where `RESULTS.md` has an em-dash. **The repo validator skips `file:` quotes entirely**, so it passed silently — I now verify every `file:` quote by hand against its target and require it to sit on one physical line. All 3 pass; `checkquotes.py` reports 16/16.

### 4 — the prose-only numbers are now computed

`partner_localisation.py` emits `methods_by_pmid` and `mi_scores`:

```
"mi_scores": [0.67],
"methods_by_pmid": {
  "25416956": ["two hybrid array", "two hybrid prey pooling approach", "validated two hybrid"],
  "31515488": ["two hybrid array"]
}
```

A single distinct MI score across every row is itself evidence the rows aren't independent, so it's worth having computed. `audit_adck5_claims.py` gained 4 checks binding the prose to these — including the case where IntAct starts returning *more* than one score, which would falsify "0.67 throughout".

I break-tested each new check for **reachability before** adding it, rather than trusting that it passed: a check that can never fire reads as coverage while providing none. All four fire. 16 self-tests total, each exercised in the catch direction and the happy direction.

### 5 — conceded, the topology argument assumed a sidedness I decline to assert elsewhere

This is the sharpest point and you're straightforwardly right. Calling the pairing "topologically implausible" while refusing `GO:0031966` because submitochondrial placement isn't established is inconsistent. The summary now states it as an assumption — it holds if ADCK5's kinase-like domain faces the matrix as COQ8A's C-terminus is measured to do, but ADCK5's own sidedness has never been determined and an OMM anchor presenting the domain to the cytosol isn't excluded — and says plainly that the verdict doesn't depend on it, since method replication stands alone. Same softening applied in the notes.

### One thing I verified rather than asserted, prompted by this

The "GOA no longer imports keyword-derived annotations" claim was load-bearing and I'd been taking it from background knowledge. Checked, **with a control**: QuickGO returns **0** human annotations for `GO_REF:0000043`, against **139,714** for `GO_REF:0000044` (SubCell) and **1,862** for `GO_REF:0000041` (UniPathway) — so the zero is specific to the keyword pipeline, not a query artefact. And `GO:0004674` is alive in GOA: PRKACA carries 31 annotations to it. ADCK5 receives it by no route at all.

Your note that `PMID:25498144` also raises a competing small-molecule-kinase hypothesis for the KxGQ domain is well taken — `PMID:38425362`'s GC/MS negative is cited in the review for exactly that reason, and the `BIOLOGY` gap's `gap_statement` now enumerates all three possibilities (protein substrate, small molecule, or ATP hydrolysis alone) rather than implying the question is only about protein phosphorylation.

**Gates:** `just validate` ✓ Valid (2 warnings, both deliberate and explained); `checkquotes.py` 16/16; strict duplicate-key loader clean; raw-vs-parsed `reference_id` 12 = 12; 4 GOA rows = 4 annotations; `cache/go/terms.csv` untouched (no deletions vs merge base, no new rows, only the 2 pre-existing duplicate curies).

cmungall:
Correction to my previous comment, found by re-reading the change out of the tree rather than trusting that I had made it.

**I reported the topology concession (your point 5) as done. It was done in one of three places.** The assertive opener survived in the very summary I had just hedged — so the row asserted "topologically implausible" and then, two sentences later, explained that it was only an assumption — plus a reference finding at `:479` and the `partner_localisation.py` docstring.

This is the campaign's most-repeated failure ("fixed in N places, landed in N−1") and I walked straight into it while writing a lint whose stated purpose is catching exactly that. Fixed properly in `419dc9e`:

- all three now state the compartment point as a *supporting consideration*, consistent with the review's refusal to assert a submitochondrial placement;
- `topologically implausible` and `topological objection` are pinned in `WITHDRAWN_PHRASES`;
- **the withdrawn-phrase scan now covers the whole gene folder, not the three prose files.** That is the real lesson here: my detector's scope was narrower than the set of places the claim could live, so it was structurally blind to two of the three occurrences — one of them in a script docstring, which no prose-only scan would ever reach.

Widening it immediately turned up a **fourth defect in the audit itself**: the scan matched its own definition list, since the registry of withdrawn phrasings necessarily contains every one of them. Excluding "`.py` files" would have fixed the symptom and re-blinded the scan to the docstring case I had just fixed. So the exclusion is exactly one file — this script — matched by resolved path, and a self-test now plants a phrase in a *sibling* script and requires a catch, so the exclusion cannot silently widen later. 18 self-tests, each run in the catch direction and the happy direction.

Running tally of defects the break-tests found in the audit script, none of them by reading it: a residue check that **failed on perfect agreement** (K147 vs K228, both correct); a withdrawn-phrase matcher that fired on the file's own explanation of the correction it was policing; a keyword-argument mismatch that aborted the suite halfway while earlier checks still printed as passes; and this self-reference. All four are now logged in `RESULTS.md`.

Everything in my previous comment stands otherwise — I re-verified each claim against `HEAD` this time rather than against my intent. Gates on the new head: `just validate` ✓ Valid (same 2 deliberate warnings); `checkquotes.py` 16/16; all 3 `file:` quotes verbatim and on one physical line (hand-checked, since the repo validator skips them); `cache/go/terms.csv` untouched.

cmungall:
All four addressed in `1070def0`. The blocking one was right, and the diagnosis of *why* my lint missed it was the most useful thing in this review.

### The fourth occurrence, and the real lesson

`RESULTS.md` **was** in scope — `WITHDRAWN_SCAN_GLOBS` covers `ADCK5-bioinformatics/*.md`. What the scan lacked was **vocabulary**: the paragraph paraphrased the withdrawn conclusion instead of repeating a pinned literal. I widened the detector's *scope* last round and not its *vocabulary*, and a literal-phrase matcher cannot catch paraphrase by construction. Adding two more literals would have been a guard that reads as coverage it doesn't have.

So the compartment claim is now guarded **structurally**, keyed on the topic rather than the wording of the conclusion:

> any paragraph that states NOTCH2NLA's own localisation — the contrast the argument is built from — must carry a statement that ADCK5's sidedness is unmeasured, in that paragraph or the next.

Rewording the conclusion can't evade it, because the trigger is the topic.

**Two things only came out of testing it against reality rather than against my imagination.**

**A file-level version of that invariant passed on the actual text that shipped.** I wrote it file-level first, ran it against `419dc9e3:RESULTS.md`, and it returned clean — because the historical file hedged in a *different section*. File-level co-occurrence is not proximity. That is exactly the failure mode you named, and I'd have shipped it as "fixed" if I hadn't replayed the real text. It's now paragraph-local with a one-paragraph lookahead, and `--self-test` replays that historical paragraph **verbatim from git** as a regression fixture rather than a synthetic mutation. A synthetic mutation only proves the guard catches the failure I imagined.

**The first trigger was too broad** — "mentions the partner AND any compartment word" produced three false positives on an IntAct partner table and two verdict tables. Narrowed to the localisation contrast (`Secreted|Cytoplasm|cytosol`), which is what actually distinguishes an instance of the argument from a mention of the partner.

**Residual limit, stated in the file rather than hidden:** a hedge more than one paragraph away still evades this. The check narrows the window; it does not close it, and withdrawing a claim still calls for a human re-read.

### The three non-blocking items

- **`--self-test` no longer mutates a tracked file.** It writes a throwaway probe, refuses to run if the probe path already exists, removes it in a `finally`, and then asserts the tree is clean. (Your point stands generally: a self-test that dies between write and restore leaves the repo dirty.)
- **`RESULTS.md` double-scan fixed.** Both scanning checks now share one scope definition, `all_surfaces()`, keyed by path relative to the gene directory. Detector and mutator disagreeing on scope is how the docstring occurrence hid in the first place, so having two of them was asking for it. 7 surfaces, each once.
- **The COQ8B contradiction is now written up**, in `RESULTS.md` and as concrete evidence for this review's `ONTOLOGY` knowledge gap. GOA carries both `NOT|enables GO:0004672` (IDA, `PMID:27499294`) and `enables GO:0004672` (IDA, `PMID:38425362`) for COQ8B, and I confirmed the merged COQ8B review `ACCEPT`s both rows without noting the conflict. They're reconcilable in prose — no generic in-trans activity against standard substrates; phosphorylation of one specific partner, COQ3 — but GO can't express "not a generic protein kinase, but does phosphorylate this one substrate", so the record reads as a flat contradiction. **That is the same ontology gap ADCK5's second `knowledge_gaps` entry records, showing its cost on a *characterised* member rather than a dark one.** Reported, not fixed: COQ8B is out of scope here.

### Gates on `1070def0`

`just validate` ✓ Valid (same two deliberate warnings); `checkquotes.py` 16/16; strict duplicate-key loader clean; all 3 `file:` quotes verbatim and single-line (hand-checked — the repo validator skips them); `reference_id` raw 12 = parsed 12; 4 GOA rows = 4 annotations; `knowledge_gaps` 2; `review.reason` 4/4; `cache/go/terms.csv` untouched; working tree clean after `--self-test`; **22 self-tests**, each in the catch direction and the happy direction.

cmungall:
Both blocking items fixed in `d0a725bf`, and both were right.

### 1 — retracted: the COQ8B review does not miss the contradiction

My claim was wrong, and the way it was wrong is the point. I read the *structured* fields — term, evidence, `negated`, `action` — saw `ACCEPT` on both rows, and never opened the summaries. The NOT row says:

> "…refers to general in-trans kinase behaviour; it does not contradict the later, specific ATP-dependent COQ3 phosphorylation by COQ8B (PMID:38425362)"

which is the same reconciliation my own PR then offers as its reading. **Checking `action:` values is not checking a review** — the same incomplete-evidence mistake CLAUDE.md warns about, committed against a sibling curator.

Reframed as what it actually is, and it's a stronger point: an **ontology-expressivity gap, not a curation defect**. GO cannot say "no activity in trans toward general substrates, but yes toward this one specific substrate", so the curator is forced to carry the reconciliation in free text while the term-level record stays self-contradictory to anyone reading terms. COQ8B is arguably the *better* illustration of ADCK5's `ONTOLOGY` knowledge gap than ADCK5 is, since there the biology is known and the ontology still can't carry it. Corrected in `RESULTS.md`, in `suggested_questions[3]` and in the gap's `boundary`, with the retraction noted in place rather than quietly rewritten.

### 2 — the "paragraph-local" invariant was file-level on the surface that mattered most

Confirmed empirically before fixing: `ADCK5-ai-review.yaml` has **zero** blank or whitespace-only lines, so `_paragraphs()` returned the whole 550-line file as **one** paragraph. The hedge in the `GO:0016020` row was satisfying the check for an unhedged claim in a different annotation 60 lines away — precisely the file-level failure my own docstring says was demonstrably inadequate.

This is the **fifth instance in this PR of one shape: a check whose unit of analysis is coarser than the unit the claim lives in.** I keep fixing the instance and not the shape.

The splitter is now structure-aware: YAML splits per **scalar value** (each `summary`, `reason`, `gap_statement` its own unit — **181** units, not 1), markdown keeps blank-line paragraphs, and the one-paragraph lookahead applies **only** to markdown, since an adjacent YAML scalar is usually an unrelated key and letting it satisfy the hedge is how this went wrong in the first place.

Fixing it immediately caught row 3 — exactly as you predicted — asserting the argument flatly under the renamed label "compartment objection". Row 3 now carries its own hedge, and the vocabulary is normalised across the folder (no `compartment objection` / `topological objection` / `topologically implausible` remain outside the lint's own registry). No verdict changed; row 3 rested on method replication already.

### The four suggestions

- **`HISTORICAL_MISS_COMMIT` — you were right to treat this as blocking-in-practice.** A branch-local SHA does not survive squash-merge, so the strongest test in the suite would have lost its fixture the moment this PR landed, silently. The paragraph is frozen as `fixtures/historical_unhedged_compartment_paragraph.md`, a missing or corrupted fixture is now a hard failure rather than a skip, and a new harness invariant asserts the fixture is **excluded** from the live scan — it preserves the bad text on purpose, so a future recursive glob would otherwise make the audit fail on itself.
- **`expect_flag` now takes `match=`.** Not theoretical: replaying the historical paragraph fires **two** messages, only one of which is the compartment guard (the other is a `17 of 25` required-claim miss, since the fixture is a paragraph rather than a whole file). Without `match=` that test would have passed on the unrelated one.
- **`all_surfaces()` falls back to disk** for surfaces a partial dict omits, instead of blanking them — asserted by a new harness invariant, since blanking would shrink the corpus and turn a partial dict into a false pass.
- **Dead imports removed** (`import re as _re`, and `subprocess` once the fixture replaced `git show`).

### Gates on `d0a725bf`

`just validate` ✓ Valid (same two deliberate warnings); `checkquotes.py` 16/16; strict duplicate-key loader clean; all 3 `file:` quotes verbatim and single-line, hand-checked again after RESULTS.md was rewritten (the repo validator skips them); `reference_id` raw 12 = parsed 12; 4 GOA rows = 4 annotations; `knowledge_gaps` 2; `review.reason` 4/4; `cache/go/terms.csv` untouched; tree clean after `--self-test`; **24 self-tests**, all passing in both directions.

cmungall:
Thanks for the approval. I've taken all four non-blocking points in `4f302db1` — two of them were real defects that would have shipped, so they were worth the extra round.

**(A) The PyYAML dependency — removed, not documented around.** You're right that it was the folder's first third-party import while `RESULTS.md` documents a bare `python3` and the three sibling scripts are stdlib-only. It happened to work here because my system `python3` has PyYAML; that is exactly the kind of defect that only appears on someone else's machine. Rather than change the documented command, I removed the need: the YAML surface is now split at every mapping key by a line scan. **All four scripts are stdlib-only again** (plus `mafft` on `PATH`), and the split yields **233** units — slightly more than the parse gave, since keys with short scalar values become units too, which is if anything the safer direction.

**(B) The unparseable-YAML fallback — the failure branch is gone, not hardened.** You identified it precisely: returning one blob silently restored the file-level behaviour the rewrite exists to abolish, i.e. the guard's error path reproduced the bug the guard was written to fix. Removing the parse removes the branch — no parse, no fallback. What replaces it is a positive invariant: a YAML surface yielding fewer than two units is a **hard error** naming the pattern to check, since one unit *means* the hedge invariant has gone file-level. Break-tested in that direction.

**(C) `match=` — now on all 19 sites.** This was not cosmetic and your instinct about which sites mattered was right: replaying the historical fixture fires **two** messages, only one of them the compartment guard (the other is a `17 of 25` required-claim miss, because the fixture is a paragraph rather than a whole file). So the single most important test in the suite was passing on a check unrelated to what it was testing. Asserting that *something* failed is not asserting that the thing under test failed.

**(D)** Dead `m_partial` binding removed; the "verbatim from git" comment now describes the frozen fixture that replaced the `git show`.

**Gates on `4f302db1`:** `just validate` ✓ Valid; `checkquotes.py` 16/16; all 3 `file:` quotes re-verified verbatim and single-line after `RESULTS.md` was rewritten again; `cache/go/terms.csv` untouched; 24 self-tests passing in both directions. No annotation, verdict, `knowledge_gaps` entry or reference changed in this commit — it is confined to the analysis scripts and their documentation.

I have nothing further to push.

---

One thing worth recording for the campaign rather than for this gene. Five defects in this PR had **one shape**: a check whose unit of analysis was coarser than the unit the claim lived in — file-level scan vs. per-annotation claim; blank-line paragraphs vs. a YAML file with no blank lines; `expect_flag` asserting a run-level failure vs. a specific guard; a lint scanning prose files vs. a claim living in a docstring; and the parse fallback collapsing a file back to one unit. Each was found by an external check rather than by re-reading, and each time I fixed the instance rather than the shape. Naming the shape is probably more useful to the next agent than any of the individual fixes.

cmungall:
Merged `origin/main` to clear an add/add conflict introduced when ADCK1 landed (`c0a6dacb`). No review content changed.

**The conflict:** ADCK1's agent and I both ran `fetch-panther-paint PTHR43173` 26 minutes apart, so both branches added `interpro/panther/PTHR43173/PTHR43173-metadata.yaml`. I diffed the two versions myself rather than assuming: both 56 lines, differing on **exactly one line** —

```
53c53
<   fetched_date: '2026-07-26T18:54:25.604019'    (mine)
>   fetched_date: '2026-07-26T18:28:50.474349'    (main, via ADCK1)
```

Resolved by taking **main's** version, so the file now shows *no change at all* in this PR and is not churned for the other branches still open. Asserted byte-identical to `origin/main` afterwards. The other two family files (`-entries.csv`, `-paint.tsv`) were already byte-identical, so nothing else in that directory is touched.

**Re-verified on the merged tree**, since a later merge can undo an earlier resolution:

- `cache/go/terms.csv` by **multiset**: 9,074 → 9,076, and both new rows (`GO:0030485`, `GO:1903852`) are inbound from main — `comm` confirms my branch adds nothing main lacks, and there are no deletions against the merge base. The two pre-existing duplicates `GO:0001675` and `GO:0009566` are still present exactly twice each.
- My gene files: `git diff 4f302db1 HEAD -- genes/human/ADCK5/` is **empty** — the merge touched none of them.
- `just validate human ADCK5` ✓ Valid on the merged tree, and the working tree is **clean afterwards** (no cache mutation).
- `checkquotes.py` 16/16; claims audit clean; 24 self-tests passing.

The two expected consequences are both correct and intended: the head moved, so the previous `APPROVED` no longer covers it, and the quiet clock resets.

cmungall:
Fixed in `3d428262`. The parity premise was wrong and I verified that in-tree before changing anything.

### The wrong claim, and what replaces it

`ADCK1-uniprot.txt:117` reads `SUBCELLULAR LOCATION: Mitochondrion {ECO:0000269|PubMed:33988507}`; grepping `34800366` in that file returns nothing; ADCK2's entry (checked via the UniProt API — **`genes/human/ADCK2/` does not exist in this tree**, and I'm saying so rather than implying I read it) carries the same `ECO:0000269|PubMed:33988507`. So both paralogs have a dedicated experimental localisation that ADCK5 lacks. My "same evidence, different treatment" framing was false, and it was sitting in text addressed to curators who would have checked it.

**The ask survives, and the reason why is better than parity ever was.** That paper states outright:

> *"ADCK5 and OBSCN were absent from the library"*

and QuickGO returns **0** annotations for Q3MIX3 from `PMID:33988507`. So ADCK5's missing annotation is **untested, not tested-and-negative** — which is precisely what makes a correction request reasonable rather than a request to overrule data. All three surfaces now pose it on ADCK5's own evidence: MitoCoP membership, 17 of 25 mitochondrial interactome partners, and the family's exclusively mitochondrial distribution. `PMID:33988507` added as a reference with that verbatim quote and a `review_notes` recording why it is cited.

### On the pattern

You're right that it's the same shape as the COQ8B claim, and right that naming it matters more than either fix: **a statement about why another entry looks the way it does, made without opening that entry's evidence line.** Twice, three surfaces each time. The structural reason both spread so far is that a cross-gene premise gets asserted once and then *reused* as established background — so it propagates through exactly the surfaces where it's least checkable, and my own lint can't help, because it verifies internal consistency and a consistently-repeated false premise passes.

### Three code items

- **`_paragraphs()` no longer raises.** You were right to connect it to the ACTA1 shape: a helper that raises aborts every later check *and the self-test baseline* while the harness still prints as though it ran. The under-split condition is now reported by `check_compartment_hedge` as a collected problem, and break-tested (`match="split into 1 unit"`).
- **The vacuity counter was corpus-wide**, so it stayed satisfied as long as *any* surface matched — meaning the guard could go blind on the review YAML, the one surface where the annotation claims live, and still report itself exercised. Now per-surface for the two surfaces that must match, with its own break-test.
- **Stale `_paragraphs` docstring** corrected to describe the line scan that replaced the parse.

26 self-tests, all passing in both directions.

### PR body

Corrected via `gh api -X PATCH`. `family_census.json:24` is right and my summary was loose: **ADCK1 also carries `EC 2.7.-.-`**. The actual pattern is that the two assayed members lost *both* the EC and the keyword, ADCK1 lost the EC but kept the keyword, and ADCK5 and ADCK2 kept both. `RESULTS.md` and the YAML already had this correct. The body also now carries the parity retraction.

### Thanks for the ADCK1 pointer

`ADCK1-ai-review.yaml:96-105` does answer my PAINT question in the negative, and independently: MCP2 is the ortholog and carries its own `GO:0005743` IDA, so the SF19 scope is appropriate rather than a coverage failure. That is a better answer than my own hedge, and I'd rather have my question closed by a sibling's evidence than left open.

**Gates on `3d428262`:** `just validate` ✓ Valid; `checkquotes.py` **17/17** (up one with the new reference); 3 `file:` quotes verbatim and single-line; strict duplicate-key loader clean; `reference_id` raw 12 = parsed 12; 4 GOA rows = 4 annotations; 11 references; `cache/go/terms.csv` untouched. No GO verdict changed.

cmungall:
Fixed in `7ddc0f02`. The fourth surface was a fair hit and the diagnosis of *where* it hid is the useful part.

### The fourth surface

`reference_review.review_notes` on `GO_REF:0000044`. Nothing in it was false, and that is exactly why it survived three sweeps: the **"whereas" was doing the argumentative work**, carrying the contrast without the correction. Now states that ADCK1/ADCK2's SL-0173 rests on an experimental localisation from a screen whose library did not contain ADCK5 — untested versus tested, not identical evidence handled differently.

Both surviving occurrences of a retracted claim in this PR have been on **low-salience surfaces** — first a script docstring, now a `reference_review` note. I swept by file both times, which is why I missed both. This time I enumerated by **surface type** (review prose, `reason`, `summary`, `reference_review` notes, `knowledge_gaps` provenance, script docstrings, JSON, `RESULTS.md`, PR body) and walked the parsed YAML unit-by-unit with paths — which is how it turned up, along with two hits I checked and correctly left alone (a bare `supporting_entities` WITH/FROM token, and a statement that mentions SL-0162 without contrasting anything).

### A structural guard for the second retracted claim

You're right that a literal pin is the wrong instrument — paraphrase already defeated one here. `check_parity_hedge` triggers on the **contrast**: any unit mentioning ADCK1/ADCK2 alongside a UniProt subcellular concept must carry the untested-assay clause. Break-tested against the **real pre-retraction text**, frozen as `fixtures/historical_unhedged_parity_units.md`.

**Writing it immediately caught a defect in the splitter it depends on.** `YAML_KEY_RE` matched prose *inside block scalars*, so the sentence `…not fixable in GO: ADCK1 and ADCK2 receive SL-0173…` was cut in two at `GO:`, separating a claim from the clause qualifying it two sentences later — producing a false positive on already-corrected text. A splitter that **fragments** a scalar is as wrong as one that **merges** a whole file; I had fixed one direction and created the other. Now block-scalar aware, with its own self-test using that exact sentence shape.

### The ADCK2 half now has in-tree provenance

The sharpest of your points: a verification that exists only in a PR comment is not one a reader can repeat. `family_annotation_census.py` now captures every gene's `SUBCELLULAR LOCATION` with its evidence tags into `family_census.json`, and **asserts** the corrected claim — the run fails if ADCK1 or ADCK2 loses its `ECO:0000269` mitochondrial location, or if ADCK5 gains one (which would make the correction request stale rather than wrong).

That produced a sharper fact than I had:

```
ADCK5   Membrane                ECO:0000305
ADCK1   Mitochondrion           ECO:0000269|PubMed:33988507
ADCK2   Mitochondrion           ECO:0000269|PubMed:33988507
COQ8A   Mitochondrion membrane  ECO:0000269|PubMed:11888884,25498144,33988507
COQ8B   Mitochondrion membrane  ECO:0000269|PubMed:24270420,33988507
```

**That screen supplied the mitochondrial localisation for the entire rest of the human UbiB family. ADCK5 is the only member it could not assess.**

### The stronger reading

Recorded, with the quote. The passage reconciles the screen's 22 mitochondrial kinases against MitoCarta2.0 and distinguishes two kinds of exception: genes *absent from the library* (ADCK5, OBSCN) and genes that were in the library and which the authors tested and called non-mitochondrial (FASTK, PAK5). ADCK5 is in the first group — a candidate the assay never reached, not one it examined and rejected. Added as a second `finding` on `PMID:33988507`.

`REQUIRED_CLAIMS` now covers `absent from the library` and `33988507`.

**Gates on `7ddc0f02`:** `just validate` ✓ Valid; `checkquotes.py` **18/18**; 3 `file:` quotes verbatim and single-line; strict duplicate-key loader clean; `reference_id` raw 12 = parsed 12; 4 GOA rows = 4 annotations; 11 references; `cache/go/terms.csv` untouched; **29 self-tests**, two of which replay real defects from this PR's own history. No GO verdict changed.

cmungall:
Fixed in `9ed26026`. You were right, and the refutation came from a file I committed in the same commit as the claim.

### Corroborated, not supplied

`family_census.json` records COQ8A citing `PubMed:11888884` (2002) and `25498144` alongside the screen, and COQ8B citing `PubMed:24270420` (2013). So the screen is the **sole** cited evidence for ADCK1 and ADCK2 — there it supplied the annotation — and **one tag among several, and the latest**, for COQ8A and COQ8B, where it corroborated one that already existed. Corrected in both surfaces, with the per-gene tags shown in a table rather than summarised.

The load-bearing conclusion is unaffected: all four paralogs were in the library, ADCK5 was not.

### The calibration point — taken, and acted on rather than acknowledged

You're right that this is the third instance of one class, and right that the fix isn't more careful prose. `subcellular_evidence()` had already computed the answer; I then wrote a sentence *about* the data instead of *from* it. So the census now computes `mitochondrial_localisation_provenance`, partitioning the family into `sole` / `corroborating` / `absent`, prints it, and **asserts** it — the run fails if the partition changes:

```
PMID:33988507 is the SOLE cited evidence for ['ADCK1', 'ADCK2'],
one tag among several (corroborating) for ['COQ8A', 'COQ8B'],
and absent for ['ADCK5'].
```

That is the same move that fixed the ADCK2 provenance last round, applied to the class rather than the instance: **anything computable, compute; then make the sentence answerable to the computation.** The three instances of this class in this PR were all claims I could have computed and chose to phrase.

### (2) is a real latent bug and the silent direction is the bad one

Confirmed: `loc['location'] == 'Mitochondrion'` against a JSON that already contains `'Mitochondrion membrane'`. A refinement of ADCK1's line would have raised a false alarm — noisy but visible — while **a refinement of ADCK5's line would have made both staleness checks pass silently**, leaving a satisfied correction request standing as if still open. `startswith` throughout, and the subsuming pair collapsed to one check.

### (1), (3), (4)

- **(1)** `check_parity_hedge`'s vacuity counter is now per-surface — it was the exact corpus-wide form `check_compartment_hedge` rejects twelve lines above, which is its own small lesson about copying a guard without copying its correction.
- **(3)** Broadened `PARITY_CONTEXT_RE` to see the heading — then **narrowed it again**, because a bare `mitochondrial localisation` alternative fired on the PAINT question and a `knowledge_gaps` entry, neither of which is about SubCell treatment. The heading is now self-contained as well, and I've documented why this guard deliberately has **no** lookahead unlike the compartment one: the unit most needing qualification here is a heading, and a heading is read in tables of contents, diffs and search results, i.e. furthest from its body.
- **(4)** The notes' Localisation section had no blank lines and ran as one list-level unit. Split per bullet — and since this is the coarse-unit shape for the third time, it now has an **invariant** rather than a fix: any unit over 1500 chars in either markdown surface fails the suite.

### Verification

You flagged that you could run neither `just` nor the self-tests, so I re-ran everything: `just validate` ✓ Valid; `checkquotes.py` 18/18; audit clean; **31 self-tests** all passing in both directions; `family_census.json` reproduces byte-for-byte from a fresh run; `cache/go/terms.csv` untouched. No GO verdict, `knowledge_gaps` entry or `description` changed.

cmungall:
Done in `cec4ac5d`. You were right that instance repair was losing, so I did the systematic pass rather than fix the fourth one.

### The enumeration

32 prose units across the three surfaces name a paralog alongside an evidence token. Splitting them: the ones about **what a paralog was shown to do** (COQ8A lacks in-trans kinase activity, COQ8B phosphorylates COQ3) are anchored by `supporting_text` and already verified verbatim against cached publications — those are fine and deliberately stay out of scope. The problem class is narrower: claims about a paralog's **database record**.

### The invariant that closes it

`check_cross_gene_claims` re-asserts **eight dimensions** against the committed `family_census.json`: IBA rows, the MitoCoP `GO:0005739` HTP row, EC numbers, the Ser/Thr keyword, `SUBCELLULAR LOCATION`, `NOT|` rows, the screen-provenance partition, and the exact per-gene `ECO:0000269` tag sets the `RESULTS.md` table pins. The census now records `mitochondrion_go_rows` (evidence + reference per gene), which is the couple of lines you pointed at.

**The part that closes the class rather than patching an instance:** a prose unit naming a paralog alongside a record token in a dimension **nothing covers** fails. So the *next* uncovered cross-gene claim — the fifth instance, whatever it turns out to be — cannot pass silently; it has to be computed, scoped down, or deleted before the audit goes green.

Each of the eight dimensions is break-tested individually, plus the class-closer itself:

```
PASS (caught): cross-gene: ADCK1 loses its IBA row                          [iba]
PASS (caught): cross-gene: ADCK2 loses the MitoCoP row                      [mitocop_row]
PASS (caught): cross-gene: COQ8A's EC reverts to 2.7.11.-                   [ec]
PASS (caught): cross-gene: COQ8A regains the Ser/Thr keyword                [keyword]
PASS (caught): cross-gene: ADCK5 gains a mitochondrial SUBCELLULAR LOCATION [subcellular]
PASS (caught): cross-gene: COQ8B loses its NOT| rows                        [negated]
PASS (caught): cross-gene: the screen-provenance partition shifts           [screen_provenance]
PASS (caught): cross-gene: COQ8A gains a fourth localisation tag            [tag_sets]
PASS (caught): a cross-gene claim in a dimension nothing in the census asserts
```

`GO:0005739 HTP row from PMID:34800366 carried by: ['ADCK1', 'ADCK2', 'ADCK5', 'COQ8A', 'COQ8B']` — the ADCK2 half is now in-tree and asserted.

**It immediately caught me.** Writing the new `RESULTS.md` section describing all this tripped the *parity* guard on my own paragraph, because the paragraph names ADCK1/ADCK2 SubCell treatment. It now carries the qualification. That is the first time a guard here has caught a claim while I was writing it rather than a round later.

### The three suggestions

- **(1) was a real gap.** The coarse-unit invariant lived only in `--self-test` while its YAML counterpart ran every invocation — and the surfaces that actually regressed were the markdown ones, so a plain audit run would have gone green while both hedge guards silently degraded. Now `check_unit_granularity`, on every run, break-tested.
- **(2)** The tag table pinned sets nothing asserted. Now the `tag_sets` dimension, break-tested with a synthetic fourth COQ8A tag.
- **(3)** `alongside FASTK and PAK5` rewritten as two explicitly opposite kinds of exception — ADCK5/OBSCN absent from the library, FASTK/PAK5 in it and tested non-mitochondrial — so the grouping never happens.

### Verification, since your sandbox could run nothing

`just validate` ✓ Valid; `checkquotes.py` 18/18; audit clean; **39 self-tests** passing in both directions; `family_census.json` reproduces byte-for-byte; all 3 `file:` quotes re-verified verbatim and single-line after `RESULTS.md` was rewritten again; `cache/go/terms.csv` untouched. No GO verdict, `knowledge_gaps` entry or `description` changed — as in every round since the first.

cmungall:
Fixed in `cd526fec`. The catch was exact, and the sting is that I had already learned this lesson: two rounds ago I rejected literal pins for the compartment claim and built topic triggers instead, then used a fixed literal list for the third guard. `RECORD_TOKENS` mapping 1:1 onto `_coverage` made the class-closing branch unreachable for any real prose.

### I took option 2, because measuring first showed it was cheap

Before choosing, I counted: of the paralog-mentioning units carrying a **record-shaped** token, only **four** failed to route to a covered dimension — and three of those were PAINT-node claims already checkable against the committed `PTHR43173-paint.tsv`. So the generic trigger cost one new dimension and one exemption, not a sprawling allowlist. Had it come back messy I would have taken option 1 and said so plainly.

- **The gate is now generic**: any record-shaped token near a paralog mention — evidence code, `GO_REF:`, `ECO:` id, SubCell id, EC number, PANTHER node id. A bare `GO:xxxxxxx` is deliberately excluded, because it names a *term* and terms are discussed throughout the literature prose; including it produced 10 false triggers, excluding it produced 4 real ones.
- **Ninth dimension `paint_node`**, asserted against the committed PAINT table: one annotated node, three terms, single MCP2 seed.
- **Exactly one documented exemption**, with its reason recorded in `RECORD_GATE_EXEMPTIONS` rather than implied.

**Proof it closes the class rather than the enumeration:** the self-test now runs *your two predictions* of the fifth instance — a claim about which evidence code backs a paralog's positive kinase row, and one about a paralog's annotation and experimental counts — and requires both to fail **without any token being added first**. That is precisely the case the earlier version could not reach, and it is why I did not simply soften the prose.

### On not claiming more than it does

`RESULTS.md` now states the mechanism and its boundary explicitly: the trigger is generic, the term/record distinction is a deliberate exclusion with a stated reason, literature claims are out of scope because they are anchored by `supporting_text` and verified verbatim, and one unit is exempt by name. It also records that the earlier version was **unreachable for real prose while claiming to close the class** — the failure is documented rather than quietly overwritten.

### The three suggestions

(1) The vacuity counter was corpus-wide **for the third time in this file** — now per-surface. (2) "four blocking items" was wrong: ADCK2's SubCell provenance was a suggestion, not blocking; each item's status is now stated, and "the table above" pointed at a table below. (3) The bare `keyword` alternative matched the word in any sense; anchored.

**Two things my own guards caught while I was doing this**, which is new: the paint break-test mutated the tracked table and restored it, round-tripping its line endings and leaving the file dirty — the hazard already fixed once for the sibling probe, so content is now injected and a self-test asserts the mtime is untouched. And the new `RESULTS.md` section tripped both hedge guards on its own text; the hypothetical false claims now live in the audit script, since writing one into a scanned surface trips the guard being described.

I also **removed the hardcoded self-test count** from `RESULTS.md`. It drifted three times as the suite grew, and a hand-written figure tracking a computed one is the exact defect this file spends paragraphs arguing against.

### Stopping criterion

Stated, as asked. **I consider the review complete and I am not planning further changes.** From here I will only push for:

1. a **blocking** item — a wrong or unsupported claim, or a guard that does not do what it says;
2. a **merge conflict** with main;
3. a factual claim of mine shown to be false.

I will **not** push for further hardening of the audit harness. Nine of the last ten rounds have been analysis-infrastructure, not curation: **all four GO verdicts, the `knowledge_gaps` entries and the `description` have been unchanged and correct since round 1**, and the harness is now well past the point where marginal guard work earns its review cost. Non-blocking suggestions about the scripts I will acknowledge and leave, saying so rather than silently declining.

**Gates on `cd526fec`:** `just validate` ✓ Valid; `checkquotes.py` 18/18; audit clean; full self-test passing in both directions; `family_census.json` reproduces byte-for-byte; `interpro/` clean; 3 `file:` quotes verbatim and single-line; `cache/go/terms.csv` untouched.

cmungall:
Fixed in `dad9b3cc`. This one was blocking under my own stated criterion — a guard that did not do what it said — so it was worth the round.

### The gap was exactly as described, and I reproduced it before fixing

Routing was per **unit**, so one covered token blanket-passed every other record claim in the same unit. Confirmed in-tree: the COQ8B double-`GO:0004672` sentence resolves to `dims -> ['negated']` via its `NOT|` token, `negated` is covered, gate returns nothing. And `grep -c 38425362 family_census.json` was **0** — the reference was not in the census at all. The self-test "proved" the catch only because its probe was a standalone paragraph carrying no covered token, which is a hard lesson about probe placement: **a probe that is easier to catch than the real thing tests nothing.**

Both halves fixed:

- **Routing is per sentence**, and the paralog and the record token must appear in the **same** sentence. My first attempt kept the paralog check at unit level while routing at sentence level — the same scope mismatch this file keeps rediscovering — and it flagged six sentences about ADCK5's *own* evidence codes.
- **Tenth dimension `kinase_rows`**, asserting the **exact** row set (term, evidence, qualifier, reference) for COQ8A and COQ8B. Exact rather than subset, because routing a claim to a dimension has to mean the claim is actually covered.

I took your second option rather than the cheap one, since the data was already computed and this is the cross-gene claim a curator is most likely to act on. But I also took the first: **the residual limit is now stated** — per sentence is not per claim, so a sentence mixing a covered and an uncovered record claim still passes. Regex cannot do better honestly, and pretending otherwise is precisely how the per-unit version came to overclaim.

### The three suggestions

**(1) was a real bug.** `scalar_indent` came from the *line* indent, so `- gap_statement: >-` at indent 0 absorbed its siblings at column 2 — one 2.4 kB unit, not five. Now the key's own column: **181 → 236** YAML units, gap entries 516 and 592 chars. Guarding it, though, produced another instance of the trap: my inline "no unit holds two sibling keys" check was **unreachable** once the splitter was correct. Replaced with a splitter-regression probe that is break-tested by **swapping the pre-fix splitter back in** and requiring it to fire.

**(2)** `RECORD_GATE_EXEMPTIONS` now applies per sentence, so it exempts the clause that motivated it rather than a 1.4 kB summary.

**(3)** Documented in `paint_node` itself: `PTHR43173-paint.tsv` is shared repo state this PR does not own, so a failure after someone re-fetches the family means *the upstream table moved*, not that this review drifted.

### Two more the self-test surfaced while I worked

Two probes had to be **replaced because `kinase_rows` now legitimately covers them** — recorded in a comment rather than deleted, so the reason is on the record. And a new IEP probe revealed `RECORD_SIGNATURE_RE` enumerated only **seven** evidence codes, so an entire class of record claim was invisible to the gate. It now carries the full GO vocabulary. Partially enumerating a controlled vocabulary is the same defect as partially enumerating the dimensions, one layer down — the third distinct level at which this PR has made that mistake.

### Stopping criterion, restated and unchanged

This was case (1) — a guard that did not do what it says. **I have nothing further planned.** I will push only for a blocking item, a merge conflict, or a factual claim of mine shown false. Further non-blocking hardening of the harness I will acknowledge and decline, saying so rather than silently.

**Gates on `dad9b3cc`:** `just validate` ✓ Valid; `checkquotes.py` 18/18; audit clean; full self-test passing both directions; `family_census.json` reproduces byte-for-byte; 3 `file:` quotes verbatim and single-line; `interpro/` and `cache/go/terms.csv` untouched. **All four GO verdicts, `knowledge_gaps` and `description` unchanged since round 1.**

cmungall:
Fixed in `57c8ca51`. This was a real regression I introduced, and the diagnosis was exact.

### The colon was the whole bug

I verified the failure before touching it: the COQ8B double-`GO:0004672` sentence gated `False`. Splitting on `:` severed the paralog from the record token in precisely the construction this review's prose favours — subject named, colon, claim. Dropping the colon restores both in-tree claims, and they route where they should:

```
The sharpest evidence of the cost is on COQ8B…   dims=['kinase_rows', 'negated']
…the identical row is carried by ADCK1 and ADCK2  dims=['mitocop_row']
```

**I implemented the optional forward-carry and then withdrew it.** With the colon gone, both target claims already sit in one sentence each, so it bought nothing — and it produced three false positives on sentences about ADCK5's *own* evidence codes. Keeping it would have been a broader rule that looked like coverage while adding none. Anaphora across a sentence boundary is now a **stated residual limit**, not a silently-handled case.

### Suggestion (4) was hiding a factual error, not just a design smell

You flagged that a bare `|anchored` marked `:47` covered while `subcellular()` said nothing about "inner". Chasing it showed the underlying **claim is wrong for both genes**: ADCK1's UniProt line is plain `Mitochondrion`, and COQ8A's is `Mitochondrion membrane` — neither says *inner*. So "COQ8A and ADCK1 are inner-mitochondrial-membrane anchored" was a fifth cross-gene overclaim, and the over-wide routing is what let it read as covered. Prose corrected to what UniProt records; the widening removed. Your framing is the durable lesson: **widening routing rather than the trigger is the only direction that silently shrinks what the gate reports.**

### The prose now says what the code does

`RESULTS.md` said "Any **unit** naming a paralog…" nine lines after saying per sentence — the load-bearing sentence for "closes the class" contradicting the code. Now "sentence", with **both** residual limits stated, including the larger one you identified: a record claim whose subject is an antecedent in a neighbouring sentence **is not gated at all**.

### (2) is the meta-failure and you were right to single it out

The vacuity counter incremented *outside* the sentence loop, so it counted units passing the pre-filter rather than sentences that routed — which is exactly why the gate's coverage on the review YAML shrank while the counter did not move. **A vacuity guard that cannot see its own guard regressing.** Moved inside the loop. (1) `RECORD_GATE_EXEMPTIONS` was unreachable while the file still described the escape as active — restricting `.py` surfaces to comments and docstrings removed the only case that motivated it (executable code is not prose; `for sym in ("ADCK1", "ADCK2")` was being read as a claim), so the mechanism is **deleted** rather than kept as decoration. (3) The comment claiming per-token routing was false and is corrected.

---

## Stopping criterion — I am done

**What I fixed this round:** the colon regression; the unsupported "inner membrane" claim; the prose/code contradiction; the vacuity counter's scope; the unreachable exemption; the false comment.

**Residual limits, stated plainly rather than claimed away:**

1. Routing is per sentence, not per claim — a sentence mixing a covered and an uncovered record claim passes.
2. A record claim whose subject is an antecedent in a neighbouring sentence is not gated.
3. The hedge guards allow a one-paragraph lookahead in markdown; a hedge further away still evades them.
4. `paint_node` depends on shared repo state this PR does not own.

**I am not planning further changes.** Per the criterion I stated two rounds ago, I will push only for a blocking item, a merge conflict, or a factual claim of mine shown false — and I will decline further non-blocking harness hardening explicitly rather than silently. Twelve rounds in, the curation has been stable and correct since round 1: **all four GO verdicts, both `knowledge_gaps` entries and the `description` are unchanged**, and every round since round 2 has touched analysis infrastructure only. The harness is well past where marginal guard work earns its review cost.

**Gates on `57c8ca51`:** `just validate` ✓ Valid; `checkquotes.py` 18/18; audit clean; full self-test passing both directions; `family_census.json` reproduces byte-for-byte; 3 `file:` quotes verbatim and single-line; `interpro/` and `cache/go/terms.csv` untouched.

cmungall:
Deleted in `ca69aa2b`. You were right that I reported it closed when it wasn't: I **added** the inner increment at `:569` and left the unit-level one at `:545`, so the guard still reported itself exercised whenever a unit tripped the pre-filter with zero sentences routing.

Verified each of the three guards now increments at its **own** routing granularity, rather than assuming:

```
check_parity_hedge        for i, unit in enumerate(_paragraphs(...))   <- per unit, by design
check_cross_gene_claims   for sentence in _sentences(unit)             <- per sentence
check_compartment_hedge   for i, para in enumerate(paras)              <- per paragraph
```

### Your diagnosis is the most useful thing in this thread and it's now in the notes

> the only committed vacuity break-test blanks the topic from the YAML entirely, driving both increments to zero, so it passes identically against the correct and the incorrect implementation — **a break-test whose mutation is coarser than the distinction it certifies.**

Written up in `ADCK5-notes.md` as a named failure mode. The discriminating probe is now a permanent self-test: `"COQ8A is a paralog of interest. The row is IDA."` — paralog and token both present, never in one sentence. It reports the guard **exercised** under the old implementation and reports **vacuity** under the new one, so it certifies the actual claim. Blanking a surface proved only that the check reads the surface at all.

**The mutation must be as fine as the claim.** This is the subtlest of the guard failures found here — subtler than the check that failed on perfect agreement, the unreachable branch that read as coverage, or the probe placed where it was easier to catch than the real thing — because this break-test genuinely ran and genuinely passed. Green is what made it feel tested.

**(2) taken.** `subcellular()` used `startswith("Mitochondrion")` while the prose now names the strings exactly, so a refinement of either UniProt line would have left the prose imprecise with a green run. Both exact-matched; the failure message prints what UniProt actually says.

**(1) declined, in the record:** noise rather than blindness by its own description, and this review is complete.

---

## Stopping criterion — done

**Fixed this round:** the leftover unit-level counter; the exact-match localisation assertions; the break-test lesson recorded.

**Residual limits I am declaring rather than closing:**

1. Cross-gene routing is per sentence, not per claim — a sentence mixing a covered and an uncovered record claim passes.
2. A record claim whose subject is an antecedent in a neighbouring sentence is not gated. Forward-carry was implemented, measured to add nothing once the colon was fixed, and withdrawn.
3. The hedge guards allow a one-paragraph markdown lookahead; a hedge further away evades them.
4. `paint_node` depends on shared repo state this PR does not own — a failure there means the upstream table moved.

**I am done and will not push again** except for a blocking item, a merge conflict, or a factual claim of mine shown false.

Thirteen rounds. **All four GO verdicts, both `knowledge_gaps` entries and the `description` are unchanged since round 1**, and every round from round 2 on touched analysis infrastructure only. The curation this PR exists for — that ADCK5's kinase label is a UniProt keyword artefact with nothing in GOA to act on, that it retains both UbiB inhibitory motifs and a complete catalytic core, and that it is the only human UbiB gene PAINT never reached — has been stable throughout.

**Gates on `ca69aa2b`:** `just validate` ✓ Valid; `checkquotes.py` 18/18; audit clean; full self-test passing both directions; `family_census.json` reproduces byte-for-byte; 3 `file:` quotes verbatim and single-line; `interpro/` and `cache/go/terms.csv` untouched.

Thank you — this was a genuinely sharp review series, and several of the findings (the class-closing invariant, the unreachable branch, and this one) improved the work well beyond what I would have reached alone.

PR #2314, opened 2026-07-27
Review human ADGRA2; GPCR activity survives in GOA via InterPro2GO and a GDB bulk TAS import

cmungall:
Reviews human **ADGRA2** (GPR124 / TEM5), UniProt **Q96PE1** — accession verified independently
against `projects/paint/human-no-IBA-simple.csv` line 6342 and the UniProt REST record
(`AGRA2_HUMAN`, Swiss-Prot, *Homo sapiens*, PE 1).

## Coverage reconciliation

`ADGRA2-goa.tsv` has 59 distinct data rows and live QuickGO returns `numberOfHits: 59`, so the
cached TSV is complete. The `fetch-gene` stub seeded **41** entries. The entire 18-row deficit is
the known `GOAValidator.seed_missing_annotations` collapse — WITH/FROM is not in the seeding key —
which folded the 21 `GO:0005515` rows into 3. All were restored one-per-partner. Final: **59 GOA
rows reviewed exactly once, plus 4 `NEW` proposals = 63 entries**, asserted mechanically by a
committed checker rather than by eye.

Verdicts: 19 ACCEPT · 15 KEEP_AS_NON_CORE · 21 MODIFY · 2 MARK_AS_OVER_ANNOTATED · 2 REMOVE · 4 NEW.

## Main finding: the fold-propagation error is here, but not by the route the campaign expects

The calibration in circulation is that the domain-name-becomes-an-activity error was swept out of
GOA by the ~April 2026 retirement of Swiss-Prot-keyword annotations and now survives only upstream
in UniProt. **Half confirmed, half refuted, on this one gene.**

*Confirmed for the keyword route.* Q96PE1's `DR GO;` block contains **no `IEA:UniProtKB-KW` line at
all**, while the flat file still carries `KW G protein-coupled receptor;`, `KW Receptor;` and
`KW Transducer;` — filed as a UniProt correction request in `suggested_questions`, not as a GO action.

*Refuted for GOA overall.* ADGRA2 still carries four GPCR rows, by **two routes the SPKW retirement
never touched**:

1. **InterPro2GO** — `GO:0004930` from `IPR000832|IPR001879|IPR017983|IPR036445`, `GO:0007186` from
   `IPR000832|IPR017983`.
2. **`TAS` assigned by GDB**, citing PMID:15203201, a 2004 adhesion-GPCR *repertoire survey*.

### Which signature licenses which term — InterPro2GO already draws this line

| entry | InterPro name | interpro2go targets |
|---|---|---|
| `IPR017981` | GPCR, family 2-like, **7TM** | `GO:0004888`, `GO:0007166`, `GO:0016020` |
| `IPR000832` | GPCR, family 2, secretin-like | `GO:0004930`, `GO:0007186`, `GO:0016020` |
| `IPR017983` | …secretin-like, conserved site | `GO:0004930`, `GO:0007186` |
| `IPR001879` | GPCR, family 2, extracellular **hormone receptor** domain | `GO:0004930`, `GO:0016020` |
| `IPR036445` | …hormone receptor domain **superfamily** | `GO:0004930`, `GO:0016020` |

`IPR017981` is the entry covering ADGRA2's transmembrane bundle, and InterPro describes it as *"the
transmembrane domain of family 2 GPCR receptor proteins **and Frizzled proteins**"* — Frizzleds being
the textbook 7TM family that does not signal principally through heterotrimeric G proteins.
`interpro2go` therefore maps it **only** to the generic `GO:0004888`/`GO:0007166`, never to
`GO:0004930`/`GO:0007186` — and that is exactly the pair ADGRA2 receives from it. A built-in negative
control showing the restraint is deliberate.

The GPCR-activity claim comes instead from `IPR001879`/`IPR036445`, which are the family-2
**extracellular hormone receptor** domain and its superfamily — InterPro calls it *"the major ligand
recognition domain"* and names the calcitonin, CRF1, diuretic hormone, GLP-1 and PTH receptors as
exemplars. ADGRA2 matches them on its GAIN/hormone-receptor module. **A peptide-hormone
ligand-recognition signature is being converted into "activates a G protein" for a receptor that has
no ligand at all.**

The projection test settles the second. Querying QuickGO by **reference** rather than by gene
(`?reference=PMID:15203201`, fully paginated, 78 of 78):

| | |
|---|---|
| annotations | **78** |
| distinct entities | **27**, all human |
| evidence / assigner | 100% `TAS` / 100% `GDB` |
| terms | `GO:0016020` → 27 entities, `GO:0007186` → 26, `GO:0004930` → 25 |

One reference giving 27 entities one identical generic triple is a **bulk classification import**,
not 27 traceable author statements — and the paper performs no functional assay on any receptor
("*…have been found by searches done in the human genome databases*"). Two recipients, **ADGRE4P and
ADGRF2P, are pseudogenes**, which should not carry a molecular function at all.

### The claim is unmeasured, not refuted — but the four rows are not equally defective

`GO:0004930` needs a ligand, an associated Gα, and nucleotide exchange. For ADGRA2: no ligand
(IUPHAR/GtoPdb target 198 returns no natural ligand, an **empty** interactions list and no
transducer); no coupling measurement anywhere in the systematic adhesion-GPCR coupling literature
for **any** ADGRA paralog; `"GPR124 Stachel"` and `adhesion GPCR GPR124 tethered agonist` each
return **zero** PubMed records; the single reported heterotrimeric-G-protein contact runs the other
way (*"Gβγ interacts with the C-terminal tail of GPR124 and promotes the formation of a GPR124-Elmo
complex"*, PMID:28600358, whose own abstract says *"the signaling properties of GPR124 remain poorly
defined"*); and UniProt states the characterised activity does not rely on the GPCR structure.

That is **unmeasured, not refuted**. The 7TM + GAIN-B architecture is genuine and intact, so per the
ACBD3 lesson an absence of reports is not a demonstration of absence. **The actions are therefore
calibrated to the route, not to the term** (this changed in round 2, on the reviewer's suggestion):

| route | rows | action | why |
|---|---|---|---|
| InterPro2GO | `GO:0004930` IEA, `GO:0007186` IEA | `MARK_AS_OVER_ANNOTATED` | the signature genuinely matches ADGRA2, so the inference has a real if over-extended basis |
| GDB `TAS` | `GO:0004930` TAS, `GO:0007186` TAS | **`REMOVE`** | `TAS` asserts a *traceable author statement* and the cited paper contains no such statement about ADGRA2 or any receptor — **the stated evidence does not exist** |

The `REMOVE` is on **evidentiary** grounds, not on the grounds that ADGRA2 provably lacks GPCR
activity; conflating those two claims is what made a uniform action look right at first. Because the
InterPro rows remain, removing the TAS rows does not erase the claim from GOA — and it is
complementary to, not a substitute for, the `suggested_questions` item about retiring the whole
78-annotation block, which is a consortium-level decision.

### PAINT reached the same conclusion independently

Across **all three** human ADGRA paralogs there is **not one molecular-function IBA**, and the BP
term PAINT chose is the parent `GO:0007166`, never `GO:0007186`. PAINT curators looking at this
family withheld both the GPCR molecular function and the GPCR-specific pathway; InterPro2GO and
GDB/TAS assert them anyway. **The two judgements sit side by side in the same GOA record**, and
PAINT's is the better founded one.

## Node-reach: both directions asked, and the answer is a negative

| node | annotations | entities | human reach | terms given |
|---|---|---|---|---|
| `PTN002914520` | 136 | 29 | **ADGRA2 only** | sprouting angiogenesis, CNS development, pos. reg. canonical Wnt, Wnt signalosome |
| `PTN001738137` | 348 | 174 | ADGRA1 + ADGRA2 + ADGRA3 | plasma membrane, cell surface receptor signaling pathway |

The node whose human reach is *exactly this gene set* gave it precisely ADGRA2's characterised
biology. No AADACL2/3/4- or ACTG2-style inversion: specific terms at the ortholog node, generic
terms at the heterogeneous family node. **Reported as a non-confirmation**, because on a family
whose other members are dark, "we looked and it is absent here" is a real distinction.

**A false positive that the anti-truncation assertion caught.** A single un-paginated `limit=200`
query on `PTN001738137` returned 200 of 348 rows, and human ADGRA1 and ADGRA3 were absent from that
page — which reads exactly like the ACTG2 "family node misses the human paralogs" finding. Full
pagination shows all three are reached. `node_reach.py` compares `numberOfHits` against
`len(results)`, never against a page-size constant, because QuickGO **clamps instead of erroring**.

## 21 `GO:0005515` rows → `GO:0030165 PDZ domain binding`

Every one is an assay of a single element — the C-terminal class-I motif `…LWKSETTV` — against PDZ
domains, by two independent methods (phage display, PMID:24550280; quantitative holdup,
PMID:36115835), corroborated by the one non-screen source (PMID:15021905). Resolving the partners
**from the GOA TSV**: 19 distinct, **19/19 reviewed Swiss-Prot at canonical length** (the ACRV1
TrEMBL/ORFeome check, run and negative), **18/19 with annotated PDZ domains** — the exception being
a genuine DLG1 splice isoform.

Deliberately **not** promoted to a core function: PMID:35649360 reports that mammalian GPR124
engages Frizzled through an **ICD-independent** mechanism, so the in vivo role of the human motif is
open. That is filed as a `knowledge_gap`.

## 4 `NEW` rows, all filling real gaps

- **`GO:0030159` signaling receptor complex adaptor activity** — ADGRA2 has *no* informative MF
  annotation. This is what the Wnt7 literature actually establishes: RECK binds Wnt7, ADGRA2
  provides the physical support that assembles the RECK-ADGRA2-FZD-LRP5/6 signalosome. Precedent is
  exact — **CD19**, a co-receptor, holds this term by EXP.
- **`GO:0005178` integrin binding** and **`GO:0005539` glycosaminoglycan binding** — human
  experimental facts (PMID:16982628) that UniProt records with `ECO:0000269` and GO lacks entirely.
  `GO:0005539` rather than `GO:0008201`, because all four GAG classes were bound.
- **`GO:0010595`** — the mouse donor holds it by IDA while only its unregulated parent transferred.

## Other checks, reported including the nulls

- **Errata/retractions**: one Author Correction, PMID:36477203 on PMID:36115835, found by reading
  `CommentsCorrections/RefType` on the article's own record. No retractions.
- **ISS/Compara donors**: mouse Adgra2 (Q91ZV8) holds its **own experimental annotation** for every
  transferred term, so `SOURCE_WEAK_OR_INFERRED` would be contradicted by the analysis. Accepted.
- **UniProt/GOA inconsistency**: `GO:0009986 cell surface` is in UniProt's `DR GO` block as
  `IEA:Ensembl` but absent from GOA. Reported, not acted on.
- **Species caveat**: UniProt attributes the Dishevelled-recruitment step to human on PubMed:30026314,
  whose central genetics is zebrafish. Filed as a UniProt correction request and a knowledge gap;
  no GO row disturbed.

## Two defects of my own, both found by scripting a comparison I had already made by eye

**A domain misassignment that had reached the prose.** I hand-wrote the five InterPro `source_label`
values and **three of five were wrong**, two substantively: `IPR001879` was labelled "GPCR
proteolysis site (GPS) domain" and `IPR036445` "GAIN domain superfamily". Both are the
hormone-receptor domain and its superfamily. The wrong names had already propagated into the
`GO:0004930` reason. This is the ACBD3 GOLD-versus-Q error exactly. Correcting it produced the
sharper argument in the section above — the fix made the case stronger, not weaker. Labels are now
computed by a committed guard (`interpro_signatures.py`) that also asserts every signature present in
the WITH/FROM is labelled (presence, not match-on-match) and that every GO term a row claims is one
`interpro2go` actually licenses.

**An off-by-one in a hand count.** I wrote that ADGRA1's PDZ panel has 20 partners; computed from
QuickGO it is 21. The paralog table is now derived rather than counted by eye.

### And one in the corrections checker, which reported success while broken

`check_corrections.py` first read the DOI with `.//ArticleId`, which also matches ArticleIds inside
`ReferenceList` — so three of fifteen PMIDs were Crossref-checked against a DOI belonging to a paper
they merely **cited** (PMID:28600358 as a JoVE DOI; PMID:35649360 and PMID:27979830 both as the same
2014 Dev Cell DOI). All three reported `crossref=none`: a **false clean**. It surfaced only because
two different PMIDs printed the same DOI. Fixed by anchoring to `PubmedData/ArticleIdList` and
asserting the DOI's Crossref title matches the PubMed title. `--self-test` now break-tests it in
five directions, including the happy path and the reference-scoped variant of the same bug in
`CommentsCorrections`.

## One reviewer point refuted rather than conceded

Round 1 flagged `directly_involved_in` as listing `GO:0060070` together with its child `GO:0090263`.
It is not a child. QuickGO `/ontology/go/terms/GO:0090263/ancestors?relations=is_a,part_of` returns
18 ancestors and **`GO:0060070` is not among them** (`GO:0060828` is). From the other end,
`GO:0060070`'s child listing gives the relation explicitly: `GO:0090263 positively_regulates`,
`GO:0090090 negatively_regulates`, `GO:0060828 regulates`, and only
`GO:0061316`/`GO:0044336`/`GO:0044337`/`GO:0044338` as `is_a`. GO keeps regulation terms out of the
subsumption hierarchy of the process they regulate, so neither entails the other and both are
separately annotated in GOA (IDA and IBA). Same shape as the ABR Rac/Rho lesson and the ACTMAP
proteolysis/modification lesson: a `positively_regulates` edge in a child listing reads like
subsumption and is not. Both terms retained.

I also declined half of one item: the suggestion to note that the `GO:0030159` imaging is "in human
cells". PMID:30026314 and PMID:28289266 are both abstract-only in the cache and **neither abstract
names the cell type or species**, so the row records that as unverified rather than asserting it.
The species caveat that *does* apply is about which domain mediates the interaction, not whether
ADGRA2 performs the adaptor role — PMID:35649360 relocates the mechanism, it does not withdraw the
function.

## Round 3: stopped asserting what an abstract-only paper's full text contains

The round-2 reviewer caught a real defect of the same class this review criticises. The `REMOVE`
rationale had said the cited paper *"contains no such statement about ADGRA2 or any other receptor"*
— but PMID:15203201 is `full_text_available: false`, so that is a full-text claim made from an
abstract. Rewritten to rest only on what is claimable: the **abstract** describes genome-database
searching, phylogenetic analysis and EST expression charting with no functional assay reported on any
receptor; and **the projection test needs no full text at all**, because a paper cannot make 25
gene-specific author statements about G-protein coupling — two of them about pseudogenes — as a
by-product of cataloguing a repertoire. Verdict unchanged, grounds now inside what the cache supports.

Also closed the loop on the asymmetry the reviewer flagged: the `GO:0016020` TAS row from the **same**
GDB block carries the identical evidentiary defect and would be a legitimate `REMOVE` on the same
reasoning. It is kept because removing it achieves nothing — membrane localisation is directly
demonstrated for ADGRA2 elsewhere in this same record (three `EXP` rows, an `IDA`, and an `IBA` to the
child `GO:0005886`), so it is redundant rather than misleading. The rule is now explicit: **remove
where the defective row is carrying the claim, flag where it merely duplicates a directly
demonstrated one.** And dropped "in a different system" for PMID:28289266, which contradicted a
sentence two lines below it.

## Round 4: the softening had landed in 4 of 7 places

The round-3 reviewer caught the campaign's most-repeated failure, and I had committed it: my
softening changed **one instance per location**, so both `REMOVE` reasons still asserted *"the paper
states no author claim…"* earlier in the same string that later disclaimed full-text knowledge, plus
three more sites and the notes table cell.

**Root cause worth recording**: my detector ran over the **builder source** while the claim lives in
the **emitted YAML**, where the same sentence is split across string-literal line breaks — so it
counted 3 where there were 4, and a second hand-fix would have been incomplete again. Re-run with
detector and mutator on the same scope: **7 detected, 7 removed, 0 remain, asserted.**

Two more accepted:

- **The pseudogene sub-claim was correct but unevidenced.** "Two of the recipients are pseudogenes"
  does not follow from a 25-of-27 distribution — that is equally consistent with ADGRE4P/ADGRF2P
  being the two that *missed* `GO:0004930`. The per-entity matrix settles it: the two that miss it
  are **ADGRG3 and ADGRV1**, both protein-coding. Now checked by name.
- **The projection test now has a committed script and JSON** (`projection_test.py`). It was the one
  load-bearing analysis here still run ad hoc, which mattered more once round 3 made it the *sole*
  support for both `REMOVE`s. Writing it exposed a defect in its own self-test: the first fixture
  omitted a pseudogene from `entities`, so the wrong branch fired — **a self-test can pass for the
  wrong reason.**

And the `GO:0016020` review now leads with the defect it **shares** with the two removed rows, says
that on evidentiary grounds alone it would also be a legitimate `REMOVE`, and only then gives the
redundancy reason for keeping it.

No action, term, or quote changed in this round.

## Gates

- `checkquotes.py`: **65 quotes, 0 problems**
- `just validate human ADGRA2`: **✓ Valid**, zero warnings
- Branch **merged into current main**: merges cleanly, validates ✓
- `check_coverage.py`: 59 = 59, no duplicate YAML keys, no anchors, nothing PENDING; `--self-test`
  passes in 6 directions
- `cache/go/terms.csv`: validate mutated it (20 ins / 21 del, collapsing main's two pre-existing
  duplicate curies). Restored from HEAD and **one** row appended (`GO:0090210`). Union asserted four
  ways after the merge — no main row dropped, no branch row dropped, nothing invented, no new
  duplicate; `GO:0001675` and `GO:0009566` remain duplicated exactly as on main.

---

> [!NOTE]
> **Low Risk**
> Curated annotation YAML, local validation scripts, and a single GO terms.csv row—no runtime application or auth/data-path changes.
> 
> **Overview**
> Adds a **complete human ADGRA2 (Q96PE1)** Gene Ontology review: every GOA row is judged (accept, modify, remove, new proposals, etc.), with emphasis on **unsupported GPCR activity/pathway** claims from InterPro2GO and a GDB bulk **TAS** block on PMID:15203201, versus the characterised **Wnt7 co-receptor / BBB angiogenesis** biology.
> 
> The review **collapses 21 generic protein-binding rows** into **PDZ domain binding**, proposes **new terms** (e.g. signaling receptor complex adaptor, integrin/GAG binding), and documents **knowledge gaps** and consortium **suggested_questions**.
> 
> Under `ADGRA2-bioinformatics/`, the PR adds **reproducible checks** (GOA coverage, InterPro labels vs `interpro2go`, QuickGO PANTHER node reach, GDB projection test, PMID corrections, action/prose consistency) plus JSON outputs and `RESULTS.md`. **`cache/go/terms.csv`** gains one term row: `GO:0090210` (regulation of establishment of blood-brain barrier).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f000a0390f57f0157f78fcbb45135767e9c59fca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
All four addressed in `396ff3597`, pushed once. No action, term, quote or verdict changed.

**(1) — and the root cause was structural, not clerical.** You are right that this is the same class as round 3, and it is the fourth surface-sweep miss here, so I went after the mechanism rather than the sentence. `GPCR_REASON` was a **single shared prose constant appended to rows with different actions**, and it ended with an action-specific sentence. When round 1 recalibrated the TAS rows to `REMOVE`, that sentence could not be edited in place without also breaking the two InterPro2GO rows where it is correct — so it survived. Split into `GPCR_REASON` (action-neutral evidence, shared) plus `OVERANN_TAIL` (the verdict sentence, attached only to the two `MARK_AS_OVER_ANNOTATED` rows). Verified out of the tree: the sentence now appears on exactly **2** rows, both with that action.

**(2)** Deleted. Round 4 added the abstract clause and the "a paper cannot make 25 gene-specific author statements" argument to `GDB_PROJECTION` without removing the round-3 versions inside `TAS_REMOVE`. Superseded sentences gone, one connective in their place, reasoning unchanged and much shorter.

**(3)** Softened to "cannot supply a gene-specific author statement" — the same abstract-only overreach you caught in the reason last round, which I had fixed in one field and not the other. Same shape as (1).

**(4)** `RESULTS.md` now records what each analysis **found**, including the negative results, kept qualitative so the figures stay in the JSON and there is no second surface to drift.

### New guard: `check_action_prose.py`

A fourth recurrence deserves an invariant, not a fifth careful re-read. It implements exactly the rule these misses kept violating — **select on the stable entity**: it iterates `review.action` and tests that block's prose, so a new row inherits the check automatically and wording drift cannot skip it. It also rejects a reason that argues the same point twice, which is how (2) happened.

Before the fix it reproduced your finding exactly — 6 problems on the 2 `REMOVE` rows, nothing elsewhere. `--self-test` exercises 5 directions, two of which are the ones worth naming:

- an **attributed** mention of another row's action (*"The InterPro rows … are marked over-annotated"*) must **not** fire, or the guard would forbid legitimate cross-referencing and get worked around rather than obeyed;
- a block with **no** `action`, or an **empty** annotation list, must fail loudly rather than pass vacuously.

### On the projection test

Thanks for checking the direction of the pseudogene result rather than just its presence — that was the sub-claim I had asserted for two rounds without evidence, and it could as easily have come back the other way and forced a withdrawal. The self-test fixture keeps both pseudogenes in `entities` deliberately: the first version omitted one, so the not-annotated-at-all branch fired instead of the withdrawal branch and the test passed for the wrong reason.

Gates on the merged tree: `checkquotes` 65/0 · `just validate` ✓ Valid, no warnings · coverage 59 = 59 · `check_action_prose` clean · all five self-tests pass · `terms.csv` no deletions, one row added (`GO:0090210`), no new duplicate curies.

cmungall:
Conflicts resolved in `5a7e73d56`, notes recorded in `1ec0bbb01`. Both premises verified independently before I touched anything.

**1. `PTHR45930-metadata.yaml` — took main's.** Confirmed for myself that the timestamp really was the only difference before blind-taking: both files are **54 lines**, and **exactly one** line differs, `_fetch_info.fetched_date` (`19:50:35` vs `19:51:02`). Asserted byte-identity to main afterwards — the file now drops out of this PR's diff entirely.

**2. `PMID_24550280.md` — took mine.** Superset asserted, not eyeballed, and re-asserted *after* resolution:

```
non-blank lines in main absent from mine: ['full_text_available: false']
```

Exactly the sole permitted exception, so the two copies do not genuinely diverge. Kept the 101-line full-text copy over main's 80-line abstract-only one. Agreed this is the one that matters — PMID:24550280 is one of the two PDZ-motif papers underpinning the `GO:0030165` call.

I generalised the check rather than stopping at the flagged file: **every** publication this PR touches that also exists on main was compared for a full-text downgrade. One file qualifies (this one), and it is an upgrade, not a downgrade.

**Re-verification after the merge**, since a later merge can undo an earlier resolution:

- all **18** gene files byte-identical to their pre-merge hashes;
- `cache/go/terms.csv` compared as a **multiset** — `GO:0001675` and `GO:0009566` each confirmed still present **twice** (a set comparison would have passed straight through a collapse); only `GO:0090210` added; no new duplicate curies; 9078 → 9079 rows;
- `just validate human ADGRA2` ✓ Valid on the merged tree, zero warnings;
- `git status --porcelain -- cache/go/terms.csv` clean afterwards — validate left no side-effect;
- all five self-tests and all four analysis scripts still pass.

**One process note worth having.** GitHub reported `CONFLICTING`/`DIRTY` after the push and then recomputed to `MERGEABLE`, as expected. But my first local probe *appeared* to confirm a conflict — and it was wrong: `git merge-tree --write-tree` does not exist in git 2.37 (added in 2.38), so the command failed to **parse** and its non-zero exit read as "conflicts". A rejected query and a real negative are indistinguishable downstream unless you check the exit status, which is the same trap as the MOD-id HTTP 400s in the brief. The authoritative probe is `git rev-list --count HEAD..origin/main` → **0**.

Head is now `1ec0bbb01`, past the round-5 approval as expected. Not merged, auto-merge not enabled.

cmungall:
Fixed in `5890d538b`, pushed once. You are right on all counts, and the blocker was a real false claim, not a wording issue.

## The blocker

Four YAML sites plus the notes bullet asserted no coupling assay exists for ADGRA1/A2/A3 or the clade. **False twice over**, and I have scoped all four to ADGRA2 and cited the sources:

- **ADGRA3** — *"We found low-level activation of Gi and Gs by ADGRA3 and slightly more by its CTF."*, and removing the stachel's first three residues gave *"abrogated G protein-mediated signaling"* [PMID:40127866].
- **ADGRA1** — *"ADGRA1 activates several G proteins, notably Gα13"*, full TRUPATH BRET2 panel [PMID:41961591], now fetched and cached.

Verified out of the tree: **0** residual clade claims in the YAML; the notes' single occurrence is the retraction notice itself, immediately followed by "That is false". I break-tested that exemption — flipping the verdict line makes the matcher fire again, so it is not a blanket pass.

**Why the sweep missed it.** Exactly as you diagnose. My query required `G protein coupling` / `coupling profile` / `constitutive activity` / `transducer` in `[tiab]`; the ADGRA3 abstract says *"G protein-mediated signaling"* and *"Gi and Gs"* and contains **none of the four**. Too narrow, not stale — no amount of re-running finds it. A keyword conjunction over abstracts is a filter whose false-negative rate is invisible to whoever runs it.

**And the distinction I want to keep.** The notes phrased it defensibly, as a corpus-level negative about the *literature*. The YAML promoted "my query returned nothing" into "no such study exists". Only the first was ever mine to assert — the same epistemic upgrade as quoting an abstract-only paper about its full text, in a new costume, and smuggled in while restating something true.

**The correction improves the review, and I have not dressed that up as a save.** ADGRA2 still has no coupling assay, and now has a *published statement* to that effect — *"Whereas ADGRA2 has never been shown to couple to heterotrimeric G proteins…"* [PMID:40127866] — which is strictly better evidence than an absence of hits. And a clade where two of three paralogs demonstrably couple makes `MARK_AS_OVER_ANNOTATED` **better** founded than `REMOVE`: a fold-based inference is not absurd for an ADGRA protein, merely unverified for this one. That reasoning is now in `OVERANN_TAIL`, on the two InterPro rows only. The two `REMOVE`s rest on the projection test and are untouched. No action or term changed.

Worth noting how it surfaced: **only because ADGRA3 merged and brought the cached publication with it.** Parallel paralog review earning its keep.

## (1) Divergence with ADGRA3

Stated on both `REMOVE` rows and in the notes. ADGRA3 `ACCEPT`s the identical `PMID:15203201` pair as *"right conclusion, weak provenance — kept because the conclusion has since been earned"*; this review `REMOVE`s it. **One rule covers both**: a defective TAS row is kept where its claim is independently established *for that gene*, removed where nothing else on the gene carries it. ADGRA3's was earned by a direct Gi/Gs measurement; ADGRA2's was not, and the same paper says so. The rule is visible *inside* this review too, which is the check that it is a rule and not a rationalisation — my `GO:0016020` TAS row survives on exactly ADGRA3's logic.

## (2), (3), (4)

- The `GO:0007186` TAS row now argues **its own** definitional requirement — ligand binding, then GDP/GTP exchange, then Gα dissociation — and notes ADGRA2 fails at the first step too, rather than borrowing `GO:0004930`'s.
- `check_action_prose.py`: you were right that `[^.]{0,200}?` was zero-length-matchable, so the exemption had collapsed to "an attribution appeared somewhere in the preceding 220 characters" irrespective of sentence boundaries. Now scoped to the sentence containing the phrase, via the `sentences()` helper that was already sitting there unused.
- `NO_DUPLICATE_SNIPPETS` comment corrected — they are distinctive fragments, not whole sentences. While there I also found a dangling reference to a never-existent `ATTRIBUTED_PREFIXES`, and added an assertion that every referenced name actually resolves in the module.

Gates: `checkquotes` 72/0 · `just validate` ✓ Valid, zero warnings · coverage 59 = 59 · `check_action_prose` clean · all five self-tests pass · retraction check clean on both new PMIDs · `terms.csv` multiset intact (`GO:0001675` and `GO:0009566` each still twice, one row added, no new duplicates).

PR #2318, opened 2026-07-27
Review human ADAM5 (Q6NVV9): pseudogene confirmed, GOA already correct, and the negative control that locates the real catalytic over-annotation

cmungall:
Review of human **ADAM5** (`Q6NVV9`, HGNC:212, 8p11.22).

## The premise, verified before reasoning from it

ADAM5 really is a pseudogene, from four independent authorities, none of which is the gene's family name:

- **HGNC:212** — `locus_type: pseudogene`, approved name *"ADAM metallopeptidase domain 5 (pseudogene)"*, RefSeq `NR_001448` (an `NR_` non-coding accession, not `NM_`).
- **NCBI Gene 255926** — same description.
- **UniProt Q6NVV9** — `PE 5: Uncertain`, `CAUTION: Could be the product of a pseudogene. Not expected to have protease activity`, and all **eleven** `DR EMBL` cross-references are `NOT_ANNOTATED_CDS`.
- **PMID:41263505** (Biol Reprod, 2026) — the mouse-knockout authors state it outright: *"While ADAM5 is a pseudogene in humans…"*.

The **mirror error was tested too**. The locus is genuinely transcribed (multiple spliced testis transcripts, two MGC cDNA clones, 16 exons), so this is a *transcribed unitary* pseudogene, not a relic. But the protein is absent: the primary paper reports no immunoreactivity on Western blots of human testis and sperm with a macaque anti-tMDC II antiserum, and the EBI Proteins API returns **HTTP 404** for mapped MS peptides on Q6NVV9 — controlled against ADAM2 (53 peptides), ADAM18 (13), mouse Adam5 (6) and ACTB (393), all HTTP 200. UniProt's `DR PeptideAtlas` / `DR TopDownProteomics` lines look like protein evidence but are database-membership cross-references.

## Verdicts — 3/3 rows reviewed, all ACCEPT

GOA carries **exactly three** annotations, all `ND` root terms against `GO_REF:0000015`. GOA TSV has 4 lines = 3 data rows; the `fetch-gene` stub seeded 3 `- term:` entries — **they reconcile exactly**, no collapsed rows. QuickGO (fully paginated, `numberOfHits == len(results)` asserted) agrees.

| term | evidence | action |
|---|---|---|
| `GO:0003674` molecular_function | ND | ACCEPT |
| `GO:0005575` cellular_component | ND | ACCEPT |
| `GO:0008150` biological_process | ND | ACCEPT |

`ND` records that a curator looked and found no data, which is exactly true here. There is nothing to modify or remove.

**No `core_functions` are asserted.** The "No core functions defined" warning is genuine for a dark gene and is deliberately left standing rather than silenced with invented content. The substantive findings are carried by three `knowledge_gaps` entries.

## The predicted catalytic over-annotation is NOT present — a non-confirmation

The brief flagged two live routes by which a pseudogene acquires a bogus molecular function. **Neither is present**, and that is reported as a non-confirmation rather than stretched into a finding:

- **Bulk TAS import** (the ADGRA2/GDB `PMID:15203201` pattern) — no TAS row of any kind on Q6NVV9, so there is nothing to run the reference-projection test against.
- **InterPro2GO fold→activity** — no `IEA:InterPro` row either. `DR PAN-GO; Q6NVV9; 0 GO annotations based on evolutionary models`, so no IBA (verified against QuickGO, not read off the worklist's name).

## Where the defect actually lives: a negative control that identifies the rule

The productive question was not "did the error hit ADAM5?" but "which relatives did it hit, and what boundary separates them?" A detector was built for **M12B fold present + no `HExxH` zinc core + `GO:0004222` nevertheless annotated**:

- **Fires** on all four ADAM5 orthologues (macaque, mouse, rat, guinea pig) and on human **ADAM2**, **ADAM18**, **ADAM7** — each with an annotated Peptidase M12B domain, **no `HExxH` substring at all**, and 2 × `GO:0004222` by `IBA` + `IEA:InterPro`.
- **Clean** on ADAM10/17/9/19 (zinc motif intact — the term is correct there) and on human ADAM5 (no fold, no term).

Mouse Adam5 is the sharpest case: the *same UniProt entry* says `CAUTION: Not expected to have protease activity` while GOA gives it `GO:0004222` from `IPR001590` and from `PANTHER:PTN000224844`.

**The boundary is the finding: the pipelines key on presence of the M12B fold, not on integrity of the zinc site.** Human ADAM5 escapes only because pseudogenisation deleted the fold the pipeline matches on — not because anything recognised it as non-catalytic. Raised in `suggested_questions` (none of those genes is in scope here), with the `GO:0004222` **definition** checked rather than its label: it does assert metal-ion-coordinated catalysis, so a protein with no zinc-binding residues cannot satisfy it.

## Structural basis (committed, reproducible)

`genes/human/ADAM5/ADAM5-bioinformatics/` — `uv run python analyze_adam5.py`, `--self-test` break-tests the guards. Every value is re-fetched from UniProt and QuickGO per run; nothing hardcoded.

Aligned to macaque ADAM5 (756 aa, `Q28483`, the species in which tMDC II was characterised), the 412-aa human putative translation is **not a diverged paralogue** — retained blocks are 96.2% and 83.6% identical. It is the same gene with a **single 246-residue internal deletion** (macaque 196–441, fused after human residue 162):

| macaque feature | fraction covered by human |
|---|---|
| Signal 1–16 | **0.000** |
| Transmembrane 699–719 | **0.000** |
| Peptidase M12B 183–380 | **0.066** |
| Disintegrin 389–478 | 0.411 |
| Propeptide 17–142 | 0.786 |
| EGF-like 630–664 | 1.000 |

So the product lacks secretory targeting *and* a membrane anchor — it could not reach the sperm surface even if translated, which is why the **cellular-component** `ND` is as well founded as the molecular-function one.

**UniProt correction filed in `suggested_questions`:** the human entry's `DOMAIN 111..199 Disintegrin` **straddles the deletion junction at 162/163**. Residues 111–162 align to macaque 144–195 (upstream of the macaque disintegrin domain) and only 163–199 to real disintegrin sequence. It is a fusion created by the deletion, not a conserved domain.

## Provider recall

`affinage_deep_research.py human ADAM5 --write` returned **`no Affinage record content for ADAM5`** — an empty record, so there is no `gates_passed` value to report. Treated as *no information*, not as evidence of absent literature: the UniProt `RN` list gave PMID:10417343, and independent PubMed searches on the **orthologue and the family** (not the human symbol) surfaced the decisive 2026 knockout PMID:41263505 and the CNV/ceRNA paper PMID:36553675. Recorded in the notes so provider recall stays measurable.

## A committed claim audit, tested against the defect that actually shipped

`ADAM5-bioinformatics/audit_adam5_claims.py`. I had described a set of consistency checks in prose while running them in a scratch buffer, which is the "a verification you performed is not a verification that exists" trap — so they are now an artifact next to the analysis they guard.

- **Numbers:** every figure the review restates (246-residue deletion, 0.066 M12B coverage, the 7-gene detector fire set, the three `ND` rows) is re-derived from `results.json`, not eyeballed.
- **A retracted claim:** the refuted single-cluster wording is kept dead across the review YAML, the notes and `RESULTS.md`, with a **required-claims half** so that deleting the correction cannot satisfy the retracted-phrase half trivially.
- **`--against-shipped-defect 4d92ca329`** runs the check against the exact blob that carried the bug: it fires on both phrasings there and is clean on the current tree. Stronger than a self-test, which only proves the guards I thought of fire.
- It **declares its own limitation** in the docstring: fixed-phrase matching cannot catch a paraphrase.

Writing it caught something immediately — the guard fired on my own notes bullet, because I had quoted the retracted phrase in order to document it. I reworded the bullet rather than teaching the guard to permit a "citation context", which would have opened exactly the bypass it exists to close.

## Checks run, including the ones that came back negative

- **Retraction / erratum / expression-of-concern** on all seven cited PMIDs via `CommentsCorrections/RefType` — none. The three load-bearing DOIs additionally resolved at **Crossref** and `update-to` / `updated-by` / `relation` inspected, to catch a correction with a null PubMed id — all empty.
- **A factual error in my own first draft, caught by computing what I had written.** I described the six human ADAM pseudogenes as one "8p11.22 reproductive ADAM cluster". An HGNC location query refuted it: only **ADAM3A and ADAM5** are at 8p11.22; **ADAM1A/ADAM1B** are at 12q24, **ADAM3B** at 16q12.1 and **ADAM6** at 14q32.33. They are independent pseudogenisation events across four chromosomes, not one deletion. Fixed at all three surfaces and re-grepped on the stable entities (gene symbols, the locus string) rather than the reworded conclusion.
- **A bug in my own checking script, caught by a value that refused to add up.** An unanchored `.//ArticleId` XPath returned DOI `10.1002/pros.23950` (*The Prostate*) for PMID:36553675, which is published in *Genes*. It had matched a DOI inside the article's own **reference list**. Anchored to `PubmedData/ArticleIdList/ArticleId` → correct `10.3390/genes13122408`.
- **Two guard formulations refuted by their own measurement.** The domain-loss check first asserted "M12B is the least-covered feature" (false — Signal and Transmembrane are 0.000), then "coverage is bimodal, place a threshold in the largest gap" (also false — the disintegrin is genuinely intermediate at 0.411 because the deletion cuts through it, so any threshold would have been invented). What shipped is **threshold-free**: every feature called lost is less covered than every feature called retained, margin 0.345, plus an assertion that no reference feature is left unclassified.
- **Self-test proven able to fail** — disabling the motif-discrimination guard in a copy makes `--self-test` report exactly the two expected failures.
- **Report reproducibility** — a fresh run reproduces the committed `RESULTS.md` and `results.json` byte-identically (`diff`).
- **Strict YAML load** with a duplicate-key-rejecting loader; raw-vs-parsed `reference_id` counts reconcile 12/12; zero anchors/aliases.
- **All 26 quotes verified verbatim** — 22 by `checkquotes.py` plus the 4 `knowledge_gaps` `provenance` quotes it does not walk, checked separately.
- **`cache/go/terms.csv` untouched** (multiset `Counter` comparison, both directions; merge-base diff clean; the two pre-existing duplicates `GO:0001675` / `GO:0009566` unchanged).
- **Pre-write hook** reported 6 `file:` reference errors; all were the known wrong-root artefact of working in a sibling worktree, verified with `grep -F` against the real files before being disregarded. No quotes were deleted to satisfy it.

Validated `✓ Valid` both on the branch and on the branch merged into `main`.

---

> [!NOTE]
> **Low Risk**
> Documentation and offline analysis scripts only; no runtime services, auth, or GO database writes. Risk is limited to review accuracy and external API dependence when re-running analysis.
> 
> **Overview**
> Adds a **complete human ADAM5** gene review package: `ADAM5-ai-review.yaml` documents the locus as a **transcribed unitary pseudogene** (`Q6NVV9`) and **accepts all three GOA root `ND` annotations** without proposing new GO terms, while recording **knowledge gaps**, literature references, and **policy questions** (pseudogene representation, InterPro/PANTHER `GO:0004222` on non-catalytic M12B folds, UniProt disintegrin feature, mouse knockout curation).
> 
> The PR also introduces **`ADAM5-bioinformatics/`**: `analyze_adam5.py` (UniProt/QuickGO fetches, macaque alignment, zinc-motif scan, fold-without-zinc detector, PTHR11905 family census) writes **`RESULTS.md`** / **`results.json`**; **`audit_adam5_claims.py`** ties prose numbers to `results.json` and blocks a **retracted “single 8p11.22 cluster”** phrasing. A small **`pyproject.toml`** pins BioPython for `uv run`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a5dab852a322b27cfd33720297464ce83cc7c9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — all three suggestions were correct, so none is refuted. All three are now addressed in 1a5dab852.

**1. Signature ids are now pinned by the artifact.** You were right that `IPR001590` and `PANTHER:PTN000224844` were the only figures in the package a re-run could not reproduce — grepping `results.json` for either did return nothing. `quickgo_count` now records `with_from`, and `audit_adam5_claims.py` asserts both ids as `GO:0004222` sources on mouse Adam5 *and* on ADAM2.

**2. Database vocabulary out of `description`.** Adopted verbatim: "reviewed entries exist for mouse, rat and guinea pig" → "the gene is conserved in mouse, rat and guinea pig".

**3. The rule is now measured, and it came back stronger than the panel suggested.** New section G runs the cross-tabulation over **all 331 Swiss-Prot reviewed members of PTHR11905**:

| reviewed members | n | carry `GO:0004222` | % |
|---|---|---|---|
| M12B fold **with** intact `HExxH` site | 204 | 204 | 100% |
| M12B fold **without** any `HExxH` site | 40 | 37 | **92%** |
| no M12B fold | 87 | — | — |

If the annotation discriminated on the catalytic site, that second row would be near zero. At 92% it is essentially indistinguishable from the intact members — so losing the zinc-binding site has almost no bearing on whether a reviewed family member receives metalloendopeptidase activity. That turns the `suggested_questions` item for InterPro/GO Central from a strongly-suggested rule into a measured one.

Three disciplines applied to that number, since a family-wide count is easy to get wrong:

- **Scope is stated, not implied.** The family total (**29,886** proteins) is fetched and printed next to every count. This is the **1.1% reviewed subset**, and it is labelled as such everywhere rather than written as a claim about the family.
- **The two measurements are reconciled, not assumed to agree.** Section G's metric is *exact* `GO:0004222` in the UniProt GO cross-references; section D's is a descendant-aware QuickGO count. Different measurements, so the family run is *required to reproduce the hand panel's verdict* on all 10 panel members present in the reviewed set, and fails if any disagrees. It reproduces 10/10.
- **An inconvenient finding reported rather than dropped.** ADAM10 and ADAM17 are **not in PTHR11905** — they are classified in another PANTHER family — so the family-wide run cannot corroborate those two controls. Recorded in `RESULTS.md` and in `results.json` as `panel_members_absent_from_family` rather than quietly skipped.

`audit_adam5_claims.py` grows 36 → 53 assertions covering the new figures, with self-tests that fire on a perturbed family count and on removal of the signature ids; `--against-shipped-defect 4d92ca329` still fires. `RESULTS.md`/`results.json` still reproduce byte-identically on a fresh run, and `just validate human ADAM5` is `✓ Valid` with only the intended "no core functions" warning.

No merge from me, per instructions.

cmungall:
Thanks — approval noted on `1a5dab852`. One correction to the bonus observation, then a scoping decision.

**Correction: `Q13443` is ADAM9, not ADAM17.** The comment reads "`UniProtKB:Q13443` (ADAM17) and `UniProtKB:Q9H013` (ADAM19)". Checked against UniProt:

| accession | entry | gene |
|---|---|---|
| `Q13443` | `ADAM9_HUMAN` | **ADAM9** |
| `Q9H013` | `ADA19_HUMAN` | ADAM19 |
| `P78536` | `ADA17_HUMAN` | ADAM17 |

ADAM17 is `P78536`, and `P78536` is one of the two accessions the family run reports as **absent** from PTHR11905 (`panel_members_absent_from_family: ["O14672", "P78536"]`), so it could not be an IBA seed on this node. The seeds are ADAM9 and ADAM19 — both in the family, both with intact zinc motifs (`HELGHNLGMNHD` and `HEMGHNFGMTHD` in section C).

**The substantive point survives the correction intact, and is arguably cleaner for it.** Both named seeds are catalytically intact ADAMs *inside* PTHR11905, so the PANTHER propagation is measurably from two zinc-site-bearing members onto members with no zinc site at all. That is a sharper statement of the defect than "IBA annotates it", exactly as you say.

**Scoping decision: not pushing again for this.** The underlying data is already committed — `with_from` is in `results.json` for every census row and pinned by four assertions in `audit_adam5_claims.py` — so the observation is reproducible from the artifact by anyone who takes up the InterPro / GO Central question. What is missing is only prose framing in `suggested_questions`, and adding it would invalidate an approval that now covers the head commit and start another review round for a 🔵 item you flagged as optional. Recording the corrected version here instead, so the accession error does not propagate from this thread into whatever picks the question up.

Per instructions I am not merging and have not enabled auto-merge.

PR #2331, opened 2026-07-27
Review human ADNP: the NAP peptide's pharmacology in the ADNP protein's GO record

cmungall:
## Summary

Full review of human **ADNP** (Q9H2P0). All **53** distinct GOA rows adjudicated, plus 2 `NEW` proposals.

**Row reconciliation, stated up front.** The GOA TSV has 53 data lines, all distinct. The `fetch-gene` stub produced **47** entries. The whole gap is `GO:0005515`: 14 IPI rows collapsed to 8, one per PMID, dropping the WITH/FROM partner. Restored to one row per (reference, partner). No distinct GO term was missing.

**The worklist's "no-IBA" name is wrong here too** — QuickGO returns **2 IBA rows** for Q9H2P0 (`GO:0005634`, `GO:0010468`). Both adjudicated; `GO:0005634`'s WITH/FROM cites `UniProtKB:Q9H2P0` itself, the valid self-referential PAINT pattern.

## Headline finding: an eight-residue peptide's pharmacology became the protein's GO record

26 of the 53 rows are Ensembl Compara projections (`GO_REF:0000107`) — 19 from rat `Adnp` (Q9JKL8), 7 from mouse `Adnp` (Q9Z103). Resolving each donor's own primary reference shows **13 of the rat-derived rows trace to experiments on the synthetic ADNP-derived octapeptide NAPVSIPQ ("NAP" / davunetide), applied exogenously, with the ADNP protein never perturbed or measured.**

Three cases are decisive rather than arguable:

- **`GO:0050805` negative regulation of synaptic transmission** — the donor (`PMID:15963648`) used **D-NAPVSIPQ**, the all-D-amino-acid enantiomer, which no gene can encode. Its own control: *"Application of D-NAP alone had no effect on LTP induction or expression."* The suppression required co-applied 50 mM ethanol. Donor evidence code is IMP, which normally implies a gene perturbation; there was none. → **REMOVE**
- **`GO:0046068` cGMP metabolic process / `GO:0080164` regulation of NO metabolic process** — from `PMID:11438390`, a paper **titled for vasoactive intestinal peptide** that doses cortical cultures with VIP/SNV/NAP. ADNP is never manipulated. → **REMOVE**
- **`GO:0042277` peptide binding** — the only binding reported is *biotin-NAP binding Abeta*, so the entity that binds is the synthetic eight-residue peptide. REMOVE rests on **entity identity alone**. *(Corrected after review: an earlier version of this row argued the direction was inverted and that RGD's own records made that explicit. Both halves were wrong — Abeta is a peptide, so for the fragment the direction is correct; and the Tubb3/Tubb4b rows carrying `WITH/FROM RGD:71030` come from PMID:16893427, the donor for `GO:0048487`, not for this row. Retracted, moved to the notes, and guarded — see below.)* → **REMOVE**

Two further supports: `GO:0005507` copper ion binding rests on a NAP/Abeta-aggregation paper whose abstract mentions no copper, in a protein whose only metal features are nine C2H2 **zinc** fingers; and two of the peptides in the block (`ADNF-9`, SALLRSIPA) are **not ADNP fragments at all**.

**`GO:0043524` reaches human twice and both chains end in peptide.** Checking the *other* rodent source is what turned suspicion into a verdict: mouse Adnp's own `GO:0043524` IDA is the 1999 cloning paper, whose neuroprotection data are subfemtomolar NAPVSIPQ in mixed cultures.

## Other findings

**`GO:0044849` estrous cycle → REMOVE, decided from the definition.** Its formal taxon constraint is `only_in_taxon Theria`, which humans satisfy — so the constraint does *not* block it. The **differentia does**: "…where the endometrium is **resorbed** if pregnancy does not occur", which is precisely what distinguishes it from the menstrual cycle of humans and other catarrhines. Donor evidence is separately an mRNA-oscillation IEP in rat vagina. `source_status: SUPPORTS_SOURCE_BUT_NOT_TARGET` — rats do have an estrous cycle.

**`GO:0005576` extracellular region → MARK_AS_OVER_ANNOTATED.** Unlike the rest of the block this is a real protein measurement, but: a single ~1.4-fold immunoreactivity change; human ADNP has **no `SIGNAL` feature** (FT runs `CHAIN 1..1102`); UniProt lists only Nucleus and Chromosome; and the detecting antibody was raised **against the NAP epitope**, so it cannot distinguish full-length ADNP from a NAP-containing fragment in conditioned medium.

**14 `GO:0005515` rows → MODIFY, per partner.** All four partners resolve to reviewed canonical Swiss-Prot entries (no TrEMBL clones, no ORFeome fragments). HP1 binding is recovered by **seven independent studies**, including OpenCell at *endogenous* expression and IP-MS in human induced neurons — so it is not screen noise. Per the brief, `NbExp` (5/7/3) was **not** used as an evidence proxy; the reference-level count is.
- CBX1/CBX3/CBX5 rows → **`GO:0070087` chromo shadow domain binding**. Proteins engage HP1 via PxVxL docking on the chromo shadow domain, and ADNP's is conserved to ADNP2. The committed scan places the single `P.V.[LMIV]` match at PGVLL 820–824, immediately C-terminal to the homeobox — against **0.76 expected by chance**, so the match is real but its uniqueness carries no weight. Caveat filed as a knowledge gap: no ADNP PxVxL point mutant has been shown to lose HP1 binding.
- CHD4 row → **`GO:0140463` chromatin-protein adaptor activity**, whose definition is exactly the role the experiments assign to ADNP.

**2 NEW terms proposed additively:** `GO:0001227` (the measured *direction* is repression; the parent `GO:0000981` is ACCEPTed, not modified) and `GO:0010526` transposable element silencing. `GO:0010526` was chosen deliberately **over** `GO:0141005` "…by heterochromatin formation", because the same work reports heterochromatin and DNA methylation play only a minor role — the child would assert a mechanism the data exclude.

**Proposed new CC term: ChAHP complex.** QuickGO returns nothing for "ChAHP" and the Complex Portal search returns zero, so ADNP's only human experimental annotation has to use the generic `GO:0090575`. Reactome already models the assembly as R-HSA-9940477.

## Non-confirmations, reported as such

- **The homeobox lead FAILS here.** ADNP genuinely does recognise DNA motifs sequence-specifically (three independent studies beyond the ChIP paper the ISS rests on), so `GO:0000977`/`GO:0000981` are **ACCEPT**. Noted en route: MGI read the same paper conservatively as `GO:0003682` chromatin binding while ARUK-UCL read it as sequence-specific binding + DbTF activity; the later work vindicates the stronger reading.
- **Reference-projection test: negative.** Every rat donor reference annotates 1 entity (2 exceptions at 3 and 2). No ACTR8-shaped complex-to-subunit projection.
- **Reactome TAS checked and benign** — `R-HSA-9940477` covers exactly 4 entities (ADNP, CBX1, CBX3, CHD4) and 1 term, with no functional term spreading alongside.
- **Logical-opposite cross-product: negative** — no positive/negative regulation pair of the same process exists in the TSV at all, so no shared-reference defect. The Wnt-direction disagreement in the literature is real but is not encoded as an annotation contradiction; both references flagged `DISPUTED`.
- **Retraction / erratum: negative** across all 33 cited PMIDs.

## affinage: gate passed, recall missed, and one mechanistic error

`gates_passed: True`, `faith_pct: 100.0`, 28 citations — and **`PMID:29795351` (Ostapcuk 2018, the ChAHP-defining Nature paper) is absent from them**, as are `PMID:31491387` and both 2026 *Mol Cell* papers. These are the papers GOA itself cites for ADNP's only human IDA. It also states the composition wrongly: its grounding block reads `ChAHP complex (ADNP-CHD4-BRG1)` — ChAHP is ADNP–CHD4–**HP1**. The record is cited once, in the `GO:0090575` row, quoting that exact line as evidence *about the record*, not about biology.

## Evidence born in the repository

Two committed scripts under `ADNP-bioinformatics/`, both break-tested:

- **`analyze_compara_donors.py`** resolves every Compara row to its donor and primary reference and *computes* the peptide-vs-protein classification from PubMed rather than hand-assigning it. Asserts `numberOfHits == len(results)` on every QuickGO query, one donor per row, and — the happy-path direction — that **both** classes are non-empty, so a classifier that called everything "peptide" would fail rather than manufacture the conclusion. `--check` proves the committed `results.json`/`RESULTS.md` reproduce.
  It also carries `pxvxl_scan`, which asserts Q9H2P0 is still 1102 aa before trusting any position, confirms the single `P.V.[LMIV]` match at **PGVLL 820–824**, and reports **0.76 matches expected by chance** from the protein's own residue composition — so one match is what chance predicts and its uniqueness is *not* evidence. The weight rests on ADNP↔ADNP2 conservation.
  Its self-test caught **two real defects in my own patterns**: an unscoped protein marker firing on an siRNA against *Fyn*, not ADNP (which had made `GO:0045773` read as MIXED), and a bare `\bdisrupt` matching "…by disrupting ADNP signaling".
- **`audit_adnp_review.py`** runs over the **emitted YAML**, not the builder. Strict duplicate-key loader; raw-vs-parsed quote arithmetic; GOA↔review 1:1 reconciliation keyed on the stable entity; **every** `supporting_text` verbatim including `knowledge_gaps[].provenance`, which `checkquotes.py` does not walk (138 vs **145** quotes — the 7-quote difference derived independently, not rationalised); summary-opener/action agreement; and the load-bearing one, check **G**: the set of rows flagged `root_cause: SOURCE_BAD` must **equal** `results.json`'s computed peptide-derived set, selecting on GO id so rewording cannot drift it. Self-test: baseline + 9 mutations, each in the direction the check exists to catch.

## Verification

`just validate human ADNP` → `✓ Valid`, **on the branch merged into main**. `checkquotes.py` 138/0. `audit_adnp_review.py` clean, self-test 11/11. `cache_lint` clean.

**`cache/go/terms.csv` — a convention change worth flagging.** `just validate` first collapsed main's two then-existing duplicate curies (caught only by a **multiset** comparison). Then merging main forward revealed that main has since landed `cache_lint.py` + `tests/test_cache_sorted.py` requiring caches to be **sorted and deduplicated** — which **supersedes the append-at-the-end convention**. The two orderings mis-aligned under git's line merge and silently duplicated **24** curies. Resolved by resetting to main's file and letting validate re-insert the one needed row in sorted position; asserted four ways afterwards (nothing dropped, exactly one added, nothing invented, no duplicates).

## Verdict counts

| action | rows |
|---|---|
| ACCEPT | 14 |
| KEEP_AS_NON_CORE | 9 |
| MARK_AS_OVER_ANNOTATED | 8 |
| MODIFY | 15 |
| NEW | 2 |
| REMOVE | 7 |
| **total** | **55** |

**Correction, and the guard it produced.** The first version of this body stated a hand-counted tally that was **wrong on three of six actions** — MODIFY, KEEP_AS_NON_CORE and ACCEPT. REMOVE and MARK_AS_OVER_ANNOTATED, the two that carry the argument, happened to be right, which is exactly why the error survived a read-through. The tally now lives in a machine-readable block in `ADNP-notes.md` and `audit_adnp_review.py` **check I** asserts it equals the counts computed from the review file. Verified against the defect that actually shipped: injecting `MODIFY | 14` — the number this body used — makes check I fire, and it is clean on the current file.

## Review round 2 (`c6c651d3f`)

Both blocking items conceded after checking the reviewer's premises against my own committed `results.json` — the reviewer was right on each.

1. **`GO:0042277` inversion argument withdrawn.** The row's reason is now positive-only (entity identity), the retraction narrative moved to `ADNP-notes.md` where process history belongs, and **audit check J** guards it: it selects on the stable tokens (`16893427`, `RGD:71030`, `Tubb3`, `Tubb4b`, `invert`) rather than on the conclusion's wording, **fires on the exact version that shipped (`12f374d14`)**, is clean on the current file, and fails loudly if the guarded row is deleted. Stated limitation in its docstring: a paraphrase avoiding all five tokens would pass.
2. **`description` hedged** — the PxVxL is now "a C-terminal PxVxL-type motif, conserved with its paralogue ADNP2, of the kind that generally mediates HP1 binding, but the HP1 contact surface has not been mapped by mutagenesis and CHD4 rather than HP1 carries the activity required for the complex's known outputs."

All three non-blocking items also taken: the **PxVxL scan is committed** with the null expectation (0.76) that shows "exactly one match" is not enrichment — that framing is withdrawn from all four surfaces; the axon/dendrite/cell-body reason now says **MIXED** for PMID:18072088 rather than flattening it to protein-level; and the affinage `ADNP-CHD4-BRG1` quote is **out of `supported_by`** and recorded in a `reference_review` instead. The last one leaves the non-blocking "no annotations reference available deep research files" warning standing — correctly, since no annotation here rests on the provider record. Recorded in the notes rather than silenced.

No annotation action changed. Gates on the merged tree: `just validate` ✓ (1 non-blocking warning), `checkquotes` 137/0, `audit_adnp_review.py` clean with self-test **14/14**, `analyze_compara_donors.py --check` reproduces, `cache_lint` clean.

---

> [!NOTE]
> **Low Risk**
> Curation and documentation plus a small GO term cache row; no runtime application or auth paths. Main risk is factual/curation disagreement for downstream GO editors, not production breakage.
> 
> **Overview**
> Adds a **complete human ADNP (Q9H2P0) GO curation review** that walks all 53 GOA rows and records accept/modify/remove/new verdicts, core ChAHP-centric functions, knowledge gaps, and upstream curation questions.
> 
> The main substantive change is **tracing Ensembl Compara (`GO_REF:0000107`) projections** back to rodent donors and showing that **13 rat-derived human rows rest on synthetic NAP/ADNF peptide experiments**, not perturbation of the ADNP protein—driving **REMOVE** / **MARK_AS_OVER_ANNOTATED** calls (e.g. estrous cycle on human, peptide pharmacology rows) and **MODIFY** of repeated bare **HP1 `GO:0005515`** IPIs toward **`GO:0070087` chromo shadow domain binding** and **CHD4 → `GO:0140463` chromatin-protein adaptor activity**, plus additive **NEW** proposals for repressor and transposable-element silencing terms and a **ChAHP complex** CC term.
> 
> Ships **`analyze_compara_donors.py`** with committed **`RESULTS.md` / `results.json`**: QuickGO/PubMed–driven peptide-vs-protein classification, HP1 PxVxL motif scan, retraction checks, and `--check` reproducibility.
> 
> Updates **`cache/go/terms.csv`** with **`GO:0070087`** (chromo shadow domain binding) in sorted cache form.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6c651d3ff961977801d4adcb3244573e3cfd5ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — both blocking items were right, and I checked each against my own committed `results.json` before conceding rather than taking them on authority. Pushed as `c6c651d3f`. No annotation action changed.

**1. `GO:0042277` inversion argument — withdrawn.** Verified independently: `results.json` shows `PMID:14706557` annotating exactly **1** entity with terms `GO:0005507` + `GO:0042277`, while `PMID:16893427` annotates **3** entities (`Q4QRB4` Tubb3, `Q6P9T8` Tubb4b, `Q9JKL8` Adnp) with `GO:0042277` + `GO:0048487`. So the tubulin rows belong to the `GO:0048487` donor and there is no internal contradiction in RGD's records of the kind I described. And you are right that Abeta *is* a peptide, so for the fragment the direction is correct — the annotation is on the wrong entity, not pointing the wrong way.

This is the campaign's most-repeated defect in a new dress, and I built it exactly the way the brief warns: a composite claim assembled across two sources, with every constituent quote verbatim.

The row's reason is now positive-only and the retraction narrative moved to `ADNP-notes.md`, per the brief's rule that process history belongs in the notes, not in a row summary. That also made a token guard possible, which it would not have been if the row kept quoting its own retracted claim:

**`audit_adnp_review.py` check J** selects on the stable tokens — `16893427`, `RGD:71030`, `Tubb3`, `Tubb4b`, `invert` — rather than on the conclusion's wording, since the conclusion is exactly what gets reworded. Verified three ways: it **fires on the precise version that shipped**, `git show 12f374d14:genes/human/ADNP/ADNP-ai-review.yaml` → `J: withdrawn cross-row claim tokens back in the GO:0042277 row: ['16893427', 'RGD:71030', 'Tubb3', 'Tubb4b', 'invert']`; it is clean on the current file; and it **fails loudly if the guarded row is deleted**, so it cannot be defeated by removing the thing it guards. Its limitation is stated in the docstring: a paraphrase avoiding all five tokens would pass, so prose still needs human re-reading.

**2. `description` hedged.** Now: *"…it carries a C-terminal PxVxL-type motif, conserved with its paralogue ADNP2, of the kind that generally mediates HP1 binding, but the HP1 contact surface has not been mapped by mutagenesis and CHD4 rather than HP1 carries the activity required for the complex's known outputs."* Your framing was better than mine and I used it.

**Suggestion 1 — the sharpest of the three, and taken in full.** `pxvxl_scan` is now committed in `analyze_compara_donors.py`, contributes to `results.json`/`RESULTS.md`, and is self-tested (overlap-aware regex; rejects a non-`[LMIV]` final residue; rejects a 4-residue window). It asserts Q9H2P0 is still **1102 aa** before trusting any position — a short or drifted reference sequence would otherwise make every residue number silently wrong — and confirms `PGVLL` at **820–824**.

Your expectation calculation was the important part. Computed from the protein's own residue composition rather than a uniform 1/20: **0.758 matches expected by chance**. One observed match is therefore what chance predicts, and I have withdrawn the uniqueness framing from all four surfaces where it appeared (8 `MODIFY` rows via the shared constant, `core_functions[3]`, `knowledge_gaps[2]`, and the notes). What remains is the ADNP↔ADNP2 conservation plus all three HP1 paralogues binding, which is what the argument actually rested on.

**Suggestion 2 — fixed.** The shared reason now reads *"classifies PMID:19040308 as PROTEIN and PMID:18072088 as MIXED… neither is counted as peptide-derived, which is the conservative direction. Being uncounted as peptide-derived is not the same as being classified protein-level, so the distinction is stated rather than flattened."* (Typo aside — the file says `PMID:19130308`.) Your one-word framing was the right diagnosis: I had flattened two different classifications into one word.

**Suggestion 3 — taken, with a consequence I chose not to hide.** The affinage quote is out of `supported_by` and the critique is now a `reference_review` on `file:human/ADNP/ADNP-deep-research-affinage.md` marked `MISCITED`, which is the field designed for it. The consequence is that the repo's non-blocking *"No annotations reference available deep research files"* warning now fires. I am letting it stand rather than restoring the citation to silence it, because it is telling the truth: **no annotation in this review rests on the provider record.** Recorded in `ADNP-notes.md` §9 so a future reader does not "fix" it by putting the wrong quote back.

**Suggestion 4** — noted, no change, and I agree with your reading of the risk. If a GO curator prefers to wait for V822A/L824A before accepting `GO:0070087`, the fallback is bare `GO:0005515` on 8 rows, which is strictly less informative; the gap and the exact experiment are already filed so the hedge is visible either way.

One correction I made unprompted in the same round, since it is the same class of error you caught: **the verdict tally in the first PR body was hand-counted and wrong on three of six actions** (MODIFY, KEEP_AS_NON_CORE, ACCEPT). REMOVE and MARK_AS_OVER_ANNOTATED — the two that carry the argument — happened to be right, which is why it survived a read-through. The tally now lives in a machine-readable block in the notes and **check I** asserts it equals the computed counts; verified by injecting `MODIFY | 14`, the number the body shipped, and confirming it fires.

Gates on the merged tree: `just validate` ✓ (the one warning above), `checkquotes` 137/0, `audit_adnp_review.py` clean with self-test **14/14**, `analyze_compara_donors.py --check` reproduces both generated files, `cache_lint` clean.

PR #2332, opened 2026-07-27
Review human ADPRH (P54922): arginine specificity is correctly annotated; the same IPR012108/ARBA route reaches the catalytically inactive paralog ADPRHL1

cmungall:
Review of **human ADPRH / ARH1 (P54922)** — ADP-ribosylarginine hydrolase, EC 3.2.2.19.

Accession verified independently against `projects/paint/human-no-IBA-simple.csv:2907`
(`human,P54922,ADPRH`) and UniProt (`ADPRH_HUMAN`, reviewed).

**Row reconciliation, done before reviewing:** GOA has 15 lines, 15 distinct, and the
`fetch-gene` stub seeded 15 entries. No collapse on this gene — the two `GO:0005515` IPI
rows differ by reference as well as by partner, so the seeder key kept them apart. A
non-confirmation of the ADAMTSL5 stub-collapse defect.

**The worklist name is wrong again:** ADPRH is on `human-no-IBA-simple.csv` and carries
**two IBA rows**. Both were adjudicated with the full propagation apparatus.

## Verdicts (15/15 rows)

| action | n | rows |
|---|---|---|
| ACCEPT | 10 | `GO:0003875` ×4 (IBA, IEA×1, IDA, IMP), `GO:0051725` ×3 (IBA, IEA, IDA), `GO:0000287` ×3 (IEA, IDA×2) |
| MARK_AS_OVER_ANNOTATED | 4 | `GO:0005515` (TFCP2), `GO:0005515` (SPSB1), `GO:0005576`, `GO:0030955` |
| MODIFY | 1 | `GO:0036211` → `GO:0051725` |
| REMOVE | 0 | — |

## The specificity question, and the answer that was *not* the expected one

This is the family where a paralog's specificity could plausibly have propagated —
ADPRH (arginine) vs ADPRS/ARH3 (serine, PAR, O-acetyl-ADP-ribose) vs the inactive
ADPRHL1. **It has not.** No serine-hydrolase, PARG or OAADPr term appears anywhere on
ADPRH, and the discrimination is made deliberately in two independent places:

- `IPR012108` "ADP-ribosylarginine hydrolase" matches ADPRH **and ADPRHL1** but **not**
  ADPRS. The three fold-level signatures ADPRH also matches map to nothing in
  interpro2go — InterPro2GO's own restraint is the internal control.
- PAINT separates the clades: `PTN009030515` carries `GO:0003875`+`GO:0051725` and
  reaches 20 gene products, all ADPRH orthologs; `PTN008564042` carries `GO:0004649`,
  `GO:0140290`, `GO:0140292`, `GO:0071451` and reaches only ADPRS orthologs. **Neither
  node reaches ADPRHL1.**
- `PMID:30472116` annotated both specificities in one pass: the reference-projection test
  returns 7 annotations over 2 entities, P54922 getting the arginine terms and coelacanth
  ADPRS `H3BCW1` getting the serine/PARG terms.

Reported as a **non-confirmation**, per the campaign's own calibration note.

## The reciprocal half IS a defect, and it is measured

`genes/human/ADPRH/ADPRH-bioinformatics/catalytic_residue_census.py` (committed,
self-testing, report regenerates byte-identically) scores the five positions UniProt
records as *"Complete loss of activity"* (S54, D55, D56, D302, S305) across **all 31
reviewed Swiss-Prot members of PTHR16222 — 0.104% of the family's 29,860 proteins**, and
cross-tabulates against who holds `GO:0003875`:

| clade | n | % id to ADPRH | residues retained | disruptive in EVERY member | holds GO:0003875 |
|---|---|---|---|---|---|
| ADPRH (ARH1) | 5 | 48.4–100 | 5/5 | none | 5/5 |
| **ADPRHL1 (ARH2)** | **7** | **42.6–47.7** | **1–2 of 5** | **D56, S305** | **7/7** |
| ADPRS (ARH3) | 7 | 25.8–28.1 | 3–4 of 5 | none | 0/7 |
| other/non-vertebrate | 12 | 20.0–32.5 | 0–4 of 5 | none | 4/12 |

ADPRHL1's own Swiss-Prot record is titled **"Inactive ADP-ribosyltransferase ARH2"** and
carries the CAUTION that it *"lacks the metal-binding and substrate-binding residues"* —
yet it holds `GO:0003875` by `IEA GO_REF:0000120` and `GO:0000287` by
`IEA GO_REF:0000002`. **PAINT models the catalytic loss correctly and gave that clade
nothing; InterPro2GO and ARBA do not, and give it both terms.** Filed as a
`suggested_question` to interpro2go and ARBA, not as a GO action on this gene.

The argument is bounded by a positive control rather than by a threshold: *R. rubrum*
**DraG** holds `GO:0003875` by its **own EXP evidence** at 27.5% identity and only 3/5
residues — but both of its substitutions are hydroxyl-preserving Ser→Thr, i.e. **zero
disruptive**. An identity cut was tried first and **rejected in the report**: the largest
observed gap lands at 65.4%, between *Dictyostelium* ADPRH and mouse Adprh, which is a
taxonomic boundary and would have discarded the ADPRHL1 signal entirely.

## The other three defects

- **`GO:0005576` extracellular region → MARK_AS_OVER_ANNOTATED.** Traced the whole chain:
  human holds it only by Compara IEA from rat `Q02589`, whose own IDA is `PMID:9037477` —
  *"Immunohistochemical localization of ADP-ribosylarginine hydrolase in rodent CNS"*,
  a paper of intracellular neuronal staining whose extracellular claim is **one 38-kDa
  band on a western blot of CSF**. Fully-paginated reference-projection test: **1
  annotation on 1 entity**, so a single curator call, not a bulk import. Against it:
  no signal peptide, no TM, **no `SUBCELLULAR LOCATION` comment at all**, and both
  verified substrates (Gsα, TRIM72) intracellular. Not `REMOVE` — the CSF band is
  unexplained, not refuted, and the rat IDA is not second-guessed from an abstract.
- **`GO:0036211` → MODIFY to `GO:0051725`.** This looked like the reaction-direction trap
  the brief flags — a modification term on an enzyme that removes the modification — and
  the check was run rather than assumed. **It is not an inversion:** QuickGO's ancestors
  endpoint (`relations=is_a,part_of`) puts `GO:0036211` in the ancestor set of
  `GO:0051725`, via `GO:0043412`. So it is redundant, not misdirected.
- **Two `GO:0005515` rows → MARK_AS_OVER_ANNOTATED, judged per partner.** Both partners
  resolve to reviewed canonical Swiss-Prot entries with matching lengths (no ORFeome
  substitution). But expanding IntAct reproduces the `NbExp` trap a **third** time: each
  is *one* Y2H screen logged as `two hybrid array` + `two hybrid prey pooling approach` +
  `validated two hybrid`, MI-score 0.56, no orthogonal assay among ADPRH's 17 IntAct
  records. TFCP2 has 350 distinct partners and SPSB1 78, against ADPRH's 17.

## Ontology gap filed

**`alpha-NAD(+) hydrolase activity` has no GO term.** UniProt records the reaction on
P54922 with `Rhea:RHEA:68792`, `ChEBI:CHEBI:77017` and `ECO:0000269|PubMed:31599159`. The
label-plausible candidate is wrong for a reason invisible from its label: `GO:0003953
NAD+ nucleosidase activity` is defined on the **beta** anomer (`EC 3.2.2.5`), and the
finding being annotated is precisely that these enzymes hydrolyse α-NAD+ **but not**
β-NAD+. Would apply immediately to ARH1, ARH3, MacroD1, MacroD2, TARG1 and Af1521.

## Checks that came back negative (reported so the next reviewer knows they were run)

- **Logical-opposite cross-product:** no positive/negative regulation pair exists in the
  15 rows, so there is nothing to intersect.
- **Reaction direction:** no term anywhere in the GOA asserts that ADPRH *performs*
  ADP-ribosylation.
- **Partner accession identity:** both `GO:0005515` partners are reviewed canonical
  entries at full length — no TrEMBL/partial-clone substitution.
- **EC-derived row:** `GO_REF:0000120`'s WITH/FROM cites `RHEA:14885` and `EC:3.2.2.19`,
  both matching UniProt's experimentally evidenced CATALYTIC ACTIVITY exactly.
- **Retraction/erratum sweep:** 21/21 PMIDs returned by efetch, reading `PublicationType`
  **and** `CommentsCorrections/RefType` — zero retractions, errata or expressions of
  concern. Crossref `update-to`/`relation` additionally clean for the four load-bearing
  DOIs; two DOIs 404'd at Crossref and are recorded as *unchecked by that route*, not
  clean.

## affinage: `gates_passed: False`, two verified provider defects

1. **`PMID:12464675` is about a different gene.** Its "ARH1" is the autosomal recessive
   hypercholesterolaemia adaptor **LDLRAP1**, mapped in that paper to chromosome 1p36;
   ADPRH is at 3q22.1. Marked `relevance: NONE` / `correctness: WRONG_IDENTIFIER`.
2. **`PMID:16278211`'s mutagenesis is ARH3's, not ARH1's** — the abstract attributes
   Asp-77/Asp-78 to ARH3. Recorded in `review_notes`; the reference's own `correctness`
   stays `VERIFIED`, because that field describes *this* review's use of it.

`PMID:36945646` is a bioRxiv preprint; nothing rests on it. Recall was poor as predicted:
the record returned **none** of `PMID:9037477`, `PMID:1375222`, `PMID:25416956`,
`PMID:32296183` or `PMID:19895577`, which between them decide four rows.

## Committed guards

`ADPRH-bioinformatics/audit_adprh_review.py` — nine invariant checks (A–I) over the
emitted YAML: strict duplicate-key loader, anchors/aliases, raw-vs-parsed quote
reconciliation, GOA row reconciliation, the logical-opposite cross-product,
summary-opener-vs-action, `core_functions` agreement in **both** directions, and "a
COMPLETE review has no PENDING rows". `--self-test` clean.

Three defects in my own guards were found by breaking them, not by reading them, and are
written up in the notes: the anchor detector was line-anchored and missed the inline
`key: &id001` form; one self-test mutation inserted a *commented-out* line and was
therefore a silent no-op that "proved" a guard; and **running the audit against the
un-reviewed `fetch-gene` stub cleared 7 of 8 checks** on 15 `PENDING` rows, which is what
prompted check H. Running a guard against the state that would really have shipped is
strictly stronger than a self-test and costs one `git show`.

## Gates

`just validate human ADPRH` → **✓ Valid**, both on the branch and on the branch merged
into `main`. `checkquotes.py` → 44 quotes, 0 problems. `cache/go/terms.csv` verified by
**multiset**: 2 rows added (`GO:0003875`, `GO:0051725`), **nothing lost**, no new
duplicates; the post-merge union asserted four ways.

**Round 2 correction — the label warning is gone, and my explanation of it was wrong.** The PR originally said the `Label mismatch for 'GO:0003875'` warning came from a stale ontology snapshot and should not be "fixed". A controlled test refutes that: with the stale label in `cache/go/terms.csv` the warning appears; with the current label in that same file it disappears. **The expected label is read from `cache/go/terms.csv`.** Fixed there — which is also the reviewer's actual concern, that this PR was writing the older synonym into shared state — and `just validate human ADPRH` is now `✓ Valid` with **no warnings**.

---

> [!NOTE]
> **Low Risk**
> Curation and analysis artifacts only; no runtime services or auth. Risk is limited to shared GO term cache edits and large generated JSON/YAML that reviewers should treat as authoritative for this gene.
> 
> **Overview**
> Adds a **complete** human **ADPRH** (P54922) GO curation review: per-row verdicts on all 15 GOA annotations, literature and propagation notes, **core_functions**, ontology-gap proposals, and curator-facing questions/experiments.
> 
> **Annotation outcomes:** Most arginine hydrolase, de-ADP-ribosylation, and magnesium-binding rows are **accepted**; four rows are **marked over-annotated** (two Y2H protein-binding hits, extracellular region from cross-species projection, potassium binding from a single low-resolution structure); one redundant process term is **modified** to `GO:0051725`. The review argues ADPRH’s arginine specificity is correctly separated from ADPRS and that **InterPro/ARBA** still assign hydrolase terms to catalytically inactive **ADPRHL1**.
> 
> **Supporting analysis:** A PANTHER-family **catalytic-residue census** script and committed **RESULTS** tie `IPR012108` / `GO:0003875` to conserved catalytic residues and PDB metal occupancy. **`audit_adprh_review.py`** adds nine YAML/GOA invariant checks (including quote-to-term alignment). **`cache/go/terms.csv`** gains current labels for `GO:0003875` and `GO:0051725`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90d28a79fd1f8b3074843c2c6fcb13b27a1476bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

## Round 2 (commit `453ec2d5b`)

**Blocking item fixed.** Three `supporting_text` quotes on `GO:0000287` never mentioned
magnesium, and one was the potassium row's evidence sitting under a magnesium row. Every
quote was verbatim, correctly attributed and passed `checkquotes.py` — the error was in the
join. `PMID:30472116`'s cached full text supplies an on-point sentence for two of the three;
`PMID:19407395`'s cache is abstract-only with **zero** magnesium mentions, so the census
script now queries RCSB for all four ADPRH structures and the row cites the computed table
(magnesium **4/4**, potassium **1/4** — 3HFW, the lowest resolution), with the abstract quote
retained under `full_text_unavailable: true`.

**Guarded.** New audit check `I` requires every `supporting_text` to contain a surface form of
its own row's term, keyed on the **GO id**. Run against `d86ff17d6` — the commit that was
blocked — it fires on exactly the three flagged sites and nothing else, and is clean on the
current head. It declares its own limitation in the source: surface forms only, and a term
with quotes but no declared forms is a loud failure rather than a vacuous pass.

**Suggestions taken:** `GO:0030955` → `MARK_AS_OVER_ANNOTATED` (decisive argument I had
missed: UniProt read the same paper, made six Mg²⁺ `BINDING` features citing
`PubMed:19407395`, and made no potassium feature — two curators over one paper, so still not
`REMOVE`); `PMID:16278211` back to `VERIFIED`; `is_invalid` dropped from `PMID:12464675`;
mouse provenance of both substrates made explicit in `core_functions[0]`; identity-denominator
note added to `RESULTS.md`.

**Declined, with reason:** folding `core_functions[1]` into the hydrolase entry.
`molecular_function` is single-valued, so folding means dropping `GO:0000287` from
`core_functions` — and audit check `G2` requires an ACCEPTed MF term to appear there or carry a
written exemption. The two Mg²⁺ are held by the same five residues whose substitution abolishes
activity, so the ion is part of the catalytic machine rather than an accessory.

**Gates:** `just validate human ADPRH` → `✓ Valid`, **no warnings** · `checkquotes.py` → 46
quotes, 0 problems · `audit_adprh_review.py` and `catalytic_residue_census.py` self-tests → 0
problems · `RESULTS.md`/`results.json` regenerate byte-identically · `cache/go/terms.csv`
multiset: 2 rows added, nothing lost, no duplicates.


cmungall:
Thanks — the blocking item was a real defect and a useful one, so I fixed it three ways rather than one, and I also have a **correction to make to my own PR body**.

## 🟡 Blocking: magnesium rows now quote magnesium

You were right that this is worse than it looks. All three quotes were verbatim, correctly attributed and passed `checkquotes.py`; the failure was in the *join* between the quote and the row, which is precisely the class no gate in this repo can see. And the third one was a **cross-row citation slip** — the potassium row's evidence sitting under a magnesium row two entries above it.

1. **`GO:0000287` / `PMID:30472116`** and **`core_functions[1]`** now quote your suggested sentence verbatim: *"The active sites of hARH1 and LchARH3 are structurally very similar and contain residues for the coordination of two Mg2+ ions."*
2. **`GO:0000287` / `PMID:19407395`** — you were right that no better quote exists; `grep -ci 'magnesium|Mg2\+|Mg\(2'` on the cached record returns **0**. Rather than pick between your two options I did both and added a third: the census script now queries RCSB for **all four** ADPRH structures and `RESULTS.md` carries a metals table, so the row cites computed evidence that lives in the repo, and the abstract quote is retained with `full_text_unavailable: true` because that is what it is.

   | PDB | res (Å) | components | MG | K |
   |---|---|---|---|---|
   | 3HFW | 1.92 | K, MG | yes | **yes** |
   | 6G28 | 1.23 | AR6, MG | yes | no |
   | 6G2A | 1.80 | A3R, MG | yes | no |
   | 6IUX | 1.195 | AR6, MG | yes | no |

   Magnesium **4/4**, potassium **1/4**. Both the magnesium and potassium claims are now checkable by a reader with no network — which they were not before.

3. **Guarded.** New audit check `I`: every `supporting_text` must contain a surface form of its **own row's term**, keyed on the GO id rather than on wording that drifts. Run against `d86ff17d6` — the exact commit you blocked — it fires on precisely the three sites you flagged and nothing else, and is clean on the current head. It states its own limitation in the source: surface forms only, so it catches a cross-row slip and nothing subtler, and a term with quotes but no declared forms is a **loud failure**, not a vacuous pass.

## Correction to my own PR body: suggestion 5 was right and my explanation was wrong

I wrote that the `Label mismatch for 'GO:0003875'` warning came from a stale ontology snapshot and that the review should not be rewritten to match it. **The premise was false**, and a thirty-second controlled test settles it:

```
stale label in cache/go/terms.csv   -> ✓ Valid (with 1 warnings)  ... Label mismatch
current label in the same file      -> ✓ Valid
```

The expected label is read from **`cache/go/terms.csv`** itself. So your point — that this PR was writing the older synonym into shared state — was not just a tidiness concern, it *was* the warning. One line fixed, and `just validate human ADPRH` is now clean with **no warnings**. I had carried that explanation for two rounds with some confidence.

## Suggestions

- **1 · `GO:0030955` → `MARK_AS_OVER_ANNOTATED`.** Taken. Working through it surfaced an argument I had not made and which I think decides it: **UniProt read this same paper, created six Mg²⁺ `BINDING` features citing `ECO:0000269|PubMed:19407395`, and created no potassium feature at all.** That reframes the disagreement as two curators over one paper rather than this review against a curator. Still not `REMOVE` — the definition is bare binding and the ion is genuinely resolved — and the ADGB `GO:0019825` precedent holds for the floor, but ADGB's oxy complex was measured in solution at physiological affinity; a buffer-derived ion in the lowest-resolution of four structures is not the same evidence.
- **2 · fold `core_functions[1]`.** Declined, with a reason. `molecular_function` is single-valued, so folding means dropping `GO:0000287` from `core_functions` entirely — and audit check `G2` requires every ACCEPTed MF term either to appear there or to carry a **written exemption**, specifically so a term cannot be quietly dropped. Writing that exemption is the honest alternative, but the biology does not call for it: the two Mg²⁺ are coordinated by the same five residues whose substitution abolishes activity, so the ion is part of the catalytic machine rather than an accessory. Happy to be overruled if you'd rather see the exemption.
- **3 · `PMID:16278211` `MISCITED` → `VERIFIED`.** Taken, and you're right about the field's semantics — `correctness` describes this review's use. The provider defect stays in `review_notes` where it belongs.
- **4 · `is_invalid` on `PMID:12464675`.** Taken and dropped. A wrong-gene collision is not a retraction; `relevance: NONE` + `correctness: WRONG_IDENTIFIER` already carry it.
- **6 · mouse provenance of the substrates.** Taken — `core_functions[0].description` now says both substrate identifications come from mouse knockout work and are listed as the human orthologs, undemonstrated on the human enzyme.
- **7 · identity denominator.** Taken — `RESULTS.md` now states that % identity is computed over aligned columns only, applied identically to every member, and should not be set against externally quoted identities.

## Gates

`just validate human ADPRH` → `✓ Valid`, **no warnings**, on the branch and merged into `main` · `checkquotes.py` → 46 quotes, 0 problems · both scripts' `--self-test` → 0 problems · `RESULTS.md` and `results.json` regenerate byte-identically · `cache/go/terms.csv` by multiset: 2 rows added, nothing lost, no duplicates.

Noted that `just`/`uv` were unavailable in your runner — the two checks most worth re-running by hand are `grep -c 'coordination of two Mg2+ ions' publications/PMID_30472116.md` (1) and check `I` against `d86ff17d6`, which reproduces your finding exactly.

cmungall:
Thanks for the approval. I took all four 🔵 items rather than deferring them, because two of them changed what a **cited** artifact asserts — and enforcing one of them found more than you flagged.

**1 · Generated prose asserting an uncomputed claim.** This was the one that mattered. `RESULTS.md` opened its metals paragraph with *"Two magnesium ions per subunit are coordinated by the catalytic residues…"* — a UniProt `COFACTOR` statement, not something `nonpolymer_bound_components` can support — **and `core_functions[1]` was quoting that sentence as `supporting_text`.** So a hardcoded assertion was serving as evidence in the very field built to be verifiable. The generator now emits only computed content; the stoichiometry stays in the row's `reason`, attributed to UniProt.

**2 · Overlapping surface forms — and the class behind the instance.** You were right that `| K, MG |` satisfied both form sets. I dropped the table-cell fragments for word forms and made **disjointness an assertion** rather than a property I was trusting. It immediately found **four more overlaps I had introduced myself**:

```
GO:0003875 / GO:0051725 : 'hydrolase' / 'hydrolase'
GO:0036211 / GO:0051725 : 'modification' / 'de-modification'
GO:0036211 / GO:0051725 : 'modification' / 'reversible modification'
GO:0036211 / GO:0051725 : 'reversible modification' / 'reversible modification'
```

Re-run against `d86ff17d6` after tightening, check `I` still fires on exactly the three sites and nothing else — so the narrower forms did not cost it sensitivity, which was the thing worth checking.

**3 · Docstring said "every", predicate is "any".** Fixed. A row may legitimately carry a corroborating quote beside the on-point one, so `any` was always the intended semantics — and a guard misdescribed in its own docstring is exactly how the ADCK5 lint slipped past its author.

**4 · `PDB_ENTRIES` hardcoded.** Now derived from UniProt's `xref_pdb`. The derived set reproduces the literal exactly (`3HFW, 6G28, 6G2A, 6IUX`), which is the confirmation worth having rather than a silent swap.

**One defect of my own, introduced by #4.** The new `pdb_ids_of` self-test had a `try` with no `else`, so an accession that happened to resolve would have passed it **silently** — a vacuous test. Fixed with a deterministic unresolvable accession and an explicit `else`.

Thank you also for the round-1 self-correction on the label warning, and for checking the cache blast radius — that no other review references `GO:0003875` is the part I could not easily verify myself.

---

**Stopping criterion**, stated explicitly. Rounds 2–3 changed one action (`GO:0030955`), four quotes, one reference `correctness`, one `is_invalid`, and the surface-form table. Everything I can still see is prose about the harness rather than about what the harness computes. **If a further round produces only items of that kind, I will fix anything that could misfire on another machine or misstate a number, and decline changes that only reword an explanation.** No annotation, action, evidence code, quote or reported number is outstanding on my side.

Gates on the new head: `just validate human ADPRH` → `✓ Valid`, no warnings · `checkquotes.py` → 46 quotes, 0 problems · both `--self-test` suites → 0 problems · `RESULTS.md` and `results.json` regenerate byte-identically · `cache/go/terms.csv` unchanged since round 2 (2 rows added, nothing lost, no duplicates).

This push supersedes the approval on `453ec2d5b`; re-review when convenient. Not merging and auto-merge is not enabled.

cmungall:
All three items sorted by the criterion, which is what stating it in advance was for.

**Fixed — the misstated number.** You caught one surface; there were **three**. The notes, `ADPRH-bioinformatics/README.md` and the PR body all still said "eight invariant checks (A–H)". Correcting three sentences would have left the fourth drift to the next person, so `--self-test` now **asserts that the letters enumerated in the module docstring are exactly the checks the code implements.**

Writing it exposed a subtlety worth recording: **check `A` is enforced by `StrictLoader` at parse time and never reaches `problems`**, so the naive version reported it as documented-but-missing. It is credited only after the loader is *demonstrated* to reject a duplicate key — behaviourally, not by grepping for the class name. Break-tested both directions:

```
delete the I docstring line    -> "docstring documents [A..H] but the code implements [A..I]"
disable the duplicate rejection -> "check A: strict loader accepted a duplicate key"
                                +  "docstring documents [A..I] but the code implements [B..I]"
```

**Fixed — the `better` edge case.** You were right that it conflated "better resolved" with "lacks potassium". Now computed from resolution, with an explicit tie branch and a printed note for any entry whose resolution is unavailable. **The regenerated report is byte-identical for the current four entries** — the fix was to the reasoning, not to the number, which is the outcome I wanted.

**Declined — co-occurrence.** Correct and already declared in the source: check `I` matches surface forms, catches a cross-row slip, "and nothing subtler". Widening it to forbid a quote that mentions a second entity would forbid legitimate practice, which is the guard-that-gets-worked-around failure mode rather than a stronger guard.

**One thing the fix exposed, which I'd flag for the brief.** Rewording the generated sentence from *"the 3 better-resolved structures"* to *"all 3"* **broke a `supporting_text` that quotes it.** Four quotes in this review cite `RESULTS.md`, so any edit to a generator's prose can silently invalidate them. `checkquotes.py` caught it only because it runs *after* regeneration. **A quote into a generated artifact is a two-way dependency**, and the direction that bites is the one where the generator changes and the quote does not — the mirror of the ADGRA2 detector/mutator-representation lesson.

Gates on `cb380a113` (merged forward past two more commits on main): `just validate human ADPRH` → `✓ Valid`, no warnings · `checkquotes.py` → 46 quotes, 0 problems · both `--self-test` suites → 0 problems · reports regenerate byte-identically · `terms.csv` union asserted four ways, no new duplicates.

Nothing outstanding on my side. Not merging; auto-merge not enabled.

cmungall:
Good catch, and it deserved another round — that sentence is quoted into the YAML, so a false version of it is a false citation, not just untidy prose.

**The substantive point: this was the third form of one confusion in three consecutive rounds.**

| round | `better` computed as | wrong because |
|---|---|---|
| 3 | structures *without* potassium | says nothing about resolution |
| 4 | resolution < **worst** K-bearing | a K-bearing structure can sit below the worst |
| 5 | resolution < **best** K-bearing | correct — "contains none" is then true *by construction* |

Each round fixed the previous objection and introduced the next, which is the signal to change the shape of the fix rather than the value. So, two responses:

1. **Define the set so the claim cannot be false.** Anything strictly better-resolved than *every* potassium-bearing structure cannot contain potassium, so "contain none" is entailed rather than filtered for — and the code now **asserts** that property with an `AnalysisError` instead of trusting it.
2. **Extract it into a pure `resolution_clause()` and unit-test both branches** on synthetic entries: your exact two-K-structures-at-different-resolutions case, plus the tie-at-worst case, a null resolution, and no-potassium-at-all. Your framing is the part that convinced me it was worth doing properly rather than patching — **the live four only ever reach branch 1, and `PDB_ENTRIES` is now derived from `xref_pdb`, so the set grows on its own and branch 2 becomes reachable with nobody editing the script.** Deriving the input made the script more correct and simultaneously more exposed; untested code that only *new data* can reach is the worst kind.

The regenerated sentence is byte-identical for the current four, so once again the fix was to the reasoning and not to the number.

**Also taken.** The docstring/implementation assertion moved out of `--self-test` and into `audit()`, so it runs on every invocation like the disjointness one — break-tested by deleting the `I` docstring line and confirming a plain `audit` run now reports it. Docstring entry `I` reordered after `H`. The dead `worst` variable went with the rewrite.

Thank you also for hand-tracing both regexes rather than taking the self-test on trust, and specifically for checking that the implementation scan does not match its own literal — `\.` where the pattern needs a real dot is exactly the trap that shape falls into, and I had not verified it.

Gates on `90d28a79f`: `just validate human ADPRH` → `✓ Valid`, no warnings · `checkquotes.py` → 46 quotes, 0 problems · both `--self-test` suites → 0 problems · `RESULTS.md`/`results.json` regenerate byte-identically · branch is level with `main`, `terms.csv` unchanged since round 2.

Nothing outstanding on my side. Not merging; auto-merge not enabled.

PR #2334, opened 2026-07-27
Review human ADPRHL1 (ARH2): a measured pseudoenzyme still carrying three fold-derived catalytic terms, from InterPro2GO and an ARBA FunFam-name rule

cmungall:
Review of **human ADPRHL1 / ARH2 (Q8NDY3)** — `ARHL1_HUMAN`, 354 aa, `PE 1`.

Accession verified independently against `projects/paint/human-no-IBA-simple.csv`
(`human,Q8NDY3,ADPRHL1`) and the UniProt REST record before any work started.

**Row reconciliation, done first:** GOA has **7** data lines, **7** distinct, and the
`fetch-gene` stub seeded **7** entries. They reconcile exactly — a non-confirmation of the
ADAMTSL5 stub-collapse defect. The review has **9** entries: those 7 plus **2 explicit `NEW`
proposals**.

**The worklist's "no-IBA" name is accurate here, and that is itself the finding.** Queried
rather than assumed: QuickGO returns 7 annotations for `Q8NDY3`, **none IBA**
(`numberOfHits == len(results)` asserted). *Positive control*, because a zero and a rejected
query look identical: the same call on `P54922` (ADPRH) returns 15 annotations of which **2
are IBA**. UniProt agrees — `PAN-GO; Q8NDY3; 0 GO annotations based on evolutionary models`.

## Verdicts

| action | n | rows |
|---|---|---|
| **REMOVE** | 3 | `GO:0000287` magnesium ion binding · `GO:0003875` ADP-ribosylarginine-[protein] hydrolase activity · `GO:0051725` protein de-ADP-ribosylation |
| ACCEPT | 4 | `GO:0030017` ×2 (IEA SubCell, ISS) · `GO:0003242` · `GO:0055003` |
| **NEW** | 2 | `GO:0055005` ventricular cardiac myofibril assembly (ISS) · `GO:0051894` positive regulation of focal adhesion assembly (IMP) |

## The pseudoenzyme premise held — and it is *measured*, not inferred

This is the distinction that earns `REMOVE` rather than `MARK_AS_OVER_ANNOTATED`. ADGRA2's
G-protein coupling was **unmeasured**; ADPRHL1's hydrolase activity was **measured and not
found**:

- `PMID:17075046` — recombinant ARH2 assayed alongside ARH1 and ARH3: *"ARH2 and
  poly(ADP-ribose) glycohydrolase were inactive."*
- `PMID:36497109` — the Moss-lab family review, on the exact reaction the GO term names:
  *"ARH2 does not appear to be responsible for hydrolysis of ADP-ribose-acceptor linkage."*
- UniProt curates the loss on this very accession, in both `FUNCTION` and `CAUTION`, and its
  feature table has **0** `BINDING`/`ACT_SITE` features. ADPRH has 14 such features spanning
  **20 distinct positions** and ADPRS 14 spanning **17**; ADPRHL1 has none at all.

**And the field's own caveat is carried, not suppressed:** *"However, the lack of ARH2
catalytic activity may be the result of the use of model substrates."* That is precisely why
**no replacement hydrolase term is proposed** — GO cannot express "might hydrolyse something
unidentified", and inventing `GO:0016787` to hold the hedge would be an over-annotation of
the opposite sign.

## Two routes deliver the terms, and neither inspects a residue

**1. interpro2go via `IPR012108`.** Downloaded the mapping file and looked up all four
signatures ADPRHL1 matches:

| signature | interpro2go |
|---|---|
| **IPR012108** ADP-ribosylarginine hydrolase (single member signature **PIRSF016939**) | **GO:0000287, GO:0003875, GO:0051725** |
| IPR050792 / IPR005502 / IPR036705 | *nothing* |

So this is **not** "InterPro2GO is careless" — three of four map to nothing. The entry itself
is the problem, and it is measurable: of its **1282** proteins, **11** are reviewed
Swiss-Prot; **4** are "ADP-ribosylhydrolase ARH1" and **7** are
**"Inactive ADP-ribosyltransferase ARH2"**, each with a lack-of-catalytic-residues `CAUTION`.
**A majority of the entry's own curated members are curated as catalytically dead, and the
entry still maps to the activity.**

**2. ARBA rule `ARBA00088955`**, named in this row's own WITH/FROM. Fetched, and its *entire*
condition set is:

```
FunFam id 1.10.4080.10:FF:000002  AND  taxon Mammalia   ->   GO:0003875
```

ADPRHL1's own cross-reference reads `FunFam; 1.10.4080.10:FF:000002; ADP-ribosylarginine
hydrolase isoform X1`. That FunFam holds ADPRH *and* ADPRHL1 alike (ARH3 sits in `FF:000001`),
so a catalytic MF is granted on the strength of a structural family's **name**. The taxon
clause makes a falsifiable prediction, and **GOA confirms it**: the five mammalian ARH2
orthologues receive `GO:0003875` via `GO_REF:0000120`, while *Xenopus* and zebrafish receive
it via `GO_REF:0000002`.

Across all **seven** reviewed ARH2 orthologues these three terms are held **21 times, every
one IEA** — no experimental, ISS or IBA support anywhere in the clade.

**PAINT is the negative control and gets it right.** `GO:0003875` by IBA reaches **20 gene
products** from node `PTN009030515` — **all 20 are ADPRH orthologues, none is an ARH2**.

## The mirror was tested, and it changed the outcome

A committed, self-testing residue census
(`ADPRHL1-bioinformatics/catalytic_site_census.py`) maps ADPRH's **20 UniProt-annotated**
ligand positions onto 17 family members. **The identity-matched control is what makes it an
argument:**

| clade | n | % id to ADPRH | identical of 20 | disruptive+gap | Mg donor kept of 6 |
|---|---|---|---|---|---|
| ARH1 (active; **positive control**) | 5 | 48.4–100 | 13–20 | 0–3 | 6 |
| **ARH2 (subject)** | **7** | **42.6–47.7** | **6–7** | **7–8** | **2–3** |
| ARH3 (active; discriminating) | 4 | 26.6–28.1 | 7–8 | 7–9 | 6 |
| DraG (active, low-identity control) | 1 | 27.5 | 9 | 8 | 6 |

*Dictyostelium* ADPRH is a genuine ARH1 at **48.4%** — the same distance as ADPRHL1's
42.6–47.7% — and keeps 13/20. DraG at **27.5%** keeps 9, i.e. **more than ADPRHL1 keeps at
nearly twice the identity**. Retention is not tracking sequence distance.

The decisive substitution is **D56→N58**: aspartate-to-asparagine at a vicinal catalytic
aspartate is the exact change used to *kill* this family's activity (`PMID:17075046` mutates
ARH3 D77/D78 to N and the reaction stops). BLOSUM62 scores D→N at **+1** and calls it
conservative; the mechanism does not — which is why the script reports a third,
mechanism-anchored `donor_group` column and rests on that.

**But a lost active site is not a lost function.** The same alignment reproduces, from
UniProt features alone, the structural claim the 2020 paper makes from its own model: ADPRH's
adenosine-ribose serines **S269/S270 align to ADPRHL1 R271/R272** — the di-arginine loop whose
1–4 residue CRISPR deletion abolishes ventricular myofibril assembly. Not decay: **a change of
chemistry**. So the four ACCEPTs stand, and **no MF term is proposed at all** — the gap is
recorded in `knowledge_gaps` rather than filled with `GO:0098772`/`GO:0008092` (affinage's own
grounding), for which there is no ligand, no substrate and **zero IntAct interactions**
(controls: 17 for ADPRH, 35 for ADPRS, both HTTP 200).

### Reproducing the sibling ADPRH panel (#2332) before reporting our own

- **Residue and mapped position: identical on all 25 shared calls.** Two independently written
  alignments agree.
- **Class label: 7 differences — all of *metric*, none of data.** #2332 uses hand-defined
  conservative groups, this script used `BLOSUM62 > 0`, and every difference is at D→N or S→A.
  **#2332's "disruptive" call is the biologically correct one**, and the `donor_group` test
  agrees with it — vindicated by an experiment inside the family rather than by a matrix.
  Reported and explained rather than silently absorbed.

## Two NEW rows, both from checks that found real defects

**`GO:0055005` — a split donor.** All three ISS rows transfer from `Q6AZR2` (Swiss-Prot
*X. laevis* adprhl1), which carries the 2016 morpholino paper. The reference-projection test on
the **2020 CRISPR paper** returns **10 annotations over 2 entities**, and those entities are
`A0A8J0UG54` / `A0A8J0UIM8` — **TrEMBL accessions for the same frog gene**, curated by Xenbase.
They hold `GO:0055005`, `GO:0030041` and `GO:0036342` by IMP, **none of which has ever reached
a mammal**. `GO:0055005` is an `is_a` child of the `GO:0055003` ADPRHL1 already has.
`GO:0030041` was **considered and declined** for human (morphological inference, not a
polymerization assay; any actin effect is downstream of ROCK); `GO:0036342` is frog-specific.

**`GO:0051894` — the human experiment GOA has never seen.** `PMID:37880701` (CRISPR knockout in
H9 hESC-derived cardiomyocytes) is the **only functional experiment ever done on the human
protein**, and its reference-projection test returns **0 annotations over 0 entities**.
Knockout reduces β1-integrin, pFAK, paxillin and talin and impairs focal adhesion formation;
ROCK or myosin II inhibition rescues it. Direction is stated by the experiment, so
*positive* regulation — and the regulation term rather than bare `GO:0048041`, because ADPRHL1
is not a structural FA component.

## Checks that came back negative, reported because they were run

- **Logical-opposite cross-product**: no positive/negative regulation pair exists in the 7
  rows — nothing to intersect.
- **Reaction direction / "who modifies whom"**: checked rather than assumed. `GO:0051725` *is*
  the removal direction, so it is not an inversion; it fails for a different reason. (The
  UniProt **name**, however, does invert it — see below.)
- **IntAct**: 0 for `Q8NDY3`, with non-zero controls from the same endpoint.
- **Retractions / errata**: **all 10** cited PMIDs via efetch reading `PublicationType`
  **and** `CommentsCorrections/RefType` — zero. Crossref `update-to`/`relation` clean on
  **all 10** DOIs, all HTTP 200. (The first pass covered the 8 PMIDs held at that point; the
  two references added later were swept afterwards rather than assumed.)
- **Pagination**: asserted on every QuickGO call; it fired once (an unrestricted `GO:0003875`
  query reports **1728** hits and would have silently returned 100).

## Filed upstream (as `suggested_questions`, not as GO actions)

1. **InterPro / GO Central** — split `IPR012108`/`PIRSF016939`, or withdraw its three GO
   mappings.
2. **UniProt (ARBA)** — `ARBA00088955` needs a negative condition excluding the ARH2 clade.
3. **UniProt (nomenclature)** — RecName *"Inactive ADP-ribosyl**transferase** ARH2"* is the
   **wrong reaction direction**; the entry's own `SIMILARITY` says glyco**hydrolase** family.
   Suggest *"Inactive ADP-ribosylhydrolase ARH2"*. This name is what GOA prints in the
   `GENE NAME` column of every ADPRHL1 row.
4. **GOA / Xenbase** — merge or cross-link the Swiss-Prot and TrEMBL copies of *X. laevis*
   `adprhl1` so the 2020 evidence can propagate.
5. **GOA** — `PMID:37880701` should also supply human IMP evidence for `GO:0055003`.

## affinage

`gates_passed: True`, 6 citations, all numeric PMIDs, and **all six verified to concern
ADPRHL1/ARH2** — checked because on the sibling gene ADPRH the same provider returned
citations resolving to different proteins. Its **recall** is the problem: it missed
`PMID:17075046` (the only direct measurement, and *titled for ARH3*), `PMID:36497109` and
`PMID:16278211`. All three were found by searching the **family**, not the symbol.

## Gates

`checkquotes.py` 61/61 clean · `just validate` ✓ Valid **on the tree merged with `origin/main`**
· `cache_lint` exit 0 · terms.csv union asserted four ways (9395 main + my 3 → 9398, nothing
lost either way, nothing invented, no new duplicates, globally sorted) · working tree clean
after validate.

Two committed self-testing scripts, both non-zero on failure:
`catalytic_site_census.py` (reproduction gate + break-tests) and `audit_adprhl1_review.py`
(7 invariants no repo gate covers — duplicate YAML keys, anchors, quotes in `provenance` and
`knowledge_gaps[].provenance` that `checkquotes.py` does not walk, GOA reconciliation,
`supporting_entities` derived from WITH/FROM, `core_functions` both directions, and
prose-vs-action). **Running the break-tests found two real defects in my own guards**: an
unanchored `"56D" in m` substring test that also matched sibling rows, and a `KeyError` that
aborted the prose check on exactly the input it exists to report.

The one standing warning — *"No annotations reference available deep research files"* — is
left standing deliberately. Nothing from the provider is quoted as `supporting_text` anywhere,
per the campaign rule that provider prose must not carry a mechanistic claim; the record is
listed under `references` with its recall failure documented instead.

## Round 2 — all seven suggestions addressed (commit `e94b91f`)

| # | change |
|---|---|
| 1 | `catalytic_site_census.py` docstring was describing the pre-measurement *plan*: corrected to **20 positions** (14 features, ranges expanded) and to *Dictyostelium* at **48.4%**, i.e. an **identity-matched** control, not the ~30% one the design assumed. The PANTHER-panel comment said "distant positive control" and now says identity-matched. **No computed value changed** — `RESULTS.md`/`results.json` regenerate byte-identically. |
| 2 | "20 features" → 14 features spanning **20 positions**, fixed in the YAML, the notes and this body. |
| 3 | `GO:0000287`'s `reason` rewritten to rest on the term's **own provenance** rather than importing the hydrolase result, and now concedes explicitly that Mg binding was never assayed, that 3 of 6 donors survive, that `MARK_AS_OVER_ANNOTATED` was defensible, and that Asp100 is why. |
| 4 | `GO:0055003` dropped from `core_functions`; the child `GO:0055005` carries it. Both stay in `existing_annotations`, where they have different donors. |
| 5 | `locations: sarcomere` removed from the ROCK/focal-adhesion core function — focal adhesions and costameres are not the sarcomere. The description now says why no location is given. |
| 6 | Stated why the same paper's sarcomere result goes to GOA while its FA result becomes a `NEW` row: a quantified four-marker result closed by pharmacological rescue, versus a qualitative supplementary-figure observation of a process already annotated. `GO:0045214` considered and **declined**. |
| 7 | Recorded the caveat that a curator may prefer a Rho-kinase-pathway term, and why this row stays on the measured, rescued phenotype. |

**Unprompted, because it is load-bearing:** the 20 `GO:0003875` IBA recipients are now broken
down **4 reviewed Swiss-Prot / 11 TrEMBL carrying an `ADPRH` gene name / 5 TrEMBL with none**.
All 20 are named "ADP-ribosylhydrolase ARH1" and none is an ARH2, but an unreviewed entry's
*name* is an automatic label, so the gene-name column carries the claim for 15 of 20 and the
remaining 5 rest on node placement — stated rather than smoothed over.

The census self-test's own mutation anchors went stale against the rewritten `GO:0000287`
reason and the **assert-before-mutate guard caught it**, which is exactly what that guard is
for. All gates re-run green: `checkquotes.py` 61/61, `audit_adprhl1_review.py` 0 problems and
`--self-test` passed, `just validate` ✓ Valid, `cache/go/terms.csv` untouched by this round.

## Round 3 — the three round-2 suggestions (commit `da57f25`)

- **Residue numbering.** `GO:0000287` mixed ADPRH and ADPRHL1 numbering in one paragraph, so
  "56" was simultaneously a retained Ser and a lost Asp. Both the summary and the reason now
  declare the scheme and write every position as *ADPRH position → ADPRHL1 residue*.
- **Denominators reconciled.** "3 of 6 donors retained" sat beside a quoted "3 of 4 aspartates
  lost". They agree: the paper's four are Asp55/56/302/304 and it scores **Asp302→Glu304 as
  lost** because the residue is no longer an aspartate, while the donor test scores it
  **retained** because glutamate still contributes a carboxylate oxygen. Same alignment,
  different denominator and criterion — now stated rather than left for the reader.
- **A guard that read as coverage it did not have.** `audit_adprhl1_review.py` check F
  advertised both directions and implemented only `cf_terms − kept` — and this branch had
  just created the unchecked case by dropping `GO:0055003` from `core_functions` while leaving
  it `ACCEPT`. Silence there was the check's *absence*, not its passing. The reverse direction
  is now implemented against an explicit `CORE_FUNCTION_EXEMPT` dict carrying a written reason
  per term, **stale exemptions are reported too**, and the orphaned comment that had drifted
  onto `QUOTE_KEYS` is gone with it.

**Writing the F2/F3 break-test immediately found a defect in the break-test.**
`import audit_adprhl1_review` from inside the script creates a **second module object with its
own globals** when the file runs as `__main__`, so patching that copy left `audit()` reading
the original and both mutations reported nothing. Fixed to mutate this module's global in
place. Found by *running* it, not by reading it.

`RESULTS.md`, `results.json` and `cache/go/terms.csv` are byte-unchanged this round.
Gates green: `checkquotes.py` 61/61, `audit_adprhl1_review.py` 0 problems and `--self-test`
passed, `just validate` ✓ Valid.

## Round 4 — the two round-3 wording notes (commit `f99c539`)

- **A quoted paraphrase is a latent fake quote.** The `GO:0000287` reason had quotation marks
  around a paraphrase of `PMID:27217161` while the verbatim string sat three lines below in
  `supported_by`. No gate can see this: the checkers only validate strings that are actually
  in `supporting_text`. The reason paragraph now contains **no quotation marks at all**, says
  so, and points at where the verbatim sentence lives. A second, unattributed scare-quote went
  with it — a quotation mark in a `reason` field reads as a citation.
- **Numbering scheme.** The paragraph declared ADPRH numbering and then closed on ADPRHL1
  Asp100; the exception is now labelled in place.

No term, action, evidence code, `supporting_text` or computed number changed this round.

---

> [!NOTE]
> **Low Risk**
> New curated review artifacts and local validation scripts only; no application runtime or auth changes. Scientific impact depends on upstream GO/UniProt adoption of the recommendations.
> 
> **Overview**
> Adds a **complete human ADPRHL1 (Q8NDY3 / ARH2) gene review** as `ADPRHL1-ai-review.yaml`, plus reproducible bioinformatics and YAML audit tooling.
> 
> The review **removes three fold-derived catalytic GO rows** (magnesium binding, ADP-ribosylarginine hydrolase, protein de-ADP-ribosylation) after direct inactivity evidence and a catalytic-site census versus ADPRH/ADPRS, and traces them to **InterPro IPR012108** and **ARBA ARBA00088955**. It **accepts** sarcomere localization and cardiac myofibril/chamber ballooning ISS transfers, and **proposes two new annotations**: ventricular cardiac myofibril assembly (from Xenopus TrEMBL donors the ISS pipeline missed) and positive regulation of focal adhesion assembly (human CRISPR knockout, PMID:37880701).
> 
> **`cache/go/terms.csv`** gains three cardiac GO terms used by those proposals (`GO:0003242`, `GO:0055003`, `GO:0055005`). **`catalytic_site_census.py`** maps UniProt ligand sites across the ARH family (cached UniProt JSON, committed `RESULTS.md`). **`audit_adprhl1_review.py`** enforces seven invariants on the review YAML (GOA reconciliation, quote verbatim checks including `provenance`, `core_functions` ↔ actions, prose vs `action`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f99c53920dab92ac7d6604d5e81931c805291158. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Both round-3 notes are fixed in `f99c539`, and I want to state a stopping criterion rather than let this run on.

**What changed:** the quoted paraphrase of `PMID:27217161` in the `GO:0000287` reason lost its quotation marks (the verbatim string is three lines below in `supported_by`, and no gate can see a quoted paraphrase because the checkers only validate strings that are *in* `supporting_text`). A second unattributed scare-quote went with it, so the paragraph now has no quotation marks at all and says so. The closing clause that slipped into ADPRHL1 numbering is labelled in place.

**Where I think the line is.** Rounds 1→2 moved a curation call (the `GO:0000287` reason's basis), a `core_functions` membership, a wrong location, and a real measurement claim. Round 2→3 moved a reported number (the donor/aspartate denominators) and closed a guard that advertised coverage it did not have. Round 3→4 moved **only prose**: no term, action, evidence code, `supporting_text`, `results.json`, `RESULTS.md` or `cache/go/terms.csv` value has changed since `e94b91f`.

So unless something turns up that could **misstate a number, misfire on another machine, or change a verdict**, I'd like to stop here. Specifically I will still fix: any quote that is not verbatim, any count that does not reconcile, any guard that passes vacuously, and any claim whose evidence does not support it. I will not keep re-polishing wording in `summary`/`reason` fields that already say the right thing.

Two things worth recording from your reviews, because both were correct and neither was cosmetic:

1. **The check-F finding was the best catch of the four rounds.** The docstring advertised both directions, only `cf_terms − kept` was implemented, and this branch had *just created* the unchecked case by dropping `GO:0055003` from `core_functions` while leaving it `ACCEPT`. Silence there was the check's absence, not its passing — and writing the break-test for the new direction immediately exposed a second defect in the break-test itself (`import audit_adprhl1_review` from inside a `__main__` script creates a second module object with its own globals, so patching that copy left `audit()` reading the original and both mutations reported nothing). Found by running it, not by reading it.

2. **The quoted-paraphrase note generalises past this PR.** `checkquotes.py`, the reference validator and the schema validator all verify strings that appear in `supporting_text`. A quotation mark inside a `summary` or `reason` is invisible to every one of them, so a paraphrase in quotes there is the cheapest possible way to ship an unfalsifiable citation. Worth a lint over the corpus, not just a fix here.

One thing I'd flag as *not* worth changing, so it doesn't come back: the standing `just validate` warning "No annotations reference available deep research files". Nothing from the affinage record is quoted as `supporting_text` anywhere, deliberately — provider prose must not carry a mechanistic claim, and on the sibling gene ADPRH that same provider returned citations resolving to different proteins entirely. The record is listed under `references` with its recall failure documented (it missed all three sources that decide the molecular-function verdicts, including the only direct measurement, which is titled for a paralogue). Silencing the warning would mean inventing a citation.

PR #2337, opened 2026-07-27
Review human ADNP2: the NAP defect did not cross the paralogue boundary, but one IBD seed carries it

cmungall:
## ADNP2 (Q6IQ32) — the paralogue test for the NAP-peptide defect

ADNP's merged review (#2331) found that **13 of its 19 rat-derived Ensembl-Compara rows rest on experiments with the synthetic octapeptide NAPVSIPQ ("NAP"/davunetide) rather than on the gene product**, one of them on the all-D enantiomer. This PR reviews ADNP2, its paralogue, and asks whether any of that propagated.

### The answer: no — and the reason is structural, not luck

**1. ADNP2 does not contain the peptide.** Human and mouse ADNP2 carry no `NAPVSIPQ` and **not one `NAP` tripeptide** in 1131 / 1165 aa. The three ADNP orthologues are asserted as positive controls in the committed script, so a broken scan raises rather than reporting a clean negative for ADNP2.

**2. Zero of the 19 rat-Compara terms reached ADNP2.** ADNP has 53 GOA rows / 33 terms, ADNP2 has 18 / 8; the intersection with ADNP's rat-Compara block is empty. The mechanism is worth stating because it predicts where this defect class can travel:

- `GO_REF:0000107` (Compara) and `GO_REF:0000024` (ISS) project along **orthologues**. ADNP's donors are rat and mouse *Adnp*; ADNP2's single sequence-similarity donor is `UniProtKB:Q8CHC8` — **mouse Adnp2, its own orthologue**. The paralogy boundary was never crossed.
- The one route spanning both clades is the IBA at `PTHR15740` node `PTN000405125`, whose reach is **44 gene products covering ADNP and ADNP2 orthologues together**. PAINT gave that node **exactly two terms** — `GO:0005634` and `GO:0010468`, both LCA-appropriate. That containment is PAINT working correctly.

Reported as a **non-confirmation**. It is worth as much as a finding: it locates the NAP problem as an *ortholog-projection* defect rather than a family-wide one, so the fix retracts cleanly on the ADNP side.

**3. The qualification, and it is the interesting part.** `GO:0010468 IBA` names two IBD seeds. Asking not "does the seed have its own experimental evidence?" but **"what molecule did that experiment use?"**:

| seed | own experimental evidence | reference title | molecule |
|---|---|---|---|
| mouse Adnp (Q9Z103) | `GO:0000981` IDA, `GO:0006357` IDA, `GO:0010468` **IMP** | ADNP…chromatin / ADNP promotes neural differentiation | gene product |
| rat Adnp (Q9JKL8) | `GO:0010629` IDA | ***NAP mechanisms of neuroprotection*** | **synthetic NAPVSIPQ** |

Separating **provenance** from **circularity**: the provenance is tainted and on its own supports no verdict; the chain is *not* empty of gene-product evidence, because the co-seed carries an IMP and two IDAs on the protein. So `ACCEPT` / `root_cause: NO_FAILURE_CORE` at the row level, `source_status: SOURCE_BAD` on the rat entity alone. That combination is deliberate. Filed as a question to PAINT: **should a seed annotation derived from a synthetic peptide fragment count toward an IBD?**

### Verdicts — 4 ACCEPT, 3 KEEP_AS_NON_CORE, 11 MODIFY, 1 NEW

*(Updated after review round 2: `GO:0006357` moved from `KEEP_AS_NON_CORE` to `MODIFY` → `GO:0000122`, and the two `core_functions` were merged with `GO:0140110` moved to `contributes_to_molecular_function`. See the [reply](https://github.com/ai4curation/ai-gene-review/pull/2337#issuecomment-5088985033).)*

**`GO:0000981` MODIFY → `GO:0140110`.** Its parent `GO:0003700` requires binding *"a specific double-stranded genomic DNA sequence … within a cis-regulatory region"*. All three clauses fail: the authors *"were unable to define a sequence motif"* explaining ADNP2's ChIP-seq; peaks are overwhelmingly repeats with few at TSSs; and recruitment tracks HP1β–H3K9me3, since PxVxL mutation nearly abolishes chromatin binding. The route is TFClass 3.1.8, a homeodomain family — a fold read as an activity. Verified via QuickGO that `GO:0140110` is a genuine ancestor, so this is a generalisation, not a sideways move. **Node reach (added round 2):** `tfclass:3.1.8` covers exactly **14** human genes (ADNP, ADNP2, HOMEZ, TSHZ1/2/3, ZEB1/2, ZFHX2/3/4, ZHX1/2/3) and gives every one the identical `{GO:0000785, GO:0000981}` pair; across the whole `GO_REF:0000113` import (1436 annotations / 727 entities / 100% ISA) **709 get that doublet and 18 get chromatin alone**. Those 18 are the pipeline's own negative control. **Corrected in round 5:** they have no shared biological characterisation — I generalised them twice ("the non-DNA-binding members", then "none is a sequence-specific Pol II transcription factor") and **both were false** (NFX1 *"Binds to the X-box motif of MHC class II genes"*). **Corrected again in round 7 — that was the third false generalisation, and the fix was structural:** the claim now has a single source, computed from the partition, and reads *"They share no property at all. 10 of the 18 sit in nodes whose other members keep the term, but the remaining 8 sit in 4 nodes where NO member keeps it, so not even the structural description holds across the set. What can be said is only per-node, which is why the table above is the claim and this sentence is not."* The per-node table is the claim; no sentence generalises over the set. Which also answers the granularity question that decides whether the ask is coherent. **Corrected in round 6:** the import operates at *both* granularities — of the 11 nodes reaching an excluded entity, **7 withhold the term from a strict subset** while the rest of the node keeps it (10 entities) and **4 withhold it from every member** (8 entities). An earlier revision said "per-entity, not per-node" and rendered only the 7, under a heading about the 18 and with no note that rows were dropped. The kind ADNP2 needs is the strict-subset one, and class 3.1 already supplies it twice: HOPX excluded alone out of 47 members of `tfclass:3.1.3`, HMBOX1 alone out of 19 of `tfclass:3.1.10`, while ADNP2's own `tfclass:3.1.8` stands at 14/14. So the ask needs no new mechanism and **would not touch ADNP**: it takes one node from 14/14 to 13/14. **Corrected in round 4:** of the 18, three carry an annotated `DNA_BIND` feature, and the fold-symmetry precedent is **HMBOX1** (`Homeobox` at 267-341 — the identical note to ADNP2's `Homeobox` at 1043-1102), *not* HOPX, whose note reads `Homeobox; degenerate`. My earlier phrasing said HOPX carried its feature "exactly as ADNP2 does", which my own committed `results.json` contradicts; a curator would rightly have replied that HOPX is excluded *because* its homeodomain is broken. HOPX still carries the biological half (UniProt: *"Atypical homeodomain protein which does not bind DNA"*). The curator ask is concrete — add ADNP2 to that existing 18-member exclusion set — but the **positive** argument for ADNP2 is the measured failure to find a motif, not symmetry of fold with any excluded protein.

**`GO:0003677` KEEP_AS_NON_CORE — same evidence, opposite verdict.** The definition makes **no specificity claim**, and the homeodomain (1043–1102) plus nine C2H2 fingers are intact across four independent signatures. Untested, not refuted. **A homeobox did not establish sequence-specific DNA binding here** — the brief's question, answered the other way from ADNP.

**Nine `GO:0005515` rows MODIFY → `GO:0070087` chromo shadow domain binding**, one verdict per partner (CBX1 ×5, CBX3 ×4), matching the merged ADNP review so the paralogues agree rather than diverging silently. For ADNP2 the CSD mechanism is **demonstrated by mutation**, not inferred: PxVxL point mutants lose HP1β. Note the in vitro reconstitution used **full-length human ADNP2, human HP1β and human CHD4**, so complex assembly is direct evidence on the human protein.

**NEW `GO:0010526`, deliberately *not* its child `GO:0141005`** (silencing *by heterochromatin formation*): *"H3K9me3 levels at ADNP and ADNP2 peaks were not reduced by either individual or combined removal of these factors."* ChAHP2 **reads** the mark, it does not deposit it. The reflexive child would have asserted a mechanism the data exclude. Missing "reader-mediated, assembly-independent silencing" term filed as an ontology gap.

### Row reconciliation

GOA has **18** rows (18 distinct); the `fetch-gene` stub seeded **15**. The three missing are the known `seed_missing_annotations` collapse — WITH/FROM is not in the seeder key, so the three PMIDs reporting *both* CBX1 and CBX3 each collapsed 2 → 1. Restored so every partner gets its own verdict, and asserted programmatically: every entry's `(term, evidence, reference, WITH/FROM)` matches a GOA row exactly. Final file has 19 entries = 18 rows + 1 `NEW`.

### Checks run, including the nulls

- **Logical-opposite cross-product**: negative — no opposed term pair exists. Automated so the null is recorded.
- **Reference-projection**: `GO_REF:0000113` carries 1436 annotations (a bulk classification import). Entity counts derivable for the three small interaction references (9 / 135 / 314); **not** for HuRI (85343) or BioPlex (9514), where the honest statement is that the test is uninformative rather than a number read off one page.
- **Retraction/erratum sweep** over all 15 PMIDs: none flagged.
- **Partner accessions**: both canonical reviewed entries at canonical length — negative for the ACRV1 ORFeome-clone failure mode.
- **PxVxL with its null**: reproduced ADNP's published figures as a precondition (1 hit @820, null 0.758) before reporting ADNP2's. ADNP2 has **2** candidates against an expectation of **2.705** — *below* chance, so the count carries no information. The discriminating test is projective: a BLOSUM62 alignment maps ADNP's motif start onto ADNP2 position **1107**, exactly where ADNP2's `PSVLL` sits. Positionally homologous.

### Two things for the brief

- **A merged UniProt accession is not the quiet zero the brief describes.** `O15507` does *not* 404 and does *not* report `entryType: Inactive` when `fields=` is supplied — UniProt answers HTTP 303 to its successor, urllib follows, and you get a complete `UniProtKB reviewed (Swiss-Prot)` record for **P56159 GFRA1_HUMAN, 465 aa**. Neither the status, nor `entryType`, nor `uniProtkbId` distinguishes it; only `primaryAccession` does. My first two guards passed against it.
- **The pre-write hook blocked 12 valid full-text quotes**, resolving `publications/` against the *other* worktree where `PMID_38960717.md` is a stale 3 KB abstract against this branch's 149 KB full text (which is what `origin/main` carries). Obeying it would have deleted the evidence four verdicts rest on.

### Gates

`just validate` ✓ on the branch **and** on the tree merged with `main`. `checkquotes.py` 52/52. The 7 `knowledge_gaps[].provenance` quotes it does not walk were hand-verified; 49 + 7 = 56 reconciles against the raw count, so no quote is covered by neither. Strict duplicate-key loader clean, no YAML anchors, raw-vs-parsed agrees for `supporting_text` (56/56) and `reference_id` (46/46). `cache_lint` exit 0; `terms.csv` untouched (all terms already cached), union asserted after the merge. Analysis script: 6 break-tested guard directions, and a fresh run reproduces both committed artifacts byte-for-byte.

The one remaining warning ("No annotations reference available deep research files") is left standing deliberately: the review cites the PMIDs affinage surfaced, not affinage's prose.

---

> [!NOTE]
> **Low Risk**
> Gene curation YAML, generated reports, and an isolated analysis script—no changes to shared application runtime, auth, or data pipelines beyond new human ADNP2 artifacts.
> 
> **Overview**
> Adds a **complete human ADNP2 (Q6IQ32) AI gene review** and a reproducible bioinformatics panel that tests whether the ADNP NAP-peptide annotation defect reached the paralogue.
> 
> The review re-adjudicates all 18 GOA rows (restoring partner-distinct `GO:0005515` entries the seeder collapsed) with verdicts such as **MODIFY** `GO:0000981` → `GO:0140110`, **MODIFY** Pol II regulation → **negative** regulation, nine HP1 interaction rows → **`GO:0070087`**, and **NEW** `GO:0010526` transposable element silencing. It frames ChAHP2-centric biology (HP1β/H3K9me3 recruitment, retrotransposon repression) and documents gaps plus a proposed **ChAHP2 complex** GO term.
> 
> **NAP propagation:** ADNP2 lacks the NAP octapeptide; none of ADNP’s rat-Compara terms appear on ADNP2. The `GO:0010468` IBA row is **accepted** but flags rat Adnp’s peptide-derived seed as **SOURCE_BAD** while mouse Adnp supplies gene-product evidence.
> 
> **`analyze_adnp2_propagation.py`** regenerates `results.json` / `RESULTS.md` (NAP scan, PxVxL alignment, IBD donor molecule audit, TFClass `GO_REF:0000113` exclusion-set analysis for the NTNU_SB ask) with extensive self-tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85d8d5c7f2539544ebfffc53739fe251db108224. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — the blocking item was right, and it is the same failure shape you name: a structured slot asserting more than the prose allowed. I verified each checkable premise before acting; all four held.

### 🟡 `core_functions` — fixed by merging, not by moving alone

`molecular_function` does mean "has the activity **independently**", and my own description ("the molecular step by which the bound complex represses is unresolved") plus `knowledge_gaps[0]` ("ADNP2 may be a pure scaffold") contradicted it.

I took your **second** option — merge — rather than only moving the slot, because the schema says that when you use `contributes_to_molecular_function` the `molecular_function` slot "*should then contain the subunit-specific activity*", and **I do not have an honest one to put there.** The two candidates both fail on this gene's own evidence:

- `GO:0140463` chromatin-protein adaptor activity — its definition is bringing a protein together with chromatin "*to establish or maintain the chromatin localization of the protein, or the complex to which it belongs*". That is what **ADNP** does (and the merged ADNP review uses it correctly). For ADNP2 the traffic runs the other way: HP1β confers the chromatin localisation *on* ADNP2, so the term would invert the mechanism.
- `GO:0030674` protein-macromolecule adaptor activity — plausible but **untested**. The published reconstitution mixed ADNP2/HP1β with separately-expressed CHD4, which does not exclude a direct HP1β–CHD4 contact. That missing control is already `suggested_experiments[3]`.

So a second entry would have needed either an empty `molecular_function` or an invented one. The merged entry says exactly what is supported: `molecular_function: GO:0070087` (independently enabled), `contributes_to_molecular_function: GO:0140110` (complex-level), `directly_involved_in: [GO:0010526, GO:0000122]`.

### 🔵 1 — `GO:0006357` → `MODIFY` `GO:0000122`: accepted, and your second argument is the stronger one

Checked three things before taking it:

| check | result |
|---|---|
| `GO:0000122` is_a-descendant of `GO:0006357`? | **yes** — a refinement within the branch, not a sideways move |
| definition matches the evidence? | **yes** — *"stops, prevents, or reduces"*; a de-repression result is directional and the undirected parent discards that |
| redundant with the `NEW GO:0010526`? | **no** — neither subsumes the other (verified both directions), so they are non-redundant statements: one about the polymerase II process, one about the transposon outcome |

The direction is also the *only* one the data show — mouse ES de-repression, and zebrafish adnp2 directly suppressing `ved/vent/vox`; nothing reports ADNP2 activating a Pol II gene. But the sharper half of your argument is provenance: as written the row's only stated antecedent was `GO:0000981`, **the term this PR modifies**, so it was hanging off a premise the same review rejects. Fixed.

### 🔵 2 — off-topic quote: agreed, dropped

The PxVxL line is about HP1 docking and says nothing about the homeodomain. Straightforward violation of "every quote must contain the entity its row is about", in a slot I filled because the citation was available rather than because it supported the claim.

### 🔵 3 — `in_complex`

Now carries an explicit comment pointing at `proposed_new_terms[0]`, so the empty slot reads as a pending ontology dependency rather than an omission.

### 🔵 4 — the `O15507` trap: **declining the issue, with reasoning**

It is already durable and in-repo in two places — `ADNP2-notes.md` and a comment on the guard itself in `analyze_adnp2_propagation.py`, next to the code it protects. Opening an issue from a gene PR fires `dragon-ai.yml` (`issues: [opened, edited]`), which queues ahead of the review jobs this campaign is bottlenecked on. I have relayed it to the campaign brief instead, where it corrects the existing entry: the brief describes a dead accession as *"the quietest possible false negative"* returning nothing, but a **merged** accession is worse than that — HTTP 303 → complete `UniProtKB reviewed (Swiss-Prot)` record for a **different protein** (`O15507` → `P56159` GFRA1_HUMAN, 465 aa). Only `primaryAccession` distinguishes it; status, `entryType` and `uniProtkbId` all look fine. Say the word and I will file it.

---

### New in this push: the TFClass node-reach measurement

Prompted by the brief's *"which node's reach is exactly my gene set, and what did it give them?"*. This turned the curator question from rhetorical into filable.

`tfclass:3.1.8` reaches exactly **14** human gene products — ADNP, ADNP2, HOMEZ, TSHZ1/2/3, ZEB1/2, ZFHX2/3/4, ZHX1/2/3 — and gives **every one the identical `{GO:0000785, GO:0000981}` pair**. So `GO:0000981` is a property of class membership, not a judgement about ADNP2; and two members of that set cannot both be right, since ADNP's is supported by a measured motif and ADNP2's is contradicted by the documented failure to find one.

Widening to the whole import, fully paginated: **`GO_REF:0000113` = 1436 annotations / 727 distinct entities / 100% ISA**, of which **709 receive that doublet and 18 receive chromatin *alone*.** Those 18 are the pipeline's own negative control, and they are exactly the non-DNA-binding members — SMAD6/7, NR0B1/2, NCOA1/2/3, ZFPM1/2, AEBP2, TFDP3, HMBOX1, DMRTC1, NFX1/NFXL1, ZC3H6/8 — and **HOPX**.

HOPX is the precedent that closes it. UniProt: *"Atypical homeodomain protein which does not bind DNA"*. It carries a `DNA_BIND` feature over its homeodomain **exactly as ADNP2 does** (HOPX 3–62, ADNP2 1043–1102), and NTNU_SB still withheld `GO:0000981` while keeping `GO:0000785`. The discrimination already exists inside this import and the criterion is already "does it bind DNA", not "does it have the fold" — so the ask is now concrete: **add ADNP2 to the existing 18-member exclusion set.** This is the interpro2go negative-control method transplanted to TFClass; the entries that behave correctly are what prove the point. It is explicitly *not* a claim that the classification is wrong — placing ADNP2 among the ZF-homeodomain proteins is correct taxonomy, and the same node's `GO:0000785` is right.

### Gates on this head

`just validate` ✓ (one warning left standing deliberately: the review cites the PMIDs affinage surfaced, not affinage's prose). `checkquotes.py` 51/51; the 7 `knowledge_gaps[].provenance` quotes it does not walk were re-verified by hand, and 51 + 7 = 58 reconciles against the raw count, so no quote is covered by neither gate. Both `file:` quotes re-verified verbatim *and* single-line. Strict duplicate-key loader clean, no anchors, raw == parsed (58/58). `cache_lint` exit 0; `terms.csv` untouched — `GO:0000122` was already cached.

One process note, since it cost me two cycles: **two of my own verification probes were wrong in the same sitting, both the same way** — an unanchored substring (`contributes_to_molecular_function`, which also appears in the comment explaining the change) and an under-scoped regex (`^- description: >-$`, which also matches `suggested_experiments`). Neither corrupted anything because both fired as *failures* rather than passing silently, but the lesson is to check structure against the parsed object and reserve regex for genuinely textual facts like a removed quote.

cmungall:
Taken, and the framing is the part that mattered: it was this campaign's own rule turned back on me. The census was the one claim held to a **lower evidentiary standard than the NAP positive control and the PxVxL null sitting immediately beside it** — and it was simultaneously the one claim leaving the repo as a concrete ask to an external curation group. That is the worst possible place for a number in prose. Fixed: *evidence born in the repository.*

### `tfclass_reach()` — section G of `RESULTS.md`, keyed on sets not counts

The point about **diffable sets rather than numbers** is the right design constraint, and it changed what I emitted:

| emitted | why a set and not a count |
|---|---|
| node reach: **14** gene products with symbols **and each member's term set** | shows the uniform doublet directly, and lets a curator see ADNP and ADNP2 side by side |
| import: **1436** annotations / **727** entities / **100% ISA** | entity counts derived as a **distinct set of gene-product ids**, never from the annotation total |
| the **chromatin-only exclusion set, as 18 accessions + symbols** | this is the payload of the ask. A curator can diff 18 accessions; nobody can act on the integer 18 |
| HOPX precedent with **both `DNA_BIND` spans computed** — HOPX 3–62, ADNP2 1043–1102 | the "carries a DNA_BIND feature exactly as ADNP2 does" claim is now measured, not asserted |

**Three assertions make it refuse to report rather than mislead**, each one a way the section could have been quietly wrong:

1. the subject **and** its paralogue must both appear in the node's reach — a census that does not contain the gene it is about is a broken query, not an empty result;
2. **HOPX must actually be in the chromatin-only set**, or the argument from it is simply false;
3. both proteins must actually carry a `DNA_BIND` feature.

Three new self-test directions, one per assertion, with the mutations kept **as fine as the claims** rather than coarser: the control test removes a **single** accession instead of blanking the query, and the precedent test points the check at **ADNP2 itself** — which is in the import but receives the doublet, i.e. exactly the plausible wrong answer a lazier implementation would accept. And the **ninth direction is the happy path**: with nothing broken, the reach must be non-empty, the exclusion set must be non-empty, the entity partition must sum, and both `DNA_BIND` features must be present. A guard that only ever fires on breakage cannot tell you it works when nothing is broken. `9/9 OK`, and a fresh run reproduces `RESULTS.md` and `results.json` byte-for-byte on the merged tree.

The YAML now **cites** those figures from `RESULTS.md` rather than asserting them (three `file:` quotes on the `GO:0000981` row). I extracted them programmatically from single physical lines rather than retyping — the HOPX line contains an em-dash, which is precisely the character that silently breaks a `file:` quote no gate validates.

And section G states in its own text that **no verdict depends on it**: `GO:0000981` → `GO:0140110` rests on the quoted three-clause failure against `GO:0003700`'s definition. That belongs in the artifact, not just in this thread, so a later reader cannot mistake the census for the argument.

### 🔵 `GO:0000122` supporting_text — fixed

`PMID:41822989` added. You were right that the reason cited "mouse and zebrafish" while `supported_by` carried only the mouse quote.

### One thing worth reporting: a numeric discrepancy that was my checker, for the third time on this gene

Verifying the new `file:` quotes, my walker reported **62 raw against 52 parsed**. I did not look for a story that made 10 acceptable — I derived the expected decomposition independently: `Finding` objects carry `supporting_text` but **no `reference_id` of their own**, inheriting the enclosing `Reference`'s id, and there are exactly **10** findings across the references. 52 + 10 = 62; the walker was wrong, not the data.

Fixed to resolve a finding's reference from its enclosing object, after which **all 62 quotes verify** (57 `PMID:`, 5 `file:`, the latter also confirmed single-line). The independent cross-check agrees: `checkquotes.py` covers **55**, the **7** `knowledge_gaps[].provenance` quotes it does not walk make **62**. Two decompositions of the same total, derived separately and agreeing.

### Gates on head `6748e5c7d`

`just validate` ✓ on the branch **and** on the tree merged with `main` (merged forward first, zero conflicted files). `checkquotes.py` 55/55 + 7 hand-verified = 62/62. Strict duplicate-key loader clean, no YAML anchors, raw == parsed. `cache_lint` exit 0. `terms.csv` union asserted against `origin/main` at merge time — 9404 rows / 9404 curies both sides, nothing lost, nothing invented, no new duplicates; working tree clean afterwards.

I have pushed everything I intend to push. **Not merging and not enabling auto-merge**, per instructions.

cmungall:
Took the 🔵, and it was not cosmetic — it was **my own committed artifact contradicting my own prose**, which is the class of defect this campaign spends most of its rounds on. Thank you for catching it.

### The defect, stated plainly

`results.json` recorded HOPX's `DNA_BIND` note as `Homeobox; degenerate` while ADNP2's is plain `Homeobox`. My renderer emitted only the spans, so the disconfirming detail sat in the JSON where nobody would read it, while the prose said HOPX carries its feature *"exactly as ADNP2 does"*. A curator would have replied — correctly — that **HOPX is excluded *because* its homeodomain is broken**, which does not transfer to ADNP2's intact one. That is not a phrasing nit; it would have sunk the filing.

### The fix produced a better precedent than the one I had picked

Rather than just render the note, I measured the whole exclusion set. Of the 18 excluded entities, **3** carry an annotated `DNA_BIND` feature:

| symbol | DNA_BIND | non-degenerate? | note identical to ADNP2's? |
|---|---|---|---|
| TFDP3 | 108-190 | yes | no |
| **HMBOX1** | **267-341 (`Homeobox`)** | **yes** | **yes** |
| HOPX | 3-62 (`Homeobox; degenerate`) | no | no |

ADNP2's own is `1043-1102 (Homeobox)`. So **HMBOX1 is the fold-symmetry precedent** — an *intact* homeodomain, annotated with the **identical note** to ADNP2's, already excluded from `GO:0000981` by this very import. HOPX carries only the **biological** half (UniProt: *"Atypical homeodomain protein which does not bind DNA"*). The two do different work, and the review now says which is which on every surface.

**I had picked the weaker of two available precedents**, and only measuring the set revealed the stronger one. That is the general lesson: a negative-control argument needs the control to be *comparable to the subject*, not merely to exist.

### And your second point, taken explicitly

Every surface now states plainly that the **positive** argument for ADNP2 is **neither precedent** — it is the measured failure to find a motif (no motif explains the ChIP-seq distribution, peaks avoid TSSs, PxVxL mutation nearly abolishes chromatin binding). The exclusion set shows only that this import *has* a mechanism for withholding `GO:0000981` and already applies it to an intact-domain homeodomain protein; it does not itself argue that ADNP2 belongs there. `GO:0000981` → `GO:0140110` still rests solely on the three-clause failure against `GO:0003700`'s definition.

### Two new self-test directions (11 total)

1. **The asymmetry must stay visible in *both* artifacts** — the degeneracy booleans in `results.json` *and* the `DOMAIN` note rendered in `RESULTS.md`. Checked against the **emitted file**, not the builder source, because dropping the note in rendering is precisely how this defect arose: a detector that read the generator could not have seen it.
2. **The fold-symmetry derivation is break-tested** by marking every excluded entity's domain degenerate — a mutation *as fine as the claim*, since it leaves the query, the entity counts and HOPX's own membership all intact — and the section must then refuse to report rather than fall back on HOPX.

### Surface sweep

I swept the withdrawn phrasing by **surface type**, not by file. It had survived in exactly the two the campaign predicts: **`ADNP2-notes.md`** and **the PR body** (fixed via `gh api -X PATCH`, verified by re-reading from the server — `gh pr edit` is broken here). Both corrected. The one remaining occurrence in the notes is the retracted wording **quoted inside its own retraction**, which is the honest way to document a withdrawal rather than a survival.

### Gates on head `51f1ea341`

`just validate` ✓ on the branch **and** on the tree merged with `main` (merged forward, zero conflicted files). Self-test **11/11**; a fresh run reproduces `RESULTS.md` and `results.json` byte-for-byte. All **62** quotes verbatim (57 `PMID:`, 5 `file:`, the latter also confirmed single-line); `checkquotes.py` 55/55 plus the 7 `knowledge_gaps[].provenance` quotes it does not walk = 62, reconciling against the raw count. Strict duplicate-key loader clean, no YAML anchors, raw == parsed. `cache_lint` exit 0; `terms.csv` union asserted against `origin/main` — 9404/9404 both sides, nothing lost, invented or duplicated. Working tree clean.

Also noting for the record: your previous review flagged that `Build and test` was still in progress when you approved. It has since **passed** on that head, and it passes on this one.

Everything I intend to push is pushed. **Not merging and not enabling auto-merge.**

cmungall:
All four taken. Two of them were my defects, and **(1) materially improves the argument** rather than just tidying it.

### (1) The repo already adjudicates HMBOX1 — and it reframes the ask better than I had it

`projects/TRANSCRIPTION_FACTORS/dbTF-discrepancy-analysis.md` lists HMBOX1 among proteins **"correctly excluded"**, reason **"Telomere binding, not gene regulation"**, under the heading **`DBD ≠ dbTF`**. (Verified those fragments are present in that file before writing a claim about it — a cross-document assertion addressed to curators is exactly where opening the record first matters.)

That changes the criterion I should have been citing. HMBOX1 is **not** excluded for failing to bind DNA — **it does bind DNA**, at telomeric repeats — but for not being a *DNA-binding transcription factor*. "Does not bind DNA" was never the right frame, and it also happens to be the frame that would have made my own review internally inconsistent: I **retain** `GO:0003677` for ADNP2 precisely because the folds are intact and unrefuted. `DBD ≠ dbTF` resolves that cleanly and is now stated in the question:

> `GO:0003677` stands because the homeodomain and zinc fingers are intact; `GO:0000981` does not, because what the ChIP-seq refutes is specifically the **sequence-specific polymerase II transcription-factor** reading.

So the ask is now "ADNP2 is a DBD-bearing protein that is not a dbTF, like HMBOX1" rather than "ADNP2 does not bind DNA", which is both true and consistent with the rest of the review. And it pre-empts the reply you flagged — HMBOX1's exclusion is a documented judgement, not a pipeline gap.

### (2) A hand-typed number contradicting my own artifact — one paragraph below the paragraph correcting that defect class

TFDP3 written as `108-341`; **341 is HMBOX1's end**. Committed artifact says `108-190`. Precisely the failure I had just finished writing a correction about, committed in the same file, twelve lines apart.

Rebuilt all three spans **programmatically from `results.json`** with an assertion, and recorded the lesson in the notes in the only form that generalises: *retyping a number you have already computed is the mistake, so the fix is to not retype it.* Not "be more careful."

### (3) Stale claim in the notes, refuted by the table below it

The notes still called the 18 **"exactly the non-DNA-binding members"** while the table twelve lines down shows three of them carrying a `DNA_BIND` feature. Notes-only residue of the round-4 sweep — the YAML was clean, which is itself informative about where sweeps leak. Corrected to what actually unites the set (none is a sequence-specific Pol II transcription factor) with the retraction **stated** rather than silently patched.

### (4) My guard passed vacuously — fifth instance, in a guard I had just written

Self-test 10 wrapped its assertions in `if md.exists()`, so a **missing `RESULTS.md` passed silently** — in the very check written to catch a rendering omission. A guard that cannot see the absence of the thing it polices reports coverage it does not have.

A missing artifact now **fails**, with the command that fixes it. Break-tested in both directions, since a fix to a vacuous pass is worthless unless you show the non-vacuous path:

```
$ mv RESULTS.md /tmp/…   && … --self-test
SELF-TEST FAILURE: …/RESULTS.md is missing, so the rendering check cannot run --
                   generate it with a plain run before self-testing
exit=1
$ mv /tmp/…/RESULTS.md . && … --self-test
self-test: 11/11 directions OK
```

### Gates on head `a1cd80c8d`

`just validate` ✓ on the branch and on the tree merged with `main` (already up to date, zero conflicted files). Self-test **11/11**, break-tested. Fresh run reproduces `RESULTS.md` and `results.json` byte-for-byte. All **62** quotes verbatim (57 `PMID:`, 5 `file:`, the latter single-line). Strict duplicate-key loader clean, no YAML anchors, raw == parsed. `cache_lint` exit 0; `terms.csv` untouched. Working tree clean.

That is everything. **Not merging and not enabling auto-merge.**

cmungall:
Both taken, and the second one caught a claim I had **introduced one round earlier that was false**. Thank you — that is two rounds running where the item was in the direction of my own prose outrunning my own data.

### The generalisation was wrong, twice. So I stopped generalising and measured.

You were right about the counterexamples. UniProt on NFX1: *"Binds to the X-box motif of MHC class II genes and represses their expression"* — sequence-specific binding at a cis-regulatory region. And DMRTC1 is literally named *"Doublesex- and mab-3-related transcription factor C1"*.

So the history is: draft one said the 18 are *"exactly the non-DNA-binding members"* (refuted by my own committed table); draft two replaced it with *"none is a sequence-specific Pol II transcription factor"* — **also false**. That is the ADCK5 pattern precisely: **substituting a second unmeasured generalisation for a retracted one is the contrast without the correction.** Withdrawn on every surface, with the retraction *stated* rather than silently patched.

The measured answer is structural, not biological: the 18 are spread across **11 distinct TFClass nodes** and share no biological property. What they share is that **each is an individually adjudicated exclusion inside a node whose other members keep the term.** A set assembled by a pipeline need not have a biological characterisation at all — and when a generalisation is refuted, the fix is a measurement, not a weaker generalisation.

### Which is also your first item — and it decides whether the ask is coherent

This was the sharper of the two suggestions, because the answer could have invalidated the request. Had `GO_REF:0000113` only been able to withhold `GO:0000981` **per node**, then excluding ADNP2 would also have excluded **ADNP** — whose sequence-specific binding *is* measured — and the correct filing would have been something else entirely.

It comes out in favour. `exclusion_granularity.per_entity_demonstrated = true`:

| excluded entity | node | members | keep the term | excluded |
|---|---|---|---|---|
| **HOPX** | `tfclass:3.1.3` | **47** | 46 | HOPX |
| **HMBOX1** | `tfclass:3.1.10` | **19** | 18 | HMBOX1 |
| TFDP3 | `tfclass:3.3.2` | 11 | 10 | TFDP3 |
| AEBP2 | `tfclass:2.3.2` | 20 | 19 | AEBP2 |
| DMRTC1 | `tfclass:2.5.1` | 8 | 7 | DMRTC1 |
| SMAD6/7 | `tfclass:7.1.1` | 8 | 6 | SMAD6, SMAD7 |
| NCOA1/2/3 | `tfclass:1.2.5` | 22 | 19 | NCOA1, NCOA2, NCOA3 |
| — | **`tfclass:3.1.8`** (ADNP2's) | **14** | **14** | — |

**Single-entity exclusion inside a populated homeodomain node is something this import already performs, twice within class 3.1.** So the ask needs no new mechanism, and it does not touch ADNP: it takes one node from **14/14 to 13/14**. Before this measurement the request was plausible but unvalidated; with it, the precedent transfers. Committed as `tfclass_reach.exclusion_granularity` with the per-entity determination derived (a node counts only if it retains the term for some members while withholding it from a strict subset), not asserted.

### Surface sweep

Both withdrawn generalisations swept across all surface types. The only surviving occurrences are **inside their own retractions**, which is documentation rather than survival. The **PR body** carried one as a live assertion — fixed via `gh api -X PATCH`, verified by re-reading from the server.

### Gates on head `1939f9942`

`just validate` ✓ on the branch and on the tree merged with `main` (merged forward, zero conflicted files). Self-test **11/11**. Fresh run reproduces both artifacts byte-for-byte. All **62** quotes verbatim (57 `PMID:`, 5 `file:`, latter single-line). Strict duplicate-key loader clean, no anchors, raw == parsed. `cache_lint` exit 0; `terms.csv` untouched.

---

### Stopping criterion

Stating this explicitly, since the loop has migrated: **the last three rounds have changed no term, no action, no evidence code and no GOA row.** The counts have been stable at 4 ACCEPT / 3 KEEP_AS_NON_CORE / 11 MODIFY / 1 NEW since round 2, and 18 GOA rows / 19 entries throughout. Rounds 3–5 have been about the artifact behind an upstream ask and about my own prose overrunning my own data — all real, all worth fixing, and now converged: the census is reproducible, the precedent is comparable to the subject, the granularity is measured, and the generalisations are withdrawn.

What I will still fix: anything that misstates a number, misattributes a quote, or would misfire on another machine. What I will not keep iterating on: further refinement of the wording around the NTNU_SB ask, which is now grounded in committed, diffable measurements and says explicitly that no annotation verdict depends on it.

Everything I intend to push is pushed. **Not merging and not enabling auto-merge.**

cmungall:
Taken, and you are right that it is **the same shape a third time** — my prose outrunning my own committed data. But the sentence was the lesser half. The real defect was a **silent display filter**, and that is worth naming precisely because of where it turned up.

### The filter, not the sentence

The renderer iterated a list filtered on `node_members_with_mf > 0`, so it printed **10 rows under a heading about "the 18"** with no statement that anything had been dropped. Ten rows beneath a claim about eighteen is a **stronger misreading than a wrong number**, because nothing cues the reader to check — a wrong number invites arithmetic, a short table invites nothing.

And the location is the interesting part: this repo's no-silent-caps rule was written for **API pagination**, and every guard I had built enforced it there (`numberOfHits == len(results)`, asserted on every query). It bit instead on a **display filter over data that was complete in the JSON all along**. The invariant "if you bound or filter what you show, say what you dropped and why" is not about network calls; I had implemented it as though it were.

Confirmed against `results.json`: `tfclass:0.4.1` (NFX1/NFXL1), `2.1.7` (NR0B1/NR0B2), `2.7.2` (ZFPM1/ZFPM2), `2.8.1` (ZC3H6/ZC3H8) all have `node_members_with_mf = 0` — **8 of the 18 entities**, exactly the omitted rows.

### The fix strengthens the argument, as you said it would

All **11** nodes now print with the kind labelled: **7 strict-subset (10 entities) + 4 whole-node (8 entities) = 18**, stated as an identity.

| node | members | keep | excluded | kind |
|---|---|---|---|---|
| `3.1.3` | 47 | 46 | HOPX | strict subset |
| `1.2.5` | 22 | 19 | NCOA1/2/3 | strict subset |
| `2.3.2` | 20 | 19 | AEBP2 | strict subset |
| `3.1.10` | 19 | 18 | HMBOX1 | strict subset |
| `3.3.2` | 11 | 10 | TFDP3 | strict subset |
| `7.1.1` | 8 | 6 | SMAD6/7 | strict subset |
| `2.5.1` | 8 | 7 | DMRTC1 | strict subset |
| `0.4.1` · `2.1.7` · `2.7.2` · `2.8.1` | 2 each | **0** | NFX1/NFXL1 · NR0B1/2 · ZFPM1/2 · ZC3H6/8 | **whole node** |

So **the import operates at both granularities**, and the ask has to name which one it relies on: the strict-subset kind, which class 3.1 already supplies twice. I have also withdrawn the universal phrasing *"per-entity, not per-node"* from the YAML question **and the PR body** — it was a live assertion in both, and 4 nodes contradict it. The conclusion never rested on it: `per_entity_demonstrated` is an **existential**, so it needs one qualifying node and has seven.

### The three 🔵, all taken

1. **`47` and `19` no longer hardcoded.** Read from the node data in the same sentence whose tail already did — the retyped-number class round 4 fixed in the notes, and, as you note, these are precisely the two figures the YAML quotes to NTNU_SB.
2. **The granularity block now has break-tests** (direction 12). If no node retained the term for any member, `per_entity_demonstrated` must go `False`; on the happy path at least one class-3.1 node must exclude exactly one member, the subject node must still be `3.1.8`, and **the kind partition must cover every node** — which is the invariant the silent filter violated, so the guard now covers the defect that shipped rather than only the one I imagined.
3. **The direction count is derived** from the numbered comments in `self_test()`'s own source. Writing that produced a bug worth recording: splitting on the literal `"def self_test() -> int:"` **matched inside the counter's own source**, so it counted the wrong region and returned `0` — a counter defeated by containing the string it searches for. Fixed with a line-anchored regex taking the last match, and it now fails loudly if it finds fewer than 12 rather than reporting a confident small number.

---

## Stopping criterion

**Done.** Stating this per the campaign's guidance, since the loop has plainly migrated from the curation to the tooling.

**Stable since round 2, across five pushes:** 18 GOA rows / 19 entries; verdicts 4 ACCEPT / 3 KEEP_AS_NON_CORE / 11 MODIFY / 1 NEW; every `supporting_text`; every term id and label. **No round since has changed a term, an action, an evidence code, a quote or a GOA row.** Rounds 3–6 were all about the artifact behind one upstream ask, and about my prose overrunning my data — every finding real, and each fix improved the ask rather than retreating from it.

**Limits I am declaring rather than closing:**
- Every functional claim about ADNP2 rests on mouse ES cells and zebrafish. The human protein has only ever been assayed biochemically. Filed as a `knowledge_gap`, not resolved.
- Whether ADNP2 contacts DNA at all is unmeasured. `GO:0003677` is retained as untested-not-refuted; a binding assay would settle it, and it is in `suggested_experiments`.
- Whether ADNP2 is *required* to bridge HP1β and CHD4, versus merely co-resident, lacks its control. Also a suggested experiment, and the reason I declined to claim an adaptor activity.
- The NTNU_SB ask is grounded in committed, diffable measurements and states in its own text that **no annotation verdict depends on it**. I am not iterating further on its wording.

**What I would still fix:** anything that misstates a number, misattributes a quote, or would misfire on another machine. **What I will not:** further refinement of prose around a request that is already measured, reproducible and explicitly non-load-bearing.

Everything I intend to push is pushed. **Not merging and not enabling auto-merge.**

cmungall:
Both taken, and you are right that patching the sentence a **fourth** time would have been the wrong fix.

### The claim now has one source

The reason it survived was structural, exactly as diagnosed: the claim lived in **two emitters** — a hardcoded string in `render()` and a hand-written paragraph in the notes. I corrected the notes last round and `RESULTS.md` stood, so the paragraph retracting two false generalisations **closed with a third**, eighteen lines below its own refuting table.

`_shared_property_statement()` now computes it once from the same partition the table iterates, asserts that partition covers the whole exclusion set, and **both `results.json` and `RESULTS.md` consume that one string**. The notes quote it verbatim from the emitted artifact rather than restating it. Emitted text:

> They share no property at all. 10 of the 18 sit in nodes whose other members keep the term, but the remaining 8 sit in 4 nodes where NO member keeps it, so not even the structural description holds across the set. What can be said is only per-node, which is why the table above is the claim and this sentence is not.

The rule I am recording, because it is the durable part: **when a claim is wrong for the third time, stop correcting the wording and remove the second place it can be written.** Every hand-written version of this one was false — "the non-DNA-binding members", "none is a sequence-specific Pol II transcription factor", "each sits in a node whose other members keep the term" — and each was refuted by a table in the same document. That is not a proofreading problem.

### Your first suggestion was the serious one, and it was right on both counts

**It was a tautology.** `set(kinds) != nodes` compared two comprehensions over the same list; it could never fail. And worse, it inspected `render()`'s **inputs** while the defect lived in `render()` itself — the inputs were complete all along, only the output was short. **The silent filter would still ship today.**

So the guard changed target rather than tightening. It now reads the **emitted `RESULTS.md`**, parses the node table back out, and asserts the rendered row set equals the node set in the data — with an inline mutation proving it can fail (deleting one table row must change the result).

**Run against the defect that actually shipped**, which is the stronger claim than any self-test:

```
nodes in data                  : 11
rows in 1939f9942:RESULTS.md   :  0   -> check FIRES, missing all 11
rows in HEAD:RESULTS.md        : 11   -> check passes
```

One honest caveat: it reads **0** there rather than 10, because that table keyed its first column on the *entity* rather than the *node*. So the guard is coupled to the table layout — and a layout change makes it **fail loudly rather than pass silently**, which is the safe direction, but it is worth stating rather than claiming the guard reconstructed the exact 10-of-11 shortfall.

**Second suggestion taken:** `same_class` is filtered once and both counted and enumerated from that same list, with an assertion, so the count and the list cannot diverge.

Swept the fourth withdrawn claim across every surface type. Only surviving occurrence is inside its own retraction; the **PR body** carried it as a live assertion and is corrected (`gh api -X PATCH`, verified from the server).

---

## Stopping criterion — done

**Curation stable since round 2, across seven pushes:** 18 GOA rows / 19 entries; 4 ACCEPT / 3 KEEP_AS_NON_CORE / 11 MODIFY / 1 NEW; 62 quotes; every term id and label. **No round since round 2 has changed a term, an action, an evidence code, a quote or a GOA row** — and you have confirmed that independently each round.

**What rounds 3–7 actually were:** one upstream ask acquiring a reproducible artifact, and then four rounds of my prose overrunning that artifact. Every finding was real and every fix improved the ask. The last one is now closed at the level that ends the class: the claim is single-sourced and computed, and the guard reads the file that ships.

**Limits I am declaring rather than closing:**
- Every functional claim about ADNP2 rests on mouse ES cells and zebrafish; the human protein has only been assayed biochemically. `knowledge_gaps`, not resolved.
- Whether ADNP2 contacts DNA at all is unmeasured. `GO:0003677` retained as untested-not-refuted.
- Whether ADNP2 is *required* to bridge HP1β and CHD4 lacks its control — which is why no adaptor activity is claimed.
- The row-completeness guard is coupled to the rendered table's layout, as noted above. I am declaring that rather than generalising the parser.
- The NTNU_SB ask is measured, diffable, and states in its own text that **no annotation verdict depends on it.**

**What I would still fix:** anything that misstates a number, misattributes a quote, or would misfire on another machine. **What I will not:** further prose refinement around a request that is already computed, reproducible and explicitly non-load-bearing.

Gates on head `d2c97f35a`: `just validate` ✓ on the branch and merged with `main`; self-test **12/12** plus the shipped-defect run; artifacts reproduce byte-for-byte; 62/62 quotes verbatim; `cache_lint` 0; `terms.csv` untouched.

Everything I intend to push is pushed. **Not merging and not enabling auto-merge.**

cmungall:
Taken — and I want to be explicit about why, since you correctly placed it **outside** the stopping criterion I declared.

**One fact changes the judgement: this script's inputs are live QuickGO queries.** So "true today" is not a property of the code, it is a property of today's TFClass assignment. In a static artifact I would have left it, exactly as you framed it. Here, a node reassignment silently misdescribes the sentence carrying the ask's precedent — which lands inside the *"would misfire on another machine"* half of what I said I would still fix.

### The fix, and it is not the narrowing you'd expect

Round 7 closed the count/list divergence by **narrowing the count** to single-exclusion nodes. You spotted that this left the **label** describing the unfiltered set — so a same-class node excluding two members would have dropped out of the count, the enumeration *and* the label together, with nothing said. The silent-filter shape one level down.

Narrowing again would have repeated the mistake. Instead:

1. **The label names its own filter** — "N of the **single-exclusion** strict-subset nodes".
2. **Anything the filter drops is disclosed**, rather than the filter being tightened until nothing is dropped.
3. **Removed `assert len(phrases) == len(same_class)`**, which had become true by construction and was therefore inert — your second point.

Break-tested in both directions, with a mutation as fine as the claim (it alters only one node's excluded-member count):

```
injected 3.1.10 excluding 2  ->  "1 of the **single-exclusion** strict-subset nodes sit in the
                                  same TFClass class as ADNP2 (`tfclass:3.1.10` excludes 2 and
                                  is not counted here). **HOPX excluded alone out of 47 …**"
unmutated                    ->  no disclosure text present at all
```

So the disclosure fires exactly when something is dropped, and is **not boilerplate** — which is the direction that usually goes untested.

### Curation untouched, as before

`ADNP2-ai-review.yaml` is byte-identical: 18 GOA rows / 19 entries, 4 ACCEPT / 3 KEEP_AS_NON_CORE / 11 MODIFY / 1 NEW, 62 quotes, and all five `file:` quotes still land on `RESULTS.md`. Gates on head `85d8d5c7f`: `just validate` ✓ on the branch and merged with `main`; self-test 12/12; artifacts reproduce byte-for-byte; `cache_lint` 0; `terms.csv` untouched.

### Stopping criterion, restated and now genuinely closed

This is my **last push**. The criterion I declared stands, and I have applied it rather than bent it: I took this item because the live-input argument put it inside the "would misfire" clause, and I am declaring the remaining limits rather than closing them —

- functional evidence is mouse ES / zebrafish only; the human protein has been assayed biochemically alone;
- whether ADNP2 contacts DNA is unmeasured (`GO:0003677` retained as untested-not-refuted);
- whether ADNP2 is *required* to bridge HP1β and CHD4 lacks its control, which is why no adaptor activity is claimed;
- the row-completeness guard is coupled to the rendered table layout (fails loudly, not silently);
- the NTNU_SB ask is computed, diffable, and states in its own text that **no annotation verdict depends on it**.

**No term, action, evidence code, quote or GOA row has changed since round 2** — eight rounds, all of them either giving one upstream ask a reproducible artifact or correcting my prose against that artifact. Thank you for the four successive catches on the same claim class; the fourth is what forced the structural fix that ended it, and the "give the claim a single source" framing is the part I'll carry forward.

**Not merging and not enabling auto-merge.**

PR #2338, opened 2026-07-27
Review human ADTRP (Q96IZ2): a mechanism/substrate category error at a pan-eukaryotic PAINT node, and a monocyte phenotype annotated to a lymphocyte term

cmungall:
## ADTRP (Q96IZ2) — androgen-dependent TFPI-regulating protein

Accession verified independently: `projects/paint/human-no-IBA-simple.csv:409` reads
`human,Q96IZ2,ADTRP`; UniProt confirms `Q96IZ2` = `ADTRP_HUMAN`, reviewed, 230 aa.

**The worklist's "no-IBA" name is stale for this gene — GOA carries three IBA rows**
(`GO:0016787`, `GO:0042758`, `GO:0005901`), two placed 2025-11-27 and one 2026-05-28.

### Row-count reconciliation (done before reviewing)

| | count |
|---|---|
| GOA TSV data lines / distinct | 27 / 27 |
| `fetch-gene` stub entries | **26** |
| final review entries | **27** |

The stub collapsed the two `GO:0005515` rows, which differ only in WITH/FROM partner — the
documented `seed_missing_annotations` behaviour. Both restored so TMED8 and CMTM7 each get a
verdict.

### The "contested enzymatic identity" is not a conflation in GOA

| claim | aspect | terms |
|---|---|---|
| hydrolyses FAHFAs | molecular function | `GO:0120573`, `GO:0016787` |
| regulates TFPI expression | biological process | `GO:0010628`, `GO:0030195` |

"Regulates the expression of TFPI" is nowhere an MF, and the vascular/thrombosis associations are
BP-only. **Function-is-not-phenotype comes back negative.**

Granularity is already optimal and reaction-anchored, the ADPRH/RHEA:14885 situation:
`GO:0120573 FAHFA hydrolase activity` (created 2026-03-14) carries 12 RHEA xrefs, UniProt curates
12 catalytic-activity lines all `ECO:0000269|PubMed:27018888`, and the IEA row lists 11 of them.

**Direction and substrate identity, checked not assumed:** the term definition and UniProt's
`PhysiologicalDirection=left-to-right` both give hydrolysis, and ChEBI classifies the substrate
9-PAHSA(1-) (`CHEBI:83670`) as *"A long-chain fatty acid anion"* — so `GO:0042758` names the right
chemistry rather than being an inversion. Mechanism is citable from the feature table:
`FT SITE 47` and `FT SITE 131` "Important for catalytic activity", `ECO:0000269|PubMed:27018888`,
= Thr47 and His131, both inside `TRANSMEM` helices.

### Main finding: a mechanism/substrate category error at one PAINT node

`PTN001659973` is scoped **taxon:2759 Eukaryota**, seeded by only human ADTRP + human AIG1, and
reaches **86 gene products** (40 Vertebrata, 25 invertebrate Metazoa, 14 Fungi, 5 Viridiplantae,
2 other Eukaryota) — carrying the *root* of the hydrolase branch as MF and a four-step-deep BP.

**I first blamed clade heterogeneity. Measurement refuted that, and the corrected version is
sharper.** Aligning all 85 other recipients onto ADTRP's own annotated SITE positions:

| clade | dyad Thr/His intact | of which ≥25% identity |
|---|---|---|
| Vertebrata | 39/39 | 39/39 |
| Metazoa (invertebrate) | 17/25 | 14/17 |
| Fungi | 11/14 | **0/0** |
| Viridiplantae | 4/5 | 1/1 |
| other Eukaryota | 2/2 | 2/2 |

Positive control: AIG1, catalytic residues independently annotated Thr43/His134, scores
dyad-intact at 36.5% identity. The dyad is **broadly conserved (73/85)**, and all 14 fungal
members fall below 25% identity so are **undetermined, not negative**.

> A conserved catalytic dyad licenses a claim about catalytic **mechanism**, not about
> **substrate**. `GO:0016787` states mechanism only and is therefore well founded family-wide;
> `GO:0042758` is a **substrate-level** claim, and the substrate is established only for the four
> characterised animal members.

The fix is cheap and asymmetric: move the process term to the subclade where the seeds sit —
PANTHER already has `PTN002591065` at Euteleostomi for `GO:0005901`. Filed as a PAINT question.
Because the MF is well founded at the node, its row is `KEEP_AS_NON_CORE` (true but redundant
against `GO:0120573` on this gene), not `MODIFY`.

**Reciprocal node question, benign answer (reported as a negative):** `PTN002591065` reaches
exactly 25 gene products, all vertebrate ADTRP orthologues, AIG1 correctly excluded; caveolae are
a vertebrate structure so scope and term agree. The one reservation is the relation — the seed's
evidence is `located_in`, the propagation asserts `is_active_in`.

### `MODIFY`: `GO:2000402` is the wrong leukocyte lineage

`PMID:28341552` cell-type words in the cached abstract: **monocyte 4, lymphocyte 0, leukocyte 0**;
the phenotype is *"transendothelial migration of monocytes"*. Both closures fetched:

| relation | holds |
|---|---|
| `GO:2000402` (lymphocyte) ancestor of `GO:2000438` (monocyte) | False |
| `GO:2000438` ancestor of `GO:2000402` | False |
| verified common ancestors | `GO:0071676`, `GO:0002686` |

**Neither contains the other → the term is wrong, not imprecise.** → `GO:2000438 negative
regulation of monocyte extravasation`. Full text is not cached, so if it does contain a lymphocyte
assay the conservative answer is `GO:0071676` (verified ancestor of both). Nothing is lost either
way — `GO:0002686` is annotated separately from the same paper. Raised as a question to BHF-UCL
rather than asserted as curator error.

### The three non-PAINT routes — all three non-confirmations, each with its reason

1. **InterPro2GO.** The signature supplying `GO:0016020` is `IPR006838`, named **"ADTRP/AIG1"** —
   family-specific, not a bare fold — and interpro2go maps it to `GO:0016020` **and nothing else**,
   no activity, even though the family is now a characterised hydrolase family. Of 5788 family
   proteins, **8 are reviewed (Swiss-Prot)**: 4 ADTRP, 2 AIG1, **2 uncharacterised UPF0641**. So a
   quarter of its curated members are uncharacterised and it still withholds the activity —
   restraint, the opposite of the `IPR012108` case.
2. **Bulk TAS imports.** **No TAS row exists.** Projection test, fully paginated: `PMID:27018888`
   8/2 entities, `PMID:21868574` 10/2, `PMID:28341552` 22/4. `PMID:32296183` has **85343**
   annotations, recorded as entity-count **UNAVAILABLE** rather than approximated from one page.
   `PMID:27018888` does give both entities identical term sets, but the paper individually mutated
   and assayed **both** proteins — parallel curation, not a projection.
3. **ARBA rules.** No WITH/FROM names an `ARBA…` rule; the automatic routes are `GO_REF:0000044`
   (naming `SL-0039`) and `GO_REF:0000116` (naming 11 RHEA ids). Neither grants activity on a name.

### Measured-and-absent vs unmeasured, kept apart

**No `REMOVE` is earned.** Measured-and-absent (no activity against phospholipids, lyso-, mono-,
di-, triglycerides) is correctly *not* annotated. The real doubts — is catalysis in caveolae? does
the TFPI effect need the active site? — are **unmeasured**, so `KEEP_AS_NON_CORE` plus experiments.

### Other items

- **`GO:0005515` ×2 → `MARK_AS_OVER_ANNOTATED`.** IntAct: one HuRI screen logged as three
  sub-methods per partner (array + prey pooling + validated two hybrid) — what `NbExp=3` counts.
  Both partners resolve to **reviewed canonical Swiss-Prot entries with matching lengths**
  (negative result, no ORFeome substitution).
- **`GO:0005886` EXP from `PMID:27018888` → `ACCEPT` on the term, flagged upstream.** That cached
  full text has **0** occurrences of `plasma membrane`/`cell surface`/`confocal`/`immunofluoresc`
  against positive controls `membrane lysates` 9, `transmembrane` 29, `HEK293T` 36. It supports
  `GO:0016020` (separately cited by IDA), not the specific compartment. UniProt's SUBCELLULAR
  LOCATION line cites both papers and GOA split it. Filed as a UniProt correction; **no GO action**.
- **Ontology gap.** GO has the MF `GO:0120573` but **no FAHFA-level process** — which is why a
  pan-eukaryotic node reaches for `GO:0042758`. One `proposed_new_terms` entry, with the
  counter-argument recorded rather than suppressed.
- **Logical-opposite cross-product: negative** (8 regulation terms, 0 opposed pairs).
- **Sibling check reported as a null:** `genes/human/AIG1/` does not exist on main, so no merged
  review to compare. AIG1 carries the **byte-identical** `GO:0016787`/`GO:0042758` IBA rows from
  the same node — it is the obvious next gene.

### affinage

`gates_passed: True`, 12 citations. Recall verified rather than assumed and it did **well** here:
all 12 numeric PMIDs, all 12 genuinely about ADTRP (no symbol collision), none retracted or
erratum-flagged, and it surfaced two papers absent from GOA that this review uses
(`PMID:32152231` in vivo knockout, `PMID:32445923` POU1F1). Its own GO grounding block is wrong
(`GO:0140098` catalytic activity acting on RNA, for a lipid hydrolase) and was not used; no
affinage sentence carries a mechanistic claim.

### Verification

`just validate` ✓ Valid on the branch **and on the branch merged into main**; `checkquotes.py`
45/45 clean, with the 4 `knowledge_gaps[].provenance[]` quotes hand-verified (known blind spot);
raw-vs-parsed `reference_id` reconciles exactly 49 = 49 under a duplicate-key-rejecting loader,
0 anchors/aliases; `cache_lint` exit 0; `terms.csv` adds exactly the 2 curies this gene needs with
nothing lost and no duplicates. Both scripts under `ADTRP-bioinformatics/` are committed, the
analysis reproduces `RESULTS.md` on a fresh run, and all five builder guards were break-tested
with each mutation asserted non-no-op — including a verbatim, correctly-attributed quote about the
wrong subject, which the reference validator passes.

---

> [!NOTE]
> **Low Risk**
> Annotation and curation artifacts only; no runtime application logic. Large YAML and scripts affect review data quality, not production services.
> 
> **Overview**
> Adds a **complete GO annotation review** for human **ADTRP** (`Q96IZ2`), including `ADTRP-ai-review.yaml`, reproducible bioinformatics (`analyze_adtrp_propagation.py`, `build_review.py`, `RESULTS.md`), and small **GO term cache** updates for `GO:0120573` (FAHFA hydrolase activity) and `GO:0042758` (long-chain fatty acid catabolic process).
> 
> The review centers on **FAHFA hydrolase activity** as the core molecular function and flags a **PAINT mechanism vs. substrate mismatch** at pan-eukaryotic node `PTN001659973`: conserved Thr/His dyad supports general hydrolase MF, while **long-chain fatty acid catabolic process** is argued to over-reach substrate claims to non-animal recipients. It proposes **`MODIFY`** of `GO:2000402` (lymphocyte migration) to monocyte extravasation from `PMID:28341552`, marks HuRI **protein binding** rows as over-annotated, and records PAINT/UniProt/BHF-UCL **follow-up questions**, ontology gaps, and suggested experiments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b9caa89977a56e44c68fa195073276b1cc2df6e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Both blocking items fixed, and thank you — item 1 was more interesting than it looked.

**1. `PMID:32152231` ×4 — fixed at the generator.** You were right that it was an accumulator bug.
The specific mechanism: `build_review.py` loads **its own previous output** as the starting document
and then appended the extra references, so every re-run appended again. I had run it four times.
Replaced the appends with a **merge keyed on reference id**; three consecutive runs are now
byte-identical, which they were not before. References 14 → **11**, matching your count.

Worth recording because it bit me: **the uniqueness assertion I wrote first was vacuous.** I
asserted `len(ids) == len(set(ids))` *after* building the list from a dict — where duplicates are
impossible by construction — so it could never fire. I only found that because the break-test
failed: the mutation that re-introduced the bug shape passed cleanly, rc=0. That is the
"unreachable check that reads as coverage" mode, which is worse than no check. Restructured into
two things that can actually fail:

- a **detector over the loaded input**, which is where duplicates arrive. Run against the
  previously pushed commit it reports exactly `['PMID:32152231']`.
- the **post-condition extracted into `_finalise_references()`** so it has a seam. Break-tested by
  appending one duplicate right before the check: fires, rc=1.

**2. `source_status` on the `GO:0042758` IBA row — fixed.** You are right that the structured field
contradicted the action, the `root_cause` and its own comment, and that the field is what a
propagation-failure query would read. Now `SUPPORTS_TRANSFER` on all three source entities, with the
46-non-vertebrate-recipient concern staying in the `comment` and in `suggested_questions`.
`SUPPORTS_SOURCE_BUT_NOT_TARGET` no longer appears anywhere in the file.

**Adopted:** the truncated mid-word provenance quote now carries the full sentence, which is more
probative for the gap as you said. `GO:0016020` dropped from `core_functions.locations`.

**Adopted but not as framed — one correction.** You suggested a note that PANTHER renumbered. I
checked, and **it did not**: both families are live and distinct.

| family | name | proteins | InterPro |
|---|---|---|---|
| `PTHR10989` | ANDROGEN-INDUCED PROTEIN 1-RELATED | 5163 | `IPR006838` |
| `PTHR12242` | OS02G0130600 PROTEIN-RELATED | 6117 | none |

So the 2016 paper points at a **sister set of AIG1/ADTRP-like proteins in a separate family**, not
at the recipients of the node under review — which is exactly why that sentence was treated as a
lead to go and measure rather than as an answer. Added as a committed check
(`check_panther_family_identity`) so it is reproducible rather than asserted from a conversation.

**Declined, with reasons:**

- **`MODIFY` on `GO:0005886` EXP / `PMID:27018888` (your #6).** There is a mechanical blocker as
  well as a judgement one: the repo validator requires all rows on one term to share an action, and
  the other two `GO:0005886` rows are correctly `ACCEPT`, so a lone `MODIFY` would trip
  same-term-same-action while the structured field asserted the gene is *not* at the plasma
  membrane. But your real point — that the summary read like a `MODIFY` — was fair, so I **reworded
  the summary to lead with the acceptance** and carry the attribution caveat second. The concern
  stays where it is actionable, as a UniProt correction request.
- **`GO:0071676` as a second `proposed_replacement_terms` entry (your #7).** Two replacement terms
  would make the machine-readable intent ambiguous rather than conservative. The fallback stays in
  the `reason` and in the `suggested_question` addressed to BHF-UCL, who are the people who would
  apply it.

**One more of the same class, found by running your item 2 as a general check over the whole file.**
`GO:0009986 cell surface` was `ACCEPT` — which the schema defines as retaining the annotation *as
representing the core function* — while its reason hedges ("retained on curator authority… the
cached abstract does not state it") and the term is deliberately absent from
`core_functions.locations`. Now `KEEP_AS_NON_CORE`. Unlike `GO:0016020`, cell surface is **not** an
ancestor of plasma membrane (verified, separate branch), so it is an independent claim, and the
catalytic residues sit inside the bilayer rather than on the external face. After the change the
only `ACCEPT` term absent from `core_functions` is `GO:0016020`, a verified ancestor of a core
location — which is your #5 rationale, so it is coherent by construction.

Gates on this head: `just validate human ADTRP` ✓ Valid, `checkquotes.py` 45/45 clean with the 4
`knowledge_gaps[].provenance[]` quotes hand-verified, references 11/11 distinct, `cache_lint`
exit 0, no `cache/**` change this round, and `RESULTS.md` reproduces on a fresh run.

cmungall:
Fixed, and the fix is structural because two attempts at a guard for it did not work.

**Restored.** `PMID:32445923` and the affinage record carry their `reference_review` again, verbatim
from the round-1 commit, each on its correct reference — the **POU1F1-is-the-DNA-binding-protein**
note and the **affinage `GO:0140098` caveat**. Re-read out of the tree: 11/11 references, 0
non-`GO_REF` references lacking a `reference_review`, and both restored claims verified present on
the right ids (not swapped).

**You were right that the mechanism matters less than what was lost, but the mechanism has a sting
in it.** My first guard for this — "assert every non-`GO_REF` reference has a `reference_review`" —
**passed, and was unreachable.** The builder loads its own previous output, so the payload was being
silently carried over from disk; the assertion could never fail. Its break-test therefore also
failed to fire, and an earlier mutation had failed for a different wrong reason (a `NameError`
before the check was reached — a mutation coarser than the claim it was certifying).

That is three forms of one predicate across three rounds, so I changed the **shape** instead of
adding a fourth assertion: **loading now strips any inherited `reference_review`**, so reviewer
judgement can only come from one place, the `reviews` dict. That makes this class of loss impossible
by construction *and* makes the assertion load-bearing.

Break-tested in both directions after the restructure:

- drop one `reviews` entry, ids intact and code valid → fires:
  `non-GO_REF references lacking reference_review: ['PMID:32445923']`
- control run on the corrected file → clean, rc=0 (a check can be wrong about success as easily as
  about failure)
- run against the **shipped defect** (`ec41395a1`) → reports exactly
  `['PMID:32445923', 'file:human/ADTRP/ADTRP-deep-research-affinage.md']`

And the point that generalises, which I have written into the notes: **an id-level check cannot see
payload loss.** On the defective file `references` was 11/11 distinct, the duplicate detector was
clean, `checkquotes` was clean and `just validate` was clean. The two checks are not redundant —
they see different things, and only the second one saw this.

**Stopping criterion.** Three rounds in, I think this should be the last. The curation has been
stable since round 0: no term, action, evidence code, qualifier, quote or reported number has
changed in rounds 1 or 2. Both substantive findings — the mechanism-vs-substrate asymmetry at
`PTN001659973` and the `GO:2000402` → `GO:2000438` lineage correction — you have independently
confirmed, and the one first-pass error in the science (the clade-heterogeneity argument) was found
and withdrawn by measurement before the first push.

Everything since has been about the *harness*: an accumulator in a generator, an enum value that
contradicted its neighbours, and a guard that could not fire. Those were all worth fixing, and I
have fixed them. What I will still fix in a further round: anything that changes a term, an action,
an evidence code, a quote, or a number, or that would misfire on another machine. What I will not
keep iterating on: further refinements to the builder's own assertions, which is where the last two
rounds went. If you see a remaining item in the first category, name it and I will take it.

Gates on this head: `just validate human ADTRP` ✓ Valid, on the branch **and merged into `main`**
(merged `origin/main` forward, zero conflicted files); `checkquotes.py` 45/45 clean with the 4
`knowledge_gaps[].provenance[]` quotes hand-verified; references 11/11 distinct and 7/7 non-`GO_REF`
reviewed; builder idempotent across three consecutive runs; `cache_lint` exit 0; no `cache/**`
change in this round.

cmungall:
Folded in your 🔵 rather than leaving it, because it is the same prose-contradicts-action class as a
round-1 blocking item and it would have shipped to `main` on a curation deliverable.

`GO:0009986`'s reason: **"Accepted with that limitation stated" → "Retained with that limitation
stated"**. One-line diff (`+1/-1`), no term, action, evidence code, qualifier, `source_status`, quote
or number touched.

I swept the **class rather than the instance**, over both `summary` **and** `reason` on all 27 rows —
the usual sweep only checks the summary opener, which is exactly why this one survived in a `reason`.
Result: one genuine contradiction, now zero.

Two mentions were deliberately left alone because they are correct prose, not defects:

- `GO:0016787` — *"The action is KEEP_AS_NON_CORE rather than ACCEPT only because it should not be
  counted as this gene's core molecular function."*
- `GO:0005901` — *"Non-core rather than accepted-as-core so that this term matches the IBA row on the
  same term…"*

Both are explanatory "X rather than Y" contrasts about the row's *own* action, which is the clearest
way to write that reasoning. A guard that forbade them would penalise correct writing and get
circumvented rather than obeyed, so the exemption is deliberate rather than an oversight — and I did
not commit a lint for this, since a fixed-phrase matcher cannot distinguish the two cases and adding
one would be exactly the harness churn I said I would stop.

**This is my last push.** The stopping criterion I stated held: rounds 1–3 moved zero curation
content, and this round moved one word of prose to match an action. Both substantive findings stand
as you independently confirmed them — the mechanism-vs-substrate asymmetry at `PTN001659973` and the
`GO:2000402` → `GO:2000438` lineage correction.

Gates on this head (`8b9caa899`): `just validate human ADTRP` ✓ Valid, branch is level with
`origin/main` so the merged tree is the same tree; `checkquotes.py` 45/45 clean with the 4
`knowledge_gaps[].provenance[]` quotes hand-verified; references 11/11 distinct and 7/7 non-`GO_REF`
reviewed; builder idempotent; `cache_lint` exit 0; no `cache/**` change in this round.

Not merging and not enabling auto-merge, per instructions.

PR #2339, opened 2026-07-27
Review human ADIRF (Q15847): a gene absent from the mouse lineage, annotated entirely from ectopic expression in mouse cells

cmungall:
### Identity check (the worklist name has been superseded)

| source | value |
|---|---|
| worklist `projects/paint/human-no-IBA-simple.csv:5942` | `human,Q15847,ADIRF` |
| HGNC | `HGNC:24043`, approved symbol **ADIRF**, 10q23.2 |
| HGNC `prev_symbol` | `C10orf116` (changed 2013-02-20) |
| HGNC `alias_symbol` | `APM2`, `AFRO` — never approved symbols |
| UniProt | `Q15847` / `ADIRF_HUMAN`, no secondary accessions, Swiss-Prot, PE 1 |

**ADIRF is current** and all three sources agree. `ADIRF-AS1` is a *different gene* (antisense
lncRNA at the same locus) carrying 10 of the 37 `ADIRF[tiab]` PubMed hits; none of ADIRF's GOA
rows cites it, so the locus/transcript hazard was checked and is **absent** from the GO record.

**The worklist's "no-IBA" name is stale here: ADIRF carries two IBA rows.**

### Row reconciliation, done before reviewing

`wc -l` on the GOA TSV = 12 (11 data rows + header); `grep -c '^- term:'` on the stub = 11;
QuickGO for Q15847 = 11 annotations. **11 = 11 = 11**, no stub collapse. `existing_annotations`
has 12 entries: the 11 GOA rows plus one `NEW` proposal.

### Verdicts

| # | term | evid | reference | action |
|---|---|---|---|---|
| 1 | GO:0005634 nucleus (`is_active_in`) | IBA | GO_REF:0000033 | ACCEPT |
| 2 | GO:0045600 pos. reg. fat cell differentiation | IBA | GO_REF:0000033 | ACCEPT |
| 3 | GO:0005634 nucleus (`located_in`) | IEA | GO_REF:0000120 | ACCEPT (circular) |
| 4 | GO:0045600 | IEA | GO_REF:0000002 | ACCEPT (circular) |
| 5 | GO:0005515 protein binding | IPI | PMID:32296183 | MARK_AS_OVER_ANNOTATED |
| 6 | GO:0005654 nucleoplasm | IDA | GO_REF:0000052 | ACCEPT |
| 7 | GO:0070062 extracellular exosome | HDA | PMID:23533145 | KEEP_AS_NON_CORE |
| 8 | GO:0070062 extracellular exosome | HDA | PMID:19056867 | KEEP_AS_NON_CORE |
| 9 | GO:0005634 nucleus | IDA | PMID:23239344 | ACCEPT |
| 10 | GO:0045600 | IDA | PMID:23239344 | ACCEPT (code should be IMP) |
| 11 | GO:0045944 pos. reg. transcription by Pol II | IDA | PMID:23239344 | KEEP_AS_NON_CORE |
| 12 | GO:0005829 cytosol | IDA | GO_REF:0000052 | **NEW** |

Nothing removed. 7 ACCEPT / 3 KEEP_AS_NON_CORE / 1 MARK_AS_OVER_ANNOTATED / 1 NEW.

### The decisive fact, and it is in none of the sources the pipeline supplies

**ADIRF is absent from the entire Muroidea clade, and every functional experiment on it is
ectopic expression in mouse 3T3-L1 preadipocytes.**

Measured with positive controls, so a zero cannot be a rejected query:

| query | hits |
|---|---|
| mouse ADIRF / rat ADIRF / Muroidea ADIRF | **0 / 0 / 0** |
| mouse ADIPOQ / mouse LEP / Muroidea ADIPOQ — CONTROLS | 1 / 1 / 30 |
| **Sciuridae ADIRF** — negative control that localises the loss | **9** |
| teleost / avian ADIRF *(symbol count — context only, see below)* | 103 / 104 |

`PMID:31945134` supplies the mechanism (a 43-kb deletion removing the promoter and first exon)
and places it in the ancestor of *mouse, rat and hamster* — converging independently on the
same clade my census found. That paper is **absent from GOA, from UniProt's reference list and
from the affinage record**, and it is the single most consequential fact about this gene.

**What this does and does not do.** It does *not* make `GO:0045600` wrong — the human protein
really did promote adipogenic differentiation. It bounds the interpretation: gain-of-function in
a background with no endogenous orthologue, so murine adipogenesis proceeds without any ADIRF
and nothing establishes that ADIRF is *required*. Row 10 is therefore **ACCEPT** with the
requirement claim filed as a `knowledge_gap`, not REMOVE or MODIFY. It also explains the
darkness: the standard model organism cannot be used, and the gene has no paralogue (confirmed —
the only other human `PTHR39227` entry, `Q5TBU5`, is an unreviewed TrEMBL duplicate with the
byte-identical 76-aa sequence).

*Where I decline to follow the paper — corrected in review round 2.* It also states ADIRF is
absent from zebrafish. I first declined that on the NCBI symbol count above, which was the
**wrong instrument**: a symbol/alias count is orthology already asserted by a pipeline, not
measured, and the only teleost sequence the UniProt family offers (carp `A0A8C1JCC4`, 938 aa)
lands in my own spurious bin at 51/76 and 19.7%. Settled now on sequence in `RESULTS.md`
section F: `NP_001373520.1` (*Danio rerio*) and `XP_085644419.1` (*Trachurus japonicus*) are
both 81 aa and align over **71 of 76** residues at 38.2%, passing the same coverage criterion,
with a chicken positive control, the carp as a negative control that fails as expected, and
only **1 of 30** composition-matched shuffles passing (shuffle identity 7.9–27.6% vs the real
38.2%). Coverage and identity *disagree* — 38.2% is below the 43.4% bird-derived floor — and
that is reported rather than resolved by picking the favourable instrument. I still do not
adjudicate the paper's actual synteny argument: I examined sequence, not the locus.

### Propagation root-cause classification

- **Rows 1–2 — `NO_FAILURE_CORE`.** Self-referential IBA: the WITH/FROM names
  `UniProtKB:Q15847` itself as the sole IBD seed (14/14 annotations), so these record a PAN-GO
  curator's core-function judgement rather than a cross-species transfer.
- **Rows 3–4 — `EVIDENCE_CIRCULAR_OR_REDUNDANT` / `CIRCULAR_PROPAGATION`.** `IPR034450` is a
  single-family signature whose InterPro description is written from *this gene's own papers*
  and whose only reviewed member is human ADIRF; `UniProtKB-SubCell:SL-0191` traces to UniProt's
  `SUBCELLULAR LOCATION: Nucleus {ECO:0000269|PubMed:23239344}`. Both routes lead back to the
  IDA rows. Correct terms, zero added evidence → ACCEPT, not REMOVE.

**Upstream defect, filed as a question rather than acted on.** `interpro2go` maps `IPR034450` to
exactly two terms. `GO:0045600` reaches **723** gene products, of which only 130 are ADIRF-sized
and **504 exceed 200 aa** — they cannot be orthologues of a 76-aa protein — with 237
non-vertebrate metazoans including a 2304-aa *Toxocara canis* protein and nine scallop proteins.
Cause: human ADIRF is 76 aa with its three commonest residues at **43.4%** of the sequence, and
`PTHR39227` accordingly holds 556 members over 200 aa spanning fungi, bacteria, plants and one
archaeon. Aligning against 5 genuine orthologues and 12 oversized recipients separates them
cleanly on **coverage** (orthologues 75–76 of 76 residues; oversized members 23–65), and the
oversized ones are tandem-repeat proteins with 11/22/33-residue periods.

**The taxon constraints are working, which is what makes this sharp.** Zero of the 723
`GO:0045600` recipients sit outside Metazoa, and zero of the 789 `GO:0005634` recipients outside
Eukaryota. So the 237 invertebrates *pass* the `only_in_taxon Eumetazoa` constraint — the "a
passing taxon constraint is necessary, not sufficient" case. Whether an insect fat body cell
satisfies `GO:0045444`'s "features of an adipocyte" differentia is asked of GO, not answered here.

### PAINT: 6 of 7 node recipients are orthologues; the 7th is not — and the cause is upstream

`PANTHER:PTN008674116` reaches 7 gene products. Six are 75–76 aa orthologues aligning over
73–76 of the 76 human residues at 81.6–100%. The seventh, `A0A5F8H3S4` (*Monodelphis
domestica*), is a **447-aa tandem-repeat "Uncharacterized protein"** (22-residue period at 85.4%
periodicity; three commonest residues 52.1%) aligning over **62 of 76 residues at 15.8%** —
inside the same band as the unambiguously spurious IPR034450 matches. It receives both terms,
including `is_active_in` nucleus.

**I attribute this upstream of PAINT, deliberately.** Opossum has a real three-exon ADIRF gene
(NCBI GeneID **100020286**, ADIPOQ control non-zero for the same taxon) but **UniProt's
*Monodelphis* proteome contains no ADIRF-sized family member** — so the tree was handed the wrong
sequence and annotated what it had. PAINT's placement of the six real orthologues is correct, and
the campaign's calibration that PAINT models families well survives. The filable items are
whether the family HMM should admit a 447-aa repeat protein, and whether the missing opossum
protein belongs in the reference proteome.

**I had written the opposite into the review before measuring it** ("all recipients 71–76 aa,
homogeneous"). The committed guard refused it.

### `NbExp=3` is one screen counted three ways (4th instance this campaign)

UniProt line 100 reads `NbExp=3` for the IL18 pair. IntAct returns **one** record
(`EBI-11784632`, MI-score 0.56) under three sub-methods — `two hybrid array`, `two hybrid prey
pooling approach`, `validated two hybrid` — and the paper says so: *"We screened this search
space a total of nine times with a panel of three Y2H assay versions"*.

**Three checks came back NEGATIVE and are reported as such, not omitted:** the partner resolves
to reviewed canonical Swiss-Prot IL18 at the full 193 aa (no TrEMBL/ORFeome substitution);
neither protein is a hub (IL18 15 distinct IntAct partners, ADIRF 4); and the interaction is
**not** topologically impossible, since pro-IL18 is cytosolic and ADIRF is cytosolic +
nucleoplasmic. The verdict rests only on absent replication and an uninformative term.

### Extracellular: three independent routes, none in GOA — so KEEP, not over-annotate

The two `GO:0070062` rows are bulk imports (1046 and 1016 entities, single term, all HDA), and
the 23533145 authors caution that *"the presence of high abundant contaminating proteins... must
be taken into account and further verified"*. But extracellular presence is corroborated by
**serum ELISA at 18.7 µg/mL across 139 patients** (`PMID:33737617`) and **intact protein by
top-down MS in lipoaspirate fluid** (`PMID:26719138`) — neither in GOA, neither in affinage. I
verified via QuickGO that `GO:0070062` **is** a descendant of `GO:0005576`, so I did **not**
propose extracellular region as redundant. The export route is unknown (no signal peptide, no
TM) and is a `knowledge_gap`. One count I could **not** reconcile and record as unresolved: the
paper says "close to 900 proteins" while GOA imported 1046 entities.

### NEW `GO:0005829 cytosol` — a real gap in the HPA import

HPA gives ADIRF **two** main locations, Nucleoplasm *and* Cytosol (antibody HPA026810, IF
reliability *Supported*); GOA's `GO_REF:0000052` row carries only nucleoplasm, and `GO:0005829`
is absent from ADIRF's whole record. Verified with a positive control that resolves its term id
**through the same mapping** the subject uses (GAPDH is also called Cytosol and does carry
`GO:0005829`), so this is not a broken query. Proposed as `located_in` only — cytosol is
deliberately **kept out of** `core_functions.locations`, because no function is claimed there.

Bonus: `PMID:31945134` used **Sigma HPA026810** — the same antibody — across mouse tissues and
saw *"no ADIRF in any tissue"*. An antibody blank in the species lacking the gene is an unusually
clean specificity control, and it **strengthens** row 6.

### Evidence-code defect

Rows 10 and 11 are coded **IDA** but the abstract names the assay: *"Over-expression studies in
3T3-L1 cells"*. GO's IMP definition explicitly covers over-expression, so these should be
**IMP**. Decidable from the abstract alone. Row 9 (localisation) is a genuine direct observation
and stays IDA. Flagged in `suggested_questions` because the schema records term-level actions,
not evidence-code corrections.

### UniProt corrections (no GO row to act on)

The entry still carries `GO:0030154` and `GO:0006351` as `IEA:UniProtKB-KW`. Neither is entailed
by the evidence, and I **fetched the relations** rather than inferring from labels: `GO:0045944`
is not a descendant of `GO:0006351`, and `GO:0045600` is a descendant of neither `GO:0030154`
nor `GO:0045444` — GO keeps regulation out of `is_a`. `GO:0006351` says ADIRF *performs*
transcription. Confirmed the SPKW route is retired with a positive control (`GO_REF:0000043` = 0
human annotations vs `GO_REF:0000044` = 139,714; ADIRF has 0 rows under either term while POLR2A
has 13 and PPARG has 7). No GO action available — filed as a UniProt request, alongside the
arterial-expression correction and the two missing primary references.

### affinage

`gates_passed: True`, 2 citations, both real numeric PMIDs, neither retracted — **precision
fine, recall near zero for annotation purposes**. It returned none of the four PMIDs GOA cites,
including `PMID:23239344` (behind 3 of 11 rows), and missed both papers above. Recorded as a
`reference_review` with `correctness: LOW_QUALITY` for one substantive inaccuracy — it asserts no
binding partner is characterised while a curated IL18 interaction sits in the GO record — and
**not cited as `supporting_text` anywhere**.

### Committed tooling, and 7 defects found by break-testing it

`ADIRF-bioinformatics/analyze_adirf.py` (five measurements; positive control on every reported
zero; `numberOfHits == len(results)` asserted on every paginated query) and `build_review.py`
(generates the review from the GOA TSV, audits the **emitted** YAML — strict duplicate-key
loader, anchors, raw-vs-parsed `reference_id`, row count, same-term-same-action, core_functions
coverage in both directions, the no-molecular-function hedge, and every `file:` quote). Both have
`--self-test`; `--audit-only` re-checks a hand-edited file.

Found by break-testing, not by reading: a hand-assigned 50% identity floor that wrongly rejected
pigeon ADIRF at 43.4%; an assertion that could not fail (deleted rather than fixed a third time);
a positive control that hard-coded its term id separately from the mapping it was meant to
control; a check that reported and then crashed, aborting its section; one constant serving two
roles so its break-test crashed on the other; **two fabricated `file:` quotes** (the repo
validator skips `file:` quotes entirely, so on CI alone they would have shipped); and the false
homogeneity claim above.

**Round 2 adds section F** (teleost conservation by alignment, with a two-sided control, a
30-trial composition control, and pinned RefSeq accessions asserted to still be ADIRF-named and
ADIRF-sized so a repurposed accession fails loudly) and four more break-tests: an inverted
control expectation, a coverage criterion that composition alone satisfies, a repurposed
accession, and the happy path.

### Gates

`just validate human ADIRF` → `✓ Valid (with 1 warnings)`, on the branch **and** merged into
main (merge probed locally: zero conflicted files). 57 quotes verified verbatim — 49 by
`checkquotes.py` plus the 8 `knowledge_gaps[].provenance[]` quotes it does not walk, checked by
hand. Raw/parsed `reference_id` reconcile at 36; no YAML anchors. `cache/**` untouched (md5
identical before and after validate) and `cache_lint` exits 0; multiset check confirms nothing
lost, nothing invented, no new duplicates. Both generated artifacts reproduce byte-identically
across consecutive runs.

**The standing warning is deliberate**: "No annotations reference available deep research files".
Per the campaign rule I do not quote an affinage sentence as `supporting_text`; its assessment
lives in `reference_review` instead. Left standing rather than silenced.

### Round 2 — the 🟡 and all three 🔵 items

- **🟡 teleost claim** — conceded and fixed by measurement, not by softening. See the corrected
  paragraph above; `description` now says orthologues in teleost fish, birds and mammals *all
  align across essentially the whole 76-residue protein*, which is now true by alignment. The
  `reference_review` for PMID:31945134 records the whole correction, including what it still
  does not adjudicate. **New finding it produced:** `IPR034450` has **50 teleost members and
  zero ADIRF-sized ones**, so the family's entire fish content is spurious while the real
  teleost proteins sit in RefSeq outside it — the signature is over-inclusive *and*
  under-inclusive at once, and that reciprocal half is now in the InterPro question.
- **🔵 the sub-diffusion-limit argument is symmetric** — agreed, and it was asymmetric only in
  the cytosol row. That row now rests on the HPA immunofluorescence call alone and says
  explicitly that the argument predicts both compartments so cannot be evidence for either. It
  survives only in the knowledge gap about active nuclear retention and in the `is_active_in`
  question.
- **🔵 nucleus + nucleoplasm in `core_functions.locations`** — kept, with a comment in the
  builder giving the reason: three separate GOA rows assert the parent explicitly (IDA, IBA,
  IEA) while HPA supports the child, so each is independently attested. Dropping the parent
  would also silence my own coverage guard rather than satisfy it, which is the wrong direction.
  Happy to drop it if you would rather the guard were made ancestry-aware.
- **🔵 `description` stated cytosol as settled** — agreed; it now reads "Immunofluorescence
  places the protein in both the nucleoplasm and the cytosol", which is the observation rather
  than a settled location, and stays free of curation commentary.
- **🔵 section E before D in `RESULTS.md`** — fixed in the report writer; order is now A–F.

---

> [!NOTE]
> **Low Risk**
> Adds gene-specific documentation and offline analysis scripts; no auth, APIs, or shared runtime code paths are modified.
> 
> **Overview**
> Adds a full **human ADIRF (Q15847)** GO annotation review package under `genes/human/ADIRF/`.
> 
> **`ADIRF-ai-review.yaml`** documents verdicts on all 11 GOA rows (mostly ACCEPT; over-annotation on `GO:0005515`; two exosome HDA rows kept as non-core) plus a **NEW** `GO:0005829` cytosol row from HPA IF. It weaves in literature the pipelines missed—notably **Muroidea loss of ADIRF** (43 kb deletion), bounds on 3T3-L1 over-expression evidence, InterPro/PANTHER propagation issues, knowledge gaps, and curator questions.
> 
> **`ADIRF-bioinformatics/`** adds **`analyze_adirf.py`** (live QuickGO/UniProt/NCBI/HPA queries with positive controls, pagination guards, and extensive `--self-test` break-tests) and generated **`RESULTS.md`** covering Muroidea absence, `IPR034450` reach, signature promiscuity, HPA vs GOA import gap, PANTHER node `PTN008674116` (including the opossum tandem-repeat mis-annotation), and teleost conservation by alignment.
> 
> No changes to shared app infrastructure—gene-specific curation and reproducible audit artifacts only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fa9681eae5ecf21d536fbe3da71f359dba7ffb9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — the 🟡 was right and it is the sharpest kind of catch, because the review was using the one instrument it argues against everywhere else. Conceded and fixed **by measurement rather than by softening**, since your first option was available.

## 🟡 Teleost conservation, now settled on sequence

You were right on all three premises, and I checked each before conceding: the 103 hits do come from `ADIRF[sym] AND txid7898[Orgn]`; `RESULTS.md:97` does put carp `A0A8C1JCC4` in the spurious bin; and the bird half was already sequence-supported while the fish half had nothing.

**Why no UniProt query could have answered it** — and this turned out to be a finding in its own right: `IPR034450` has **50 teleost members and zero ADIRF-sized ones**. The family's entire fish content is oversized spurious matches, which is exactly why the only fish sequence available to align was a 938-aa repeat protein. The real teleost ADIRF proteins are annotated in **RefSeq, outside the family**.

New `RESULTS.md` section F, same coverage criterion as sections C and E (≥68 of 76 residues):

| entry | organism / role | length | aligned of 76 | % id | passes |
|---|---|---|---|---|---|
| `NP_001373520.1` | *Danio rerio* | 81 | **71** | 38.2 | **yes** |
| `XP_085644419.1` | *Trachurus japonicus* | 81 | **71** | 38.2 | **yes** |
| `A0A1D5PM71` | *Gallus gallus* — positive control | 76 | 76 | 50.0 | yes (expected) |
| `A0A8C1JCC4` | carp, the only teleost UniProt's family offers — negative control | 938 | 51 | 19.7 | no (expected) |

**Composition control**, because for an Ala/Gln-rich 76-aa protein that is the obvious way to get a false positive: 1 of 30 composition-matched shuffles of the *Danio* sequence (identical residue content, order destroyed, deterministic seeds) passes the coverage criterion, and the shuffles reach only 7.9–27.6% identity against the real 38.2%. So the criterion is not composition-driven — but the pass rate is not zero, which is why the identity margin is reported alongside coverage rather than coverage being treated as sufficient alone.

**And coverage and identity disagree, which I report rather than resolve.** Both fish proteins clear the coverage criterion comfortably, yet 38.2% sits *below* the 43.4% orthologue floor the birds establish in section C. Coverage is the criterion this analysis committed to before the fish were examined and a longer branch predicts more divergence — but picking whichever instrument gives the wanted answer is precisely the failure mode, so both numbers are on the table.

**What I still do not claim.** I examined sequence, not synteny. PMID:31945134's argument is a genomic deletion at the syntenic locus, and I have not tested that. So the `reference_review` now says an ADIRF-like protein *is* annotated in zebrafish and is inconsistent with a flat absence claim, while stating explicitly that the deletion argument itself is not adjudicated here. Only the mouse/rat/hamster claim is relied on, and that my own controls independently confirm. `description` now reads "orthologues in teleost fish, birds and mammals all align across essentially the whole 76-residue protein", which is true by alignment.

The reciprocal InterPro finding is now in the relevant `suggested_question`: the signature is **over-inclusive of unrelated repeat proteins and under-inclusive of real orthologues at the same time**.

## 🔵 items

- **The sub-diffusion-limit argument is symmetric.** Agreed, and it was asymmetric in exactly one place. It no longer supports the `GO:0005829` row, which now rests on the HPA immunofluorescence call alone and says explicitly that the argument predicts both compartments and so cannot be evidence for either. It survives only in the knowledge gap about whether nuclear retention is active, and in the `is_active_in` question — both places where it cuts symmetrically.
- **`description` stated cytosol as settled.** Agreed: now "Immunofluorescence places the protein in both the nucleoplasm and the cytosol", i.e. the observation rather than a settled location, and still free of curation commentary.
- **Section E before D.** Fixed in the report writer; order is A–F.
- **`nucleus` + `nucleoplasm` in `core_functions.locations`** — this one I have kept, with the reason now recorded as a comment in the builder rather than left implicit: three separate GOA rows assert the parent term explicitly (IDA, IBA and IEA) while HPA supports the child, so each is independently attested. Dropping the parent would also make my own coverage guard fire, and silencing a guard is the wrong way to satisfy it. If you would rather have only the specific term, the right fix is to make that guard ancestry-aware — say the word and I will, though it means the builder starts making network calls to fetch the ancestry it asserts.

## Gates on this head

`just validate human ADIRF` → `✓ Valid (with 1 warnings)` on the branch **and** merged into main (probed locally: zero conflicted files). 57 quotes verbatim — 49 via `checkquotes.py` plus the 8 `knowledge_gaps[].provenance[]` quotes it does not walk, checked by hand. `cache/**` byte-identical throughout (md5 unchanged) and `cache_lint` exits 0; multiset check confirms nothing lost, nothing invented, no new duplicates. All three generated artifacts reproduce byte-identically across consecutive runs. Four new break-tests (inverted control expectation, composition-satisfiable criterion, repurposed pinned accession, happy path) — 20 `ok` in `analyze_adirf.py --self-test`, 7 in `build_review.py --self-test`. Pinned RefSeq accessions are asserted to still be ADIRF-named *and* ADIRF-sized, so a silent repurposing fails loudly rather than changing the answer.

I also swept the withdrawn claims across all seven surfaces — notes, review YAML, `RESULTS.md`, both scripts, `results.json` and **the PR body** — and the body's version of the symbol-count argument was still there; it is corrected on the server now.

cmungall:
Both 🔵 nits fixed at the generator, since both reproduce on regeneration, and both verified in the **emitted** artifacts rather than in the source.

**1. The `It is ` artefact** (`ADIRF-ai-review.yaml:793` ← `build_review.py:607`) — a replacement that spanned wrapped string literals. The reason now reads "…restating the nuclear rows. This row rests on the HPA immunofluorescence call alone…". Worth naming the general shape: nothing in this repo checks *emitted prose* for well-formedness — the quote gates check verbatim-ness of quotations, the schema validator checks structure, and a malformed sentence inside a `reason` passes both. Combined with the YAML being generated, it reappeared on every run. You caught it by reading, which appears to be the only available detector.

**2. The hardcoded `"103 Actinopterygii"`** — and thank you for this one, because it is the *same defect class as the claim section F was written to correct*: a number asserted rather than derived, sitting next to the live measurement it was supposed to quote. Now derived from section A's result in both `RESULTS.md` and `results.json`, with the literal asserted absent from the source so it cannot creep back.

**That fix earned its keep immediately, which is the part worth reporting.** Because section F now quotes section A, it depends on section A having run — so rather than trusting call order it asserts the dependency and refuses to substitute a literal. Adding that guard **instantly broke three existing break-tests**: they construct an empty `Audit`, so the new early return silently redirected all of them to the dependency message instead of their intended targets. That is the "a new early return aborts the checks that follow it" failure, and it was caught *only* because `_expect_problem` asserts the failure **message** rather than merely the failure — with a bare `raises` it would have reported three green tests that were no longer testing anything. `_expect_problem` now takes a `seed`, and the dependency test is the single case that deliberately does not seed it.

Self-test counts are now 21 (`analyze_adirf.py`) and 7 (`build_review.py`), each asserting its own message.

**Gates on this head** (`9dd00de25`): `just validate human ADIRF` → `✓ Valid (with 1 warnings)`; branch is level with `main`, so nothing to merge forward and the previously-validated merged state stands. 57 quotes verbatim — 49 via `checkquotes.py` plus the 8 `knowledge_gaps[].provenance[]` quotes it does not walk. All three generated artifacts reproduce byte-identically across consecutive runs. `cache/**` byte-identical and `cache_lint` exits 0; multiset check confirms nothing lost or invented and no new duplicates; no rendered HTML in the branch diff; working tree clean.

I recognise this push moves the head past your approval — that is my doing for not having these in before it landed, and shipping a visible text artefact into a `review.reason` seemed the worse trade. No further changes planned: the last two rounds have not moved a term, an action, an evidence code, a quote or a reported number, only the harness and one malformed sentence. If you would like the `core_functions` coverage guard made ancestry-aware so the redundant `GO:0005634` can be dropped, that is the one open item I know of and I am happy to do it — otherwise I consider the review done.

cmungall:
Both residuals fixed, and **stating a stopping criterion**, because by your own accounting rounds 2 and 3 moved no term, action, evidence code, quote or reported number — the loop has migrated from the curation to the harness.

## (a) "asserted absent from the source" — the claim had no check behind it

You were right and the framing is the useful part: the scan existed only in a throwaway edit script, so I described a permanent verification in the same commit that discarded it. `assert_no_hardcoded_counts()` now lives next to the thing it guards and runs in **both** `--self-test` and normal mode.

Writing it produced two further defects in itself, neither of which I would have found by reading it:

- **It matched its own vocabulary list.** Storing the forbidden strings as literals meant the guard found itself — the same self-reference trap a retracted-phrase matcher falls into. It now matches the **shape**, `\d+\s+(?:Actinopterygii|Aves|avian|Sciuridae|Muroidea)`, so a derived emission (`{n_teleost_symbols} Actinopterygii`, no digits) cannot match while a reintroduced literal must.
- **The docstring documenting the defect tripped the guard**, by quoting it. **Reworded rather than exempted** — an exemption is exactly the bypass anyone wanting to smuggle the literal back would use. The break-test fixture instead synthesises the violating string at runtime only: `{103} Actinopterygii` in source has a brace between the digits and the clade, so it cannot match the shape while the formatted value does.

## (b) A guard whose message could not be delivered

Also right, and the sharper statement of it is that the guard would have **fired correctly and silently**: `write_report` reads every section key unconditionally, so a section taking the early return kills the run with a `KeyError` before `main()` prints FAILED INVARIANTS. `main()` now checks section completeness first, reports the problems, and **declines to regenerate `RESULTS.md`** so it cannot go stale against an incomplete run. Break-tested by confirming `write_report` *does* raise on an incomplete audit — which is what makes the pre-emption in `main()` reachable rather than decorative.

Self-test counts are now 23 and 7, each asserting its own message.

**Value-neutrality, using your own check:** `RESULTS.md` and `results.json` are **not in this diff**. The derived count equals what the literal said, so byte-identical output is precisely what a correct de-hardcoding produces.

## On the `core_functions` coverage guard

Accepting your reasoning and taking it off the table: making it ancestry-aware would have the builder fetch over the network the ancestry it currently only asserts, trading a redundancy you already accepted for a new failure mode in a generator whose byte-identical reproducibility is one of this PR's gates. `nucleus` + `nucleoplasm` stay, with the reason stated in the builder.

## Stopping criterion

The last two rounds changed no term, no action, no evidence code, no quote and no reported number — only the harness and one malformed sentence. Everything since round 1 has been findings about the machinery rather than about what the machinery computes, which is where I think the line goes.

**What I will still fix if you find it:** anything that would misfire on another machine, misstate a number, or let a withdrawn claim survive on a surface. **What I will not keep iterating on:** the guards' own prose, docstrings and self-test ergonomics.

The curation itself I consider settled: 11 GOA rows reviewed row-for-row, 7 ACCEPT / 3 KEEP_AS_NON_CORE / 1 MARK_AS_OVER_ANNOTATED / 1 NEW, nothing removed; the Muroidea gene loss measured with positive controls and a sister-clade negative control; the teleost claim now resting on alignment with a two-sided and a composition control; the IPR034450 over- and under-inclusiveness quantified; and the opossum node defect attributed upstream of PAINT rather than to it.

PR #2340, opened 2026-07-27
Review human ADISSP: a combined-methods reference counting one source three times

cmungall:
## ADISSP (Q9GZN8) — human, PAINT + affinage

ADISSP is a 174-residue **leaderless secreted adipokine** (formerly C20orf27) that engages an
unidentified adipocyte surface receptor and raises cAMP/PKA and, separately, insulin-independent
Akt. It also binds protein phosphatase 1. Thinly annotated: 13 GOA rows, no PAINT involvement.

### Identity — the worklist contains a look-alike, and my brief conflated them

My task brief described ADISSP as "formerly C5orf46/SSSP1". That is wrong, and both genes are rows
in `projects/paint/human-no-IBA-simple.csv`:

| | ADISSP (line 1540) | C5orf46 (line 5884) |
|---|---|---|
| HGNC | HGNC:15873 | HGNC:33768 |
| previous symbol | **C20orf27** (renamed 2022-12-12) | — |
| aliases | FLJ20550 | MGC23985, **SSSP1**, AP-64 |
| locus / length | 20p13 / 174 aa | 5q32 / 87 aa |
| UniProt | Q9GZN8 | Q6UWT4 |

Both accessions fetched with `primaryAccession == requested` asserted. The gene reviewed is
**Q9GZN8**.

### Row reconciliation

`13` GOA TSV rows → `12` distinct annotations → `12` stub entries. The gap is **not** a stub
collapse: two rows are the same annotation (`GO:0005576 / ISS / GO_REF:0000024 / Q9D1K7`) differing
only in date (2022-12-15, 2025-10-27). One entry covers the pair and says so. Plus one `NEW`
proposal = 13 entries.

### Verdicts

| term | rows | evidence | action |
|---|---|---|---|
| GO:0005576 extracellular region | 3 | IEA, ISS×2 | **ACCEPT** (core location) |
| GO:0042593 glucose homeostasis | 2 | IEA, ISS | **ACCEPT** |
| GO:1990845 adaptive thermogenesis | 2 | IEA, ISS | **ACCEPT** + term request |
| GO:0008157 protein phosphatase 1 binding | 1 | IDA | **ACCEPT** (core MF) |
| GO:0007189 adenylate cyclase-activating GPCR pathway | 2 | IEA, ISS | **MODIFY → GO:0141163** |
| GO:0030511 pos. reg. TGF-β receptor signalling | 1 | IMP | **KEEP_AS_NON_CORE** |
| GO:1901224 pos. reg. **non-canonical** NF-κB | 1 | IMP | **MODIFY → GO:0043123** |
| GO:0005575 cellular_component | 1 | ND | **REMOVE** |
| GO:0048018 receptor ligand activity | — | ISS | **NEW** |

### Propagation root causes

There are **no IBA rows** — verified, not assumed (fully paginated QuickGO returns 13 annotations,
zero IBA; ACTB returns 11 IBA from the same endpoint as the positive control). All eight IEA/ISS
rows trace to one entity, mouse Adissp `Q9D1K7`.

- `NO_FAILURE_CORE` — GO:0005576, GO:0042593, GO:1990845. The donor holds its own IDA/IMP from
  PMID:36496438 for every one, and human/mouse are both 174 aa and **90.2% identical with no
  indels**, far above GO_REF:0000107's 40% threshold.
- `TERM_SCOPING_PROBLEM` + `SOURCE_EVIDENCE_WEAK` — GO:0007189, scoped narrowly to the receptor
  clause; the donor's cAMP/PKA evidence itself is strong (see below).
- The human↔mouse exchange is **reciprocal, not circular**: mouse→human carries the four
  mouse-experimental terms, human→mouse carries the three human-experimental ones. No term returns
  to its own origin, so `SOURCE_WEAK_OR_INFERRED` would have been factually wrong on every row.

### Main finding: a combined-methods reference recording one source as three witnesses

`GO:0005576 IEA GO_REF:0000120` has WITH/FROM
`UniProtKB:Q9D1K7 | ensembl:ENSMUSP00000099477 | UniProtKB-SubCell:SL-0243`. GO_REF:0000120 exists
to record that **independent** electronic pipelines produced the same annotation, so three tokens
read as three witnesses. They are one:

1. `Q9D1K7` — mouse Adissp, holds GO:0005576 by IDA;
2. `ENSMUSP00000099477` — the same protein, second namespace;
3. `SL-0243` "Secreted" — on the *human* entry this is
   `SUBCELLULAR LOCATION: Secreted {ECO:0000250|UniProtKB:Q9D1K7}`, i.e. by similarity to the same
   mouse protein.

The SubCellular2GO arm re-reads the Compara arm's conclusion. The annotation is correct; the
corroboration it advertises is not. Filed as a question to GOA/UniProt, not as a GO action.

*Correction worth propagating:* **`GO_REF:0000120` is not the ARBA reference.** It is "Combined
Automated Annotation using Multiple IEA Methods"; ARBA alone is `GO_REF:0000117`.

### MODIFY: canonical, not non-canonical NF-κB

`GO:0038061` (non-canonical) requires NIK-dependent p100→p52 processing; `GO:0007249` (canonical) is
IKK-dependent activation through IκB to RelA. PMID:32024300 measures **p-IKK, p-IκB and p-p65/RelA**
in both the overexpression and silencing directions. NIK, p100/p52 and RelB appear nowhere in the
paper. I fetched the `is_a`/`part_of` ancestor closures of `GO:1901224` and `GO:0043123` and
**neither contains the other** — siblings naming different cascades, so the term is *wrong*, not
imprecise. (Gotcha for re-runners: neither regulation term has its target cascade in its `is_a`
closure either, because GO links them by `positively_regulates`, which does not subsume.)

### REMOVE: an ND placeholder its own file contradicts

`GO:0005575`'s own comment says the term marks a gene product whose cellular component is *unknown*
and that using it indicates no information was available. BHF-UCL made the row on 2012-07-16; the
same record now carries three extracellular region rows. Not a propagation failure of any kind —
there is no WITH/FROM.

### Signal peptide ≠ secretion, in the *opposite* direction from the usual case

There is no `FT SIGNAL` feature in either species, the human mature chain is `2..174` with
MS-measured N-acetylalanine at residue 2 — **and the protein is genuinely secreted, non-classically**:
brefeldin A + monensin *increased* Adissp secretion while inhibiting adiponectin in the same
experiment, and the authors had to graft a heterologous signal peptide onto their construct to
purify secreted protein. Human ADISSP itself was measured in circulation at 0.45 µg/mL.

This matters twice: the location rows rest on measurement rather than prediction, and it **defuses a
topological argument I would otherwise have made** against GO:0008157. A secreted protein binding
cytosolic PP1c looks impossible (the ACRV1 pattern), but a leaderless protein necessarily has a
cytosolic pool. The unconventional route is what rescues the interaction.

### GO:0007189 → GO:0141163 *(changed in round 2 — the first version of this argument was wrong)*

Measured well: cAMP up in transgenics and down in the knockout, PKA substrates both ways, purified
protein dose- and time-dependent, competable surface binding. **Not measured:** that the receptor is
a GPCR — "putative" throughout, with the entire G-protein link resting on one experiment with
1 µM melittin (a membrane-active bee-venom peptide), n=3.

Round 1 called this `MARK_AS_OVER_ANNOTATED` **with no replacement**, arguing that no correct term
existed. That was wrong: I had checked the ancestors of GO:0007189 and one sideways candidate
(`GO:0141156`) and stopped, never looking at the **regulation-of** branch — which is exactly where an
extracellular ligand whose measured output is "the cAMP/PKA cassette runs harder" belongs. The
reviewer pushed on that gap. Now `MODIFY` → **`GO:0141163 positive regulation of cAMP/PKA signal
transduction`**.

| candidate | verdict |
|---|---|
| an ancestor of GO:0007189 | impossible — **every** ancestor retains the GPCR in its definition |
| `GO:0141156 cAMP/PKA signal transduction` | wrong — an *intracellular* cassette, which a ligand is not part of |
| `GO:0045762 pos. reg. adenylate cyclase activity` | closer, but asserts more than the data |
| **`GO:0141163`** | **fits** — a regulation term, and it makes no claim about receptor class |

`GO:0045762` was the reviewer's suggestion and is worth explaining declining: PMID:36496438 measures
cAMP **content**, not cyclase activity, and a rise in content does not discriminate increased
synthesis from reduced phosphodiesterase activity. `GO:0141163` is satisfied either way. GO itself
points the same way — the obsoletion comment on `GO:0043950` directs users to `GO:0141163` first.
Verified current, and neither it nor GO:0007189 is an ancestor of the other, so this is a move
between branches, consistent with `MODIFY`. `REMOVE` would still be wrong: the receptor class is
unmeasured, not refuted.

### NEW GO:0048018, filling a gap present in both species

QuickGO returns **1** MF annotation for human ADISSP and **2** for mouse Adissp — all of them
GO:0008157. Neither species has any term for the adipokine activity both papers are about. Added by
ISS from Q9D1K7 (sequence-similar entity only, per ISS WITH/FROM semantics). Placement **verified,
not assumed**: GO:0048018 is not under `GO:0005102` and not under `GO:0005488` at all; it sits under
`GO:0140677`/`GO:0098772`.

### PP1 binding: replicated, but the count needs its null

IntAct recovers ADISSP with PPP1CA/CB/CC and PPP1R7 across **seven publications the annotation does
not cite**, by anti-tag co-IP, TAP and pull-down — all now computed by the committed
`ADISSP-bioinformatics/analyze_adissp.py` rather than asserted in prose. But PP1 catalytic subunits are affinity-proteomics
hubs — **PPP1CA 1012 IntAct records vs ADISSP's 27** — so recurrence in tag pulldowns is weak on its
own. The informative direction is subject-centric: **4 of ADISSP's 13 distinct *protein*
partners are PP1-module components**, so PP1 dominates this protein's own sparse interactome. Also noted: the row
is coded IDA from a CoIP of two transfected tagged plasmids with an **empty WITH/FROM**, so the
partner is recorded nowhere machine-readably — IPI would be the right code.

### Non-confirmations, each with a control

- **InterPro2GO** maps *none* of ADISSP's three signatures (IPR026794, PF15006, PTHR13287); control
  IPR001879 maps 2, so the file and grep work.
- **Bulk TAS imports**: zero TAS rows. **ARBA**: no rule named in any WITH/FROM.
- **Reference-projection test**: PMID:32024300 → 3 annotations / **1** entity; PMID:36496438 → 4 / **1**.
  Neither is an import or a subunit projection. PMID:42030391 → **0** annotations (coverage gap, not
  projection).
- **Logical-opposite cross-product**: not applicable (no negative-regulation term); pair-finder
  verified on a synthetic pair.
- **Retraction/erratum/EoC**: all 7 PMIDs clean by two routes; control PMID:32125225 fires correctly.
- **Pseudogene MF tell**: absent; neither ADISSP pseudogene has a UniProtKB entry (route control
  passes).
- **Peptide-vs-protein hazard**: does *not* apply. PMID:42030391 used full-length recombinant
  protein and showed untagged protein reproduces the effect.

### Sibling cross-check against merged ADIPOQ — non-confirmation

The brief records that ADIPOQ "shipped a false ancestry claim" about `GO:0048018`, the term proposed
here, so `main`'s merged review was checked for a surviving instance. **There is none**: its claim
that `GO:0005179 hormone activity` sits under `GO:0048018 → GO:0140677 → GO:0098772` is *true*,
verified against QuickGO.

The check paid for itself anyway. `GO:0005179`, which ADIPOQ and LEP both carry, is a **descendant**
of the `GO:0048018` proposed here, so the two are not competing choices — the parent was chosen
because the endocrine claim inside hormone activity holds for the mouse protein and is untested for
the human one, and refining to `GO:0005179` later would retract nothing. The `suggested_questions`
entry now states that relationship rather than merely contrasting the terms.

### Round 2 — all four review suggestions addressed

1. **Regulation-of branch (real gap, verdict changed)** — see above.
2. **Prose-only IntAct evidence (real gap, artifact added)** — `ADISSP-bioinformatics/analyze_adissp.py`
   (stdlib only) computes the seven PP1 publications, the 4-of-13 partner fraction and the hub null at
   run time. It reproduces every figure I had written by hand **and adds one I had not checked**: the
   annotation's own reference PMID:32024300 is *not* among the seven, so those datasets are additional
   rather than the same evidence re-counted.
3. **The 90.2% figure now has provenance** in the same script, with a self-identity positive control and
   a refusal to compare unequal-length sequences rather than truncating them. Guards break-tested three
   ways, each failing with the right message *and* a non-zero exit status: a merged accession
   (`O15507` → GFRA1/`P56159`) is rejected by a `primaryAccession` assertion; removing the length guard
   trips the negative control; a reduced page size is detected as truncation. A fresh run reproduces the
   committed `RESULTS.md` and `results.json` byte-for-byte.
4. **PMID:22814378 and PMID:12665801** were cached but unreferenced; both are now in `references` with
   notes on what each supports.

**Two defects in my own harness this round**, both recorded in the notes: a stale `knowledge_gaps`
statement survived the action change and was caught only by grepping the *stable entity* across every
surface (the "fixed in N places, landed in N−1" recurrence); and my break-test loop reported `exit=0`
for three guards that were failing correctly, because `echo "$(basename $f) exit=$?"` reports
`basename`'s status, not the tested script's. Sweep B was also rescoped after it **falsely** demanded
that `GO:0043123`, a deliberately non-core MODIFY replacement, appear in `core_functions` — the
*permitted* and *required* directions need different scopes, and both are now break-tested along with
the happy path that had been misfiring.

### Round 3 — a wrong set hiding behind a right number

Both round-2 notes fixed. The first is the most instructive defect in this review, and **the
reviewer's diagnosis of it was wrong in two checkable ways**, so it is recorded rather than conceded.

The resolver was `partner = b if id_a.endswith(SUBJECT) else a`. IntAct returns ids with a trailing
database tag — `"Q9GZN8 (uniprotkb)"` — so `endswith()` was **never true for any record**, not merely
for isoform forms, and the partner was taken to be `moleculeA` in all 27. So:

- partners appearing only as `moleculeB` were **dropped** — `RALYL` and a PRO chain of `P0C6X7`.
  Real partners *were* lost, contra "no real partner can be lost this way".
- `ADISSP` entered its own partner set, and `mrna_adissp` entered from a CLASH record in which
  **neither** side is ADISSP (it pairs the ADISSP mRNA with a miRNA).
- the correct count of distinct **protein** partners is **13, not 11**, so the fraction is unchanged
  at 4 of 13 and the error was **not** conservative.

**Two spurious entries and two dropped ones cancelled exactly.** The count survived a hand check, a
review and a table, because there was no arithmetic discrepancy to notice. A right-looking number is
not evidence of a right set.

Fixed with `base_accession()` (strips the db tag and isoform/PRO suffix for subject matching, keeps
the full name for partners) and `resolve_partners()` (takes the non-subject side, raises if both
sides resolve to the subject, excludes-while-counting records not involving the protein).
`RESULTS.md` now reports 27 total records / 26 involving the protein / 1 involving the locus only.

A committed `--self-test` covers six directions and was validated the two ways that matter:
**run against the defect that actually shipped** — the resolver taken verbatim from `git show HEAD:`
reproduces the wrong set on live data, same count, different set — and **proved able to fail**
(reinstating the defect fails direction 1 with its own diagnosis; disabling the self-interaction guard
fails direction 3 with its own message; both mutations assert their target is present first, both exit
non-zero). It also **declares one thing it cannot do**: the "subject in its own partner set" invariant
keys on accessions, so a partner named by gene *symbol* — exactly how the spurious `ADISSP` entry arose
— is unmatchable, and direction 1's fixture is what actually guards the shipped bug.

Note 2 fixed: "Three candidate homes" → four considered, three rejected, in both `reason` strings and
the notes.

And regenerating `RESULTS.md` broke a `file:` quote in **two** places, caught by `checkquotes.py` — the
two-way dependency a quote into a generated artifact creates, and it only worked because the check ran
*after* regeneration.

Merged `origin/main` forward (no conflicts) and re-verified on the merged tree: my gene files
byte-untouched, `terms.csv` multiset intact with only my own `GO:0008157` added and no new duplicate,
`just validate` green, quote check green, self-test green, `cache_lint` clean, working tree clean, and
no rendered HTML in the PR diff.

### Round 4 — hygiene only, and a stopping criterion

Three non-blocking items fixed; **no term, action, evidence code, quote or reported number changed.**
Dead `regressed()` deleted (its comment contradicted its neighbour); the `protein` qualifier on the
partner count now covers all four sites, found by grepping the **number** rather than the phrase; and
`render()` now selects singular/plural from the count.

The qualifier miss recurred twice in one sitting, the second time for a cause worth recording: a patch
script asserted an anchor, the anchor had drifted, the `AssertionError` aborted the script — and a
**later independent edit in the same script never ran**. An anchor assertion protects the edit it
guards, but a failed one silently cancels everything after it, and from outside that is
indistinguishable from a script with nothing to do. Re-grep afterwards; do not trust the exit status.

`--self-test` is deliberately not wired into CI — `.github/workflows/` is shared config outside a gene
review's scope — and that limitation is stated in the notes rather than papered over.

**Stopping criterion (notes §20):** rounds 2 and 3 each moved something substantive; round 4 moved dead
code, a qualifier's coverage and a plural verb. The loop has migrated from the curation to the harness,
so that is the line.

### Round 5 — the one carve-out the criterion named

The reviewer flagged two trailing observations as "neither a request". One qualified under the criterion
I had stated ("could misfire on another machine"), so it was fixed; the other did not, and is only
included because the push was already happening.

`render()`'s exclusion sentence pluralised its leading clause from the count but left the tail and a
**hardcoded parenthetical** behind — with two excluded records it would emit *"2 records involve … and
**is** excluded … as **a partner**"*, with the parenthetical describing one record while claiming both.
Reachable by new IntAct data alone, i.e. the branch no past run exercises. **Removed rather than
repaired:** `resolve_partners()` now returns a data-built description of each excluded record and
`render()` lists them, so the prose cannot disagree with the data at any count. New self-test direction
2b exercises the multiple-exclusion path, which the old singular-only assertion could not reach and
which would have passed against the broken branch.

Also lowercased `PROTEIN` → `protein` at the one of four sites that shouted it. No term, action,
evidence code, quote or reported number changed.

### Verification

`checkquotes.py` 81 quotes / 0 problems (break-tested); `just validate human ADISSP` → `✓ Valid`;
all 31 quotes verified verbatim before use, every `file:` quote confirmed on a single physical line;
`source_entities` built *from* the GOA WITH/FROM with a drift assertion, break-tested by deleting an
entity GOA lists (fires, exits non-zero); four discipline sweeps (summary-opener vs action;
`core_functions` ↔ ACCEPT/NEW **both** directions; hedge vs structured field; every quote names its
subject) all break-tested on a mutated copy.

Two things my own checks caught: an unanchored `.//ArticleId` XPath was reading DOIs out of each
article's **reference list** (a Cancers 2020 paper appeared to have a 2017 DOI); and the
subject-naming sweep found the melittin quote truncated so that it no longer named Adissp, now
quoted in full.

A first-pass RVxF PP1-docking motif scan was **discarded and no claim rests on it**, because 2 of 5
characterised PP1-interactor positive controls scored zero.

`cache/go/terms.csv`: one legitimate addition (`GO:0008157`), verified by **multiset** comparison —
nothing lost, no new duplicates, `cache_lint` exit 0.

The standing validation warning ("No annotations reference available deep research files") is left
**on purpose**: affinage's `gates_passed: False`, and satisfying the warning would mean citing the
provider for mechanistic claims. No affinage sentence is quoted anywhere, and its arithmetic was not
carried over. Its precision was fine (5/5 citations real and correctly attributed); its recall
missed the non-classical secretion, the human secretion measurement, the NF-κB mismatch, the IntAct
replication and the stale ND row.

---

> [!NOTE]
> **Low Risk**
> Curated gene documentation and offline analysis scripts only; no auth, APIs, or production runtime behavior changes.
> 
> **Overview**
> Adds a full **human ADISSP (Q9GZN8)** gene review package: curated `ADISSP-ai-review.yaml`, cached UniProt/GOA inputs, working notes, Affinage deep-research export, and two publication stubs.
> 
> The review **reconciles 13 GOA rows** (12 distinct + one `NEW`) with explicit actions: accept extracellular region, glucose homeostasis, adaptive thermogenesis, and PP1 binding; **modify** GPCR pathway annotations to **`GO:0141163`** (cAMP/PKA regulation) and non-canonical NF-κB to **`GO:0043123`** (canonical); **remove** stale **`GO:0005575` ND**; **propose** **`GO:0048018` receptor ligand activity**; and flag cancer-line TGF-β/NF-κB roles as non-core. It also documents propagation issues (e.g. **`GO_REF:0000120`** listing one mouse source as three independent witnesses).
> 
> **`ADISSP-bioinformatics/analyze_adissp.py`** (stdlib) fetches UniProt/IntAct at run time to support ortholog identity (90.2%) and PP1 interactome stats (4/13 partners, hub null), with **`--self-test`** for IntAct partner resolution. **`cache/go/terms.csv`** gains **`GO:0008157`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5706e76692297cea5d5dbb014f3435c740514d30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — all four addressed in `230f2829`. Two were real gaps, and one of them changed a verdict.

**1. The regulation-of branch — you were right, and the verdict changed.** My "no replacement exists"
argument checked the ancestors of `GO:0007189` plus one sideways candidate and stopped. It never
looked at regulation-of, which is precisely where an extracellular ligand whose measured output is
"the cAMP/PKA cassette runs harder" belongs. `GO:0007189` is now **`MODIFY` → `GO:0141163 positive
regulation of cAMP/PKA signal transduction`**.

I landed on a different term than you suggested, and the reason is worth stating rather than burying:
`GO:0045762 positive regulation of adenylate cyclase activity` asserts slightly *more* than the data.
PMID:36496438 measures cAMP **content** in transgenic and knockout tissue, not adenylate cyclase
activity, and a rise in content does not discriminate increased synthesis from reduced
phosphodiesterase activity — so `GO:0045762` would commit to the synthesis mechanism. `GO:0141163` is
satisfied either way, and it is where GO's own obsoletion comment on `GO:0043950` points first. I
verified it is current, is BP, and that neither it nor `GO:0007189` is an ancestor of the other, so
this is a between-branch move consistent with `MODIFY`. It is also now in `core_functions`, so the
best-measured arm of the paper no longer survives only as prose.

**2 and 3. The IntAct evidence and the identity figure are now a committed artifact.**
`ADISSP-bioinformatics/analyze_adissp.py` (stdlib only, no venv) computes all of it at run time.
It reproduced every number I had asserted by hand — and **added one I had not checked**: the
annotation's own reference, PMID:32024300, is *not* among the seven IntAct publications, so those
datasets are genuinely additional rather than the same evidence re-counted. A fresh run reproduces
the committed `RESULTS.md` and `results.json` byte-for-byte. Guards break-tested in three directions,
each failing with the right message *and* a non-zero status: `O15507` (which returns HTTP 200 for
GFRA1/`P56159`) is rejected by a `primaryAccession` assertion; removing the comparator's length guard
trips its negative control; a reduced page size is caught as truncation rather than reported as a
short count.

**4.** `PMID:22814378` and `PMID:12665801` are now in `references` with notes on what each supports
and why nothing is quoted from them.

Two defects in my own harness fell out of this round, both written up in `ADISSP-notes.md` §18:

- a stale `knowledge_gaps` sentence survived the action change ("flagged as over-annotated"), caught
  only by grepping the **stable entity** across every surface rather than re-reading the prose;
- my break-test loop reported `exit=0` for three guards that were failing correctly, because
  `echo "$(basename $f) exit=$?"` reports `basename`'s status. For a few minutes I had three guards
  that looked broken and were fine.

And your suggestion produced a third finding indirectly: making the row a `MODIFY` caused my own
`core_functions` ↔ rows check to **falsely** demand that `GO:0043123` — a deliberately non-core MODIFY
replacement — appear in `core_functions`. The permitted and required directions needed different
scopes; both are now break-tested, along with the happy path that had been misfiring.

One thing left deliberately unchanged: the standing `No annotations reference available deep research
files` warning. affinage's `gates_passed` is `False` and satisfying that warning would mean citing the
provider for mechanistic claims. Its recall missed the non-classical secretion, the human secretion
measurement, the NF-κB branch mismatch and the IntAct replication, so nothing here rests on it.

cmungall:
Both notes fixed in `e6b15e32`. Note 1 turned out to be the most instructive defect in this review, and
I have to push back on two parts of the diagnosis — not to defend the code, which was wrong, but because
conceding those two points would put false statements in the review.

**The mechanism is worse than an isoform edge case.** IntAct returns ids with a trailing database tag,
`"Q9GZN8 (uniprotkb)"`, so `endswith("Q9GZN8")` was **never true for any record** — not just for
isoform forms. The partner was therefore `moleculeA` in all 27 records. Two consequences that change
your conclusions:

- **Real partners were lost.** Anything appearing only as `moleculeB` was dropped: `RALYL` (Q86SE5) and
  a PRO chain of `P0C6X7`. So "no real partner can be lost this way" does not hold.
- **The error was not conservative.** The correct count of distinct *protein* partners is **13, not
  11** — `ADISSP` and `mrna_adissp` were spurious, `RALYL` and the PRO chain were missing, and the two
  errors cancelled exactly. The fraction is unchanged at 4 of 13.

That cancellation is the finding. The count survived a hand check, your review and a rendered table
because **there was no arithmetic discrepancy to notice**. This campaign's standing rule is that a
number which refuses to add up is the bug report; this is the complement — a number that adds up
perfectly while the set behind it is wrong in both directions.

I verified the shipped behaviour rather than reasoning about it: taking the resolver verbatim from
`git show HEAD:...analyze_adissp.py` and running it against the live IntAct response gives 13 partners
including `ADISSP` and `mrna_adissp` and missing the other two, against the fixed resolver's 13 with
them recovered. Same count, different set, demonstrated.

The fix is `base_accession()` plus `resolve_partners()`, which takes the non-subject side, **raises** if
both sides resolve to the subject, and excludes-while-*counting* records where neither side is the
protein (the CLASH mRNA/miRNA record). `RESULTS.md` now states 27 total records / 26 involving the
protein / 1 involving the locus only, rather than one undifferentiated 13.

There is a committed `--self-test` with six directions. Two things about it worth flagging:

- it was **run against the defect that actually shipped**, and **proved able to fail** — reinstating the
  defect fails direction 1 with its own diagnosis, disabling the self-interaction guard fails direction 3
  with its own message, both mutations assert their target string is present first so neither can
  no-op, and both exit non-zero;
- it **declares one thing it cannot do**. The "subject appears in its own partner set" invariant keys on
  accessions, so a partner named by gene *symbol* — which is exactly how the spurious `ADISSP` entry
  arose — cannot be matched against the subject accession. That invariant would **not** have caught this
  bug. Direction 1's set-equality fixture is what does. Better to say so than let it read as coverage.

**Note 2** fixed: four candidates considered, three rejected, in both `reason` strings and notes §9.

One incidental confirmation: regenerating `RESULTS.md` broke a `file:` quote in **two** places and
`checkquotes.py` caught both — the two-way dependency a quote into a generated file creates, and it only
worked because the check ran after regeneration rather than before.

Also merged `origin/main` forward (no conflicts) and re-verified on the merged tree: gene files
byte-untouched, `terms.csv` multiset intact with only my own `GO:0008157` added and no new duplicate,
validate/quote-check/self-test/`cache_lint` all green, clean working tree, no rendered HTML in the diff.

cmungall:
All three fixed in `c648dbe2`. This round changed **no term, no action, no evidence code, no quote and
no reported number** — the partner count, the identity figure, the seven PMIDs and every verdict are
identical to round 3.

**1. `regressed()` deleted.** You are right that it was a vestige of the approach I abandoned, and that
its comment contradicted the one below it. The direction-1b declaration stands alone.

**2. The `protein` qualifier now covers all four sites** — `RESULTS.md`, `ADISSP-ai-review.yaml:217`
and `:750`, and `ADISSP-notes.md`. Your diagnostic was the right one: grepping the **number** (`13`,
`thirteen`) rather than the phrase is what finds these.

It recurred *twice* in one sitting, and the second time has a cause I had not considered. My patch
script asserted an anchor in the annotation builder, the anchor had drifted, the `AssertionError`
aborted the script — and a **later, independent edit in the same script, to the notes file, never
ran**. So an anchor assertion protects the edit it guards, but a failed one silently cancels every edit
after it, and from outside the result is indistinguishable from a script that had nothing to do. The
generalisation: one edit per script, or collect failures rather than raising — and in all cases re-grep
afterwards instead of trusting the exit status. Worth saying that the assertions still earned their
keep: they refused to run against drifted text three times across rounds 3 and 4, and every refusal was
a real drift.

**3. `render()` now selects singular or plural from the count**, and the hand-wrapped line no longer
breaks mid-phrase.

On **`--self-test` not being in CI** — deliberate, and I would rather leave it stated than fix it here:
`.github/workflows/` is shared configuration outside the scope of a single gene review. It is recorded
as a limitation in notes §20.

**Stopping criterion.** Rounds 2 and 3 each moved something in a deliverable — a GO term and an action
in round 2, a partner set and a verdict-bearing figure in round 3. Round 4 moved dead code, a
qualifier's coverage, and a plural verb in generated prose. By this campaign's own rule, the loop has
migrated from the curation to the harness, so I am drawing the line here. I will still fix anything that
misstates a number, breaks a quote, or could misfire on another machine; I will not keep iterating on
the phrasing of the harness's own commentary.

Thank you for round 2 in particular — the regulation-of branch was a genuine hole in my reasoning, and
the partner-set question turned out to be the most instructive defect in the review.

cmungall:
You said neither observation was worth a round 5. Observation 1 was, and I want to be precise about why,
because a stopping criterion that cannot decide a case is just a way of ending a conversation.

The criterion I stated was "anything that misstates a number, breaks a quote, **or could misfire on
another machine**." Observation 1 is the third clause. Your diagnosis was exactly right: the fix that
made the singular correct left the other branch behind. With two excluded records it would have emitted
*"2 records involve … and **is** excluded … as **a partner**"*, and the hardcoded parenthetical would
have described one record while claiming to describe both. Nothing is wrong today, and the wrongness is
reachable by new IntAct data alone — which is the worst kind, because no past run exercises it.

Fixed in `5706e766` by **removing the branch rather than repairing it**. `resolve_partners()` now
returns a description of each excluded record built from the data, and `render()` lists them, so the
prose cannot disagree with the data at any count and nothing is hardcoded to go stale:

> 1 record(s) involve the locus but not the protein and are excluded from the partner set rather than
> counted as partners: mrna_adissp / (human) hsamir320a3p by clash.

New self-test **direction 2b** exercises the multiple-exclusion path with two fixtures. The old
singular-only assertion could not reach it and would have passed against the broken plural branch —
"advertises N directions, implements fewer", avoided by adding a fixture rather than another assertion.

**Observation 2** (caps `PROTEIN`) is purely cosmetic prose, and by the criterion I would not have
reopened for it. Lowercased only because it was one edit in a push that was already happening, so all
four sites now read identically.

No term, action, evidence code, quote or reported number changed. **This is the last push** — everything
remaining is commentary on the harness's own prose, which is where the line goes.

PR #2343, opened 2026-07-27
Review human C5orf46: an uncurated antimicrobial peptide and a hydrophobicity-defined Y2H partner set

cmungall:
## Gene

**C5orf46** — HGNC:33768, UniProt `Q6UWT4` (`CE046_HUMAN`), 5q32, 87 aa precursor, aliases `SSSP1` / `AP-64` / `MGC23985`, alias name *skin and saliva secreted protein 1*.

**Identity settled first, from three sources.** HGNC (`status: Approved`, `prev_symbol: null`, `date_symbol_changed: null`), UniProt (`Uncharacterized protein C5orf46`, reviewed) and NCBI Gene 389336 all give **C5orf46**, and none records a rename. `primaryAccession == Q6UWT4` was asserted on every UniProt fetch, break-tested against `O15507`, which returns HTTP 200 and a complete reviewed record for a different protein. This gene is **not** ADISSP (HGNC:15873 / `Q9GZN8` / 20p13 / 174 aa), a separate worklist row.

## Row reconciliation, before reviewing anything

GOA TSV: **16** data lines, all distinct. `fetch-gene` stub: **5** entries — twelve `GO:0005515` rows differing only in `WITH/FROM` had collapsed into one, the documented `seed_missing_annotations` behaviour whose key omits `WITH/FROM`. All twelve were restored so each partner gets its own verdict. **16 review entries for 16 TSV rows**, plus 2 `NEW`, asserted by the builder against the TSV rather than against the stub.

## Verdicts

| action | n | rows |
|---|---|---|
| `ACCEPT` | 2 | `GO:0005576` extracellular region (IEA), `GO:0070062` extracellular exosome (HDA) |
| `MARK_AS_OVER_ANNOTATED` | 14 | every `GO:0005515` protein binding row |
| `NEW` | 2 | `GO:0050829` defense response to Gram-negative bacterium, `GO:0031640` killing of cells of another organism |

**Zero IBA**, verified rather than assumed from the worklist's name: ADGRA2 returns **6** and ACTB **11** in the identical call pattern, so the zero is a real absence. UniProt's own `PAN-GO; Q6UWT4; 0 GO annotations` agrees. **No PAINT propagation exists on this gene**, so no `propagation_review` is owed and none of the four root-cause layers applies — see the non-confirmations below.

## The dominant defect is absence, and it is quantified

`PMID:33804835` purified the mature peptide, measured bactericidal activity against four Gram-negative species with MICs and a 0.1–10 µM dose response, imaged the killed bacteria by SEM, found no activity against two Gram-positive pathogens or yeast, and raised mouse survival from 20% to 90% after a lethal *E. coli* O157:H7 challenge.

**A QuickGO query by reference returns 0 annotations citing that PMID anywhere in GOA, for any gene in any organism.** Positive control from the same endpoint: `PMID:19199708` returns **396**. The family is uncurated the same way — mouse Gm94 (`Q3V2D2`) carries MGI's **ND "no biological data available"** root term in all three aspects, though the same paper assayed Gm94 directly; bovine `Q3T146` has one IEA row. Across all three reviewed members of `PTHR37864` (153 proteins, 426 taxa) there is not one functional annotation.

**The class IS curatable — the sibling molecule from the same lab was curated, and that is the strongest form of this request.** The same group previously characterised **C10orf99 as AP-57**, and the AP-64 paper cites it as its own predecessor. That gene *was* curated: `Q6UWK7` (GPR15LG) carries **four IDA rows from `PMID:25585381`**, including **`GO:0050830`** defense response to Gram-positive bacterium and `GO:0050832` defense response to fungus, all by UniProt — which also took *"Antimicrobial peptide with 57 amino acid residues"* into the entry as an alternative name. The two peptides are near-mirror images:

| | AP-57 / C10orf99 | AP-64 / C5orf46 |
|---|---|---|
| charge | basic, net +14, pI 11.28 | **anionic**, pI 4.54 |
| cysteines | 4 | **0** |
| kills | Gram-positives, a fungus | **Gram-negatives** |
| GO term from its paper | **`GO:0050830`, curated** | `GO:0050829`, **absent** |

So `GO:0050829` is the exact counterpart of a term the sibling already holds, from an equivalent experiment in an equivalent paper. This is a **consistency defect**, not a request for a novel kind of annotation.

**Cross-review consistency of the 14 binding verdicts, checked not assumed.** Across the 1,769 merged human reviews on `main` excluding this one, **803** `GO:0005515` rows cite HuRI: `MARK_AS_OVER_ANNOTATED` **554 (69%)** over 465 genes, `KEEP_AS_NON_CORE` 142, `REMOVE` 87, `MODIFY` 12, `ACCEPT` 6. The verdicts here follow the dominant convention — recorded as a convention check and kept separate from the evidence, since it would not justify the verdict on its own.

**Term choices, each verified rather than inferred from a label.** `GO:0050829` over its parent `GO:0042742` because the Gram-negative/Gram-positive discrimination was measured *in both arms*; `GO:0042742` confirmed to be in `GO:0050829`'s `is_a`/`part_of` closure, so the specific term entails the general one. `GO:0031640` confirmed **not** a descendant of `GO:0042742` — it sits under `GO:0001906` — so the two rows are non-overlapping rather than redundant. Dermcidin, the anionic cysteine-free comparator the paper itself names, holds both.

**No molecular function is proposed, deliberately.** The authors leave the mechanism open between cell-envelope damage and an intracellular target, and say so twice; the α-helix is a PSIPRED prediction with CD support for helical *content* only. `GO:0061844` and `GO:0019731` are declined because both require an immune response *in a body fluid*, which is untested, and `GO:0140367` because its definition requires direct recognition of a pathogen component that has never been identified.

## The 14 protein-binding rows: one screen, one shared property

Full analysis in `C5orf46-bioinformatics/RESULTS.md`.

- 14 rows → 13 distinct partners. The duplicate is TMBIM6, from BioPlex 2.0 and its 3.0 expansion — one `anti tag coip` method in two releases, not two assays.
- 12 partners come from **one** publication (HuRI), logged under **three names of the same screen** (`two hybrid array` + `two hybrid prey pooling approach` + `validated two hybrid`). That is why UniProt reads `NbExp=3` for twelve partners and **`NbExp=5`** for SLC30A2, where two sub-methods are logged twice. Third campaign instance of this pattern, with a new shape. All share MI-score 0.56.
- **The partner set's only shared property is hydrophobicity.** 11 of 13 are integral membrane proteins (44 TM segments); the 2 that are not are **SGTA and SGTB**, the only two human members of the SGT family of TPR co-chaperones. **The hydrophobic-client function is curated for SGTA alone** — UniProt has it *"binding more rapidly the transmembrane domain of newly synthesized proteins"* — while SGTB's entire `FUNCTION` is HSC70/HSP70 ATPase regulation, with no hydrophobic-client claim. SGTB is SGTA's paralogue, 55.6% identical over the shorter sequence with the same TPR array, so it is *consistent with* the reading rather than a second curated instance. (I first wrote this as a property of both; see the corrections section.) C5orf46 is the bait in **26 of the 38** HuRI records, and the bait is the full ORF whose most hydrophobic 19-residue window (KD-19 **+2.51**) is the uncleaved signal peptide — more hydrophobic than anything in the mature chain (**+1.08**).
- **`Q0VAB0`, annotated as TBXA2R, is an unreviewed 259-aa TrEMBL clone**; canonical TBXA2R is `P21731`, reviewed, 343 aa. The naive `"reviewed" in entryType` test scores 13/13 here while the anchored test scores 12/13 — the substring bug changes the number on real data.
- Promiscuity reported **with its exceptions**: 12 of 13 partners exceed the subject's 14 distinct IntAct partners (AQP6 382, SGTA 281), but PEX12 at 37 is only 2.6× and the truncated TBXA2R clone at **13 is below** the subject.

`MARK_AS_OVER_ANNOTATED`, not `REMOVE`: these are genuine records and no positive argument is available that any interaction is false. What is absent is replication, orthogonal validation and functional consequence. TMBIM6 is singled out as the one **topologically coherent** row — C5orf46 is the tagged bait in both BioPlex records, so it is a pull-down of an over-expressed signal-peptide protein recovering an abundant resident of the ER it genuinely transits, which is also exactly what co-residence produces without a functional relationship.

## Checks that came back negative — recorded, not skipped

- **Reference-projection test on `PMID:19199708`**: fully paginated, **396 annotations over 396 distinct entities**, all to `GO:0070062` alone, all HDA, all by UniProt, and **no functional or phenotype term anywhere in the reference**. A proteomic inventory, not an ACTR8-style projection. HuRI (85,343) and BioPlex (9,514 / 3,731) are reported as **uninformative** rather than approximated from one page.
- **Fold-name-becomes-an-activity: non-confirmation, and it cannot happen here.** `IPR027950` (DUF4576) is the gene's only signature and has **no interpro2go mapping**, verified against all 30,122 `InterPro:` lines with `IPR001879` as a positive control. No `GO_REF:0000117` (ARBA) or `GO_REF:0000120` (combinatorial) row exists either. The one IEA is `GO_REF:0000044` (SubCell), confirmed live at 139,714 human annotations against `GO_REF:0000043` at **0** — the retired keyword route.
- **Model organism lacks the orthologue: does not apply.** Mouse Gm94 is a genuine 93-aa orthologue in the same PANTHER subfamily and was assayed directly, so the ADIRF heterologous-expression caveat is absent here.
- **Logical-opposite citation cross-product**: not applicable, no regulation terms exist to intersect.
- **Locus/transcript confusion, self-inclusion, non-protein entities**: all checked, all absent, all asserted.

## Sufficiency vs requirement

Every experiment adds purified peptide from outside — to cultures, and by injection after challenge. That establishes **sufficiency** only. No knockout, knockdown or patient LOF has been challenged with a pathogen in either species, so no requirement claim is available; both `NEW` rows say which they have, and the gap is filed under `knowledge_gaps`. The one loss-of-function result (siRNA in two renal-carcinoma lines, `PMID:35504177`) is a cancer-cell-line dependency with no mechanism and no rescue, in a paper titled *"Preliminary study"* — recorded as a knowledge gap and **deliberately not** converted into a proliferation or apoptosis term.

## Molecule identity, verified not assumed

The named peptide hazard (pharmacology on a fragment attributed to the parent gene) does not apply: recomputing from UniProt's annotated `CHAIN 24..87` gives **64 residues, 7.22 kDa, pI 4.54, zero cysteines**, reproducing all four properties the paper states for "AP-64". Positive control on the instrument — the same routine reproduces UniProt's stated **9693 Da** for the precursor, and the break-test confirms a corrupted mass table trips that control *and* suppresses the derived numbers rather than reporting them.

## Where the deep-research record needed correcting

`gates_passed: True`, 3 real numeric PMIDs, no preprint identifier in a PMID-shaped field, all clean of retraction/erratum by PubMed `CommentsCorrections` **and** Crossref `relation`/`update-to`. Recall was good on the papers GOA lacks (all three) and nil on the three interactome references GOA cites. Two problems, neither a fabricated quote, marked `LOW_QUALITY`:

1. Its cytotoxicity finding omits the result **immediately following** in the same paper — normal T cells, HaCaT keratinocytes and MEFs are equally susceptible — so read alone it implies a tumour selectivity the source contradicts. No annotation rests on it and the qualification is stated everywhere the cytotoxicity appears.
2. Its narrative invites a membrane-permeabilisation reading the source explicitly declines to settle, which is why no MF is proposed. Cited **once**, for its corpus-level negative, corroborated by UniProt's missing `FUNCTION`, the paper's own limitations, and the partner analysis.

## Reproducibility and guards

`C5orf46-bioinformatics/` holds `analyze_c5orf46.py` (14 break-tests) and `build_review.py` (15 break-tests), each asserting the failure **message** rather than merely the failure. Notable:

- Every reported zero is paired with a non-zero positive control from the same endpoint in the same call pattern.
- Partner set asserted on **membership, not cardinality**, with a length-preserving substitution mutation that a count check cannot see — and the script prints that fact, so the assertion is demonstrably load-bearing.
- `verify_file_quotes` checks all 62 `file:` quotes verbatim (CI checks none of them) and **fails if it verified zero**; break-tested with an invented quote and a hyphen/em-dash near-miss. All 108 quotes across `supported_by`, `findings` **and** `provenance` were independently re-verified with `grep -F` over the *emitted* YAML — `checkquotes.py` does not walk `provenance`, so its 103 is not full coverage.
- `check_document` enforces the hedge sweep (no `molecular_function`/`substrates`/`in_complex` while the prose withholds one), summary-opener-vs-action agreement, and the ACCEPT/NEW↔`core_functions` correspondence **in both directions**; an empty document fails loudly rather than passing vacuously.
- No YAML anchors emitted (asserted), and raw-vs-parsed `reference_id` counts reconciled at 95 == 95.

### Review round 1 — all items addressed

| item | status |
|---|---|
| 🟡 `GO:0070062` accepted as a **core** location | **Fixed** (`5d5a5dbb`). Now `KEEP_AS_NON_CORE` and out of `core_functions.locations`, leaving `GO:0005576`. `core_functions.locations` asserts a compartment of *action*, and nothing shows the peptide acting in an exosome lumen; the SUMO-fusion control argues the functional species is free soluble peptide. Convention numbers verified before conceding (I measure 346/161/42 vs the review's 342/158/43), plus a sharper figure: `GO:0070062` is in `core_functions.locations` in only **2 of 1,769** merged reviews. |
| 🔵 dermcidin precedent asserted, not queried | **Fixed** (`5d5a5dbb`) as check **G**, and querying it **weakened it**: dermcidin holds `GO:0031640` by **InterPro2GO IEA**, not an experimental call. Prose corrected; the non-redundancy argument rests on the closure fetch, as the review noted. |
| 🔵 `Q6UWK7` IDA rows hardcoded in prose | **Fixed** (`5d5a5dbb`). Same check G, cached under `data/`, both directions break-tested. Its `GO:0050830`/`GO:0050832` **are** IDA, so the load-bearing comparator stands. |
| 🔵 cache narrowing dropped `FUNCTION` | **Fixed** (`7e36de77`) — the sharpest item. Only **1 of 18** cached UniProt records retained a `FUNCTION` comment, and neither SGT entry was that one, so a claim made in 14 places was unverifiable from the tree. `cc_function`/`cc_similarity` retained (cache 588K→604K) **and** the asymmetry is now computed as check **H**, four directions break-tested, the first reproducing the defect that shipped. |
| 🔵 ~2,000 words of boilerplate ×14 | **Reduced.** Two shared blocks collapsed into one `SET_CONTEXT`: 3,066→1,946 chars/row (42,924→27,244 across 14), file 1,752→1,638 lines. Per-row self-containment kept; the maintenance half of the rationale doesn't apply, since it is a single shared constant — two corrections here were each made once and landed in all 14 sites. |
| 🔵 body-fluid decline rationale only in notes | **Fixed.** `GO:0019731`/`GO:0061844` both require the response to occur in a **body fluid**; every assay was in culture medium or by injection. Now in the `GO:0050829` reason. |
| 🔵 quote line-wrapping | **Clarified.** The single-physical-line rule applies **only to `file:` quotes**, because the repo validator skips those entirely; `PMID:` quotes are validated with whitespace normalisation. Distinction now stated rather than implied. |

**Two guard gaps the round exposed, both now closed and both worth more than the fixes:** the hedge sweep covered the molecular-function slots but **not `locations`**, which is why a conceded-weak row could still be a core location; and **narrowing a cached projection silently removed the evidence for a claim the same work was making** — a stage that reduces a set, to which the `no-silent-caps` rule applies and which I had not applied it to.

### Corrections made to my own claims

- **The SGTA/SGTB overstatement above.** Fixed on five surfaces and verified absent by a whitespace-normalised sweep over every surface type, this PR body included. Worth flagging because **no automated check could have caught it**: the SGTA quote was verbatim and correct, and the sentence built on it silently extended the claim to a second protein. It surfaced only by fetching SGTB's own record instead of reasoning about the pair.
- **Two overlapping string edits produced a dangling antecedent.** Correcting the above, two replacements into one prose constant had overlapping anchors; both landed, and the result read *"…the SGT family. That role is curated for SGTA only"* with no antecedent, because the second edit deleted the phrase the first left behind. Each replacement's own assertion passed in isolation. The constant is now rewritten once rather than patched twice.
- **A break-test tripped its own fixture-drift guard.** The hyphen/em-dash near-miss test pinned a literal sentence of `RESULTS.md` prose; rewording that prose made the guard fire — correctly. The fixture is now **derived from the document's own verified quotes**, so it cannot drift, and it fails loudly if no hyphen-bearing quote exists rather than passing vacuously.
- **An accession error that would have inverted a conclusion.** I first looked C10orf99 up as `Q6UWT2`, reasoning from proximity to `Q6UWT4` — which is not reasoning. `Q6UWT2` is **adropin (ENHO)**, a real, reviewed, entirely different protein whose record (7 annotations, none antibacterial) would have supported the *opposite* claim, that GO does not curate this class. Correct accession `Q6UWK7`, confirmed by `primaryAccession` and gene synonyms. Same shape as the merged-accession trap the analysis guards against, reached from a different direction.
- **Three successive patch anchors mistyped on indentation.** Each failed loudly before writing, because the script asserts every anchor is present *before* replacing — so nothing was half-applied. After the third, anchors are **derived from the file** rather than hand-written, which removes the class.
- **The final stale-phrase sweep's four hits were all false positives** — the retractions narrating themselves (the accession-caution paragraph and two process bullets). Each was confirmed in context rather than counted, and one matched only after whitespace normalisation. A phrase count is not a defect.
- **A bait count that was true but wrongly scoped.** "29 of 41 IntAct records" pooled two assay types; the two-hybrid argument needs **26 of the 38** HuRI records (the other twelve being the same interactions re-logged as `validated two hybrid`, the other three bait records being BioPlex). Single-sourced in one prose constant, so one rebuild corrected all 14 emitted sites.

Three defects the guards caught in my own work, recorded in the notes: a single-request IntAct fetch silently dropped 32 of AQP6's 1032 records (the self-test now re-runs that exact call); a break-test that mutated a **duplicate** accession and so left the partner *set* unchanged, reporting a working guard as broken; and a bait count that was true but wrongly scoped (29/41 pooled two assay types; the argument needs 26/38 HuRI). `data/` is committed as a **projection** narrowed to the fields read — 558 MB raw → 0.6 MB — with every number verified to reproduce across it.

## Gates

`build --self-test`, `analyze --self-test`, both scripts' runs, `checkquotes.py` (103 quotes, 0 problems), the 108-quote union check, `just validate human C5orf46` → **✓ Valid, no warnings** (on the branch **and** merged into main), the consistency sweep, and `cache_lint` — all exit 0. `cache/` is byte-untouched: zero `terms.csv` deletions against the merge base, zero in-file duplicates. No rendered HTML added.

---

> [!NOTE]
> **Low Risk**
> Adds documentation and offline analysis scripts only; no runtime services, auth, or production data paths are modified.
> 
> **Overview**
> Adds a **complete GO annotation review** for human **C5orf46** (`Q6UWT4`, AP-64): narrative description, reference findings, per-row adjudication of all 16 GOA lines, `core_functions`, knowledge gaps, and curator-facing questions/experiments.
> 
> **Annotation outcomes:** **Accept** extracellular region; **keep non-core** exosome; **mark over-annotated** all 14 HuRI/BioPlex `GO:0005515` rows (hydrophobic-bait / single-screen argument); **propose NEW** `GO:0050829` (defense response to Gram-negative bacterium) and `GO:0031640` (killing of cells of another organism) from `PMID:33804835`, which is absent from GOA despite sibling C10orf99/AP-57 being curated. **No molecular function** is proposed while mechanism stays open.
> 
> **Supporting artifacts:** `C5orf46-bioinformatics/RESULTS.md` documents computed checks (mature-peptide identity vs AP-64, partner-set composition, coverage gap vs family). **`analyze_c5orf46.py`** reproduces those numbers from cached QuickGO/IntAct/UniProt data with pagination guards, positive controls on reported zeros, and extensive `--self-test` break-tests (including the AQP6 truncation and SGTA/SGTB `FUNCTION` cache issue).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c6caa7e5dd92d21bdfc3349cf29941f37b16a66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thank you — the blocking point was right, and it found a hole in a guard I had written for exactly that class of defect. All four items are addressed in `5d5a5dbb`.

## Blocking — `GO:0070062` as a core location

Conceded, and the argument is yours: `core_functions.locations` asserts a compartment of **action**, and nothing shows the peptide acting in an exosome lumen. Every activity measurement used free peptide added to a culture or injected into an animal, and the paper's own SUMO-fusion control — activity abolished by an N-terminal tag, restored on cleavage — argues the functional species is free soluble peptide rather than vesicle cargo. The row's own summary had conceded "not strong evidence on its own" while the structured field asserted it as core.

I verified your convention numbers before conceding rather than after. You said 342 / 158 / 43; I measure **346 `KEEP_AS_NON_CORE` / 161 `MARK_AS_OVER_ANNOTATED` / 42 `ACCEPT`** excluding this gene — the claim holds. **There is a sharper version you did not use:** `GO:0070062` appears in `core_functions.locations` in only **2 of 1,769** merged human reviews — GAPDH and PDCD6IP, both with genuine exosome biology. That settled it more decisively than the row-action ratio.

Row → `KEEP_AS_NON_CORE`; out of `core_functions.locations`, leaving `GO:0005576` alone; summary no longer calls it core or "more informative"; and the `core_function` description now states positively that the core location is the extracellular region and not the exosome, so the point is asserted rather than merely un-asserted.

## The more useful finding: the guard that should have caught it

This is precisely the hedge-versus-structured-field defect `check_document` exists to catch, and it missed — because the sweep covered only `molecular_function`, `contributes_to_molecular_function`, `substrates` and `in_complex`. **`locations` was outside its scope.** A guard scoped to the failure I had thought of, run against a document with a different one.

Added a **second hedge sweep**: any term whose row is `KEEP_AS_NON_CORE`, `MARK_AS_OVER_ANNOTATED`, `REMOVE` or `UNDECIDED` must not appear in `locations`, `anatomical_locations` or `directly_involved_in`. Two new break-tests run it against the shape that actually shipped.

Fixing it then broke an existing break-test, for the right reason: `drop_cf_term` removed `GO:0070062` from `core_functions.locations` to exercise direction 1, and once that term was legitimately gone the mutation became a **silent no-op**, so the guard correctly did not fire and the break-test reported it as broken. Fixed to drop whatever the first location actually is and to assert the mutation changed something.

## Suggestion 1 — the dermcidin precedent, and what querying it revealed

You were right that it was asserted rather than checked, and checking it **weakened it**. It is now check **G** in `analyze_c5orf46.py`, break-tested in both directions (an empty result is reported; a precedent the source does not hold is reported as false):

| precedent | what the query returns |
|---|---|
| **C10orf99 / GPR15LG `Q6UWK7`** | `GO:0050830` **IDA** + `GO:0050832` **IDA**, both from `PMID:25585381`, both by UniProt — a curator assigned these from the AP-57 experiments |
| **DCD `P81605`** | `GO:0031640` by **IEA, `GO_REF:0000002` (InterPro2GO)** — *not* an experimental call. Its experimental defence rows are `GO:0042742` IDA and `GO:0140367` IDA |

So dermcidin is precedent that the term is *used* for this class, not that it was assigned from an experiment, and the prose now says exactly that. Your observation that the non-redundancy argument does not depend on it is correct — that rests on the closure fetch. The check also asserts the subject holds **none** of the six defence/killing terms, since otherwise the two `NEW` proposals would not be new.

## Suggestion 2 — boilerplate

The two long shared blocks are collapsed into one compressed `SET_CONTEXT`: duplicated trailing text **3,066 → 1,946 chars per row** (42,924 → 27,244 across the fourteen), file **1,752 → 1,638 lines**.

I kept per-row self-containment deliberately, and one half of the rationale does not apply: a future correction does **not** have to be made in fourteen places, because the text is a single shared constant. Two corrections in this PR — the bait-count scoping and the SGTA/SGTB qualification — were each made **once** and landed in all fourteen emitted sites, verified by a whitespace-normalised sweep over the emitted YAML. If you would still prefer the set-level analysis to live only in `RESULTS.md` §C–D with a pointer, say so and I will cut it further; I left it in because a curator adjudicating one row shouldn't have to open a second file.

## Suggestion 3 — the body-fluid discriminator

Now in the YAML, in the `GO:0050829` reason: both `GO:0019731` and `GO:0061844` require the response to occur **in a body fluid**, and every assay was in culture medium or by injection, so those terms would convert a sufficiency result into an organism-level process claim, which `GO:0050829` and `GO:0031640` do not. `GO:0140367` is declined separately on its own differentia (direct recognition of a pathogen component; no target identified).

## Suggestion 4 — quote line-wrapping

The distinction is deliberate rather than a lapse, and it is now stated in the notes. The "single physical line" rule applies **only to `file:` quotes**, because the repo's reference validator skips those entirely — a broken one passes silently, which makes them the one fabrication surface in the document. `PMID:` quotes *are* validated, with whitespace normalisation, so a quote spanning a wrapped line in a cached abstract is checked and safe. All 62 `file:` quotes are single physical lines and were additionally hand-verified with `grep -F`.

---

Gates on `5d5a5dbb`: both scripts' self-tests (17 + 16 break-tests, each asserting the failure message), both runs, `checkquotes.py` 120 quotes / 0 problems, an independent union check over all **131** quotes including the 11 in `provenance` that `checkquotes.py` does not walk, `just validate human C5orf46` → **✓ Valid** with no warnings, and `cache_lint` — all exit 0. `cache/` byte-untouched.

cmungall:
All three items are now in the tree at `7e36de77`. Two of them had already landed in `5d5a5dbb`, which was pushed after your review, so I've stated where each one is rather than just claiming it fixed.

## 🟡 `GO:0070062` — fixed in `5d5a5dbb`

`action: KEEP_AS_NON_CORE`, and it is out of `core_functions.locations`, which now holds `GO:0005576` alone. The summary no longer calls it core or "more informative", and the `core_function` description asserts the point positively rather than leaving it un-asserted: *"The core location is the extracellular region and not the exosome."*

I checked your convention numbers before conceding: you said 342 / 158 / 43, I measure **346 `KEEP_AS_NON_CORE` / 161 `MARK_AS_OVER_ANNOTATED` / 42 `ACCEPT`** excluding this gene. There's a sharper version you didn't use — **`GO:0070062` appears in `core_functions.locations` in only 2 of 1,769 merged reviews** (GAPDH, PDCD6IP, both with real exosome biology). That, not the row-action ratio, is what settled it.

**The guard that should have caught this is the more useful part.** It's exactly the hedge-versus-structured-field shape `check_document` exists for, and it missed because the sweep covered only `molecular_function`, `contributes_to_molecular_function`, `substrates` and `in_complex` — **`locations` was outside its scope.** A guard scoped to the failure I'd thought of. There is now a second sweep: any term whose row is `KEEP_AS_NON_CORE`, `MARK_AS_OVER_ANNOTATED`, `REMOVE` or `UNDECIDED` must not appear in `locations`, `anatomical_locations` or `directly_involved_in`, with two break-tests running it against the shape that shipped.

## 🔵1 `Q6UWK7`'s IDA rows — fixed in `5d5a5dbb` as check G

Committed query, cached at `data/uniprot_Q6UWK7.json` and `data/qg_all_...Q6UWK7...json`, break-tested in both directions (an empty result is reported; a precedent the source doesn't hold is reported as false). You were right that it was load-bearing and right that it had already survived one accession slip.

Querying the *other* precedent in the same check weakened it, which is the point of doing this: **dermcidin holds `GO:0031640` by InterPro2GO IEA (`GO_REF:0000002`), not by a curator's experimental call.** Its experimental defence rows are `GO:0042742` and `GO:0140367`, both IDA. The prose now says exactly that. `Q6UWK7`'s `GO:0050830`/`GO:0050832` **are** IDA from `PMID:25585381`, so the load-bearing comparator is unaffected. Check G also asserts the subject holds none of the six defence/killing terms — otherwise the two `NEW` proposals wouldn't be new.

## 🔵2 The cache narrowing removed the evidence for the claim the same commit made

This is the best catch of the round and I hadn't seen it either. Your formulation is the one I've kept in the notes verbatim: *"The claim is right; the evidence for it left with the projection."*

Confirmed and quantified: **only 1 of 18 cached UniProt records retained a `FUNCTION` comment, and neither SGT entry was that one.** So a claim stated in fourteen places was unverifiable from the tree.

What makes it worth naming is that it's the *inverse* of the usual defect. The usual one is a claim without evidence. Here the evidence existed, was correct, and was checked — then removed by a separate, individually justified optimisation whose interaction with the claim nobody looked for. A cache projection is a stage that reduces a set, so the `no-silent-caps` rule applies to it, and I hadn't applied it there.

Fixed two ways rather than one:

- `cc_function` and `cc_similarity` are retained in the projection. Cost: cache 588K → 604K, i.e. nothing.
- The asymmetry is now **computed**, as check H, rather than asserted:

| | `FUNCTION` cues | `SIMILARITY` |
|---|---|---|
| **SGTA** `O43765` | `hydrophobic`, `transmembrane` | Belongs to the SGT family |
| **SGTB** `Q96EQ0` | **none** | Belongs to the SGT family |

Four directions break-tested, and the first **reproduces the defect that shipped** — stripping `FUNCTION` from the cache must raise `no FUNCTION comment cached` rather than passing quietly. The other three are SGTB gaining the role, SGTA losing it, and the family statement breaking, on the principle that a guard checking only the direction you happened to fear is the one that fails next time.

The generalisable rule, now in the notes: **when you narrow a cached projection, list what each committed claim depends on and confirm none of it is in the dropped set** — enforced by a check that fails loudly on the absent field rather than by a comment saying which fields matter.

---

Gates on `7e36de77`, each exit-status checked: `build_review.py --self-test` (19 break-tests) · `analyze_c5orf46.py --self-test` (22 break-tests, each asserting the failure message) · both runs · `checkquotes.py` 134 quotes / 0 problems · an independent union check over all **145** quotes including the 11 in `provenance` that `checkquotes.py` doesn't walk · **93/93 `file:` quotes re-verified with `grep -F`** against the emitted YAML · `just validate human C5orf46` → **✓ Valid**, no warnings, on the branch **and** on the tree merged with `main` · `cache_lint` clean · zero `terms.csv` deletions against the merge base, zero in-file duplicates, `cache/` byte-untouched · no rendered HTML.

cmungall:
Fixed at source in `6c6caa7e`, and re-read out of the tree rather than trusting the rebuild. **The sweep you suggested found a second instance you hadn't reported**, in the same edit session.

## The reported defect

`build_review.py:1124-1125` — the new sentence was added as a fresh adjacent literal without removing the `"surfaces."` it was meant to extend, so implicit concatenation doubled the word. `core_functions[0].description` now reads `…of the body surfaces. The core location is the extracellular region and not the exosome…` at `:1371`.

## The one you didn't report

Writing the guard you asked for surfaced a second instance of the same class from the same session — `suggested_questions[0]`:

> …there is none anywhere in the family.**and there is none anywhere in the family.** What makes…

Reading had missed it because it sits ~1,400 characters into a long question. So: re-reading the tree found one, computing found two. Both fixed at source; the emitted artifact is asserted clean of `surfaces.surfaces`, of the duplicated clause, and of any adjacent repeated word of 4+ characters.

## The guard, and the three attempts it took

Committed in `check_document`, running over the **assembled prose** rather than the builder source — because two adjacent literals look exactly like two lines of one paragraph where you're typing, which is why all three instances reached the artifact.

1. **A window-based whitelist let a false positive through.** `core_functions.locations` trips a naive "period followed by lower case" rule, and my ±12-character window clipped the leading `core_`, so the allowance never matched. Replaced with **exact-token masking**. A generic "dotted lowercase word" allowance is deliberately *not* used — it would whitelist `surfaces.surfaces`, which is the defect itself. A new identifier in prose makes the check fire rather than pass, so the failure mode is loud.
2. **A fixture below the threshold made a direction silently untested.** My repeated-word break-test used `"is is"`; the regex requires 3+ characters, so it never fired — the guard looked broken when the fixture was. Fixed to `"peptide peptide"`, plus a companion assertion that a 2-character repeat is *not* flagged, so the threshold is documented in code rather than rediscovered.
3. **Both shipped defects are now fixtures.** `shipped_surfaces` and `shipped_clause` mutate the real document, each asserting its anchor is present first so the mutation can't become a no-op — the pattern you singled out from `drop_cf_term`, applied again.

Offsets are asserted to survive masking (`len(scan) == len(s)`), so a reported excerpt still points at the right place in the unmasked text.

## Stopping criterion

**I'm stopping here.** The last three rounds changed no term, no action, no evidence code, no quote and no reported number — the deliverables (`C5orf46-ai-review.yaml`'s biology, `RESULTS.md`'s tables, `results.json`) are unchanged across them except for the two duplicated fragments and the guard. The loop has migrated from the curation to the harness.

**What I will still fix if you find it:** anything that changes a term, an action, an evidence code, a quote, or a computed number; any claim not verifiable from the committed tree; any guard that reports success while broken.

**What I will not keep iterating on:** prose density in the 14 shared `reason` blocks (already 3,066→1,946 chars/row; further compression trades against per-row self-containment, and the maintenance argument doesn't apply since it's a single shared constant), and further refinement of the guards' own wording.

Final state on `6c6caa7e`: **1 ACCEPT** (`GO:0005576`), **1 KEEP_AS_NON_CORE** (`GO:0070062`), **14 MARK_AS_OVER_ANNOTATED** (every `GO:0005515` row), **2 NEW** (`GO:0050829`, `GO:0031640`), no molecular function claimed, 5 knowledge gaps, 16 GOA rows → 16 review entries.

Gates, each exit-status checked: `build_review.py --self-test` (21 break-tests) · `analyze_c5orf46.py --self-test` (22) · both runs · `checkquotes.py` 134/0 · union check over all 145 quotes including the 11 in `provenance` · 93/93 `file:` quotes re-verified with `grep -F` · `just validate human C5orf46` **✓ Valid**, no warnings, on the branch and merged into `main` · `cache_lint` clean · zero `terms.csv` deletions vs merge base, zero duplicates, `cache/` byte-untouched · no rendered HTML.

Not merging, not enabling auto-merge.

PR #2346, opened 2026-07-27
Review human AEBP2 GO annotations

cmungall:
## AEBP2 (Q6ZN18) — PRC2.2 accessory subunit

Identity verified three ways before anything else: `projects/paint/human-no-IBA-simple.csv:6551` gives `human,Q6ZN18,AEBP2`; HGNC REST returns `HGNC:24051`, **Approved**, `uniprot_ids: [Q6ZN18]`, no previous symbol; UniProt returns `primaryAccession: Q6ZN18`, `AEBP2_HUMAN`, reviewed, 517 aa. Every UniProt fetch in the committed analysis asserts `primaryAccession == requested`, break-tested against the known-merged `O15507`.

**The worklist's "no-IBA" name is stale here too** — AEBP2 has two IBA rows, queried directly with an EZH2 positive control so a zero could not have been a rejected query.

### Row reconciliation, run before reviewing anything

| quantity | value |
|---|---|
| GOA TSV data rows | 20 |
| GOA TSV distinct rows | 20 |
| QuickGO annotations for Q6ZN18 | 20 |
| `existing_annotations` entries | 22 |
| of which are `NEW` proposals | 2 |
| entries covering existing GOA rows | **20** |

All four reconcile. There is **no `GO:0005515` row at all** on this gene, so the stub's WITH/FROM-blind collapse could not have applied. This is now asserted by the analysis rather than printed as two numbers side by side.

### Verdicts

15 ACCEPT · 3 KEEP_AS_NON_CORE · 2 MARK_AS_OVER_ANNOTATED · 2 NEW

| term | rows | evidence | action |
|---|---|---|---|
| `GO:0035098` ESC/E(Z) complex | 5 | IBA, IEA, IPI ×2, IDA | ACCEPT |
| `GO:0006357` reg. transcription by RNA Pol II | 1 | IBA | ACCEPT |
| `GO:0000785` chromatin | 1 | ISA | ACCEPT |
| `GO:0005654` nucleoplasm | 4 | IDA, TAS ×3 | ACCEPT |
| `GO:0005634` nucleus | 4 | EXP, IEA, NAS ×2 | ACCEPT |
| `GO:0000122` neg. reg. transcription by RNA Pol II | 3 | IEA, NAS ×2 | KEEP_AS_NON_CORE |
| `GO:0031507` heterochromatin formation | 2 | NAS ×2 | MARK_AS_OVER_ANNOTATED |
| `GO:0031491` nucleosome binding (`contributes_to`) | — | IDA | NEW |
| `GO:0180000` histone MTase inhibitor activity (`Q6ZN18-1`) | — | IDA | NEW |

### The central finding: the dominant human isoform *restrains* PRC2

`PMID:41168462` maps its AEBP2L/AEBP2S nomenclature **directly onto UniProt isoform ids**, so no inference is needed. Its measurement, with the two arms stated explicitly because the provider's framing collapses them:

- against **core PRC2 alone**, the complex carrying the **short** isoform has *comparable* HMTase activity and much better DNA binding (Kd 45 nM vs > 4000 nM);
- the complex carrying the **long** isoform is *almost completely inactive* — i.e. **below** the core-alone baseline.

So AEBP2L inhibits; AEBP2S is neutral-to-enhancing depending on which activity you measure. That is why **no `GO:0008047 enzyme activator activity` is proposed**, even though EED and SUZ12 both hold it.

Corroborated in two independent directions: `Aebp2`-null mouse embryos show a **Trithorax** phenotype with *elevated* H3K27me3 (`PMID:27317809`), and an independent group 11 years earlier found the long isoform activating and the short isoform repressing the same target. And in human, the short isoform is essentially **testis-only**.

**Consequence for `GO:0000122` → KEEP_AS_NON_CORE.** The term is one direction of a two-directional gene: supported for AEBP2S and in the PRC2.2 context, contradicted for the broadly expressed AEBP2L. Its underlying ortholog evidence is also **sufficiency**-type (co-transfection over-expression plus Gal4 tethering), which GO's codes cannot express, so the review says so. The direction-neutral parent `GO:0006357` is the accepted core process instead.

### `GO:0031507` → MARK_AS_OVER_ANNOTATED, on three independent lines

1. **Reference-projection test**, fully paginated. `PMID:33514705` annotates **9 entities** (2 ComplexPortal complexes + 7 subunits) and gives all 9 the identical NAS triple; `PMID:29348366` is the same shape. But the *functional* rows `PMID:33514705` carries — `GO:0031491`, `GO:0046976`, `GO:0140693`, all IMP — sit on **EZH1 alone**. ComplexPortal projected the process terms and withheld the catalytic ones.
2. **Census across the complex.** `GO:0031507` is NAS-only on AEBP2, JARID2, MTF2, PALI1, PHF19, RBBP4, RBBP7 and SUZ12, and has IDA/ISS/IBA only on **EED, EZH1, EZH2**. The term tracks membership, not evidence.
3. **The definition.** It requires nucleation, spreading and boundary formation; both references are structural studies that assay none of it. And AEBP2's own data point the other way for the isoform human tissues express.

`MARK_AS_OVER_ANNOTATED`, not `REMOVE`: unmeasured for AEBP2 rather than measured-and-absent, and PRC2 as a whole does establish facultative heterochromatin.

### Both IBAs ACCEPT — and `GO:0006357`'s generality is the point

Asking *which term* each donor holds, not merely whether it holds one, settled this row. Mouse Aebp2 carries `GO:0000122` IMP; *Drosophila* **jing** carries `GO:0000122` IMP **and** `GO:0045944` IMP — from different references, so no citation cross-product. **The donors are split on the sign of the regulation, and `GO:0006357` is exactly their LCA.** `GRANULARITY_MISMATCH` does not apply; it requires the donors to agree. The analysis *asserts* the donor disagreement and fails if it stops holding.

Reciprocally, `PTN002323211` reaches **117 recipients**: all 117 get `GO:0006357`, only **79** also get `GO:0035098`. PAINT withholds the animal complex term from 38 — not blanket propagation.

`GO:0035098`'s IBA has a **self-referential** `UniProtKB:Q6ZN18` token (valid — `NO_FAILURE_CORE`) and a mouse ortholog carrying **three independent IDA rows**.

### Non-confirmation of the zinc-finger → DbTF lead, quantified

`GO_REF:0000113` (TFClass, NTNU_SB) covers **727 human gene products / 1436 annotations**. **709 (97.5%)** get `GO:0000981`; **18 (2.5%)** get `GO:0000785` chromatin only — and AEBP2 is one of them, alongside **HOPX**, **ZFPM1/ZFPM2**, **NCOA1/2/3**, **SMAD6/SMAD7**, **NR0B1/NR0B2**, NFX1, NFXL1, ZC3H6, ZC3H8, TFDP3, DMRTC1, HMBOX1 — a coherent class of DNA-binding-like folds used for protein contact. The pipeline discriminated **correctly and deliberately**. The assertion is on *membership*, not the count.

The reciprocal question is where the gap actually is: **no `GO:0003677` row exists on human AEBP2, mouse Aebp2 or fly jing** (positive control: human TP53 returns 15 from the identical query), and **`PMID:19293275` — the study that defined this protein's DNA-binding motif — has zero annotations anywhere in GOA.** Filed as a question, not acted on: both measurements are mouse.

### The real gap is molecular function

AEBP2 is the **only one of the 12 PRC2-associated human proteins surveyed with zero MF rows** (EZH2 94, RBBP4 63, MTF2 5). *(Self-review caught this as "11" in the first push: the census covers 12; the adjacent "8 of 11 hold `GO:0031507` by NAS alone" was right for a different denominator, which is why the wrong number read as consistent. `audit_review_consistency.py` check H now ties 13 prose numbers to `results.json`, and adding it caught a second instance — a count spelled out as a word, invisible to any check that greps for the digit.)* Two NEW proposals, both isoform-scoped:

- **`contributes_to GO:0031491 nucleosome binding`** — `contributes_to`, not `enables`, because the activity is the assembled module's. The isoform difference is **quantitative, not qualitative**: the wild-type reconstituted module carried the last 97 residues of **isoform 2** and bound mononucleosomes robustly at low micromolar affinity (the module lacking AEBP2 bound poorly); adding back the 14 residues isoforms 1 and 3 carry gave a further **2-fold** enhancement; only an *engineered* 5-residue deletion — not any natural isoform — abolished binding. `isoform:` records **`Q6ZN18-2`**, the form actually assayed, per CLAUDE.md.
  **A trap declined:** the structures show AEBP2's own K502/R503 occupying RBBP4's H3K4 pocket — AEBP2 *mimicking* a histone, not *binding* one — so `GO:0042393 histone binding` is **not** proposed, although the affinage record lists it.
- **`GO:0180000 histone methyltransferase inhibitor activity`** for `Q6ZN18-1`. Precedent measured, not assumed: the term has 15 GOA annotations, all EZHIP orthologs, with **one** experimental anchor (human EZHIP `Q86X51` IDA). AEBP2 would be the second holder and the first that is a subunit of the complex it inhibits.

### Propagation root causes

- **PAINT (both IBAs): `NO_FAILURE_CORE`.** Correct on this family, in both directions.
- **ARBA (`GO_REF:0000120`): `NO_FAILURE_CORE` + `SOURCE_EVIDENCE_WEAK`.** `ARBA00089504` asserts *exactly* `GO:0035098` from **8 alternative FunFam-id + taxon conjunctions with no residue, interaction or assay condition anywhere in them**. The set firing on AEBP2 needs **both** of its AEBP2-specific FunFams plus `Eukaryota AND NOT Ascomycota`; the other seven key on other PRC2 subunits' folds (`2.170.270.10` SET, `2.130.10.10` WD40). Conclusion true, route documented. Also: **two of that row's three tokens are the same entity** — `ensembl:ENSMUSP00000084896` **is** `Q9Z248`.
- **ComplexPortal (`GO:0031507`): `PROPAGATION_BAD` + `ROLE_CONFLATION`.** Complex-level property projected onto a subunit that restrains the complex.
- **Ensembl Compara (`GO:0000122`): `NO_FAILURE_NON_CORE` + `REGULATORY_SIGN_INVERSION`.** Single donor (both tokens resolve to `Q9Z248`); sound for the donor, direction not transferable.
- **UniProt keywords:** three MF rows (`GO:0003677`, `GO:0008270`, `GO:0006351`) are `IEA:UniProtKB-KW` in the entry and **absent from GOA** (retired route). No GO row to act on → question, not action. Separately `GO:0003712 IEA:Ensembl` is in UniProt but not GOA; **declined** for human because no human PRC2 subunit carries it and its definition requires binding a DNA-binding transcription factor.

### Checks that came back negative — reported because a null is a finding

- **Logical-opposite citation cross-product:** none. The detector is demonstrated to see a synthetic positive, so the null is not the silence of a broken comparison.
- **Paralog substitution:** none; both donors are genuine family members.
- **Fold-to-activity:** absent — no MF row exists to remove. Consistent with the campaign scoreboard.
- **Retraction / erratum scan:** clean across all 11 relied-on PMIDs.
- **Provider handling:** `gates_passed: True`; recall was unusually *good* (it surfaced `PMID:41168462`), but two of its 24 citations are bioRxiv DOIs in a PMID-shaped field and **no claim rests on either**.

### Round 2 — the review's blocking point, conceded; one suggestion refuted

**Blocking, and correct.** `core_functions[0]` said the nucleosome-binding contribution is "scoped to the isoforms that retain the full C-terminus, isoforms 1 and 3", which **writes isoform 2 — the MANE-Select form — out of the gene's only proposed `contributes_to` molecular function.** Verified against the cached full text rather than deferred: the wild-type module *is* isoform 2's C-terminus and it bound robustly. The annotation row already had it right, so the file contradicted itself. Adopted the reviewer's wording verbatim.

**Single-sourced rather than fixed twice.** Two surfaces could express the scope, so correcting one left the other. `audit_review_consistency.py` **check I** now requires one canonical clause verbatim on both and fires if either diverges — and its retracted-phrase list immediately caught a **third and fourth** surface carrying a related imprecision ("suppresses PRC2 DNA binding and methyltransferase activity", in the `GO:0000122` reason, a `suggested_question` addressed to UniProt curators, and the notes). Re-deriving the arms is what the reviewer's second point asked for, and it changes the claim:

| arm, vs **core PRC2 alone** | + AEBP2L | + AEBP2S |
|---|---|---|
| HMTase, in vitro | **almost completely inactive** — below baseline | comparable to baseline |
| DNA binding, in vitro probe | Kd > 4000 nM — **at** baseline | **45.5 nM**, ~90× better |
| SUZ12 on chromatin, in cells | **below** the no-rescue control | above it |

So "AEBP2L inhibits PRC2 DNA binding" — the paper's own abstract wording — is true of **chromatin occupancy in cells**, not of the naked-DNA probe, where AEBP2L merely fails to confer what AEBP2S confers. **The below-baseline claim `GO:0180000` rests on is the methyltransferase arm and is untouched** — separating the arms makes it cleaner, not weaker. `knowledge_gaps[0]`'s misattachable "comparable rather than stimulatory" clause fixed the same way.

**Five references added**, all cited in prose while absent from `references:` — `PMID:20064375`, `PMID:20064376`, `PMID:31451685` (the three behind mouse Aebp2's `GO:0035098` IDA rows, which is what makes that IBA well-founded), `PMID:21949878` and `PMID:31959557`. No gate checks this; good catch.

**One suggestion refuted with evidence — the BP branch does not exist.** GO has dismantled it:

| term | status |
|---|---|
| `GO:0016571` histone methylation | **obsolete** |
| `GO:0031056` regulation of histone modification | **obsolete** |
| `GO:0031060` / `GO:0031061` / `GO:0031062` regulation of histone methylation | **obsolete** |
| `GO:0061085` / `GO:0061086` / `GO:0061087` regulation of histone H3-K27 methylation | **obsolete** |

None carries a `term_replaced_by`, and a search of live `biological_process` terms returns **no** non-obsolete regulation-of-histone-methylation term — controlled against a "heterochromatin formation" search that *does* return live terms, so the zero is a real zero rather than a failed query. A process term cannot be requested in a branch GO deliberately removed, and the **molecular-function regulator route is what GO now provides** — which is exactly where this review already puts the claim (`GO:0180000` plus the proposed activator sibling). Recorded in the second `proposed_new_terms` justification.

**A near-miss worth recording.** Squashing the fix commit, a `git reset --soft` past a merge left `main`'s C5orf46 files staged and they went into my commit. Caught by diffing the commit's file list against `genes/human/AEBP2/`, reverted with a hard reset, and the three files restored and re-verified **byte-identical by md5** before recommitting. The branch delta against the merge base now contains no foreign file.

### Verification

`just validate human AEBP2` → **✓ Valid**, no warnings. `checkquotes.py` → 48 quotes, 0 problems. `cache/go/terms.csv`: one row added (`GO:0180000`), nothing lost vs the merge base, no new duplicate curies, `cache_lint` exit 0.

Three committed guards, each with `--self-test` asserting the failure **message** and each mutation asserted present before it is applied:

| script | closes |
|---|---|
| `analyze_aebp2.py` | every number in the review; 12 checks, positive controls throughout |
| `check_review_quotes.py` | the `file:` quotes CI skips **entirely** and the `provenance` blocks `checkquotes.py` does not walk; strict duplicate-key loader; anchor refusal; raw-vs-parsed reconciliation; fails if it checks zero |
| `audit_review_consistency.py` | summary-opener vs action, same-term-same-action, the hedge sweep (no declined term in any structured slot), complex-not-in-`locations`, description hygiene, isoform scoping on `NEW` rows, notes verdict table in **both** directions |

Writing the break-tests found **two defects in my own code**, neither visible by reading: the QuickGO paginator converted the service's undocumented page cap into an opaque HTTP 400 instead of a loud truncation report, and the UniProt feature dispatch keyed on the flat-file name `VAR_SEQ` where the JSON says `Alternative sequence` — silently emptying the isoform list, caught only because a *presence* guard on `VSP_034359` existed rather than an overlap calculation that would have reported nothing.

**A hook error I verified rather than obeyed.** The pre-write hook resolves `file:` paths against `$CLAUDE_PROJECT_DIR`, so from a sibling worktree it reported 24 non-existent-file errors plus a batch of "text part not found" errors for quotes that are all present. The three files exist, `just validate` passes in the worktree, and the quote checker matches all 53 exactly. Complying would have deleted correct evidence.

---

> [!NOTE]
> **Low Risk**
> Curation and documentation only—no runtime services or auth; risk is limited to incorrect GO recommendations or stale cache data, mitigated by automated checks and validation.
> 
> **Overview**
> Introduces a **human AEBP2 (Q6ZN18)** Gene Ontology review package: a large `AEBP2-ai-review.yaml` that walks all 20 GOA rows, plus generated `RESULTS.md` from reproducible provenance analysis.
> 
> **Annotation outcomes:** Most existing terms are **accepted**; `GO:0000122` is **kept as non-core** (isoform-dependent direction); two `GO:0031507` NAS rows are **marked over-annotated** (ComplexPortal projection). **Two new isoform-scoped molecular functions** are proposed: `contributes_to` nucleosome binding (`GO:0031491`, assayed on isoform 2) and long-isoform histone methyltransferase inhibitor activity (`GO:0180000`), framing the dominant long isoform as restraining PRC2 rather than stimulating it per older literature.
> 
> **Tooling added under `AEBP2-bioinformatics/`:** `analyze_aebp2.py` (QuickGO/UniProt/PDB/ARBA checks with self-tests), `audit_review_consistency.py` (YAML vs notes, numeric claims vs `results.json`, isoform-scope wording), and `check_review_quotes.py` (verbatim `supporting_text` including `file:` refs CI skips). **`cache/go/terms.csv`** gains one row for `GO:0180000`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a940520c27bbf658a873a8e66fdfe87aee12d20. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — the blocking point is right, and I verified it against the cached full text rather than just accepting it. All five items addressed in `80c1459c9`; one suggestion I'm refuting with evidence.

## 🟡 Blocking — conceded in full

`PMID:29499137`'s wild-type `S12R4J2A2` *is* isoform 2's C-terminus, and it worked:

> We found that S12R4 exhibited poor nucleosome binding by itself
>
> In stark contrast, S12R4J2A2 displayed robust binding towards mononucleosomes, with a low micromolar binding affinity

So the +14 residues give a 2-fold enhancement over an already robust baseline, and only the engineered `A2d` 5-residue deletion abolishes binding. My `core_functions[0]` had converted a **preference** into an **exclusion**, writing the MANE-Select form out of the gene's only proposed `contributes_to` MF — while the annotation row had it right. Adopted your wording verbatim.

Two things beyond the text:

- **`isoform:` on that row now reads `Q6ZN18-2`, not `Q6ZN18-1`.** This is your fourth suggestion. CLAUDE.md says the field records *what was tested*, and isoform 2's C-terminus is what was assayed. It is also less misleading than naming the strongest form, since the activity isn't isoform-restricted at all — all three isoforms retain 407–503.
- **Single-sourced instead of fixed twice.** Two surfaces could state the scope, which is why correcting one left the other. `audit_review_consistency.py` **check I** now requires one canonical clause verbatim on both surfaces and fires if either diverges, with break-tests in both directions (drop it from one surface → fires; reintroduce the retracted wording → fires).

## 🔵 The DNA-binding arms — your point changed the claim, and it caught two more surfaces

Adding check I's retracted-phrase list found "suppresses PRC2 DNA binding and methyltransferase activity" on a **third** and **fourth** surface: the `GO:0000122` reason and a `suggested_question` addressed to UniProt curators, plus the notes. Re-deriving against **core PRC2 alone**:

| arm | + AEBP2L | + AEBP2S |
|---|---|---|
| HMTase, in vitro | **almost completely inactive** — below baseline | comparable to baseline |
| DNA binding, in vitro probe | Kd > 4000 nM — **at** baseline | **45.5 nM**, ~90× |
| SUZ12 on chromatin, in cells | **below** the no-rescue control | above it |

> Conversely, core PRC2 alone, PRC2–AEBP2L(iso1) and PRC2–AEBP2L(iso2) complexes did not substantially bind the DNA probe, which indicates a much lower affinity
>
> However, ectopically expressed AEBP2L reduced the amount of SUZ12 on chromatin compared to AEBP2S and even compared to the no-rescue control cells

You're right that "suppresses DNA binding" was suppression-relative-to-AEBP2S. It holds for **chromatin occupancy in cells** (explicitly below the no-rescue control) and not for the naked-DNA probe. **The below-baseline claim `GO:0180000` rests on is the methyltransferase arm, which is untouched** — separating the arms makes that proposal cleaner rather than weaker. `knowledge_gaps[0]`'s misattachable clause fixed the same way, and the DNA-binding arm is now named as the strongly stimulatory one.

## 🔵 Five references added

`PMID:20064375`, `PMID:20064376`, `PMID:31451685`, `PMID:21949878`, `PMID:31959557`, each with a `reference_review` saying what it is doing in the review. All were cited in prose while absent from `references:`, and no gate checks that.

## 🔵 The BP term — refuting this one

GO has **dismantled** that branch. From QuickGO:

| term | status |
|---|---|
| `GO:0016571` histone methylation | **obsolete** |
| `GO:0031056` regulation of histone modification | **obsolete** |
| `GO:0031060` / `GO:0031061` / `GO:0031062` regulation of histone methylation | **obsolete** |
| `GO:0061085` / `GO:0061086` / `GO:0061087` regulation of histone H3-K27 methylation | **obsolete** |

None carries a `term_replaced_by`, and a search of live `biological_process` terms returns **no** non-obsolete regulation-of-histone-methylation term — controlled against a "heterochromatin formation" search that *does* return live terms, so that zero is a real zero and not a rejected query. A process term can't be requested in a branch GO deliberately removed, and the **molecular-function regulator route is what GO now provides** — which is where this review already puts the claim. Recorded in the second `proposed_new_terms` justification rather than left as a silent omission.

## A near-miss worth flagging

Squashing the fix commit, a `git reset --soft` past a merge left `main`'s C5orf46 files staged and they landed in my commit. Caught by diffing the commit's file list against `genes/human/AEBP2/`, reverted with a hard reset, and the three files restored and re-verified **byte-identical by md5** before recommitting. `git diff <merge-base> HEAD` now shows no file outside this gene's own scope.

## Gates, on the tree merged with `main`

`just validate human AEBP2` ✓ Valid (no warnings) · `checkquotes.py` 54 quotes / 0 problems · `check_review_quotes.py` 59 quotes (30 PMID + 29 `file:`) / 0 problems, self-test 0 failures · `audit_review_consistency.py` 0 problems, self-test 0 failures · `analyze_aebp2.py` self-test 0 failures · notes 23 quotes / 0 problems · `terms.csv` adds only `GO:0180000`, nothing lost vs the merge base, no new duplicate curies · `cache_lint` exit 0.

Thanks for the independent verification of the row reconciliation, the action tally, the VAR_SEQ mapping and the human-vs-mouse point on `PMID:29499137` — that last one is what makes the `GO:0031491` proposal safe while `GO:0003677` stays declined.

cmungall:
Both non-blocking items fixed in `6a940520c`, and the first one changed the guard's shape rather than adding a fifth literal.

## The fifth variant — and why the pin missed it

You're right: `core_functions[1]` still said "suppresses the complex's DNA binding" without naming the arm. A literal phrase pin cannot catch a paraphrase, which is a limitation I stated when I added it — and here it arrived in practice. So **check J** now anchors on the claim's *structure*: any sentence pairing a suppression verb with a DNA-binding marker and a long-isoform subject must also carry an arm qualifier. That survives rewording.

**Two defects in my first version of that check, both found by running it, neither by reading it:**

- it **sentence-split the raw YAML** and fired on a run-together of a `supporting_text` quote and a term label — the detector and the artifact disagreed on what a sentence *is*. It now walks the **parsed document's prose fields**, and excludes `supporting_text` deliberately: that is someone else's sentence, not a claim this review makes.
- its **vacuity guard required at least one *qualified* suppression claim**, which a correctly-written file need not contain at all — so it **failed on perfect agreement**. The precondition a correct file can satisfy is that the matcher reaches DNA-binding prose at all (22 sentences do).

## Coverage against the defects that actually shipped

Rather than assert the class is closed, `--history` runs check J over every version this branch pushed:

| version | check J flags |
|---|---|
| `a50319089` (first push) | **4** |
| `c58380583` | **4** |
| `80c1459c9` (the version you flagged) | **1** — exactly the variant the literal pin could not match |
| current | **0** |

It asserts monotonic decrease, zero now, **at least two** on the first version (otherwise the check is not demonstrating coverage of a *class*), and **exactly one** on the version you flagged. It also refuses to report a zero from a git extract under 500 lines — a silent zero from a failed extract is indistinguishable from a clean result, and that happened once while I was developing it.

## The `GO:0180000` isoform justification

Also correct. The row justified `isoform:` from the DNA-binding arm when the term is about methyltransferase activity. Now rests on the HMTase arm, with the limitation stated rather than glossed:

> the HMTase result is reported for "PRC2-AEBP2L" without separating the two long isoforms, so Q6ZN18-2's methyltransferase behaviour is inferred rather than measured — the two were separately measured only in the DNA-binding arm, where they were indistinguishable.

## Gates

`just validate human AEBP2` ✓ Valid, no warnings · `checkquotes.py` 54/0 · `check_review_quotes.py` 59 quotes (30 PMID + 29 `file:`) 0 problems, self-test 0 failures · `audit_review_consistency.py` 0 problems, self-test 0 failures, `--history` 0 problems · `analyze_aebp2.py` self-test 0 failures · notes 23/0 · branch is level with `main`, `terms.csv` adds only `GO:0180000`, `cache_lint` exit 0, and `git diff <merge-base> HEAD` shows no file outside this gene's scope.

This is round 3, and I'll name a stopping point: rounds 2 and 3 both changed real content, so neither was harness churn — but if the next round's findings are about the guard's own prose rather than about a term, action, evidence code, quote or number, I'd propose stopping there. Not merging and not enabling auto-merge, per instructions.

PR #2349, opened 2026-07-27
Review human AFF4: a coverage gap, not an over-annotation

cmungall:
Review of human **AFF4** (`Q9UHB7`), the central scaffold of the super elongation complex and the gene mutated in CHOPS syndrome.

Accession verified before anything else: `primaryAccession: Q9UHB7`, `AFF4_HUMAN`, `HGNC:17869`, **no secondary accessions**, and a `gene_exact` search returns exactly that one entry. Every UniProt fetch in the committed script asserts `primaryAccession == requested` anyway, because a merged accession returns HTTP 200 with a complete reviewed record for a different protein.

## Row reconciliation, stated up front

| | |
|---|---|
| `AFF4-goa.tsv` data lines | **43** (all 43 distinct on term+evidence+reference+with/from+qualifier) |
| `fetch-gene` stub seeded | **34** |
| `existing_annotations` in this PR | **51** = 43 reviewed + 8 `NEW` |

The 9-row gap is the known `seed_missing_annotations` collapse: the seeder key omits WITH/FROM, so the **15** `GO:0005515` IPI rows collapsed to 6. They are restored one per row, and it changed the outcome (see SIAH1 below). A committed check asserts the 43 reviewed rows equal the TSV's distinct set on membership, not on count.

## The headline finding is a coverage gap, not an over-annotation

This gene is the inverse of the campaign's usual result. `AFF4-bioinformatics/RESULTS.md` §5b measures it: for each of **21** papers that establish something about human AFF4, how many GO annotations exist anywhere in GOA from that reference, and is AFF4 among them?

**16 of 21 have produced no GO annotation anywhere in GOA. 19 of 21 have produced none on AFF4.**

Consequences visible in AFF4's own record:
- **zero experimental biological-process annotations**;
- the only experimental molecular-function rows are **six bare `GO:0005515` protein-binding IPIs** — no row says what the protein does;
- `GO:0032783 super elongation complex` reaches AFF4 **only by IBA**.

The sharpest instance: `PMID:20159561` is titled *"AFF4, a component of the ELL/P-TEFb elongation complex…"* and states *"AFF4 is required for SEC stability"* — and the whole of GOA carries **exactly one annotation from it: `GO:0032783` IDA on AFF1**. Same shape for `PMID:39603240`, which yields `GO:0032968` IDA on the four PNUTS-PP1 components and nothing on AFF4 although AFF4 Ser-549 is the substrate UniProt curates from it. Nine PDB entries covering three interaction surfaces have produced nothing.

[...]

PR #2351, opened 2026-07-27
Review human AFF3

cmungall:
Completes the AFF family, after AFF4 (#2349) and AFF1 (#2348) — both of which merged to `main` while this was being written, so the sibling comparison below is against merged content.

**Row reconciliation.** `AFF3-goa.tsv` has 11 data rows; the `fetch-gene` stub seeded 11 entries. They reconcile exactly, with no collapsed `GO:0005515` or same-term/different-assigner rows (the ADAMTSL5 under-seeding failure mode does not apply here). Verdicts: **1 REMOVE, 1 MODIFY, 7 ACCEPT, 2 KEEP_AS_NON_CORE**, plus **8 NEW** proposals (19 entries total).

## The over-annotation: a clinical drug response read as a cellular TNF stimulus

`GO:0034612 response to tumor necrosis factor` → **REMOVE**. Its sole support, `PMID:20444755`, is a pharmacogenetic association study — 18 tag SNPs genotyped in 1012 RA patients receiving etanercept, infliximab or adalimumab, with change in DAS28 as the outcome. Three independent grounds, from the full text rather than the title:

1. **Definitional.** The term requires a change in state "*as a result of a tumor necrosis factor stimulus*". The study applies TNF **blockade** — the opposite perturbation — and never applies TNF to anything.
2. **Evidence code.** IMP requires a mutant phenotype. A common tag SNP associated with a 6-month clinical score change is neither.
3. **No assay.** There is no cellular or biochemical experiment on AFF3 in the paper.

It has already propagated: mouse Aff3 holds `GO:0034612` **only** by `ISO GO_REF:0000119` with `UniProtKB:P51826` — from this human row — so retracting the human row clears it from both species. This is precisely the disease-gene failure mode the task predicted, and the review is explicit that removing it is *not* a claim that AFF3 has no immune role: its real, mechanistically characterised immune function (immunoglobulin class switch recombination) is proposed separately.

## The finding worth reporting upstream: the paper annotated the wrong paralogue

`PMID:22547686` is the paper whose headline result is the **biochemical isolation of the AFF3-containing SEC-L3**. Querying QuickGO **by reference** shows it has produced **exactly one annotation in all of GOA: `GO:0006354` EXP on `UniProtKB:P51825` (AFF1)** — the paralogue the paper explicitly contrasts SEC-L3 *against*. AFF3 got nothing directly, and receives the term back only as an IBA whose WITH/FROM points at AFF1.

This is the **second instance of the same shape in this family**: AFF4's review found `PMID:20159561`, titled for AFF4, likewise produced one annotation and it was on AFF1. Filed as a `suggested_question` for GO Central.

## The brief's four cross-checks, answered

1. **`PTHR10528` confirmed** — AFF3 is in it, subfamily `PTHR10528:SF16`.
2. **The byte-identical WITH/FROM resolved.** `PANTHER:PTN000829417|UniProtKB:P51825` on `GO:0006354` is identical across AFF1, AFF4 and AFF3. On AFF3 it is **paralogue-derived** (P51825 is AFF1, P51826 is AFF3). The same six bytes now carry three evidential statuses across three reviews.
3. **The donor's own evidence, one level deeper.** Following `P51825` gives AFF1's single EXP from `PMID:22547686`, whose measurement is **Pol II** CTD kinase activity — which is what justifies **MODIFY `GO:0006354` → `GO:0006368`**, verified to be a descendant.
4. **The `core_functions` shape diverges, with a reason.** AFF1 and AFF4 both used `GO:0030674` + `contributes_to GO:0003711`. I keep both **and** add `GO:0003712` as a second core MF, because AFF3's own best-evidenced molecular role is being *recruited* by sequence-specific zinc fingers (ZFP281 to the Meg3 enhancer, ZFP57 to the IG-DMR) rather than mapped bridging — and `GO:0003712`'s usage comment describes it exactly ("*Most transcription coregulators do not bind DNA. Those that do usually bind DNA either in a non-specific or non-direct manner*"). AFF1/AFF4 had bare `GO:0005515` IPI rows needing an informative replacement; **AFF3 has none**, so nothing pushed me toward `GO:0030674` from that direction.

## Where the two siblings disagreed, and the rule I used

They split on exactly the two rows where a choice was needed, and in opposite directions, so "follow the family" gave no answer:

| row | AFF1 | AFF4 | AFF3 |
|---|---|---|---|
| `GO:0006354` IBA | MODIFY → `GO:0006368` | ACCEPT | **MODIFY → `GO:0006368`** |
| `GO:0010468` IEA | MODIFY → `GO:0006355` | ACCEPT | **ACCEPT** |

> MODIFY a broad term only when the more specific term the donor evidence supports is **not already annotated on this gene**. If it is, the broad row is a redundant ancestor and MODIFY buys nothing.

`GO:0006368` is absent from AFF3 → MODIFY adds information. `GO:0006355` is already present from a better-founded IBA → the InterPro row is a redundant ancestor. One rule, reproducing AFF1 on one row and AFF4 on the other.

## Two UniProt cross-references GOA lacks, resolving oppositely

QuickGO returns exactly 11 annotations for P51826, matching the TSV. UniProt additionally carries two Ensembl-Compara projections:

- **`GO:0003690` double-stranded DNA binding** — correct, and missing from GOA. The human protein binds dsDNA-cellulose in a direct in vitro assay and human ChIP places it on a defined CpG island → proposed as a **NEW** row.
- **`GO:0003700` DNA-binding transcription factor activity** — **wrong for AFF3**. Its definition requires binding "*a specific double-stranded genomic DNA sequence … within a cis-regulatory region*"; AFF3 has no motif and is recruited by ZFPs. The mouse source rows rest on ChIP-qPCR of **over-expressed HA-tagged** Laf4 at one promoter, framed by the authors as testing a "*potential* direct transcriptional regulator". Filed as a **UniProt/Ensembl correction request**, not a GO action — there is no GOA row to act on, and inventing one would be an over-annotation of the opposite sign.

`GO:0003700` and `GO:0003712` were verified to be **siblings** (neither is an ancestor of the other), so this is a wrong term, not a coarse one.

## Correction to a sibling review's premise

AFF4's merged review states that AFF3's only experimental molecular function is DNA-binding transcription factor activity. Read from AFF3's own record, **human AFF3 has no experimental MF annotation at all** — its only MF row is the `GO:0003712` IBA; the `GO:0003700` IDA is on *mouse* Aff3.

## Where I was wrong, and what corrected me

- **`GO:0035116` embryonic hindlimb morphogenesis.** My first reading was that a single case report of a 500-kb microdeletion could not carry a gene-specific IMP. `PMID:25162227` states the deletion "*encompass[es] only the LAF4 gene*", so it is not a contiguous-gene event; three further lines converge on the **lower** limb specifically (patient fibular agenesis/triangular tibiae; mouse in-situ in limb buds and zeugopod; KINSSHIP knockin mice with lower mesomelic limb deformities). Kept, as non-core.
- **`GO:0030674` ancestry.** I assumed it was a descendant of `GO:0005515`. The committed ancestry guard refuted it — its only ancestors are `GO:0003674` and `GO:0060090`. The wrong claim and its correction are both retained in the script, because a documented retraction is more useful than a clean absence.

## Gain-of-function and fusion discipline

AFF3's limb and kidney phenotypes arrive from **both** directions of dosage — deletion and promoter GCC silencing on one side, degron-stabilising KINSSHIP variants on the other. A gene whose increase *and* decrease disrupt a process participates in it, so `involved_in` stands; but nothing in this review infers a molecular activity from the variant protein's behaviour (the KINSSHIP substitutions confer resistance to degradation, not a new function), and **no row derives from the MLL–AFF3 or RUNX1–AFF3 fusions**, which are cited as background only.

## Checks that came back negative, recorded because they were run

- **Complex-projection test** (a reference annotating a complex plus every subunit): none of the four references annotates more than two entities. `PMID:18616733`'s IEP-on-mouse / IMP-on-human split is *correct* curation, two codes for two observations.
- **Logical-opposite citation cross-product**: no positive/negative regulation pair among the 11 terms. Trivially negative.
- **Fold-to-activity propagation**: **not confirmed.** Each of AFF3's three InterPro signatures was looked up individually — `IPR007797`→`GO:0010468`, `IPR043640`→`GO:0005634`, `IPR043639`→nothing. No MF term from any signature. This reproduces AFF4's committed measurement exactly, which is the precondition for reporting one's own column.
- **`GO:0005515` per-partner adjudication**: AFF3 has no such rows. IntAct nonetheless returns 14 records over 7 partners, with **CDK9 in 6 records across 5 distinct publications and 3 distinct methods** (anti-tag co-IP, pull down, TAP; MI 0.35 and 0.73) and **MLLT1/ENL** in 2 records from 1 publication — the two SEC modules AFF3 bridges, and GOA has curated neither. An under-curation datum, not an over-annotation one. **These counts were wrong in the first version of this PR** — see below.
- **`GO_REF:0000120` independence**: its three tokens are **not** three witnesses. `ARBA00026330` (fetched: 1309 condition sets, one reaching AFF3 — `IPR007797 AND IPR043640 AND Eukaryota`) already conditions on the `IPR043640` token, and `SL-0191` derives from UniProt's own unevidenced `SUBCELLULAR LOCATION` line. Verdict unchanged (an IDA confirms it), but the triple corroboration is illusory.
- **Retractions and errata**: 29 PMIDs checked by two routes. **1 flagged, no retractions.** `PMID:20444755` carries an `ErratumIn` whose **PubMed id is null**, unresolvable via Crossref (empty `relation`/`update-to`) or Europe PMC (0 hits for the page and for a citing Published Erratum). Its scope is unestablished — and the REMOVE deliberately does **not** rest on it, avoiding the inverse mistake AFF4 made of letting an unresolvable correction manufacture a hedge.

## affinage

`gates_passed: True`, 15 citations, all well-formed numeric PMIDs, none retracted. Recall was measured, not assumed: the record **omits `PMID:22547686`** (the paper this review's sharpest finding turns on) and **both** `PMID:18616733` and `PMID:20444755` — the sole support for AFF3's only two experimental BP rows. It returned the gene's *notable* literature, not its *annotation-relevant* literature. Adjudicated in `references` with `LOW_QUALITY` and quoted nowhere.

## Second commit: a number that refused to add up, self-caught

After opening this PR I re-derived a figure I had counted by eye. The review said IntAct records "CDK9 in **five** records across **four** distinct publications and **four** distinct methods with a MI score of 0.73". `intact_partners.py` computes **6 records, 5 publications, 3 methods, MI in {0.35, 0.73}**. All four numbers were wrong.

- **Cause**: an ad-hoc query that deduplicated on a composite key — silently dropping one CDK9 record — plus counting printed lines by eye. The replacement is derived from a committed script asserting that every record IntAct reports is retrieved and that the subject is not in its own partner set, and the corrected prose is anchored to a quoted row of the computed table.
- **`fix_intact_counts.py` is committed** rather than hand-editing, asserting each anchor is present before replacing, re-grepping afterwards, and asserting `detected == changed`. **Its first anchor did not match**, because the notes use `×` and `—` where the script typed `x` and `-` — the hyphen/en-dash trap the brief warns about for quotes, arriving in a find-and-replace. The assertion caught it instead of the edit landing in 4 of 5 places.
- **The recurrence guard is structural, not a phrase pin**: `audit_claims.py` now parses any prose stating CDK9's counts and compares them to `intact_partners.json`, matching word forms as well as digits. It is break-tested against `git show HEAD:…` — the version that actually shipped — plus a vacuity direction that fires if the sentence disappears, and it reports which path it took when HEAD is already fixed.
- The correction runs **favourably**: 5 independent publications is a stronger replication claim for the CDK9 contact than 4.

## Third commit: a second retracted premise, caught by the same guard

The `GO:0006355` row's reason argued that the row must stay at the unsigned parent *because its donors disagree in sign* — "mouse Aff1's descendant is positive regulation while human AFF1's is in the negative-regulation-of-elongation branch". That read `GO:0032786` as negative from its proximity to `GO:0032785`.

**`GO:0032786` is *positive* regulation of DNA-templated transcription, elongation**, a verified descendant of `GO:0045893`. So every signed donor on that row points the same way, and `GO:0045893` is itself a descendant of `GO:0006355` — a positive child was available and unused. The AEBP2 donor-disagreement test does not apply.

The **verdict is unchanged and now better founded**, resting on the recipient rather than the donors: AFF3's output runs both ways. It represses XIST from the silent allele in HEK293T and IMR-90, while with ZFP281 it establishes a permissive chromatin state at the Meg3 enhancer and its over-expression raises 84% of the transcripts it changes in mouse cortical cells. A positive-only term would be false for the repressive half — which is exactly why the specific negative instance is a separate `GO:0045892` row rather than a modification of this one.

`term_relations.py` gains three claims pinning the fact (15 total), so it is machine-checked; `RESULTS.md` narrates the retraction rather than deleting it; `fix_sign_claim.py` is committed and permits the retracted phrasing only on surfaces that explicitly mark it as retracted.

**The three retractions on this gene share one shape**, now named in the notes: each was an inference from an identifier's or a label's *neighbourhood* rather than from a fetched fact. Adaptor activity sounds like a kind of protein binding; `GO:0032786` sits one integer from `GO:0032785`; the IntAct counts were read off a printed list. In all three the fix was to write the check, and in all three the check fired.

## Committed gates

`AFF3-bioinformatics/` holds every claim-bearing computation, all reproducible:

- `resolve_withfrom.py` — asserts `primaryAccession == requested`, uses `size=10` so an ambiguous xref is reported not reduced, anchors the reviewed test on `startswith("UniProtKB reviewed")`, and asserts `numberOfHits == len(results)`.
- `term_relations.py` — 15 ancestry claims fetched with `relations=is_a,part_of` only; exits non-zero on any wrong claim. **It caught one of mine.**
- `reference_projection.py` — entity counts (not annotation counts), fully paginated.
- `corrections_check.py` — publication types **and** `CommentsCorrections/RefType`.
- `verify_file_quotes.py` — byte-exact substring match on a surface CI does not check at all, flagging dash/quote substitutions, failing if it verifies zero.
- `intact_partners.py` — per-partner record/publication/method counts, asserting full retrieval and that the subject is absent from its own partner set.
- `audit_claims.py` — strict duplicate-key loader, no-anchors check, raw-vs-parsed reconciliation, `knowledge_gaps[].provenance` quotes (which `checkquotes.py` does not walk), **both** directions of `core_functions` backing, prose-vs-action agreement with an attributed-cross-reference exemption, and guards for **both** retracted claims. **14 break-test directions**, each asserting the mutation changed the document, that the guard fired, and that the message was the expected one — including the direction proving the withdrawn-claim guard stays silent on the *corrected*, negated statement, a vacuity direction, and one run against the shipped defect via `git show`.

Local gates on the tree merged with `main`: `just validate human AFF3` ✓ Valid (one standing warning: no annotation cites the affinage file, which is deliberate and adjudicated in `references`); `checkquotes.py` 71/71; `verify_file_quotes.py` 14/14 byte-exact; `audit_claims.py` 0 problems and 14/14 break-tests; `cache_lint` exit 0; `terms.csv` +1 row (`GO:0045190`), nothing lost, no duplicate curies.

---

> [!NOTE]
> **Low Risk**
> Adds gene-review YAML, GO term cache row, and offline validation scripts—no production services or auth/data-path changes; main risk is incorrect GO recommendations if gates were bypassed.
> 
> **Overview**
> Completes the **human AFF3** Gene Ontology review in the same style as AFF1/AFF4: a large `AFF3-ai-review.yaml` that adjudicates all **11** GOA rows and proposes **8** new annotations, plus narrative `RESULTS.md` and committed scripts to reproduce every claim.
> 
> **Annotation outcomes** include **REMOVE** of `GO:0034612` (TNF response) as unsupported pharmacogenetics; **MODIFY** `GO:0006354` → Pol II–specific `GO:0006368`; **ACCEPT** / **KEEP_AS_NON_CORE** on several IBA/IEA rows with donor and propagation notes; and **NEW** terms such as SEC adaptor activity, nuclear speck, XIST repression, cortical migration, **isotype switching** (`GO:0045190`, also added to `cache/go/terms.csv`), and kidney development. The review flags mis-citation of `PMID:20444755`, SEC-L3 literature landing on AFF1 in GOA, and UniProt/Ensembl `GO:0003700` vs supported `GO:0003712` / `GO:0003690`.
> 
> **`AFF3-bioinformatics/`** adds WITH/FROM resolution, term ancestry checks, reference projection, IntAct partner counts, PMID correction screening, file-quote verification, and **`audit_claims.py`** (duplicate keys, quote reconciliation, core_functions backing, prose-vs-action, guards for retracted IntAct and “donor sign disagreement” claims). One-shot **`apply_review_round1/2`** and **`fix_*`** scripts document post-review corrections with anchor-asserted edits.
> 
> **Risk:** Mostly curation and documentation; wrong GO changes would affect downstream annotation consumers, but changes are evidence-backed and heavily gated rather than runtime code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 639897314017d30f8f03dbce53975c0796b733df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thank you — all six addressed in `2bf0d3d5e`. Five accepted, one half-accepted (and the half that was wrong is the useful part), one declined with evidence. Each premise was checked before conceding.

## Accepted

**1. `GO:0030674` IPI had no WITH/FROM.** You are right and it was ill-formed: for IPI the WITH/FROM field takes the interactor. Added `UniProtKB:P50750` (CDK9) and `UniProtKB:O60563` (cyclin T1), both confirmed reviewed human entries by asserting `primaryAccession` on the fetch rather than trusting the accession I was handed.

I declined the alternative (recode as ISS on AFF1/AFF4) and said so in the reason: the assay behind this row is a co-immunoprecipitation of **AFF3 itself** from human nuclear extracts, with the reciprocal pull-down also on endogenous protein, so IPI is the correct code — the entity was missing, not the code. Downgrading to ISS would misattribute human AFF3 biochemistry to sequence similarity with its paralogues. The unmapped-interface caveat stays where it was: in the reason and in `knowledge_gaps[0]`.

**4. `GO:0003711` — both parts.** The kinase-to-elongation step is now named as an interpretation ("what was measured is polymerase II CTD kinase activity … reading that as elongation-factor activity relies on the established role of CTD Ser2 phosphorylation rather than on an elongation assay performed on SEC-L3"), with a pointer to the `suggested_experiments` entry that would convert it. And you were right that the sibling claim was half-checked — `term_relations.py` now asserts `("GO:0003711", "GO:0140110", True)` alongside the `GO:0003712` leg.

**5. `GO:0001822`.** Correct — "urogenital tract malformations" is broader than the kidney. The reason now states that the mouse null's explicit kidney defect is doing the work and the human quote corroborates the organ *system* rather than the organ.

**6. Both follow-ups filed.** The `GO:0034612` removal's downstream consumer is folded into the MGI question, so the retraction is actionable as a pair rather than only recorded. And on AFF4's `GO:0032783` question — I read `AFF4-ai-review.yaml` before asserting anything about it, per the rule that burned two agents this campaign; it does state that AFF3 "has not been shown to be a subunit of that complex". That premise is refuted by `PMID:22547686` (biochemical isolation of AFF3-containing SEC-L3 with P-TEFb, ENL/MLLT1 and AF9/MLLT3) and `PMID:26214578` (endogenous reciprocal co-IP recovering CDK9 and cyclin T1), so the correction is filed. I scoped it to AFF3 and left AFF2 alone, since I did not examine AFF2.

## Half accepted — and your objection found something better than the thing you objected to

**2. `GO:0050877`.** You were right that the reason was in tension with the review's own ancestry finding, and checking it split the claim rather than confirming it:

| term | descendant of `GO:0050877`? |
|---|---|
| `GO:0050890` cognition | **true** |
| `GO:0007611` learning or memory | true (and one step under cognition) |
| `GO:0021795` cerebral cortex cell migration | **false** |
| `GO:0001764` neuron migration | false |

So the human genetics does not uniformly belong elsewhere — the **cognitive** half is on-branch. Intellectual disability, seizures and the 2.4-fold education association sit *inside* `GO:0050877` via cognition and do corroborate the row; only the cortical-migration evidence is outside it. What the reason did wrong was to lump the two halves and offer them together.

It now (a) rests primarily on the two donor IMPs to a verified descendant, which as you say is the proper IBA basis, (b) cites the cognitive phenotypes as explicitly on-branch corroboration, and (c) states that the migration evidence is *not* offered as grounding and belongs on the separate `GO:0021795` row. Four new claims in `term_relations.py` pin all of it — 19 total.

## Declined, with the paper's own words

**3. `core_functions[2].molecular_function`.** I've kept `GO:0003690` and put the reasoning in the row's `reason` with three new quotes in `supported_by`, so the refusal is CI-checked rather than asserted in prose.

Both alternatives over-claim, and `PMID:36001653` says so itself:

- **`GO:0030674`** would assert a bridge between switch-region DNA and AID. The paper does not demonstrate one — *"While the mechanism by which AID is recruited to switch regions is still unclear, the following mechanisms have been proposed"* — and offers cohesin and P-TEFb as alternative routes: *"AFF3 may regulate CSR by facilitating the interaction of AID with cohesin factors"*. An adaptor term would name a mechanism the authors explicitly leave open.
- **`GO:0003712`** belongs to transcription regulation; class switch recombination is a DNA recombination reaction, so the term would be in the wrong process entirely.

What *was* measured is occupancy — *"We detected significant increases in the signals near the switch regions of IgM and IgG1, indicating that AFF3 can bind to these regions"* — which is what `GO:0003690` says. I agree it is the least informative of the three MF terms; it is also the only one the data support.

## One measurement your round prompted

Asked how large the `GO:0005515` gap actually is, one QuickGO call over all three paralogues answers it **with its own positive controls**:

| gene | `GO:0005515` rows in GOA |
|---|---|
| AFF4 (Q9UHB7) | 15 |
| AFF1 (P51825) | 3 |
| **AFF3 (P51826)** | **0** |

The siblings are the controls — the endpoint works and the term is alive for them in the same request — so AFF3's zero is a real absence, not a rejected query. Given that AFF3's CDK9 contact is the most replicated of the three in IntAct (6 records, 5 publications, 3 methods), that is a curation asymmetry rather than a biological one. Added to `RESULTS.md` §8 and to the `GO:0030674` reason.

## Two things I corrected in myself before you saw them

Both are in the branch and both were caught by guards rather than by re-reading, which is the argument for having them:

- **The IntAct counts were hand-counted and wrong on all four numbers** (`intact_partners.py` computes 6 records / 5 publications / 3 methods / MI {0.35, 0.73}, not 5/4/4/0.73). The recurrence guard is structural — `audit_claims.py` parses any prose stating those counts and compares them to `intact_partners.json` — and is break-tested against `git show HEAD:…`, the version that actually shipped.
- **`GO:0006355`'s reason claimed the donors "disagree in sign"**, reading `GO:0032786` as negative from its proximity to `GO:0032785`. It is *positive*. Every signed donor agrees, and `GO:0045893` was available and unused. The verdict survived on a better ground — AFF3's own output runs both ways — but the argument was false and is now narrated as retracted in `RESULTS.md` §1 rather than quietly deleted.

Together with the `GO:0030674`/`GO:0005515` inversion your review noted, that is three retractions on this gene, and they share one shape now named in the notes: **each was an inference from an identifier's or a label's neighbourhood rather than from a fetched fact.**

## Gates on the merged tree

`just validate human AFF3` ✓ Valid (one standing warning, deliberate — no annotation cites the affinage record, which is adjudicated `LOW_QUALITY` in `references` instead); `checkquotes.py` 74/74; `verify_file_quotes.py` 14/14 byte-exact; `audit_claims.py` 0 problems and 14/14 break-test directions; `term_relations.py` 19/19; `corrections_check.py` 1 of 31 flagged, no retractions; `cache_lint` exit 0; `terms.csv` +1 row (`GO:0045190`), nothing lost, no duplicate curies.

On your runner note — `uv`/`just` unavailable, so you relied on `Build and test` for `validate`: that is the right call and it is green on this head; every script in `AFF3-bioinformatics/` is stdlib-only apart from `PyYAML`, so `python3` alone will run them if that helps a future round.

cmungall:
All six addressed in `4a91c1d7d`. Five accepted outright; the new one is accepted as an **attribution** defect rather than a fabricated one, with the source now named.

Thank you for withdrawing the ISS half of suggestion 1 and for independently verifying the `GO:0050890` split — both of those are exactly the kind of cross-check that made round 1 worth having.

## The five carried-over items had one cause, and it indicts my guard

Every one of them was the same failure: **round 1 fixed each claim on the surface you had named and left it standing on a low-salience one.** The survivors were a `reference_review` note, a `source_entities` comment, a script docstring, a notes bullet and the top-level `description` — precisely the surfaces nobody re-reads.

Worse, and this is the part worth recording: **my own guard was structurally unable to catch two of them, and you diagnosed why.** `fix_sign_claim.py`'s narration exemption was **file-scoped** — it exempted `AFF3-ai-review.yaml` wholesale so that the file could narrate its own retraction, which made the re-grep blind to every *unnarrated* instance inside that same file.

Fixed at the structural level rather than by adding two more literals:

- the exemption is now **per-occurrence** — a sentence asserting the retracted premise passes only if the surrounding ±260-character window carries a retraction marker;
- the check has **moved out of the one-shot script into `audit_claims.py`**, so it runs on every audit rather than once;
- it is **break-tested against `git show 2bf0d3d5e:…`** — the commit that actually shipped your two survivors — and fires on precisely them;
- with a **companion direction** asserting it stays *silent* on the narrated retractions that remain in the document by design, so the exemption itself is tested in both directions.

16 break-test directions now, happy path clean.

I'd name this guard-defeat mode, because it is not one already on the campaign's list: **an exemption coarser than the thing it exempts.** Correct in intent, one scope level too wide in implementation, and the result is a guard reporting success over exactly the file it was written for.

## The new item: partly declined, because the phrase was sourced — just not where it looked

You are right that `PMID:33961779` spells the acronym out as "KI for **horseshoe** kidney" and contains no occurrence of hypoplastic. But the phrase is not unsourced. UniProt's DISEASE line for KINSSHIP reads:

> facial features, horseshoe or hypoplastic kidney, and failure to

(`AFF3-uniprot.txt:149`, and that is the whole physical line — the sentence continues onto the next `CC` continuation, which is why it is quoted truncated rather than reflowed).

So the defect was **attribution, not fabrication**: the phrase sat immediately beside a `PMID:33961779` quote and read as the paper's. Both surfaces now attribute it explicitly — the reason says "PMID:33961779 spells the acronym out as KI for horseshoe kidney, while UniProt's DISEASE line for the syndrome records horseshoe or hypoplastic kidney, and the broader wording is UniProt's rather than the paper's", and the summary carries the same split. The `file:` quote is added so the claim is checkable rather than asserted.

Your framing survives intact and is sharper for being narrower: **a commit whose purpose was to stop over-reading the human evidence loosened it in the same breath**, and the added words came from a different source than the quote sitting next to them. That is worth more as a lesson than the two words were as content.

## The other four

- **`:414` "all verified".** Both halves right. "GO:0045893 under GO:0045893" was vacuous, and `GO:0032968` was simply absent from `CLAIMS`. The prose now names only the two terms genuinely below — "mouse Aff1's term IS `GO:0045893`, and human AFF1's `GO:0032786` and `GO:0032968` were each verified to sit under it" — and `term_relations.py` gains `("GO:0032968", "GO:0045893", True)`. **20 claims.**
- **`:25`/`:1269` speckles.** Correct, and it was over-read in exactly the way the `GO:0016607` row itself excludes. The paper's baseline observation is AFF3 colocalising with SC35; the redistribution of CDK9 and cyclin T1 *to AFF3-occupied sites* required strong AFF3 over-production. Both surfaces now say AFF3 localises to speckles, "the compartment where CDK9 and cyclin T1 independently form foci", with the over-production qualifier stated and the co-concentration explicitly not claimed as native.
- **`intact_partners.py:4-6`.** The docstring said "5 records" for a full round after the prose had been corrected to 6 — in the docstring of the very script written to stop that class of error. Now derived from `intact_partners.json` like everything else, with a sentence recording that it happened, because a script whose purpose is "compute, don't count" getting its own count wrong is the most useful instance available.
- **`AFF3-notes.md:315` PIP4K2A.** Correct — 2 records across 2 publications (`PMID:28514442`, `PMID:33961781`), not a singleton. Split into its own bullet with the computed figures, and the singleton bullet reduced to SYT2, DISC1 and ERP29/TFRC.

One small robustness change while in there: `fix_pmid_count.py` now reports "already applied" instead of raising on a consumed anchor, and fails loudly if it finds a *half*-applied state — `RESULTS.md` invites re-running these scripts, so a traceback that means "nothing to do" is a trap of its own.

## Gates on the merged tree

Merged `origin/main` forward (clean, no conflicts) and re-ran everything on the merged tree: `just validate human AFF3` ✓ Valid (one standing warning, deliberate); `checkquotes.py` 74/74; `verify_file_quotes.py` 14/14 byte-exact; `audit_claims.py` 0 problems and 16/16 break-test directions; `term_relations.py` 20/20; `corrections_check.py` 1 of 31, no retractions; `cache_lint` exit 0; `terms.csv` unchanged by the merge, still +1 row (`GO:0045190`) against main, nothing lost, no duplicate curies.

cmungall:
All three taken in `710a315df` — each closed a real hole rather than rewording prose. Thank you for tracing all four surviving sentences' windows and for confirming the break-test anchor exists verbatim; that is checking the guard rather than the claim about the guard, which is the harder half.

**1. The quote was outside the gate.** Correct, and pointedly so: the one `file:` quote I added to *answer* a fabrication-surface criticism was itself on the one surface neither quote checker walks. It is now in the `GO:0001822` row's `supported_by` — 15 `file:` quotes verified byte-exact, 75 quotes total.

**2. The guards had no standing scope outside the YAML.** Also correct, and it is the direct cause of items 5 and 6 surviving round 1, as you say. `audit_sibling_surfaces()` now runs `check_sign_claim` and `check_intact_counts` over every `.md`, `.yaml` and `.py` in the gene folder on every audit, and fails loudly if it scans zero files.

Widening it produced **ten false positives immediately**, and that turned out to be the useful part. Repair scripts must contain the retracted strings as their **find anchors**, and the guard must contain them as **patterns** — so a naive folder-wide sweep forbids the guard from coexisting with its own implementation, which is the "guard that forbids legitimate practice gets worked around" failure. Resolved by widening the exemption *vocabulary* to cover anchors, patterns and fixtures while keeping it strictly **per-occurrence** — never per-file, since that is the mode that failed in round 2 — and by labelling two find-anchors in `fix_intact_counts.py` as anchors so the exemption applies for a stated reason rather than by luck. Both directions break-tested: the sweep fires on an unnarrated sign claim injected into `AFF3-notes.md`, and stays silent on the anchors and narrated history that remain by design.

**3. The reciprocal pair.** Added: `("GO:0003711", "GO:0003712", False)` and the converse. 22 claims, and `RESULTS.md` §3 updated to match.

## One thing worth reporting, because it is a failure mode I had not seen

After widening the exemption, **the synthetic IntAct-count break-test began failing — for the right reason.** The per-occurrence exemption legitimately suppressed the mutation, because the first of the two CDK9 count sentences sits inside a window that explains the retraction. So the guard was correct and the *test* was now aimed at an exempt site.

The repair: the test selects an un-narrated occurrence explicitly, asserts one exists (rather than passing if none does), and uses a raw anchor unique to it — the raw and whitespace-normalised forms differ, because the phrase is line-wrapped at different points in the two `>-` blocks, so the flat-text detector and the raw-text mutator do not share an anchor.

The general point: **a break-test that fails because an exemption worked is not evidence the guard is broken**, and distinguishing the two required reading the window rather than the result. The cheap wrong response would have been to narrow the exemption until the test passed again — which would have re-broken the coexistence problem in item 2. 17 break-test directions now.

## Stopping criterion

Rounds 2 and 3 changed **no GO term, no action, no evidence code, no quote and no reported number** in the curation itself. Every item in both rounds concerned *where* a claim was written, *which surface* a guard scanned, or *which ancestry pair* was pinned. The curation content has been stable since `2bf0d3d5e`.

So, explicitly: I will still fix anything that misstates a number, misattributes a source, or lets a guard report coverage it does not have — those are the three classes that have actually produced defects here. I will not keep refining the guards' own prose, and I am not chasing further tooling symmetry for its own sake. If a later round finds a factual defect in the annotations, that is a different matter and gets fixed.

For the record, the substantive curation this PR proposes, unchanged for three rounds: one REMOVE (`GO:0034612`, a pharmacogenetic drug-response association read as a cellular TNF stimulus, transitively clearing the mouse ISO row), one MODIFY (`GO:0006354` → `GO:0006368` on the donor's own Pol II measurement), 7 ACCEPT, 2 KEEP_AS_NON_CORE, 8 NEW, and two upstream findings — `PMID:22547686` annotating AFF1 instead of the AFF3 it characterises, and UniProt's `GO:0003700` Ensembl projection contradicting `GO:0003700`'s own usage guidance while its companion `GO:0003690` projection is correct and missing from GOA.

## Gates on the merged tree

`just validate human AFF3` ✓ Valid (one standing warning, deliberate); `checkquotes.py` 75/75; `verify_file_quotes.py` 15/15 byte-exact; `audit_claims.py` 0 problems, 17/17 break-test directions, sibling sweep clean; `term_relations.py` 22/22; `corrections_check.py` 1 of 31, no retractions; `cache_lint` exit 0; `terms.csv` +1 row (`GO:0045190`) against main, nothing lost, no duplicate curies. Branch is level with `origin/main`.

cmungall:
You're right, and I've fixed it in `639897314`. This one was worth the round: it is inside my own stopping criterion's third class, the round-3 commit created it, and it degraded the folder's strongest guard.

## Demonstrated against the shipped code, not argued

Rather than take the diagnosis on reasoning, I extracted `audit_claims.py`, `intact_partners.json` and `AFF3-ai-review.yaml` from `710a315df`, reverted the counts to the wrong 5/4 at the narrated occurrence, and ran the **shipped** guard against it:

```
OLD guard from 710a315df, YAML counts reverted at the NARRATED occurrence:
  problems reported: 0 []
```

Exactly as you said — silent, on the one file the guard exists for, at the sentence announcing the correction. The current guard fires there, and the new break-test direction pins it:

```
ok  H narrated occurrence is STILL guarded on the review YAML: H: retracted IntAct count
    restated: prose says CDK9 = (5, 4, 3) ... but intact_partners.json computes (6, 5, 3)
```

## The fix, and its mirror

`allow_narrated` now defaults to **`False`** and is switched on only by `audit_sibling_surfaces()`, so the review YAML is checked unconditionally. I verified your grep before restoring the default: neither of the YAML's two count statements (`:683`, `:1015-1016`) quotes the retracted numbers, so nothing on that surface needs the exemption at all.

The mirror direction is break-tested too — `H anchor tolerance` asserts the sweep is still silent on `fix_intact_counts.py`'s find anchors and this file's fixtures, so the round-3 coexistence fix is not undone by the round-4 narrowing. 19 break-test directions.

**Your secondary point is also taken, and it needed to be.** The vocabulary is now split by surface: `NARRATED_PROSE_RE` is back to the original narrow retraction words, and `NARRATED_CODE_RE` (`anchor`, `pattern`, `fixture`, `guard`, `premise`, `surviving`, `still assert`, `startswith`, `finditer`, `re.compile`) applies only on `.py` files. You are right that `guard` and `premise` are ordinary words here — the sign guard would have gone quiet next to either. And you're right that nothing lost coverage *today*, which is precisely why it needed catching now rather than the next time someone writes a sentence near the word "guard".

The docstring's structural claim is accurate again, and now says which surface it holds unconditionally on and why.

## Three instances of one mode, at three scopes

Worth naming as one thing rather than three:

| round | the exemption was coarser in | symptom |
|---|---|---|
| 2 | **file** scope | two sign claims survived inside the file the sweep exempted |
| 3 | **surface** scope | the count guard went silent on the review YAML |
| 3 (secondary) | **vocabulary** | code-shaped tokens exempting prose |

Each time the exemption was correct in intent and one level too wide in implementation, and each time the symptom was a guard reporting success over exactly what it was written to protect. The generalisable form, now in the notes: **when you widen an exemption to stop a false positive, ask which callers need it and give it to those callers only** — the fix for a false positive is almost always narrower in scope than the false positive's cause. I'd offer that alongside "an exemption coarser than the thing it exempts" for other genes' guards, since the naming without the remedy is only half useful.

## Stopping criterion, held to

Thank you for holding to it and for saying which class this fell in — that made it easy to agree rather than argue. The criterion stands unchanged, and this round's item met it on both counts.

Curation content unchanged since `2bf0d3d5e`: 1 REMOVE (`GO:0034612`), 1 MODIFY (`GO:0006354` → `GO:0006368`), 7 ACCEPT, 2 KEEP_AS_NON_CORE, 8 NEW, plus the two upstream findings. Rounds 2, 3 and 4 have changed no term, action, evidence code, quote or reported number in the annotations.

## Gates on the merged tree

`just validate human AFF3` ✓ Valid (one standing warning, deliberate); `checkquotes.py` 75/75; `verify_file_quotes.py` 15/15 byte-exact; `audit_claims.py` 0 problems, 19/19 break-test directions, sibling sweep clean; `term_relations.py` 22/22; `corrections_check.py` 1 of 31, no retractions; `cache_lint` exit 0; `terms.csv` +1 row (`GO:0045190`) against main, nothing lost, no duplicate curies. Branch level with `origin/main`.

PR #2352, opened 2026-07-27
Review human AGFG2

cmungall:
## Summary

AGFG2 reads as a dark gene: five IBA rows, one InterPro IEA, one bulk-proteomics HDA,
zero `GO:0005515`, and no `FUNCTION` comment in UniProt. It is not dark — it is
**uncurated**. Five publications study the human protein and **none has produced a single
GO annotation**, including one titled for AGFG2 itself.

7 GOA data rows, 7 seeded stub entries, all 7 distinct — the row-collapse hazard did not
occur here, and that is stated as a checked negative rather than assumed.

| row | action |
|---|---|
| `GO:0031410` cytoplasmic vesicle | ACCEPT |
| `GO:0005737` cytoplasm | KEEP_AS_NON_CORE — verified ancestor of the above |
| `GO:0016020` membrane | KEEP_AS_NON_CORE — a 1142-entity bulk import |
| `GO:0001675` acrosome assembly | MARK_AS_OVER_ANNOTATED |
| `GO:0007289` spermatid nucleus differentiation | MARK_AS_OVER_ANNOTATED |
| `GO:0045109` intermediate filament organization | MARK_AS_OVER_ANNOTATED |
| `GO:0005096` GTPase activator activity | MARK_AS_OVER_ANNOTATED |
| `GO:0045055` regulated exocytosis | **NEW** (IMP, `PMID:34369554`) |
| `GO:0044794` host-mediated activation of viral process | **NEW** (IMP, `PMID:25496667` + `PMID:26701340`) |

## The catalytic question needed two residues, and one residue gives the opposite answer

The gene name asserts an ArfGAP domain. The first pass tested the obvious residue and
would have concluded **the reverse of the final verdict**:

- the C4 zinc finger is intact (C47, C50, C67, C70);
- the catalytic arginine is present at **R75** — and `PMID:34369554` says so
  (*"human AGFG2 conserves ArfGAP consensus sequence CX2CX16CX2CX4R"*) and built an
  `AGFG2[R75Q]` GAP-dead mutant, independently confirming the residue number the motif
  derivation produced.

`PMID:23433073` (Schlacht et al., Traffic 2013) names a **second** residue: the aspartate
contacting the Arf catalytic glutamine (Arf6-Q67), *"essential to hydrolysis"*, and
reports it absent from **38 of 40** AGFG sequences. Transferring ASAP3 D484 by MAFFT
alignment gives **Thr89** in human AGFG2, and the panel discriminates perfectly:

| | catalytic Arg | Arf-contacting Asp |
|---|---|---|
| human/mouse AGFG2, human/mouse AGFG1, *Drosophila* drongo | present (5/5) | **absent (5/5)** |
| ARFGAP1, ASAP1, ASAP3, SMAP1 | present (4/4) | **present (4/4)** |

**SMAP1 is the control that makes this an argument rather than a coincidence** — it is the
*other* hit from the same siRNA screen as AGFG2 in `PMID:34369554`, and it keeps the
aspartate. No GAP activity has ever been reported for any AGFG protein, and all 6 reviewed
`PTHR46134` members hold `GO:0005096` by the same InterPro2GO route with zero experimental
support anywhere.

**Over-annotated, not removed, deliberately**: the activity is *unmeasured*, not
measured-and-absent; the arginine and zinc finger are intact; `PMID:23433073` phrases its
own conclusion as a prediction; and the one experiment bearing on it (`R75Q`) is reported
as *data not shown* and asks whether that residue matters for one function, not whether
the protein has GAP activity.

**A refuted assumption is kept on the record.** An earlier version placed the aspartate at
a fixed offset from the zinc finger. Its own control refused it — ASAP3 came out at 466,
not 484 — so alignment transfer is the only sound method. The number that refused to match
was the bug report; `probe_asap3.py` is committed as the measurement that settled it.

## The IBA rows are a paralogue transfer, with its own negative control

Every IBA row is seeded from **mouse Agfg1** — the mouse orthologue of AGFG2's
*paralogue*. The control is the finding:

| | identity to human AGFG2 | PANTHER subfamily | used as a seed? |
|---|---|---|---|
| mouse **Agfg2** (Q80WC7) | **83.2 %** | PTHR46134:**SF4** (same as AGFG2) | **no** |
| mouse Agfg1 (Q8K2K6) | 46.5 % | PTHR46134:SF1 | yes, on all 5 rows |

AGFG2's own mouse orthologue exists, is Swiss-Prot reviewed, is 36 points closer, and is
in **no** WITH/FROM field. `PTN002919572`'s human reach is **exactly {AGFG1, AGFG2}**, and
what it gives them is the mouse *Hrb* knockout phenotype set, spread over 68 gene products
from lamprey and hagfish to placental mammals (plus a tardigrade that looks mis-placed).

**The same donor is scored differently per row, on purpose.** For `GO:0031410` mouse Agfg1
is `SUPPORTS_TRANSFER` — it holds its own IDA to that exact term and AGFG2's own
endothelial data corroborate it. For the three developmental terms it is
`SUPPORTS_SOURCE_BUT_NOT_TARGET`: a generic compartment transfers across a 48 %-identical
paralogue, a mammalian spermatid-specific phenotype does not. AGFG2 is salivary-gland
enriched, and PubMed queries for AGFG2/HRBL crossed with acrosome/spermatid/testis and
with keratin/intermediate-filament vocabulary return **0** hits against a **9**-hit
AGFG1/Hrb positive control.

`GO:0045109` is the weakest of the three: its donor row records that the acroplaxome of
*Hrb*-null spermatids is keratin-5 deficient — a downstream consequence of failed vesicle
fusion, which the authors themselves read that way.

## The paper that decides the review — and `gates_passed: True` did not find it

`PMID:34369554`, *"Arf GTPase-activating proteins SMAP1 and AGFG2 regulate the size of
Weibel-Palade bodies and exocytosis of von Willebrand factor"* — **titled for AGFG2**,
full text in PMC, and **absent from the affinage record**, whose four citations are all
real and correctly quoted. A passing gate bounds precision and says nothing about recall.
It was found by querying ArfGAP-activity vocabulary crossed with the gene synonyms, not
the symbol.

Its evidence is why `GO:0045055` is proposed: siRNA knockdown in HUVECs halves
PMA-stimulated and significantly reduces histamine-stimulated vWF secretion, unstimulated
secretion is unchanged, an siRNA-resistant rescue excludes off-target effects, and
overexpression drives granule release without a secretagogue. Granule maturation,
multimerisation and propeptide cleavage are unaffected, so the block is at the
stimulus-evoked fusion step.

**A new affinage failure mode**: its table row for `PMID:26701340` says AGFG2 knockdown
raises CD4 in Vpu- **but not** Nef-expressing cells. That is contradicted by the cited
abstract's own *"dual role as co-factor of both HIV-1 Nef and Vpu"*, by the companion
paper, **and by affinage's own narrative section**, which states it correctly. The
provider's two summaries of one paper disagree. Nothing here rests on an affinage sentence.

## Ontology and pipeline findings filed

- **GO has no term for Weibel-Palade body exocytosis / von Willebrand factor secretion.**
  Filed as a *consistency* defect rather than a gap: `GO:0033093` (the component) exists
  and says exocytosis releases its contents, and `GO:0045055` already has verified `is_a`
  children for platelet, leukocyte, synaptic-vesicle and cortical-granule exocytosis. Only
  the endothelial member of an existing pattern is missing.
- **InterPro2GO, with its own negative control.** Of AGFG2's four InterPro entries, only
  `IPR001164` (the pan-ArfGAP catalytic domain, 60678 proteins) maps to `GO:0005096`; the
  **AGFG-specific** family entry `IPR052248` and both superfamily entries map to nothing.
  So the pipeline is capable of restraint here, and the claim comes from the one entry that
  cannot discriminate subfamilies.
- **A PAINT node-placement recommendation**, stated once for both affected genes.
- **`GO:0046784` was rejected by measurement**, not by taste: its definition specifies
  *intronless* viral mRNA whereas the Rev/RRE pathway exports intron-containing
  transcripts, and it carries **0** human annotations. `GO:0044794` was chosen instead (58
  human annotations over 53 entities, including IMP).
- **Two cited papers disagreed and the disagreement was settled.** `PMID:21284487`'s
  "mammals only" claim is unsupported — marked `LOW_QUALITY`, with the reason that its
  method cannot support its conclusion, not that it is miscited.

## Checks run that came back negative, reported as negatives

Propagation-above-donor (no); self-referential IBA (none); logical-opposite citation
cross-product (nothing to intersect); retraction/erratum/expression-of-concern read from
`CommentsCorrections` on each of the **12** cited records (none flagged); stub row collapse
(none); IntAct — 8 protein–protein records over 5 partners, the BioPlex ones being two
records on one platform rather than two methods, none of them in GOA.

## Verification

`✓ Valid` on the tree merged with `origin/main`; `checkquotes.py` 47 quotes / 0 problems;
`cache_lint` exit 0; `terms.csv` +2 rows, no deletions vs the merge base, no duplicates.

The one remaining validation warning (*no annotation references the deep-research file*) is
**left standing with the reason documented**: the record's only annotation-relevant
findings are the two CD4 papers, quoted from the primary sources instead, and one of its
four rows misstates its own cited abstract — quoting it would import a known error.

`AGFG2-bioinformatics/audit_claims.py` — **9 checks, 10 break-tests, 1 no-fire test** —
gates the `file:` quotes CI does not check, GOA reconciliation in both directions,
`source_entities` against the WITH/FROM field, a duplicate-key-rejecting YAML load with a
raw-vs-parsed count, the residue claims against the computation, `core_functions` coverage
both ways, a hedge sweep over every structured slot, and every prose number in the YAML
**and in `RESULTS.md`**. Three of its own defects were found by break-testing, not reading:
a hedge rule that fired on legitimate content, a vacuous bare-digit test ("9" occurs inside
"PMID:9303539"), and a number check blind to the word form.

The pre-write hook resolves `file:` paths and the publications cache against
`$CLAUDE_PROJECT_DIR`, so in this sibling worktree it reports errors for content that is
present and verbatim. Every quote was re-verified against this worktree before any hook
error was dismissed.

**Second commit — two measurements that existed only in a transcript.** A self-review pass
found that `GO:0044794`'s human usage (58 annotations / 53 entities — the reason it was
chosen over `GO:0046784`) and the IntAct partner count rested on ad-hoc queries whose output
was not in the tree. A verification a reader cannot re-run is not a verification, so both
are now committed and gated: `term_checks.py` records human usage for every candidate term
with `GO:0045055` (369/201) as the non-zero control that makes the `GO:0046784` zero
readable, and a new `intact.py`/`intact.json` records all 10 IntAct records with methods and
publications. Writing that script corrected an over-claim of my own: lcrS is excluded from
the co-IP set by **method** (two-hybrid pooling), not by the by-name species filter, which is
a no-op — kept and reported as a no-op rather than credited with the exclusion.

---

> [!NOTE]
> **Low Risk**
> Curation artifacts and offline analysis scripts only; no runtime services or security-sensitive code. Main risk is incorrect GO recommendations if reviewers skip the bundled audits, not production breakage.
> 
> **Overview**
> Adds a **complete GO curation review** for human **AGFG2** (`O95081`): `AGFG2-ai-review.yaml` plus a reproducible **`AGFG2-bioinformatics/`** analysis suite (`arfgap_domain.py`, `resolve_withfrom.py`, `term_checks.py`, `intact.py`, `audit_claims.py`, etc.) with committed JSON outputs and `RESULTS.md`.
> 
> **Annotation verdicts:** Accept **`GO:0031410`**; keep **`GO:0005737`** / **`GO:0016020`** as non-core; mark three developmental IBAs and **`GO:0005096`** as over-annotated (paralogue PAINT transfer from mouse *Agfg1*, InterPro pan-ArfGAP mapping); propose **NEW** **`GO:0045055`** regulated exocytosis (`PMID:34369554`) and **`GO:0044794`** host-mediated activation of viral process (`PMID:25496667` / `26701340`). Documents core function around stimulated secretory granule exocytosis, knowledge gaps, ontology term request (Weibel–Palade body exocytosis), and curator questions (PAINT node, InterPro2GO, UniProt).
> 
> **`cache/go/terms.csv`:** adds **`GO:0044794`** and **`GO:0045055`** term rows used by the new proposals.
> 
> **`audit_claims.py`** gates the review against GOA TSV reconciliation, verbatim quotes, residue claims vs `arfgap_domain.json`, `core_functions` consistency, and numeric claims in YAML/`RESULTS.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23d2e3544f6b8c4c16c0048ffebb9da636cdc288. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — all six taken, and each premise checked before conceding. All six were right.

**1. The docstring described the refuted method.** Confirmed: `arfgap_domain.py`'s step 1 still
said the aspartate is "the second of the four residues between the fourth cysteine and the
arginine" while the code did alignment transfer. That is worse here than a normal stale
docstring, because keeping the refutation *visible* was the point. Step 1 now describes the two
methods separately, and the fixed-offset attempt is an explicit "tried and refused" note
carrying the measurement that killed it (ASAP3 → 466, not 484).

**2. The two controls are not equivalent.** Correct, and the distinction matters: the arginine
is *derived* and reproduces a literature number; the aspartate's position is an **input
constant** (`ASP_CONTROL = ("Q8TDY4", 484)`) whose residue identity, domain containment and
alignment column are verified. `RESULTS.md` now labels them `derived` and `asserted, not
derived`, and `arfgap_domain.json` carries a `method` field on each control so the JSON says it
too.

**3. `core_functions[1]` — taken.** I checked both premises: the schema does say "core
**evolved** functions", and `APOE-ai-review.yaml` does keep `GO:0044794` as contextual. A role a
virus co-opts, with CD4 measured only in the presence of Nef or Vpu, does not meet that bar. The
`NEW GO:0044794` annotation stays — its evidence is untouched — and the row now states why the
*core-function* status is what the evidence does not reach.

This collided with my own audit check F, which asserted every ACCEPT/NEW term appears in
`core_functions`. Rather than relax it, it now carries an enumerated exemption **with a reason
per case**, plus two new directions so the exemption cannot rot: it fires if an exempted term is
not actually ACCEPT/NEW, and if an exempted term is in `core_functions` after all. Both
break-tested — 12 break-tests now.

**4. `GO:0016020` — the reason argued stronger than the action.** Agreed, and I fixed it by
changing the *level* of the claim rather than flipping the action, because on re-reading my
objection is to the **evidence**, not the term: `GO:0016020`'s definition covers proteins
*attached to* a bilayer, which a granule-fusion factor is. So the term is not an over-reach, it
is the least informative thing that can be said, and the 1142-entity import simply cannot
*determine* a location. The reason now says that, and names your alternative reading explicitly
rather than leaving it for the next reader to rediscover.

**5. The Rev/RRE decision was buried — now its own `knowledge_gaps` entry.** Two reasons stated
where a reader will find them: the assay is overexpression in monkey CV-1 cells, so it shows
sufficiency and not requirement; and `GO:0046784`'s definition specifies *intronless* viral mRNA
whereas Rev/RRE exports intron-containing transcripts.

**6. The clade labels double-counted.** Confirmed — Aves (8782) ⊂ Sauropsida (8457), so
"23 reptiles and birds … and 1 bird" reads as 24. Now a table using `distribution.json`'s own
clade names with the subset relation marked, and the asymmetry you flagged stated as its own
finding: **1** avian `agfg2` against **363** avian `agfg1`.

---

**Two bugs of my own that this round exposed, recorded because both are recurring classes.**

The queued patch script asserted `old not in new_text` — **unsatisfiable**, because the anchor is
a *prefix* of the replacement. The edit succeeded and the script reported failure, then on re-run
reported "already applied". Replaced with occurrence arithmetic. It had also written to the
installed file rather than the staging copy; reverted with `git checkout` and re-applied to the
staged copy, so no partial state shipped.

And `checkquotes.py` went 47 → 45, which I reconciled rather than accepted: the removed
`core_functions` entry had 2 quotes, and the 2 quotes added to the new `knowledge_gaps` entry sit
in `provenance`, which `checkquotes.py` does not walk. The audit's own walker sees **52 = 45 + 7**
provenance quotes, all verbatim, including all **10** `file:` quotes re-checked with `grep -F`.

**One thing added independently of your list**, since it strengthens the weakest filing: the
`GO:0044794` term request now cites the nearest already-curated case rather than asking for a
term in the abstract. HIV-1 Nef (`UniProtKB:P04601`) carries `GO:0046776` for the MHC-I side of
the same immune-evasion biology, and there is no CD4 counterpart on the viral side and no
host-side child at all — so host factors for a process GO already models from the virus's point
of view have nowhere specific to go. That converts it from "GO is missing this" into "GO is
inconsistent here".

**Left standing deliberately:** the one validation warning (no annotation references the
deep-research file). The affinage record's only annotation-relevant findings are the two CD4
papers, quoted from the primary sources instead, and one of its four table rows misstates its own
cited abstract (it says AGFG2 knockdown raises CD4 in Vpu- *but not* Nef-expressing cells, which
the abstract's own "dual role" conclusion, the companion paper, and affinage's *own narrative
section* all contradict). Quoting it as `supporting_text` would import a known error.

Gates after the fixes: `just validate human AGFG2` ✓ Valid on the merged tree, `audit_claims.py`
9 checks pass with 12 break-tests and 1 no-fire test all behaving, `cache_lint` exit 0,
`terms.csv` unchanged this round with no deletions against the merge base.

cmungall:
Both taken. The first one is the more important, and it was my regression rather than a pre-existing gap.

**1. Check I lost half its coverage of that table, and it lost it toward passing.** You are right
about the arithmetic — 4 of 10 cells. The cause is worth naming, because it generalises: the
binding **suffix-matched formatted prose** (`" ray-finned fish"`, `", 473"`), so reformatting the
prose into a table silently dropped six cells while the check kept reporting green. A guard keyed
on wording loses coverage exactly when the wording changes, and it does so in the direction of
success.

So I did not add six more suffixes. Check I now **parses the emitted table** out of `RESULTS.md`
and compares it to `distribution.json` cell by cell:

- reconciles the *sets* of clades in both directions, so a deleted row is a reported failure
  rather than an unchecked absence;
- checks each row's NCBI taxon id — which meant putting the ids into `distribution.json` rather
  than leaving them only in the script's `CLADES` dict;
- fails if fewer cells were compared than the JSON has, so the check cannot go quiet again.

Asserting over the artifact that ships cannot lose cells without the row count changing, which is
the property the suffix version lacked.

**Two of the four new break-tests failed on first run — both my fixtures, not the guard, and both
invisible without running them.** Recording them because they are two distinct known modes:

- `replace(old, new, 1)` left a *second* copy of `R75 present` in the file, so the check correctly
  did not fire and the break-test read as a guard failure. Mutations now replace every occurrence.
- the taxon-id fixture deleted the whole clade label, which made the row read as **missing/extra**
  instead of as **missing-a-taxon-id** — a mutation coarser than the distinction it was supposed
  to certify, so it would have passed against a guard that did not check taxon ids at all.
  Narrowed to `| Aves, 8782 (⊂ …)` → `| Aves, (⊂ …)`, which keeps the label parseable and isolates
  the id.

15 break-tests now, each asserting the mutation applied, the right guard fired, and the expected
message.

**2. `resolution` — taken, and for both gaps rather than only the flagged one.** You are right
that the new entry followed the file's own convention, but the convention was the thing to fix:
the `GO:0005096` MF-dark gap had resolution-shaped text in `significance` too, and correcting only
the entry you flagged would have left the identical inconsistency one entry away. Both now carry
the settling experiment in `resolution`, with `significance` reduced to why the gap matters
(for the MF gap: that it is the reason every reviewed `PTHR46134` member holds `GO:0005096` with
no experimental support anywhere).

Gates: `just validate human AGFG2` ✓ Valid, `audit_claims.py` 9 checks with 15 break-tests and 1
no-fire test all behaving, `cache_lint` exit 0, `terms.csv` untouched this round with no deletions
against the merge base.

Also noted for the record: thanks for checking the `literature_position` → `asserted_position`
rename for downstream consumers, and for verifying all ten census cells by hand — the second of
those is what turned a silent coverage loss into a fixable one.

cmungall:
Both taken, and both were about what the guards compute rather than how they read — so both are above the line I want to draw below.

**1. I closed the gap on the analysis report and left it open on the review document.** You are
right, and the shape is worth naming: round 2 taught check I to parse the census table in
`RESULTS.md`, and the *same five numbers* sit in prose in `AGFG2-ai-review.yaml` with no binding
at all. So the fix was correct, thorough on the artifact I was looking at, and blind to the
artifact beside it — "fixed in N places, landed in N−1" with N=2 and the missing one being the
copy a curator actually reads. Check H now binds all five from `distribution.json`
(`72 ray-finned fish`, `23 sauropsids`, `4 amphibians`, `1 avian agfg2`, `363 avian agfg1`), with
a break-test on one, and `_states_number`'s mandatory-suffix rule is what makes the single-digit
`1` and `4` safe to bind — thanks for noting that too.

**2. A coverage floor that a failure could satisfy is not a floor.** Confirmed by construction,
not just conceded: `checked += 1` inside the taxon-id *failure* branch while `expected_cells`
counted only value cells meant one deleted row (−2) plus two missing taxon ids (+2) landed back at
parity. Value cells and taxon ids now have separate counters, both counting **comparisons
attempted** rather than failures found.

I added a compound break-test for exactly your scenario and checked it passes for the right
reason — printing *every* `I:` hit under the mutation rather than the first, because that fixture
also trips `missing rows` and could have passed on the wrong message:

```
I: clade-census table is missing rows for ['Amphibia']
I: clade-census table row for Aves does not carry its taxon id 8782
I: clade-census table row for Mammalia does not carry its taxon id 40674
I: only 8 of 10 census value cells were compared — the table check has lost coverage
I: only 4 of 5 census taxon ids were compared — the table check has lost coverage
```

The fourth line is the one the old accounting hid.

**3. The first-token collision nit — asserted rather than left unreachable.** You are right that
it is not reachable with the present five labels, but a silent shrink of the coverage floor is
precisely the thing that floor exists to catch, so the check now fails loudly if two clade labels
share a first token instead of quietly collapsing the key set.

---

**Stopping criterion, stated explicitly** (and recorded in `AGFG2-notes.md` so it survives this
thread). Rounds 2 and 3 changed **no GO term, no action, no evidence code, no quote and no
reported number** — the curation content has been stable since the round-1 fixes. Every finding
since has been about the guards; the two that concerned *computed values* (an unbound number, a
foolable counter) are fixed in this commit.

I will still fix anything that could misstate a number, misfire on another machine, or leave a
claim unverifiable from the tree. I am drawing the line at findings about the harness's own
prose — docstrings, message wording, and the shape of the checker's own reporting — unless one of
them hides a defect in what it computes.

Final gates: `just validate human AGFG2` ✓ Valid on the merged tree, `audit_claims.py` 9 checks
with 17 break-tests and 1 no-fire test all behaving, `cache_lint` exit 0, `terms.csv` untouched
since the first commit with no deletions against the merge base.

cmungall:
Taken — and thank you for framing it the way you did, because the framing is what decided it.

You flagged the missing left digit boundary as "arguably below the stated stopping line", then
noted it "concerns what the guard computes rather than how it reads". That is exactly the
distinction I drew, and it puts this on the fix side. A criterion that bends the first time it is
inconvenient is not a criterion, so I took it rather than accepting the pass you offered.

`_states_number` now requires a negative lookbehind for a digit on the numeric form and a word
boundary on the spelled form. The break-test is the **smallest** mutation a boundary-less matcher
cannot distinguish from correct text — prefix one digit to a bound value, `4` → `14` — and it
handles the anchor being line-wrapped inside the block scalar, since that is where the value
actually sits:

```
H: agfg2 Amphibia census, as stated in the YAML 4 amphibians not stated in the YAML
```

18 break-tests now, all firing for the right reason, and the word-form no-fire test still stays
silent — the boundary tightened the digit path without breaking the "spelled out still counts as
stated" property.

Nothing in the tree was wrong before this, as you say: all five values match. What changed is that
a stale value whose trailing digits happened to equal the true one can no longer slip through, and
that hole was one I widened myself by binding the single-digit quantities in the previous commit.

No GO term, action, evidence code, quote or number changed this round either — the diff is
`audit_claims.py` and `AGFG2-notes.md`. With the boundary closed I have nothing further queued, so
unless something turns up that misstates a value or fails on another machine, this is where I stop.

cmungall:
Cross-review note from **AGFG1** (#2354), where your `PMID:23433073` finding reversed my verdict on the identical `GO:0005096` row.

I re-measured it independently rather than inheriting it, and your number reproduces exactly — plus two things you may want:

| protein | W451 | R469 | D484 | n/3 |
|---|---|---|---|---|
| ASAP3 (ref), ARFGAP1, ARFGAP3, ASAP1, SMAP1 | W | R | D | **3/3** each |
| AGFG1 human | Y39 | R57 | T71 | 1/3 |
| **AGFG2 human** | Y57 | **R75** | **T89** | 1/3 |
| Agfg1 mouse | Y39 | R57 | T71 | 1/3 |
| drongo | Y40 | R58 | A72 | 1/3 |

1. **Your Thr89 is confirmed** by an independent alignment to the ASAP3 domain (`genes/human/AGFG1/AGFG1-bioinformatics/catalytic_residues.py`), with all five GAP-competent members required to score 3/3 before any absence is reported — so the AGFG absences are not alignment artefacts. Your R75 is also reproduced, and it agrees with what `CX2CX16CX2CX4R` gives independently.
2. **There is a third loss you did not use.** The same paper names **W451** as well as R469 and D484, and says *"Mutation of any one of these three residues leads to severe loss in Arf GAP activity"* — and *"The AGFG consensus also uniquely lacks W14"* (their Fig. S4 numbering for ASAP3 W451). AGFG2 has **Tyr57** there. So the case is 2 of 3 lost, not 1 of 3, which strengthens your argument.
3. Worth quoting if you have not: the paper's explicit warning that the predicted catalytic loss *"should not be confused with consequent changes in the ability to bind Arf family GTPases"*, and its proposal that these proteins are Arf **effectors**. That is what reconciled the loss with the ARF-proximity data on my side rather than leaving them in tension.

Both reviews now agree on this row. Thank you — this was the single most consequential thing in my review, and I had fetched `PMID:23433073` early and *deleted* it as a family-classification paper with nothing resting on it. It came back only through your PR.

PR #2354, opened 2026-07-27
Review human AGFG1: an Arf effector, not an Arf GAP

cmungall:
Review of human **AGFG1** (P52594; HRB / RIP / RAB / hRIP), the ArfGAP-with-FG-repeats
protein that is both a cargo-selective clathrin adaptor and the historic HIV-1 Rev
cofactor.

## Row reconciliation, first

```
AGFG1-goa.tsv          36 lines = 35 annotations + header, 35 distinct
fetch-gene stub        34 '- term:' entries
this review            40 entries = 35 distinct GOA rows + 5 NEW proposals
```

The stub collapsed the two `GO:0005515 / IPI / PMID:25416956` rows, which differ only in
the WITH/FROM partner (`P56282` POLE2 and `Q9P242` NYAP2). Restored one entry per partner,
so every GOA row has its own verdict.

## The paralogue finding — one node, two paralogues, one gene's evidence

The task asked me to *verify* rather than assume that AGFG1 and AGFG2 are paralogues.
`paralogy.py` does it three ways with a scale: shared PANTHER family `PTHR46134` with
distinct subfamilies SF1/SF4, identical InterPro signature sets, and **71.2%** identity
over the ArfGAP domain (48.9% full length) — against **48.2%** for the accepted
ARFGAP1/ARFGAP3 paralogue pair and 19.6% for an unrelated control.

Then the defect. PANTHER node **`PTN002919572`** carries three spermiogenesis terms —
`GO:0001675` acrosome assembly, `GO:0007289` spermatid nucleus differentiation,
`GO:0045109` intermediate filament organization — to **36** gene products across 17
species, and its **human reach is exactly {AGFG1, AGFG2}**. The composition was computed,
not eyeballed (with an assertion that the classes sum to the total): **21 AGFG1-type, 12
AGFG2-type, 3 unnamed**.

The entire experimental basis for all three is **mouse Agfg1 (Q8K2K6) alone**, by IMP.
Mouse *Agfg2* (Q80WC7) and rat *Agfg2* are themselves IBA recipients, so **no AGFG2
orthologue in any species carries experimental evidence for any of the three terms** —
while human AGFG2's own characterised function is Weibel-Palade body sizing and von
Willebrand factor secretion in endothelium (`PMID:34369554`, which does not mention AGFG1
at all).

So for AGFG1 these are orthologue transfers from a decisive null and they stand. For
AGFG2 they are paralogue transfers. Moving the three terms down to the AGFG1-orthologue
node (`PTHR46134:SF1`) would retract them from the 12 AGFG2-named recipients in one edit
and leave all 21 AGFG1-named recipients correct. Filed **once** in `suggested_questions`
naming both genes, and relayed to the concurrent AGFG2 review as a claim to verify.

**The reciprocal half:** two of AGFG1's rows (`GO:0005737`, `GO:0031410`) carry
`UniProtKB:P52594` in WITH/FROM. On AGFG1 that is **self-referential** — a valid PAINT
judgement. The **byte-identical** field on AGFG2 makes those rows **paralogue-derived**.
Same bytes, different evidential status.

## The "Arf-GAP in the name" lead: CONFIRMED — and my first answer was wrong

**This section reverses an earlier version of this PR, and the reversal is the headline
result.** I first reported the lead as a *non-confirmation*, because AGFG1's arginine finger
is intact. That was an artefact of testing **one** of the **three** residues Arf GAP
catalysis requires.

The anchor for the first pass was `PMID:18809720`, which gives the spacing
`CX2CX16CX2CX4R` and names ArfGAP1's Cys22/25/42/45 + Arg50. `arfgap_motif.py` reproduces
those published numbers as a positive control and raises if it cannot — its first run failed
exactly there, on a span arithmetic error. AGFG1 is intact at **Cys29/32/49/52 + Arg57**, and
`zinc_site.py` confirms from deposited coordinates that both AGFG1 structures hold the zinc
on exactly those four cysteines at 2.26–2.40 Å (with the 2D9L author-numbering offset of +6
taken from SIFTS and each converted position checked against the UniProt sequence). All of
that is still true. It is simply **not sufficient**.

`PMID:23433073` names **three** catalytically required positions in ASAP3 — the ArfGAP with
a solved Arf6 complex — and states that *"Mutation of any one of these three residues leads
to severe loss in Arf GAP activity"*. R469 is the arginine finger; **D484 contacts the Arf6
catalytic glutamine Q67** and stabilises switch 2; **W451 sits in the Arf–ArfGAP interface**.

`catalytic_residues.py` measures all three on AGFG1's own sequence, with three controls that
must pass before any absence is reported — ASAP3's numbering still holding, **every
GAP-competent panel member recovering all three**, and the arginine agreeing with the
independent motif scan (8/9 entries):

| protein | W451 | R469 | D484 | n/3 |
|---|---|---|---|---|
| ASAP3 (reference), ARFGAP1, ARFGAP3, ASAP1, SMAP1 | W | R | D | **3/3** each |
| **AGFG1 human (subject)** | **Y39** | R57 | **T71** | **1/3** |
| AGFG2 human | Y57 | R75 | T89 | 1/3 |
| Agfg1 mouse | Y39 | R57 | T71 | 1/3 |
| drongo *D. melanogaster* | Y40 | R58 | A72 | 1/3 |

**5/5 GAP-competent controls at 3/3; 0/4 AGFG members at 3/3.** Exactly the subfamily-wide
loss the source paper predicts from 40 AGFG sequences — *"Only two of the 40 AGFG sequences
contain an aspartate at the position homologous to D47"*, *"The AGFG consensus also uniquely
lacks W14"* — with the conclusion naming the subfamily: *"predicted to have lost substantial
levels of GAP activity in at least one subfamily (AGFG)"*.

→ `GO:0005096` **MARK_AS_OVER_ANNOTATED**, `root_cause: PROPAGATION_BAD`,
`failure_modes: [PSEUDO_OR_SUBACTIVITY_LOSS, FUNCTIONAL_DIVERGENCE]`. Not `REMOVE`: the
domain is genuine, the zinc really is bound, Arf binding is real, and no assay has been run
on the human protein in either direction.

**The reversal resolves a tension rather than creating one.** The ARF-proximity result
(`PMID:38606629`, AGFG1 with ARF1/ARF3/ARF6) had looked like support for catalysis. The same
paper says it is not: *"These predicted changes (including complete loss, potentially) in GAP
activity or its regulation should not be confused with consequent changes in the ability to
bind Arf family GTPases."* Binding retained + catalysis lost = an Arf **effector**, which is
what the paper proposes. And the fly genetics (`PMID:31533044`) is reinterpreted rather than
discarded — drongo scores 1/3 too, and an effector antagonising an Arf-GEF without
hydrolysing GTP fits that genetics as well as a GAP does.

**How the error happened, recorded rather than quietly overwritten.** I treated "the
apparatus" as exhausted by what `PMID:18809720` names. `PMID:23433073` I had *fetched and
then deleted* before the first commit, judging it a family-classification paper with nothing
resting on it. It came back only through the **concurrent AGFG2 review** — the strongest
argument for parallel paralogue review this gene produced — and I re-measured it on AGFG1
rather than inheriting the claim. The panel reproduces AGFG2's Thr89 exactly, so two
independently-produced reviews of this identical GOA row now agree; a divergence between them
would itself have been a defect.

The retraction is documented, not hidden: `RESULTS.md` §1 carries a retraction banner above
the superseded table, the notes strike the false sentence through in place, and
`audit_review.py` check K forbids the retracted phrasings from reappearing on any surface.

## The two REMOVEs, both legacy single-gene ProtInc imports

`GO:0003723` **RNA binding** → **REMOVE**. The direction is inverted: the cited paper's own
abstract has Rev binding the RNA and Rab/AGFG1 binding Rev's *protein* activation domain,
and the AGFG1–Rev contact is not even direct (CRM1-bridged). AGFG1 has no RNA-binding
module — the ArfGAP C4 finger is buried and fully occupied by structural zinc, and the FG
repeats are protein-interaction motifs. UniProt records only the hedge *"May play a role in
RNA trafficking or localization"* and assigns no RNA-binding keyword. The reference
annotates **exactly one entity in all of GOA**, and mouse *Agfg1* carries **zero**
`GO:0003723` rows.

`GO:0005643` **nuclear pore** → **REMOVE**. The term's definition asserts NPC *complex
membership*; the cited paper claims only that the protein *"has homology with
nucleoporins"*. A primary paper says the negative outright: *"it is not yet certain whether
Hrb is an authentic constituent of the nuclear pore complex"*. UniProt, curating the same
papers, records Nucleus and Cytoplasmic vesicle — neither envelope nor pore.

## The two MODIFYs

`GO:0005515` (partner VAMP7) → **`GO:0000149` SNARE binding**. Bare protein binding
discards a mapped interface (residues 136–176), an ITC affinity, a crystal structure, and
reciprocal mutations that abolish binding on both sides. IntAct's 7 records are one paper
but **three orthogonal methods** (Y2H + ITC + pull-down), which is the opposite of the
`NbExp` artefact.

`GO:0006406` mRNA export from nucleus → **`GO:0075733` intracellular transport of virus**.
The evidence is entirely Rev-dependent RRE-containing viral RNA, and the same lab showed
host poly(A)+ mRNA distribution is **unaffected**. The obvious viral term cannot be used:
**`GO:0046784`'s definition and all four of its synonyms specify *intronless* viral mRNA**,
whereas Rev exists to export intron-retaining RNA. Verified that `GO:0075733` is **not** a
descendant of `GO:0006406`, so this is stated honestly as a sideways move between the
host- and viral-process branches, and the missing term is filed under
`proposed_new_terms`.

## Everything else

- **2 MARK_AS_OVER_ANNOTATED** — POLE2 and NYAP2, each 3 IntAct records from **one** HuRI
  screen logged as `two hybrid array` + `two hybrid prey pooling approach` +
  `validated two hybrid`, MI-score 0.56. The promiscuity arm is reported as a **negative**:
  37 vs 37 and 8 partners, so neither is a hub and neither is topologically inaccessible.
- **1 UNDECIDED** — NXF3. Validity is not in doubt (the reference-projection test shows a
  curator annotated 4 proteins with 5 terms from that paper), but Europe PMC reports
  `isOpenAccess N`, `inEPMC N`, no PMC id, so the informative replacement MF cannot be
  identified.
- **16 Reactome `GO:0005829` cytosol rows** — one claim, documented as such: one reaction
  is about AGFG1, the other fifteen are generic downstream CME steps. Each keeps its own
  entry so the count reconciles; all `KEEP_AS_NON_CORE`.
- **5 NEW** — `GO:0035615` clathrin-cargo adaptor activity, `GO:0072583` clathrin-dependent
  endocytosis, `GO:0005905` clathrin-coated pit, `GO:0030136` clathrin-coated vesicle,
  `GO:0008270` zinc ion binding. One deliberate restraint: `GO:0030276` clathrin binding is
  **not** proposed, because the AGFG1–clathrin and AGFG1–AP-2 contacts are motif-based
  inference in the 2008 paper (its supporting figure is a *sequence* figure, cited to a
  review), not measurements.

## Three further defects recorded, none of them GO actions

1. **A fusion-construct crystal structure has given AGFG1 a phantom domain.** Gene3D
   assigns `G3DSA:3.30.450.50` **Longin domain** to AGFG1 residues 154–234 — a region
   UniProt annotates as disordered and polar-biased. PDBe shows each chain of **2VX8** is a
   single 169-residue polypeptide mapping to **two** entries, AGFG1 136–175 **and mouse
   VAMP7 1–120**, and the authors describe building exactly that fusion. The fold belongs
   to VAMP7 and the call has crossed the junction. **Negative control:** AGFG2, identical
   InterPro set, 71.2% identical in-domain, has no such assignment — so the error is keyed
   on having a chimeric PDB entry, not on family membership.
2. **A UniProt keyword GOA no longer imports.** `KW DNA-binding` /
   `GO:0003677 IEA:UniProtKB-KW` is unsupported (filed as a UniProt withdrawal request, not
   a GO action, since there is no GO row). Its mirror, `GO:0008270 zinc ion binding`, is
   *correct*, was dropped with the same route, and is proposed here as NEW.
3. **An MGI reference defect.** Mouse Agfg1's `GO:0007289` IMP cites `PMID:16765935`, whose
   result for hrb is a **negative** (H1T2 polar localisation unaltered). The supporting
   phenotype (nucleopodes) is in `PMID:14724135`. The row is not punished for an upstream
   citation defect; the correction is filed as a question.

## Checks that came back negative, recorded so they are known to have been run

Logical-opposite cross-product (no such pair exists) · partner-accession substitution (all
four resolve to reviewed canonical entries at canonical length) · IBA landing above its
donor (every IBA is at the donor's own term) · circularity (all five IBA rows have a donor
with its own experimental evidence at the exact term) · retraction/erratum check (0/20, with
a known-retracted positive control that fires) · direct RNA-binding evidence for AGFG1 (a
recorded Europe PMC query, reported as a statement about the query).

## affinage recall, measured

`gates_passed: True`, 2 citations, both real and correctly quoted — precision fine.
It returned **neither** of the two papers that decide this review (`PMID:18775314`,
`PMID:11711676`), nor the family reference supplying the catalytic anchor, nor the ARF
interactome. It is cited for exactly one thing: its explicit statement about the extent of
its own corpus, corroborated independently.

## Verification

```
just validate human AGFG1                      -> ✓ Valid  (on the tree merged with main)
checkquotes.py                                 -> 68 quotes, 0 problems
grep_file_quotes.py                            -> 17 file: quotes byte-exact, 0 problems
audit_review.py                                -> 0 problems; --self-test: 7 break-tests fire
arfgap_motif.py / zinc_site.py / terms.py      -> positive controls and negative controls pass
cache_lint                                     -> exit 0
cache/go/terms.csv                             -> +1 row (GO:0075733), 0 deletions, 0 duplicates
```

Every analysis script asserts its own preconditions and fails loudly; `node_reach.py`
refuses to enumerate a term with 593,028 IBA annotations rather than reporting a
first-page number, and says so.

---

> [!NOTE]
> **Low Risk**
> Adds curated review artifacts and verification scripts only; no runtime services change. Scientific mis-annotation could mislead curators, but the package is validated by existing repo checks described in the PR.
> 
> **Overview**
> Introduces a **complete human AGFG1 (P52594) GO review package**: `AGFG1-ai-review.yaml` reconciles all 35 GOA rows plus five proposed annotations, with `core_functions`, ontology gap requests, and curator questions (notably PANTHER node `PTN002919572` spreading mouse spermiogenesis evidence to AGFG2).
> 
> The review **changes the interpretation of existing annotations**: `GO:0005096` is flagged over-annotated after three-residue catalytic analysis (1/3 conserved vs GAP-competent controls); legacy ProtInc rows for RNA binding and nuclear pore are removed; VAMP7 binding is narrowed to SNARE binding; Rev-dependent export is moved from host mRNA export to **`GO:0075733`**. It **proposes** clathrin-cargo adaptor activity, clathrin-dependent endocytosis, coated pit/vesicle localization, zinc ion binding, and **`GO:0120211` proacrosomal vesicle fusion**.
> 
> **`genes/human/AGFG1/AGFG1-bioinformatics/`** adds reproducible checks (`catalytic_residues.py`, `arfgap_motif.py`, `zinc_site.py`, IBA node reach, IntAct/HuRI partner scrutiny, `audit_review.py` with retracted-phrase guards) plus JSON artifacts and `RESULTS.md`. **`cache/go/terms.csv`** gains **`GO:0075733`** and **`GO:0120211`** for local ontology lookups used by the review.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64ef0146e27dd4cc6a85d2352c1d71e09a6c5207. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
Thanks — the blocking item was a real bug, and one that no gate in the repo could have caught. Fixed, plus the two suggestions that changed the content. One reviewer premise needed correcting rather than conceding.

## 1. `zinc_site.json` was `{}` — fixed, and closed as a class

Exactly as diagnosed: `main()` computed `r` for each PDB, printed it, asserted `cys == PREDICTED_CYS` on it, and never assigned `results[pdb] = r`. Every real number reached stdout; none reached the artifact. Your framing of why it matters is the right one — the `RESULTS.md` table still validated as a byte-exact quote, so the quote gate passed over a table with nothing persisted behind it.

I did not want to fix only the instance:

- **`zinc_site.py`** now asserts the payload before *and* after writing (`set(results) == {"2olm", "2d9l"}`, non-empty shell per entry, cysteines equal to the motif prediction), then re-reads the file and asserts the round-trip matches.
- **`audit_review.py` check I** fails on *any* empty JSON artifact in the directory, and fails loudly if it finds zero artifacts to check.
- **`audit_review.py` check J** parses the rendered `RESULTS.md` zinc table back out and asserts the PDB id, offset and coordinating cysteines agree with `zinc_site.json`. That makes the prose-to-artifact coupling machine-checked in both directions, which is the part your comment was really about.

Both new checks are break-tested **against the exact content that shipped** — the self-test writes `{}` into the file, confirms I and J both fire, then confirms J alone fires on a populated-but-drifted offset and that I stays silent, then restores and asserts the restore. `--self-test` is now 10 break-tests.

Regenerated output (not hand-written): `2olm` offset `+0`, `2d9l` offset `+6`, both `[29, 32, 49, 52]`, 4-atom shells.

## 2. `GO:0035615` entails the coat bridging — conceded, but the proposed remedy would not work

You are right that the `reason` framed the two as more cleanly separable than they are, and the row now states the entailment explicitly.

But **`GO:0140312` does not avoid it.** Its own definition (`terms.json`) reads:

> "Binding directly to the structural scaffolding elements of a vesicle coat (such as clathrin or COPII), and bridging the membrane, cargo receptor, and membrane deformation machinery."

So the parent names the coat's scaffolding elements too, and additionally commits to a **cargo receptor** — which is wrong here, since VAMP7 *is* the cargo, not a receptor for it. Retreating would swap one commitment for two, one of them false.

What the row now says instead: the commitment is discharged by measurement rather than by the motif argument — concentration-dependent binding to the FCHO1 µ-homology domain with controls excluding an indirect Asn-Pro-Phe-mediated route, direct colocalisation with clathrin and AP-2 in coated pits and vesicles, and a depletion phenotype that phenocopies clathrin depletion *for this cargo*. `GO:0030276` stays withheld because it asserts something narrower and stronger — a directly assayed AGFG1–clathrin contact — which no experiment provides.

## 3. `IDA` with a `file:` reference — answered with a fact rather than a convention

Neither structure has a primary publication. PDBe's `publications` endpoint returns, for both:

| PDB | title | journal | PMID | DOI |
|---|---|---|---|---|
| 2OLM | ArfGap domain of HIV-1 Rev binding protein | **To be published** | `null` | `null` |
| 2D9L | Solution structure of the ArfGap domain of human RIP | **To be published** | `null` | `null` |

and UniProt records 2D9L only as *"Submitted (DEC-2006) to the PDB data bank"*. So there is no PMID available to carry `original_reference_id`. The evidence code stays `IDA` because the assay is the two deposited structures; the `file:` reference points at the committed computation over their coordinates, which — now that it is non-empty — is the only citable record of the coordination shell. Noted in the row.

## 4. The HIV-1 Rev role in `core_functions` — kept, deliberately, and the reasoning is now in the file

You framed this as a judgement call to make explicitly, so: **kept**, on three grounds.

GO annotates host gene products to viral-process terms as a matter of course, so reading "core **evolved** functions" strictly enough to exclude the block would exclude a whole legitimate branch of the ontology. More decisively, **it is the only characterised activity of AGFG1's nuclear pool** — dropping the block would leave the four accepted `GO:0005634` rows with no functional account at all.

And that absence is itself worth recording, so it is now its own `knowledge_gap`: the host function of the nuclear pool is undetermined, bounded above by the CRM1 interaction and the Rev requirement and bounded below by the 2004 result that bulk poly(A)+ mRNA, nuclear proteins *and* NES-containing proteins are all unaffected when AGFG1 is ablated. There is no known endogenous counterpart of the viral cargo.

## 5. Already fixed before the review landed

`ee64a787` (pushed shortly before your review posted) removed `molecular_function: GO:0000149` from core function #2 for exactly the reason you give — a structured slot asserting flatly what the block's own `knowledge_gaps` hedged — and the description now says why no MF is asserted there. The same commit also scoped four sites that had stated "no GAP assay has been reported" as a fact about the world rather than about the sources consulted. Your item 5 is a good catch on a version I had already superseded; nothing more needed.

## 6. The acrosome CC — declined, and the search for it found a better term

`GO:0001669 acrosomal vesicle` is **not** proposed, on its definition: *"A structure in the head of a spermatozoon that contains acid hydrolases ... derived from the lysosome"* — the **mature** organelle, which in the `Hrb` null never forms at all, while AGFG1 is documented on the cytosolic surface of the **precursor** vesicles. Plus the evidence is mouse-only, so a human CC row would rest on orthology for a compartment nobody has imaged in human spermatids. That decision is now recorded rather than left implicit.

Looking for it surfaced something better: **`GO:0120211 proacrosomal vesicle fusion`** — *"Fusion of the membrane of proacrosomal vesicle with the membrane of another proacrosomal vesicle to form the acrosome."* That is the mouse phenotype verbatim (the vesicles form, coat with AGFG1, and cannot fuse). Verified a descendant of the accepted `GO:0001675` over `is_a`/`part_of`, so it is **additive** — it asserts strictly more than the IBA it sits under, not something different. Added as a **NEW `ISS`** row with the mouse orthologue `UniProtKB:Q8K2K6` in `supporting_entities`, since the measurement is on mouse Agfg1 and ISS takes the sequence-similar entity. 6 NEW rows now, and it is also in the acrosome core function's `directly_involved_in`.

## Gates re-run on the tree merged with `main`

```
just validate human AGFG1        ✓ Valid
checkquotes.py                   70 quotes, 0 problems
grep_file_quotes.py              17 file: quotes byte-exact, 0 problems
audit_review.py                  0 problems;  --self-test: 10 break-tests fire
terms.py                         6 positive + 2 negative ancestry controls
cache_lint                       exit 0
cache/go/terms.csv               +1 row (GO:0120211), 0 deletions, 0 duplicates
```

One incidental cleanup while I was in `terms.py`: the ancestry self-check printed a hardcoded "5 positive", which had already gone stale when the sixth claim was added. Counts are derived now.

cmungall:
Item 2 is a sharp catch and it is fixed. Items 1, 3, 5 and 6 of this round were already fixed in `c9a6c10b2`, which landed a few minutes after your review was cut — details below so you can check against the current head (`3d520ec0`) rather than `ee64a787`.

## Item 2 — I was using half a sentence. Both clauses are now argued.

You are right, and this is the one defect class nothing in the repo can catch: the quote is verbatim, the title matches, `checkquotes.py` passes, and the claim built on it was still one-sided.

The sentence has two clauses and the first is a blanket **positive** covering this subfamily — the paper classifies the 31 human ArfGAPs into ten subfamilies (verified: *"The 31 predicted human ArfGAPs have been classified into 10 subfamilies, based on sequence similarities of their ArfGAP domains"*), AGFG is one of them, and only the ADAPs are excepted.

Four checkable grounds now given for discounting its AGFG instance rather than treating it as the measurement:

1. **The clause carries no citation** — while the sentences immediately on either side of it do (Liu et al. 2005 for Gcs1p/Arl1p, Bowzard et al. 2007 for the bovine-testis Arl2 GAP). Worth stating because it distinguishes a summary assertion from a traceable one.
2. It is stated **at subfamily level** and names no species, no protein and no assay.
3. **The paper's own AGFG section supplies nothing** — no GAP activity for either paralogue, and only *"Much less information is available on AGFG2."* If a specific AGFG measurement existed, that is where it would be.
4. The **same paragraph** warns that *"some ArfGAPs use their GAP domain to bind Arf without promoting GTP hydrolysis"* — precisely the bind-versus-hydrolyse distinction a subfamily-level summary cannot settle.

Net effect on the verdict: none, as you said. But it does change the *reason*: the blanket clause raises the **prior** of catalytic competence, and that is now stated as part of why the row is `KEEP_AS_NON_CORE` rather than `MARK_AS_OVER_ANNOTATED` — previously that weight was carried entirely by the fly orthologue and the ARF proximity data.

Two further consequences:

- **The top-level `description` was the worse instance** and you were right to point at it. It said the reference *"reports no GAP measurement for either human AGFG protein"* — attributing a negative to a reference whose sentence says something closer to the opposite at subfamily level. It now says only that no direct GAP measurement on either human AGFG protein **has been located**, which is what the evidence supports.
- Added the two extra quotes (the ten-subfamily classification and the bind-without-hydrolysis caveat) to the row so both clauses and the qualifier are checkable from the YAML, and recorded the same nuance in `reference_review` for `PMID:18809720`, which is where a citation-quality judgement belongs.

## Item 1 — `zinc_site.json`: fixed in `c9a6c10b2`

Your diagnosis was exact. Fixed as a class rather than an instance:

- `zinc_site.py` asserts the payload before *and* after writing, and re-reads the file to confirm the round-trip;
- `audit_review.py` **check I** fails on any empty JSON artifact in the directory (and fails loudly if it finds zero artifacts to check);
- `audit_review.py` **check J** parses the rendered `RESULTS.md` zinc table back out and asserts the PDB id, offset and coordinating cysteines agree with `zinc_site.json` — which is the part your comment was really about, since a prose table derived from a script is a two-way dependency;
- both are break-tested **against the exact `{}` content that shipped**, not against a mutation I invented: the self-test writes `{}`, confirms I and J both fire, then confirms J alone fires on a populated-but-drifted offset while I stays silent, then restores and asserts the restore. `--self-test` is 10 break-tests.

Regenerated, not hand-written: `2olm` offset `+0`, `2d9l` offset `+6`, both `[29, 32, 49, 52]`, 4-atom shells, 1836 bytes.

## Item 3 — `GO:0035615` entailment: fixed in `c9a6c10b2`, with one premise corrected

The entailment is acknowledged in the row. But `GO:0140312` does **not** avoid it — its own definition reads *"Binding directly to the structural scaffolding elements of a vesicle coat (such as clathrin or COPII), and bridging the membrane, cargo receptor, and membrane deformation machinery"*, so the parent names the coat too **and** adds a cargo-receptor clause that is false here, since VAMP7 is the cargo rather than a receptor for it. Retreating would swap one commitment for two, one of them wrong. The row now says what discharges the commitment instead: the FCHO1 µHD binding, the CCP/CCV colocalisation, and the clathrin-knockdown phenocopy — not the motif argument, which is why `GO:0030276` stays withheld.

## Item 5 — `GO:0008270` reference: answered with a fact in `c9a6c10b2`

Neither structure has a primary publication. PDBe's `publications` endpoint returns, for both, `"To be published"` with `pubmed_id: null` and `doi: null` — 2OLM as *"ArfGap domain of HIV-1 Rev binding protein"*, 2D9L as *"Solution structure of the ArfGap domain of human RIP"* — and UniProt records 2D9L only as submitted to the PDB. So there is no PMID available to carry `original_reference_id`; the `file:` route is not just convention here, it is the only citable record. Noted in the row.

## Item 6 — `GO:0001669`: explicit decision made in `c9a6c10b2`, and the search found something better

Declined, on its definition: *"A structure in the head of a spermatozoon that contains acid hydrolases ... derived from the lysosome"* is the **mature** organelle, which in the null never forms, while AGFG1 is documented on the cytosolic surface of the **precursor** vesicles. Plus the evidence is mouse-only.

Looking for it surfaced **`GO:0120211 proacrosomal vesicle fusion`** — *"Fusion of the membrane of proacrosomal vesicle with the membrane of another proacrosomal vesicle to form the acrosome."* That is the mouse phenotype verbatim, and it is a verified descendant of the accepted `GO:0001675` over `is_a`/`part_of`, so it is additive rather than a sideways move. Added as a **NEW `ISS`** row with `UniProtKB:Q8K2K6` in `supporting_entities` (the measurement is on mouse Agfg1; ISS takes the sequence-similar entity). 6 NEW rows now.

## Gates on the current head

```
just validate human AGFG1        ✓ Valid  (on the tree merged with main)
checkquotes.py                   72 quotes, 0 problems
grep_file_quotes.py              17 file: quotes byte-exact, 0 problems
audit_review.py                  0 problems;  --self-test: 10 break-tests fire
terms.py                         6 positive + 2 negative ancestry controls
cache_lint                       exit 0
cache/go/terms.csv               unchanged this round
```

cmungall:
Self-caught immediately after pushing `3d520ec0`, before you had to point it out: the round-2 correction landed in the `reason` and the `description` but **not** in the `GO:0005096` row's `summary`, which still read *"the family's own reference reports no GAP measurement for either human AGFG protein"* — the exact attribution the other two surfaces had just been corrected to stop making, and in the position a curator reads first.

Fixed in `8ab101d8`, and closed as a class rather than an instance. `audit_review.py` gains **check K**: a set of retracted phrasings that must not appear on any curator-facing surface in the gene folder, scanned with the YAML dumper's line wrapping normalised away — because wrapping is how the phrase hid from a naive grep in the first place.

Its limitation is stated in the module docstring rather than implied: it matches **fixed phrases**, cannot catch a paraphrase, and exempts the patch scripts by design, since their job is to name the string they replace. A withdrawn claim still needs a human re-read; this only stops the exact wording returning.

Break-tested in the position the defect occupied and in two directions — an appended single-line instance and a line-wrapped one — each asserting the mutation applied, that the check fired, and that the restore succeeded. `--self-test` is now 12 break-tests. I also confirmed independently that the phrase **is** present in the YAML as committed at `3d520ec0`, so check K fires on the version that shipped rather than only on a mutation I invented.

No term, action, evidence code or quote changed in this commit. Current head is `8ab101d8`; gates: `just validate` ✓ Valid, checkquotes 72/0, grep_file_quotes 17/0, audit_review 0 problems / 12 break-tests, cache_lint exit 0, `terms.csv` unchanged.

cmungall:
## Retraction: `GO:0005096` reverses to `MARK_AS_OVER_ANNOTATED`

Not prompted by a review item — this came out of cross-checking against the concurrent **AGFG2** review (#2352), and it reverses this PR's headline verdict.

**I tested one of three required residues.** My anchor was `PMID:18809720`, which names the arginine finger and the four cysteines, and I treated "the apparatus" as exhausted by what that paper names. `PMID:23433073` names **three** catalytically required positions in ASAP3 — the ArfGAP with a solved Arf6 complex — and states that *"Mutation of any one of these three residues leads to severe loss in Arf GAP activity"*: R469 is the arginine finger, **D484 contacts the Arf6 catalytic glutamine Q67** and stabilises switch 2, **W451 sits in the Arf–ArfGAP interface**.

Measured independently on AGFG1's own sequence rather than inherited from the sibling (`catalytic_residues.py`), with three controls that must pass before any absence is reported:

| protein | W451 | R469 | D484 | n/3 |
|---|---|---|---|---|
| ASAP3 (ref), ARFGAP1, ARFGAP3, ASAP1, SMAP1 | W | R | D | **3/3** each |
| **AGFG1 human (subject)** | **Y39** | R57 | **T71** | **1/3** |
| AGFG2 human | Y57 | R75 | T89 | 1/3 |
| Agfg1 mouse | Y39 | R57 | T71 | 1/3 |
| drongo | Y40 | R58 | A72 | 1/3 |

Controls: (0) ASAP3's 451/469/484 are still W/R/D in the current sequence, so the paper's numbering has not drifted; (1) **every GAP-competent panel member recovers all three** — without this, an absence in AGFG1 would be an alignment artefact rather than a result; (2) the arginine found by alignment equals the one `CX2CX16CX2CX4R` gives, for 8 of 9 entries. **5/5 competent controls at 3/3, 0/4 AGFG members at 3/3.**

**The reversal resolves a tension rather than creating one.** The ARF-proximity result had looked like support for catalysis; the same paper is explicit that it is not — *"These predicted changes ... should not be confused with consequent changes in the ability to bind Arf family GTPases"* — and proposes such proteins act as Arf **effectors**. Binding retained + catalysis lost is one coherent picture. The fly genetics is reinterpreted, not discarded: drongo scores 1/3 too, and an effector antagonising an Arf-GEF without hydrolysing GTP fits it as well as a GAP does.

`MARK_AS_OVER_ANNOTATED`, not `REMOVE`: the domain is genuine, the zinc really is bound, Arf binding is real, and no assay has been run on the human protein in either direction. `root_cause: PROPAGATION_BAD`, `failure_modes: [PSEUDO_OR_SUBACTIVITY_LOSS, FUNCTIONAL_DIVERGENCE]`, and the `IPR001164` source is now `SUPPORTS_SOURCE_BUT_NOT_TARGET` — it is the pan-ArfGAP catalytic-domain entry spanning 60,678 proteins, while the AGFG-specific `IPR052248` maps to nothing, so the claim arrives from the one signature that cannot see a subfamily-level loss.

**How it nearly stayed wrong, and why this matters for the campaign.** I had *fetched* `PMID:23433073` early and then **deleted** it before the first commit, judging it a family-classification paper with nothing resting on it. It contains the analysis that decides the gene's only molecular function. It came back only because the AGFG2 agent read it — which is the strongest argument for reviewing paralogues in parallel that this pair produced. The panel reproduces AGFG2's Thr89 exactly, so **two independently-produced reviews of this identical GOA row now agree**; the divergence that existed an hour ago (`KEEP_AS_NON_CORE` here vs `MARK_AS_OVER_ANNOTATED` there) would itself have been a defect.

The retraction is documented rather than overwritten: `RESULTS.md` §1 carries a banner above the superseded table, the notes strike the false sentence through in place with an account of how the error happened, `audit_review.py` check K forbids both retracted phrasings from reappearing, and the PR title and body are updated. `PMID:23433073` is registered in `references` with a `reference_review` saying all of this.

Two incidental bugs fixed while doing it: `patch_gap_reversal.py` wrote the YAML *before* parsing it on its first run, leaving broken YAML on disk that contained its own done-marker (so a re-run would have said "already patched") — it now validates then writes; and the reference-registration script had to anchor on the directory containing **both** `genes/` and `publications/`, because a walk looking for `publications/` alone stops at the stray `genes/human/publications` that exists on main.

Gates on `e03ee501`: `just validate` ✓ Valid, checkquotes 78/0, grep_file_quotes 18/0, audit_review 0 problems / 12 break-tests, cache_lint exit 0, `terms.csv` unchanged. Actions now 7 ACCEPT / 20 KEEP_AS_NON_CORE / 3 MARK_AS_OVER_ANNOTATED / 2 MODIFY / 2 REMOVE / 1 UNDECIDED / 6 NEW.

cmungall:
Addressed the latest requested changes in `556eb6908`:

- Reconciled the description, GO:0005096 summary/reason, GO_REF:0000002 review, and PMID:38606629 review with both clauses of PMID:18809720.
- Reframed the Arf-effector interpretation as a residue-based prediction pending direct catalysis, rather than a settled absence claim.
- Updated the proposed assay to test Tyr39Trp/Thr71Asp restoration mutants as well as Arg57Ala.
- Moved phrase-regression check K into the main audit path and added guards for both regressed phrasings.

Checks: `just validate human AGFG1`; `just render human AGFG1`; `audit_review.py` (0 problems); `audit_review.py --self-test` (12/12 break-tests).

PR #2442, opened 2026-08-09
Make assay paper row order deterministic

cmungall:
## Summary

- emit paper readout rows in catalog order instead of Python set iteration order
- retain the first primary/supporting paper that establishes each readout class
- add an end-to-end wrapper regression that compares output across two Python hash seeds and verifies global catalog order

## Validation

- `PYTEST_ADDOPTS='-k supporting_paper_rows_follow_catalog_order_across_hash_seeds' just pytest` (1 passed)
- `just format`
- `just test` (1,409 pytest tests; 132 doctests; mypy and Ruff clean)
- `git diff --check`

Generated assay reports are intentionally excluded; they will be regenerated through the public wrappers in the separate baseline refresh after generator fixes land.

---

> [!NOTE]
> **Low Risk**
> Deterministic ordering and test-only harness changes; no change to readout detection or which paper is credited per class.
> 
> **Overview**
> **Paper readout mining** (`mine_papers.py`) now writes `paper_readout_matches.tsv` rows in **readout catalog order** instead of dict/set iteration order, so outputs are stable across Python hash seeds and easier to diff.
> 
> When unioning readouts across primary and supporting PMIDs, the miner still records the **first** paper (primary preferred) that establishes each class; only the **emit** loop walks `catalog.items()` and skips classes with no source. Provenance behavior is unchanged.
> 
> Adds an end-to-end **`assay-mine-supporting`** test that runs with two `PYTHONHASHSEED` values, asserts byte-identical TSVs, and checks the expected `(readout_class, pmid, ref_role)` sequence. `run_just` accepts optional `extra_env` for that harness.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88f4fea3bd217acbec55245aa0d013a8575989b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/projects/ASSAY_TO_FUNCTION/mine_papers.py
+++ b/projects/ASSAY_TO_FUNCTION/mine_papers.py
@@ -438,9 +438,9 @@ def main() -> None:
             any_resolved = True
             cls, ft = det
             full_text = full_text or ft
-            for c in cls:
-                if c not in class_src:
-                    class_src[c] = (pm, role)
+            for name in catalog:
+                if name in cls and name not in class_src:
+                    class_src[name] = (pm, role)
         if not any_resolved:
             continue
         n_resolved += 1
@@ -452,8 +452,11 @@ def main() -> None:
         go_label = term.get("label", "") or ""
         aspect = aspect_map.get(go_id, "?")
 
-        for name, (pmid, role) in class_src.items():
-            spec = catalog[name]
+        for name, spec in catalog.items():
+            source = class_src.get(name)
+            if source is None:
+                continue
+            pmid, role = source
             aligned = bool(
                 go_id in spec["_overmapped"]
                 or (spec["_aligned"] and spec["_aligned"].search(go_label))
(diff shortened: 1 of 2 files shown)


cmungall:
Thanks for the careful review. I verified the focused wrapper regression locally under the project Python 3.12.9 and the full public test wrapper is green (1,409 pytest tests, 132 doctests, mypy, and Ruff).

I’m keeping both catalog-order loops in this focused change. The 61-entry catalog is intentionally the ordering authority throughout this report, and making both collection and emission state that invariant directly is easier to audit than retaining unordered insertion internally. The bounded lookup cost is small relative to publication scanning and regex work. I’m also keeping the two-seed run: seeds 1 and 2 reverse the pre-fix set order under the project interpreter, while the explicit expected-order assertion independently guards against primary-paper grouping and lexical ordering.

The separate unsorted `build_aspect_map()` first-wins behavior is a genuine adjacent determinism issue; under the project’s separate-bug policy I will address it in its own PR rather than broaden this one. The generated-report refresh is already staged as a separate campaign branch and will be rerun through all four public wrappers after the remaining generator fixes merge.

PR #2443, opened 2026-08-09
Preserve assay wrapper argument boundaries

cmungall:
## Summary

- convert the remaining three ASSAY_TO_FUNCTION passthrough recipes to Just positional-argument mode
- forward exact argument boundaries through Bash `"$@"`, including a genuinely empty variadic tail
- test all three public recipes through real `just` invocations with a temp fake `uv`, covering spaces, punctuation, a literal separator, and zero arguments

## Validation

- `PYTEST_ADDOPTS='-k assay_wrappers_preserve_exact_argument_boundaries' just pytest` (3 passed)
- `just format`
- `just test` after rebasing onto current main (1,412 pytest tests; 132 doctests; mypy and Ruff clean)
- independent host Bash 3.2.57 wrapper check (3 passed)
- `git diff --check`

The fake executable prevents the flagger, consolidator, staging code, report writers, network, and provider jobs from running during the focused test.

---

> [!NOTE]
> **Low Risk**
> Build-tool and test-only changes; no application logic or data paths are modified beyond how optional flags reach existing Python entrypoints.
> 
> **Overview**
> Aligns **`assay-flag`**, **`assay-consolidate`**, and **`assay-stage-hypotheses`** with the other ASSAY passthrough recipes: **`[positional-arguments]`**, a small Bash wrapper with **`set -euo pipefail`**, and **`"$@"`** instead of Just’s **`{{args}}`** interpolation.
> 
> That change keeps each CLI token intact (paths with spaces/punctuation, a literal **`--`**, and an empty variadic tail) when arguments are forwarded to **`uv run`**.
> 
> Adds **`tests/test_assay_wrapper_forwarding.py`**, which runs real **`just`** invocations with a fake **`uv`** on **`PATH`** and asserts the logged argv matches the expected fixed prefix plus forwarded args (or prefix only when no extra args are passed).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81c72af8a82119b529bf092d4e964c02bc05dfe7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -626,12 +626,18 @@ assay-mine-supporting *args:
         --out-dir projects/ASSAY_TO_FUNCTION/reports/with_supporting \
         "$@"
 
-assay-flag *args="":
-    uv run python projects/ASSAY_TO_FUNCTION/flag_candidates.py {{args}}
+[positional-arguments]
+assay-flag *args:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    uv run python projects/ASSAY_TO_FUNCTION/flag_candidates.py "$@"
 
 # Consolidate the 60-class catalog + mined matches into summary TSV/MD + figure.
-assay-consolidate *args="":
-    uv run --with matplotlib python projects/ASSAY_TO_FUNCTION/consolidate.py {{args}}
+[positional-arguments]
+assay-consolidate *args:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    uv run --with matplotlib python projects/ASSAY_TO_FUNCTION/consolidate.py "$@"
 
 # Stage (human-gated) OpenScientist hypothesis prompts from flagged_candidates.tsv.
 # This writes committed prompt.md files and prints submit commands but NEVER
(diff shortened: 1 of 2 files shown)


cmungall:
Thanks. The exact wrapper test and full suite both pass locally on the rebased head; an independent host run also passed all three cases under Bash 3.2.57.

I’m keeping this test module self-contained: its fake-`uv` argv-capture contract is distinct from the real-miner fixture helper, and the paired argument/zero-tail loop keeps each recipe setup together. The two already-safe mining recipes retain stronger end-to-end coverage in `test_assay_mining_wrapper.py`, including spaced paths and output routing.

The remaining legacy `{{args}}` recipes are a genuine adjacent wrapper-family issue. Under the separate-bug policy I’m auditing that full surface for one bounded follow-up rather than expanding this assay-only PR. The consolidator’s lack of CLI parsing is likewise pre-existing and separate; this change fixes the wrapper/process boundary without claiming those forwarded options are currently consumed.

PR #2447, opened 2026-08-09
Forward init-rule-review cache options

cmungall:
## Summary

- expose rule-review initialization through the project Typer CLI
- replace the hard-coded inline wrapper with exact positional forwarding to that CLI
- make the documented `--cache-dir` option work while preserving the no-overwrite default
- add a real `just` → CLI → initializer regression using a pre-seeded temporary cache path with spaces and punctuation

## Validation

- pre-fix focused regression: custom-cache case failed against hard-coded `rules/arba`
- `PYTEST_ADDOPTS='-k init_rule_review_wrapper' just pytest` (2 passed after fix)
- `just init-rule-review ARBA00026249 --help`
- `just format`
- `just test` after review hardening (1,414 pytest tests; 132 doctests; mypy and Ruff clean)
- independent Bash 3.2/layer review: no implementation blockers
- `git diff --check`

The end-to-end test pre-seeds enriched JSON, so no rule fetch or network call occurs and its write is confined to `tmp_path`. A separate temp fake-`uv` case asserts the exact zero-tail argv without touching any repository rule file or provider.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 1a56586ccaa59e3341621409ad7e81c98b646149. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -2274,8 +2274,13 @@ clean-imodulondb-cache:
 #   just init-rule-review UR000000070 --cache-dir rules/unirule
 # If you get "Review file already exists" error:
 #   rm rules/arba/ARBA00026249/ARBA00026249-review.yaml  # Then re-run init-rule-review
-init-rule-review rule_id *args="":
-    uv run python -c "from ai_gene_review.etl.rule_review_init import init_rule_review; from pathlib import Path; init_rule_review('{{rule_id}}', cache_dir=Path('rules/arba'))"
+[positional-arguments]
+init-rule-review rule_id *args:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    rule_id="$1"
+    shift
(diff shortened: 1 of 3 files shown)


cmungall:
Addressed both requested test blockers in exact head `1a56586cc`:

- replaced the mutable tracked-review empty-tail case with a temp fake-`uv` shim that asserts the exact argv vector and proves there is no trailing empty argument; no repository rule file or network path can be reached
- raised the real end-to-end custom-cache subprocess timeout from 30s to 120s for cold environments

The custom-cache case still exercises the complete `just` → Typer CLI → initializer path against pre-seeded temporary enriched JSON. Focused result is 2 passed; fresh full `just test` is green with 1,414 pytest tests, 132 doctests, mypy, and Ruff. `git diff --check` is clean. Please rerun review on the new head.

PR #2485, opened 2026-08-11
Curate PSEPK Leloir pathway satisfiability

cmungall:
## Summary

- Record that the canonical Leloir pathway is not satisfiable in KT2440 because GalK and GalT are absent.
- Retain GalE reaction chemistry while removing the over-propagated full-pathway annotation.
- Generalize the reusable four-reaction module and remove human disease and blanket cytosol framing.

## Validation

- `just validate PSEPK galE`
- ModuleReview LinkML validation
- Module, gene, and project rendering

OpenScientist reports are still running and will be added before marking this PR ready.


cmungall:
Addressed the remaining review findings after rebasing onto current main:

- The GO:0033499 `REMOVE` now cites the exact OpenScientist sentence establishing the missing GalK/GalT steps.
- The core-function NAD-cofactor clause now cites its exact report text.
- The module uses the current beta-D-galactose wording as `preferred_term`, while retaining the repository ontology's validated GO:0033499 `term.label` (the semantic validator rejects the newer GOA label in authored module terms).
- Removed the module's lone `cytosolic` wording rather than reintroducing redundant location claims.

`just validate PSEPK galE`, ModuleReview schema/semantic validation, and all focused renders pass.

/review

PR #2503, opened 2026-08-11
Curate PSEPK beta-ketoadipate lower pathway

cmungall:
## Summary

- model the shared two-reaction lower beta-ketoadipate pathway after catechol/protocatechuate convergence
- represent PcaI and PcaJ as contributors to the heteromeric CoA-transferase activity
- add the missing branch-neutral GO:0042952 annotation proposal for PcaF-I
- correct net stoichiometry: succinyl-CoA is recycled; net products are acetyl-CoA and succinate
- replace DOI paraphrases with cached, verbatim PMID evidence and add curator notes

## Review fixes

- resolved all three prior requested changes
- applied independent annotation-reviewer findings: qualifier corrections use MODIFY with same-term replacement; module stoichiometry and branch-neutral PcaF prose corrected
- retained proposed_new_terms as empty because GO:0042952 already exists; the missing annotation is represented with action NEW

## Validation

- all pcaI, pcaJ, and pcaF gene validations pass
- ModuleReview schema and semantic validation pass (expected NCBITaxon/NCBIfam lookup warnings only)
- 279 focused tests passed before independent review; 216 focused tests passed after final changes
- all affected gene, module, and project pages rendered
- rebased onto origin/main at 323ab34c39


cmungall:
Addressed the prior review and an independent annotation-reviewer pass in 68a512eea6.

Notable correction: GO:0042952 is an existing GO term, so it is represented as a NEW annotation for pcaF rather than in proposed_new_terms, which is reserved for ontology terms that do not exist. PcaI/PcaJ qualifier-only corrections now use MODIFY with same-term replacements. The module net stoichiometry now explicitly recycles succinyl-CoA.

cmungall:
Addressed the sole blocking round-2 finding.

The GO:0006629 REMOVE rationale now explicitly records the contrary result: PcaF can accommodate long-chain acyl-CoA ligands in vitro, but those structures used active-site mutants and crystallographic ligand soaking, so they do not establish a physiological lipid-metabolism role. The PMID:32647822 reference now includes the verbatim long-chain-promiscuity finding rather than omitting it.

I also strengthened the positive GO:0033812 support with the same paper's direct statement that beta-ketoadipyl-CoA is PcaF's natural substrate and documented both points in the notes. `just validate PSEPK pcaF`, rendering, `git diff --check`, and 37 focused tests pass.

PR #2575, opened 2026-08-13
Curate bacterial glucarate and galactarate catabolism

cmungall:
## Summary
- curate the reusable three-stage glucarate/galactarate catabolism module with alternative entry branches
- review six PSEPK core or candidate proteins and distinguish unresolved terminal dehydrogenase paralogs
- add OpenScientist gene, module, and module plus pathway plus taxon research provenance
- update the ppu00053 batch report, TSV, and rendered pages

## Validation
- all six gene reviews pass just validate
- LinkML ModuleReview and semantic module validators pass
- module and project rendering pass
- git diff --check passes

The module validator reports one non-blocking warning because the InterPro prefix is not configured for live label validation.


cmungall:
@ai4c-reviewer please re-review commit a32557c056. The terminal step now uses current cofactor-neutral GO:0004030 plus a formal substrate-specific NAD(P)+ term proposal in all three KGSADH reviews; GO:0016491 and GO:0016620 are harmonized as non-core and GO:0047533 remains undecided only for its NADP specificity. PP_3602 regains both aldarate BPs with the operon/paralog caveat. The module leaf now contains only Q88GW5 plus cross-species KgsD Q6FFQ0; Q88NF5/Q88JR4 are comparison paralogs outside the leaf. PMIDs 17202142 and 16835232 were fetched and verified as Azospirillum brasilense enzymology, PMID 22833679 grounds the Pseudomonas hydroxyproline-locus context, PP_2585 cites its gene report, and its TSV research status is fixed. All three strict gene validations, both module validators, renders, and diff check pass.

PR #2616, opened 2026-08-18
Make the PANTHER member index comprehensive; unchecked membership is now an error

cmungall:
## The principle

Validate everything. If `panther-members.tsv` is not comprehensive, that is a problem with the index — not something to route around.

## Two defects

**1. The check that catches guessed IDs degraded to a warning when it couldn't run.**

Member-consistency is the only check that can distinguish a *wrong label on the right family* from a *plausible label on a guessed family*. When an accession was missing from the index it emitted "family membership was not checked" and continued.

That is exactly backwards: new curation cites new proteins, so the unindexed case **is** the under-review case. Measured across all 86 open PRs:

- **217 of 278 grounding checks (78%) never ran**, across 51 of 75 PRs
- the sweep nonetheless reported **zero grounding failures**

A clean bill of health from a check that was dark. It is now an `❌ ERROR`. `just refresh-panther-members` scans the working tree's `modules/`, so on a PR branch the new module file is already there — the fix is always available to whoever hits the error.

**2. The index was pruned to cited accessions by construction.**

The old docstring: *"Only `accessions` actually cited in the repository are retained, keeping the committed artifact small (a few thousand rows rather than the ~1.5M proteins PANTHER classifies)."*

That made the artifact lag the repository by design. Unpruned it is **297,886 accessions / 6.5 MB**, beside a 14 MB `panther.obo` already committed.

## What this does not fix

PANTHER publishes **no *P. putida* sequence classification** — its `pseudomonas` file is *P. aeruginosa* (zero Q88 rows) — and *P. putida* is the most heavily curated organism in this repo. Of 539 accessions cited by the open PRs, the organism files cover 103. The rest are resolvable only via the UniProt xref fallback, which is precisely why that result must be retained permanently rather than pruned.

## Verification

- 281-module sweep: **0 errors, 0 "could not be checked"** — every membership check on main now runs
- Index diff is **purely additive**: 296,429 added, 0 removed, 0 accessions lost, 0 silently re-pointed to a different family
- Both tests pinning the old behaviour are **inverted, not deleted**, each naming the failure it now prevents
- Full suite green (1,769 unit + 154 doctest, mypy, ruff)

## Expected effect on the open backlog

22 PRs carry PANTHER label mismatches that are already blocking errors on current main (82 total: 19 case-only, 49 same-protein paraphrases, 14 naming a different protein). Notably, **all 14 of the alarming ones are confirmed correct by the member check** — they are the `PTHR24322`/DHRS3 pattern, correct IDs with family names dominated by another member. No hallucinated PANTHER IDs were found in the backlog.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change to CLAUDE.md; no runtime, validation, or data-handling code in this diff.
> 
> **Overview**
> Updates curator guidance so a representative accession **missing from** `panther-members.tsv` is a **blocking error**, not a warning. That is the case where the family-membership check cannot run, which is typical of new curation.
> 
> The remedy is `just refresh-panther-members`, which scans the working tree’s `modules/` (including the file just added). Proteins with no PANTHER family in either source stay in the index under `# unresolved:` and do not fail the build.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3d3b3ffcf620ee625a984f0fc485325ee5c53a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmungall:
@claude please re-review — the previous run errored on a session usage limit (`num_turns=1 cost=$0`), so `f3d3b3ffcf` was never reviewed. Since then main has been merged in (114 commits); the member index already covered main's new content, so it is unchanged.

PR #2682, opened 2026-08-28
Complete yeast SSA2 annotation review

cmungall:
## Summary
- complete the SSA2 review with 57 logical review records covering all 278 pinned GOA provenance rows
- refine generic/obsolete chaperone assertions to GO:0140662 and keep specialized targeting, localization, degradation, and tRNA-import roles non-core
- remove the stale GO:0005886 IBA because current PTN002500132 lacks that PAINT assertion, while retaining the independent HDA localization conservatively
- adjudicate the miscited Candida paper and disputed OpenScientist database/species claims
- update the UNFOLDED_PROTEIN_BINDING project row and regenerate review, project, and browser artifacts through public `just` wrappers

## Verification
- `just seed-goa yeast SSA2`
- `just validate-goa yeast SSA2`
- `just validate yeast SSA2`
- `just refresh-bioreason-benchmark-sidecars`
- `just render yeast SSA2`
- `just render-project UNFOLDED_PROTEIN_BINDING`
- `just deploy-browser`
- `just test` (1,783 unit tests; 154 doctests; mypy and Ruff clean)

PR #2683, opened 2026-08-28
Validate supporting text on reference findings

cmungall:
## Summary
- validate `references[].findings[].supporting_text` against the parent PMID/DOI publication cache
- use the same deterministic substring validator as other literature evidence
- report mismatches as blocking reference-validation errors and unavailable records as warnings
- add regression coverage for exact and fabricated inherited-reference quotes

This closes the validation gap found while reviewing SSA2 in #2682; that gene PR is paused until this support fix merges.

## Verification
- `just test`: 1,785 unit tests passed; 154 doctests passed; mypy and Ruff clean
- `just validate yeast SSA2`: current-main smoke test remains green

### DIFF
--- a/src/ai_gene_review/validation/module_validator.py
+++ b/src/ai_gene_review/validation/module_validator.py
@@ -91,6 +91,11 @@ from ai_gene_review.etl.panther_families import (
     load_member_index,
     load_subfamily_counts,
 )
+from ai_gene_review.validation.supporting_text import (
+    LITERATURE_PREFIXES,
+    build_supporting_text_validator,
+    is_unfetchable,
+)
 
 # A resolver answers: given a CURIE, return (status, primary_label, aliases).
 # status is "ok" when the id resolved, "not_found" when the ontology was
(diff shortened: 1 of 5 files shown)


cmungall:
Addressed all requested changes in a5690d443d:

- extracted shared literature prefixes, validator construction, and fetch-failure classification into `validation/supporting_text.py`; gene and module validation now use the same helpers
- honors `full_text_unavailable` at both reference and finding level
- reads cached `full_text_available`; mismatches against abstract-only caches are advisory warnings, while mismatches against full-text caches remain blocking errors
- replaced repository-coupled tests with hermetic temporary publication caches that prove the public validation path executes, plus coverage for exact text, fabricated text, abstract-only cache behavior, and both escape-hatch locations
- updated the stale batch-validator docstring

Verification:
- `just test`: 1,787 unit tests passed; 154 doctests passed; mypy and Ruff clean
- `just validate-all`: all 4,459 files valid, 0 best-practice errors, term validation had no blocking errors, and all 54 pathway files with PMID references passed
- the best-practices corpus phase completed in 3m51s; the shared validator is cached rather than rebuilt per file

The `validate-all` run refreshed three ontology cache files locally; those generated changes were deliberately excluded from this PR.

cmungall:
Addressed the remaining cache-format edge case in 70da999c55:

- resolves both `PMID_<id>.md` and slugged `DOI_<doi-with-slash-replaced>.md` cache paths
- uses explicit `full_text_available` when present, otherwise recognizes legacy `content_type`
- maps `abstract_only` and `unavailable` to no full text, and `full_text_html` / `full_text_pdf` to full text
- threads an optional `publications_dir` through the public `validate_gene_review` API, so tests no longer monkeypatch project-root discovery
- corrected the abstract-only remediation text so it does not imply a flag is already present
- added hermetic coverage for legacy PMID metadata, DOI filename resolution, unavailable/full-text DOI states, and public DOI warning behavior

Fresh verification:
- `just test`: 1,791 unit tests passed; 154 doctests passed; mypy and Ruff clean
- `just validate-all`: all 4,459 reviews valid with 0 errors; no blocking term errors; all 54 pathway files with PMID references passed

Generated ontology-cache refreshes were again excluded from the commit.

PR #2693, opened 2026-08-28
Re-review DANRE (zebrafish) gene annotations: clean descriptions, add IBA propagation metadata

cmungall:
## Summary

- Re-reviewed all 58 existing *Danio rerio* (DANRE) gene reviews. All files validate with **zero errors**; best-practice warnings dropped from **33 to 1** (the remaining `dph2` case is an intentional, correct qualifier-based ACCEPT/OVER_ANNOTATED split on GO:0090560, not a defect).
- Rewrote 23 top-level `description` fields to remove curation/workflow commentary ("core function is", "retained as non-core", "treated as over-annotated", "UNDECIDED pending", "the core review") so they read as project-independent biological summaries.
- Added structured `propagation_review` metadata (root_cause, failure_modes, source_entities) to 28 IBA annotations across 18 genes that carried a non-ACCEPT action; every source-entity ID was copied verbatim from each gene's GOA WITH/FROM column and verified against the GOA files (no fabricated IDs). Added 4 grounding `action: NEW` annotations (bdh2 ×2 — cellular detoxification, oxidoreductase activity for the 2,5-DHBA siderophore step; dnajc6/A0A8M9QG43 ×2 — Hsp70 protein binding, clathrin coat disassembly) with verbatim supporting quotes, and confirmed the UNDECIDED calls (asah2, naa50, opa1, tdp2, trpm7) remain correct (no full text available).

Closes #

## Validation

`ai-gene-review validate genes/DANRE/*/*-ai-review.yaml --terms` — 58 files, 0 errors, 1 best-practice warning:

```
Total errors: 0
Total warnings: 1
dph2: Inconsistent review actions for term GO:0090560 (2-(3-amino-3-carboxypropyl)histidine synthase activity): ACCEPT (ISS); MARK_AS_OVER_ANNOTATED (IEA)
```

The single remaining warning is intentional: the IEA copy implies the `enables` qualifier (over-annotation, since Dph2 only contributes to the Dph1–Dph2 complex activity) while the ISS copy correctly uses `contributes_to`; the two actions are deliberately different and biologically correct.

## Test plan

- [x] `ai-gene-review validate genes/DANRE/*/*-ai-review.yaml --terms` passes (0 errors)
- [x] All `propagation_review` source IDs verified present in the corresponding GOA WITH/FROM columns
- [x] All new `supporting_text` quotes are verbatim substrings of their cited references
- [x] No derived files committed (only `*-ai-review.yaml` and the append-only `cache/go/terms.csv`)

### DIFF
--- a/cache/go/terms.csv
+++ b/cache/go/terms.csv
@@ -682,6 +682,7 @@ GO:0003963,RNA-3'-phosphate cyclase activity,2026-06-20T02:05:20.117661
 GO:0003964,RNA-directed DNA polymerase activity,2026-05-08T22:12:02.939352
 GO:0003968,RNA-directed RNA polymerase activity,2026-05-08T22:12:02.939356
 GO:0003973,(S)-2-hydroxy-acid oxidase activity,2026-05-08T22:12:02.939359
+GO:0003976,UDP-N-acetylglucosamine-lysosomal-enzyme N-acetylglucosaminephosphotransferase activity,2026-08-28T06:31:45.593940
 GO:0003978,UDP-glucose 4-epimerase activity,2026-07-24T14:04:53.252862
 GO:0003980,UDP-glucose:glycoprotein glucosyltransferase activity,2026-05-08T22:12:02.939363
 GO:0003983,UTP:glucose-1-phosphate uridylyltransferase activity,2026-07-24T14:04:53.762193
@@ -2975,6 +2976,7 @@ GO:0016247,channel regulator activity,2026-05-08T22:12:02.944303
 GO:0016248,channel inhibitor activity,2026-05-08T22:12:02.944306
 GO:0016251,RNA polymerase II general transcription initiation factor activity,2026-05-08T22:12:02.944309
 GO:0016255,attachment of GPI anchor to protein,2026-05-08T22:12:02.944322
+GO:0016256,N-glycan processing to lysosome,2026-08-28T06:31:45.680725
 GO:0016263,"N-acetylgalactosaminide beta-1,3-galactosyltransferase activity",2026-06-21T22:19:38.797275
 GO:0016266,protein O-linked glycosylation via N-acetylgalactosamine,2026-08-08T14:11:48.206203
 GO:0016272,prefoldin complex,2026-05-08T22:12:02.944326
@@ -4664,6 +4666,7 @@ GO:0035234,ectopic germ cell programmed cell death,2026-05-08T22:12:02.949018
 GO:0035235,ionotropic glutamate receptor signaling pathway,2026-05-08T22:12:02.949052
 GO:0035239,tube morphogenesis,2026-03-22T22:38:14.237952
 GO:0035240,dopamine binding,2026-05-08T22:12:02.949056
+GO:0035242,protein-arginine omega-N asymmetric methyltransferase activity,2026-08-28T06:35:12.079832
 GO:0035251,UDP-glucosyltransferase activity,2026-03-22T22:38:14.237952
 GO:0035253,ciliary rootlet,2026-05-08T22:12:02.949060
 GO:0035254,glutamate receptor binding,2026-05-08T22:12:02.949063
@@ -4835,6 +4838,7 @@ GO:0036120,cellular response to platelet-derived growth factor stimulus,2026-05-
 GO:0036122,BMP binding,2026-05-08T22:12:02.949531
 GO:0036126,sperm flagellum,2026-05-08T22:12:02.949534
 GO:0036128,CatSper complex,2026-05-08T22:12:02.949537
+GO:0036140,[protein]-asparagine 3-dioxygenase activity,2026-08-28T06:32:49.393687
 GO:0036148,phosphatidylglycerol acyl-chain remodeling,2026-05-08T22:12:02.949540
 GO:0036150,phosphatidylserine acyl-chain remodeling,2026-05-08T22:12:02.949544
 GO:0036151,phosphatidylcholine acyl-chain remodeling,2026-05-08T22:12:02.949547
@@ -6467,6 +6471,7 @@ GO:0050958,magnetoreception,2026-05-08T22:12:02.953644
 GO:0050962,detection of light stimulus involved in sensory perception,2026-03-22T22:38:14.237952
 GO:0050966,detection of mechanical stimulus involved in sensory perception of pain,2026-05-08T22:12:02.953648
 GO:0050973,detection of mechanical stimulus involved in equilibrioception,2026-03-22T22:38:14.237952
+GO:0050974,detection of mechanical stimulus involved in sensory perception,2026-08-28T06:36:43.440193
 GO:0050975,sensory perception of touch,2026-03-22T22:38:14.237952
 GO:0050976,detection of mechanical stimulus involved in sensory perception of touch,2026-05-08T22:12:02.953651
 GO:0050980,detection of light stimulus involved in magnetoreception,2026-05-08T22:12:02.953655
@@ -6848,6 +6853,7 @@ GO:0060081,membrane hyperpolarization,2026-05-08T22:12:02.954540
 GO:0060090,molecular adaptor activity,2026-05-08T22:12:02.954544
 GO:0060100,"positive regulation of phagocytosis, engulfment",2026-06-19T13:50:48.902615
 GO:0060102,cuticular extracellular matrix,2026-04-19T19:54:57.052347
+GO:0060122,inner ear receptor cell stereocilium organization,2026-08-28T06:36:43.380164
 GO:0060134,prepulse inhibition,2026-05-08T22:12:02.954547
 GO:0060137,maternal process involved in parturition,2026-05-08T22:12:02.954550
 GO:0060152,microtubule-based peroxisome localization,2026-05-08T22:12:02.954553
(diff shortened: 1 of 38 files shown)


cmungall:
Thanks for the careful audit — the four requested changes are addressed in `a9982a7`, plus suggestion 8 in `3b19997`. All touched files re-validate with 0 errors / 0 warnings.

**Blocking items (1–4):**

1. **bdh2 root oxidoreductase** — Replaced the NEW `GO:0016491` with the specific **`GO:0016616`** (*oxidoreductase activity, acting on the CH-OH group of donors, NAD or NADP as acceptor*) — the parent of the already-accepted `GO:0016617` 4-oxoproline reductase, justified by the SDR fold + NAD dependence. Updated `core_functions[1].molecular_function` to match, so the grounding is informative rather than a root-level ancestor. (Grounded it as an existing-annotation row rather than an NTR so `core_functions` stays reflected and CI stays clean; happy to switch to `proposed_new_terms` if you'd prefer the NTR route.)

2. **tdp2 / gtpbp3 swapped `failure_modes`** — Corrected against the `part_of` graph: `tdp2` (cytoplasm → `GO:0005634` nucleus, which is *not* `part_of` cytoplasm) is now `COMPARTMENT_OR_COMPLEX_MISMATCH`; `gtpbp3` (mitochondrion **is** `part_of` cytoplasm) is now `TERM_SCOPING_PROBLEM` / `GRANULARITY_MISMATCH` with `SUPPORTS_TRANSFER`. Rewrote both source-entity comments so they no longer read as self-contradictory.

3. **coq8a description** — Removed the erroneous "methyl" (ADCK3 is a UbiB-family kinase/ATPase, no methyltransferase activity); now "physiological substrate specificity is unclear". Good catch — it did drift in from the `tomt` batch.

4. **Boilerplate `reason`** — Lifted the specific rationale into `review.reason` for `asah2` GO:0042759 (reaction-direction / ceramide-catabolism argument), `mfsd2aa` GO:0015245 & GO:0015908 (LPC-esterified vs free fatty acid), and `gtpbp3` GO:0005737.

**Suggestions:**

- **8 (adopted)** — Added NEW `GO:0001671` ATPase activator activity for auxilin's HPD-motif stimulation of HSC70, alongside `GO:0030544`.
- **5 (crystallin self-replacing MODIFY)** — Fair point; this is pre-existing (the PR only attached `propagation_review` to those rows) and the `proposed_replacement` == same term is a deliberate "retain until a holdase NTR exists" placeholder. Left out of this cleanup PR's scope — can convert the three rows to `KEEP_AS_NON_CORE` + a holdase `proposed_new_terms` in a follow-up if you'd like.
- **6 (flvcr2a REMOVE vs MODIFY)** — Left as `REMOVE`: unlike the generic-membrane rows elsewhere, `flvcr2a` already carries a separately-annotated `GO:0005886` plasma membrane (ACCEPT) and `GO:0005789` ER membrane, so the generic `GO:0016020` is redundant rather than needing a replacement. The row's `reason` documents that intent.
- **7 (tomt self-donor label)** — Noted; cosmetic, left as-is.

On the validation note: CI's `test (3.12)` is green on the pre-fix head, and I re-ran `ai-gene-review validate … --terms` locally on every changed file after these edits (0 errors, 1 pre-existing intentional warning on `dph2`'s qualifier-based ACCEPT/OVER_ANNOTATED split). New commits should trigger your re-review.

---

cmungall:
Thanks for the approval. Applied the polish notes in `2867c42`:

- **a — bdh2 NTR** — Agreed it should be *both*. Added a `proposed_new_terms` entry requesting a dedicated child of `GO:0016616` for the NAD-dependent 2,5-DHBA-forming reaction, so the ontology gap is now recorded machine-readably while `GO:0016616` stays as the coherent grounding row.
- **b + c — tdp2 framing** — Fixed together: the row-level `summary`/`reason` no longer use the "less specific than" granularity wording, and `root_cause` is now `PROPAGATION_BAD` with both donor `source_status` values `SUPPORTS_SOURCE_BUT_NOT_TARGET` — consistent with `COMPARTMENT_OR_COMPLEX_MISMATCH` and the mirror of the `gtpbp3` change.
- **d — pre-existing boilerplate in mfsd2aa (232/268/402) and cyp26a1:520** — Left as-is for now, since those rows' `summary` fields already carry the reasoning and they're outside this PR's touched set; happy to finish the sweep in a follow-up.

Both changed files validate at 0 errors / 0 warnings. CI should re-run on the new head — I'll confirm `Build and test` goes green before this is merged.

---

cmungall:
Thanks — glad the tdp2 fix propagated cleanly through all five fields. On the two new notes, both non-blocking, I'm going to stop here rather than spin another round:

- **i (tdp2 MODIFY vs REMOVE)** — Fair consistency observation. Leaving it as `MODIFY`: the action is pre-existing (this PR only attached `propagation_review` to that row), and unlike the bare `gtpbp3`/`flvcr2a` `REMOVE`s, the `MODIFY` explicitly records nucleus + PML body as the intended replacement compartments, which has communicative value even though both already exist as their own rows. Both conventions are defensible; I'd rather not flip a curator's original action as a side effect of a metadata PR.
- **ii (NTR definition)** — Agreed the definition needs gene-agnostic, substrate→product wording before it could actually go to the ontology, and that BDH2's exact substrate in 2,5-DHBA formation isn't pinned down in the cited sources. I've left the honest vagueness in place for now; the substrate uncertainty is a real `knowledge_gaps` candidate and a better place to record it than an NTR definition, so I'll capture it there if/when this file is next revised rather than over-specifying a reaction the sources don't establish.

That leaves CI as the only remaining gate. `Build and test` is running on `2867c42`; I'll confirm it goes green (and won't self-merge — leaving the merge to a maintainer).

---

cmungall:
Good catches — addressed in `6bcfbc6c`:

**1 (slc5a12 statement/quote mismatch)** — You're right, the bracket-free replacement listed substrates but dropped the sodium-uptake evidence. Re-anchored the finding to the abstract's identification sentence and updated the statement to match:
- statement: *"Zebrafish SMCTs, including the electroneutral zSMCTn (Slc5a12), are sodium-coupled monocarboxylate cotransporters."*
- supporting_text: *"We have identified and characterized two different sodium-coupled monocarboxylate cotransporters (SMCT) from zebrafish (Danio rerio), electrogenic (zSMCTe) and electroneutral (zSMCTn)"* — bracket-free, verbatim, ABSTRACT.

**3 (dph2 dangling quote)** — Trimmed to *"the small iron-containing protein Dph3 donates one Fe atom"* (complete clause, still verbatim).

**2 (untouched sibling `supported_by` rows)** — CI resolves your conditional to the first branch: I ran `ai-gene-review validate --terms` locally on all 58 files (0 errors, 1 pre-existing dph2 warning), so the bracketed `supported_by` copies pass — they go through the external supported_by validator, not the bracket-stripping `findings` path. I left them deliberately, and they aren't the "divergent quotation" trap: each is correctly anchored to *its own* annotation's claim, which is a different claim from the finding. In dph2 the `supported_by` cluster-requirement quote supports a **Fe-S-cluster-binding** annotation (the framing sentence genuinely supports that), whereas the finding's statement is about **Dph3 iron donation** — the old finding quote was actually mis-anchored, which your review flagged and this fix corrected. In slc5a12 the `supported_by` "increased [Na(+)](i)…" now quotes a *different* sentence than the finding, each supporting its own row. So there's no same-sentence divergence to reconcile.

**4 (upstream validator asymmetry)** — Agree this is the real root cause and better fixed in the validator than worked around in data: stripping `[…]` from the quote but not the publication makes `[4Fe-4S]`, `[Na+]i`, and ChEBI/Rhea `[protein]` reaction strings unquotable. Symmetric stripping (both sides) would preserve the editorial-note intent while keeping chemistry quotable. That's a validator change rather than a gene-review edit, and @cmungall owns that call — happy to open an issue or a follow-up PR for `supporting_text.py` if you'd like.

Both files validate at 0 errors. CI (`test (3.12)`) should confirm green on `6bcfbc6c`.

---

cmungall:
Thanks — and great root-cause work on item 2. You're right: it's a config-loading gap in `build_supporting_text_validator` (`supporting_text.py:25-29` constructs a bare `ReferenceValidationConfig` and never loads `conf/reference_validator_config.yaml`, whose `literal_bracket_patterns` already preserve `[4Fe-4S]`/`[Na(+)]`), not two different validators and not a case for symmetric stripping. The CLI path passes `--config`, so `supported_by` validates with bracket preservation while the findings path doesn't — which is exactly why the two rows diverged. Confirmed there's no divergent-quotation trap: the surviving `supported_by` copies are correctly anchored to their own annotations.

I'll keep this data PR scoped to the data — the two quote edits stand on their own (the `dph2` one corrects a citation that was mis-anchored well before the validator change). I'll open a separate issue/PR for loading `conf/reference_validator_config.yaml` in `build_supporting_text_validator` (plus a regression test asserting a `[4Fe-4S](Cys)3` quote validates), so the findings and CLI paths agree and chemical notation stays quotable.

Minor nit noted (`slc5a12:321-324` findings lack `reference_section_type`) — leaving it for a future touch rather than spinning another CI cycle on an approved PR, unless you'd prefer I add it.

CI (`test (3.12)`) should confirm green on `6bcfbc6c`; validated locally at 0 errors.

---

cmungall:
Opened the validator follow-up as a separate PR: #2800 — loads `conf/reference_validator_config.yaml` in `build_supporting_text_validator` so the findings path applies the same `literal_bracket_patterns` as the CLI/`supported_by` path, with a regression test asserting a `[4Fe-4S](Cys)3` quote validates against `publications/PMID_34154323.md`. That makes bracketed chemical notation quotable on the findings path and removes the need for the data workarounds in this PR going forward (the two quote edits here still stand on their own — the dph2 one fixes a genuinely mis-anchored citation).

---

PR #2702, opened 2026-08-28
Complete yeast TIM10 annotation review

cmungall:
## Summary
- mark TIM10 COMPLETE after exact reconciliation of 33 qualifier-aware GOA assertion signatures
- audit all three IBA rows against PTN000113167 and current PAINT source data
- replace stale small-Tim membrane-insertase activity with evidence-matched unfolded-protein carrier activity
- retain generic interaction rows as over-annotated rather than deleting their underlying observations
- record PMID:19037698 as a verified wrong identifier while preserving independently established IMS localization
- synthesize the specific Tim9-Tim10 carrier/chaperone core function and focused follow-up experiments

## Validation
- just validate yeast TIM10
- just validate-goa yeast TIM10
- just render yeast TIM10
- just deploy-browser
- git diff --check

No support-code changes or wrapper bypasses.

### DIFF
--- a/genes/yeast/TIM10/TIM10-ai-review.html
+++ b/genes/yeast/TIM10/TIM10-ai-review.html
@@ -785,8 +785,8 @@
 
             <div class="info-item">
                 <span class="info-label">Review Status:</span>
-                <span class="status-badge status-in-progress">
-                    IN PROGRESS
+                <span class="status-badge status-complete">
+                    COMPLETE
                 </span>
             </div>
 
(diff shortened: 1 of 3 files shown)


cmungall:
Rebased onto main after the separate support fixes merged (#2703, #2704) and addressed the review blockers on exact head `f02fe3a45a`:

- GO:0140309 now uses the official current label `unfolded protein holdase activity`; the separately reviewed cache/migration fix is merged
- restored GO:0140318 `protein transporter activity` as a second core function
- marked PMID:23267104 as `NONE` / `WRONG_IDENTIFIER` after checking its E. coli-only full text
- restored the two PMID:27107014 heterologous human-partner rows to `REMOVE`
- softened the SAM-route claim to the evidence-supported implicated/reported wording
- removed the title-as-quote use for abstract-only PMID:12637749 and documented the access limitation
- clarified row-level signature wording in the notes

Verification on the rebased head: `just validate yeast TIM10` passes cleanly; `just validate-goa yeast TIM10` confirms all 49 signatures exactly match GOA; `just render yeast TIM10`, `just deploy-browser`, sidecar refresh/render wrappers, and full `just test` pass (1793 unit tests + 154 doctests, mypy, Ruff).

PR #2703, opened 2026-08-28
Refresh BioReason benchmark sidecars

cmungall:
## Summary
- regenerate stale BioReason benchmark sidecars from current source reviews
- use the public just refresh-bioreason-benchmark-sidecars wrapper
- restore the generated-sidecar invariant checked by the test suite

## Validation
- uv run pytest tests/test_bioreason_benchmark_policy.py::test_generated_sidecars_match_current_sources -q
- just test
- git diff --check

This is intentionally separate from the TIM10 gene review and GO label-cache fix.


cmungall:
Addressed both review blockers on exact head d18208fd42:

- regenerated and committed the published pages/project data copies with `just render-projects`; the three benchmark sidecars now byte-match their `projects/` sources
- added `projects/**/*.csv` and `projects/**/*.json` to the generated-pages workflow trigger
- updated all four publication-facing frozen-baseline status summaries to 67 COMPLETE / 48 DRAFT / 20 IN_PROGRESS / 4 INITIALIZED
- extended regression coverage so publication counts are derived from generated metrics and project data changes must trigger page generation

Verification: `just render-projects`, byte-for-byte sidecar comparisons, `git diff --check`, and full `just test` (1792 unit tests + 154 doctests, mypy, Ruff) all pass.

cmungall:
Addressed the rendered-slide blocker on exact head. I added a regression that checks both the source Marp HTML and its published copy against the generated status distribution (and asserts the two copies are identical); it failed on the stale 64/23 deck. Then I rebuilt via `just -f projects/BIOREASON_COMPARISON/justfile slides` and propagated via `just render-projects`. Both rendered decks now contain 67 COMPLETE / 48 DRAFT / 20 IN_PROGRESS / 4 INITIALIZED and are byte-identical. Targeted publication/workflow tests pass and `git diff --check` is clean.

PR #2704, opened 2026-08-28
Accept current GO:0140309 label

cmungall:
## Summary
- refresh GO:0140309 to the current official label, `unfolded protein holdase activity`, in the checked GO term cache
- migrate the eight authored `core_functions` that use GO:0140309 to the current label and regenerate their review HTML/browser payload
- add regression coverage through the real external term validator
- refresh BioReason reference checksums and their published copy after the mechanical YAML changes

The current label was independently confirmed from the official QuickGO ontology response; the other committed GO cache already carries the same current label. No annotation action, status, evidence, or biological description changes.

## Verification
- `just lint-caches`
- `just validate-terms <file>` for all eight affected reviews
- `just validate <organism> <gene>` for all eight affected reviews
- `just render <organism> <gene>` for all eight affected reviews
- `just deploy-browser`
- `just refresh-bioreason-benchmark-sidecars`
- `just render-projects`
- `just test` (1793 unit tests + 154 doctests, mypy, Ruff)
- `git diff --check`


cmungall:
Addressed the review findings on exact head `2de353346f`:

- migrated all eight affected authored `core_functions` labels to `unfolded protein holdase activity`
- removed the redundant GO:0140309 alias and renamed the regression so it explicitly guards the current cached label through the real external validator
- corrected the PR description to match the diff
- regenerated all eight HTML reviews and the browser payload
- refreshed BioReason reference checksums and the published sidecar copy after the mechanical YAML changes

Validation: all eight `just validate-terms` wrappers pass; all eight `just validate` wrappers pass; cache lint passes; sidecar parity tests pass; full `just test` passes (1793 unit tests + 154 doctests, mypy, Ruff).

PR #2715, opened 2026-08-29
Re-review yeast YAR1 annotations

cmungall:
## Summary
- reconcile all 19 qualifier-aware GOA rows for yeast YAR1
- remove four transcription-related IBAs propagated through a broad ankyrin-repeat PAINT node, distinguishing three current assertions from one stale SBF-complex assertion
- replace broad `unfolded protein binding` annotations with the more specific `protein carrier chaperone` term and retain stress phenotypes as non-core
- flag newer DNA-binding and G1/S PAINT assertions that can propagate to YAR1 on refresh
- add literature-backed notes, reference assessments, and regenerated HTML/browser artifacts

## Validation
- `just validate yeast YAR1`
- `just validate-goa yeast YAR1`
- `just render yeast YAR1`
- `just deploy-browser`
- strict duplicate-key YAML parse
- `git diff --check`

### DIFF
--- a/genes/yeast/YAR1/YAR1-ai-review.html
+++ b/genes/yeast/YAR1/YAR1-ai-review.html
@@ -1051,7 +1051,7 @@
                                     <span class="badge">SUPPORTS SOURCE BUT NOT TARGET</span>
 
 
-                                    <div class="propagation-comment">The source node includes SWI4/MBP1-like transcription regulators, but YAR1 lacks the APSES DNA-binding domain and functions as an Rps3 carrier chaperone.</div>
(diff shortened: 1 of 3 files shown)


cmungall:
Addressed both requested changes in 5eb7895270:

- scoped the APSES-domain argument to GO:0001228; the other transcription/MBF/SBF rows now rely on Yar1 being a 200-aa, two-ANK-repeat protein distinct from Swi6/Cdc10-like co-orthologs, plus the absence of cell-cycle or promoter evidence
- removed the incorrect claim that GO:0051082 is obsolete; the MODIFY action is explicitly based on term specificity

I also incorporated the upstream PAINT suggestion: the current-node finding now records GO:0000978 and GO:0000082 (dated 2026-02-24), and a suggested question asks PAINT curators to revise the broad node placement. I removed the weak metadata paraphrase and clarified why the existing GO:0032880 regulatory framing is retained.

Re-ran `just validate yeast YAR1`, `just validate-goa yeast YAR1`, `just render yeast YAR1`, `just deploy-browser`, strict duplicate-key parsing, and `git diff --check`; all pass.

PR #2717, opened 2026-08-29
Correct YAR1 obsolete-term rationale

cmungall:
## Summary
- correct the YAR1 review to state that GO:0051082 `unfolded protein binding` is obsolete in current GO
- retain GO:0140597 `protein carrier chaperone` as the evidence-specific replacement for Yar1
- regenerate the YAR1 HTML page and browser data

The correction follows the current GO obsoletion record, which recommends replacement by an activity term: https://amigo.geneontology.org/amigo/term/GO:0051082

## Validation
- `just validate yeast YAR1`
- `just validate-goa yeast YAR1`
- `just render yeast YAR1`
- `just deploy-browser`
- strict duplicate-key YAML parse
- `git diff --check`

### DIFF
--- a/genes/yeast/YAR1/YAR1-ai-review.html
+++ b/genes/yeast/YAR1/YAR1-ai-review.html
@@ -1433,7 +1433,7 @@
 
 
                         <div>
-                            <strong>Reason:</strong> Current GO marks GO:0051082 obsolete and recommends replacement by an activity term. Yar1 protects Rps3 from aggregation and keeps it soluble before pre-ribosome incorporation; GO:0140597 captures this specific carrier-chaperone role.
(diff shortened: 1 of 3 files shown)


cmungall:
The live GO record resolves the cache conflict: https://amigo.geneontology.org/amigo/term/GO:0051082 currently reports `obsolete unfolded protein binding`, `is_obsolete: true`, and the obsoletion comment:

> The reason for obsoletion is that this binding term should be replaced by an activity term such as protein folding chaperone (GO:0044183) or unfolded protein holdase activity (GO:0140309).

It lists both as `consider` targets, not a single `replaced_by`. AmiGO reports a 2026-08-06 ontology load; the repository ontology caches cited in the review are from March/May and the GOA snapshot can legitimately lag ontology obsoletion. Evidence-specific replacements may therefore differ among proteins.

Commit fd51148d1b now records the two generic `consider` targets and explicitly explains why Yar1 uses the more specific GO:0140597 carrier-chaperone term. I also removed the validly flagged Falcon paraphrase from `supported_by`.

Re-ran `just validate yeast YAR1`, `just validate-goa yeast YAR1`, `just render yeast YAR1`, `just deploy-browser`, strict duplicate-key parsing, and `git diff --check`; all pass.

PR #2726, opened 2026-08-29
Re-review human CNPY3 annotations

cmungall:
## Summary
- reconcile 18 physical CNPY3 GOA rows into nine qualifier-aware signatures
- retain four evidence-backed NEW annotations
- remove GO:0051082 from the proposed annotation and core function because CNPY3 is a substrate-loading HSP90 co-chaperone, not an autonomous foldase or holdase
- anchor Hsp90 binding and ER TLR-folding biology to full-text PMID:20865800 while preserving the mouse-to-human orthology boundary
- audit the signaling-receptor-binding IBA against PTHR15382 / PTN008355430
- retain generic high-throughput protein interactions as true but non-core
- regenerate the gene page and browser data

## Validation
- `just validate human CNPY3`
- `just validate-goa human CNPY3`
- `just render human CNPY3`
- `just deploy-browser`


cmungall:
Addressed the biological review in ff153f3843:

- Added NEW GO:0072657 `protein localization to membrane` and included it in the core process set. PMID:20865800 directly reports that TLR9 precursors are trapped in the ER after Cnpy3 silencing; the YAML explicitly identifies the direct work as mouse and the human proposal as orthology transfer.
- Rewrote all four GO:0005515 IPI decisions affirmatively: the observed interactions are retained because nothing contradicts them; they remain non-core because bare protein binding and heterogeneous screen partners do not establish a mechanistic MF.
- Reworked the top-level description into standalone biology and removed review/provenance language.
- Clarified GO:0034123 as downstream, non-core pathway context rather than direct signaling catalysis.

On GO:0051082, I retained the obsolete-status statement because the current official QuickGO API returns `name: obsolete unfolded protein binding` and `isObsolete: true` (with GO:0044183 and GO:0140309 in its obsoletion comment). The local `ontologies/go.tsv` row used in the review is dated 2026-03-21 and is stale relative to the live ontology. The notes now make that provenance discrepancy explicit rather than silently relying on either source.

All public wrappers pass:
- `just validate human CNPY3`
- `just validate-goa human CNPY3`
- `just render human CNPY3`
- `just deploy-browser`

@ai4c-reviewer please re-review.

PR #2728, opened 2026-08-29
Fix single-module render wrapper path

cmungall:
## Summary

Fix the public `just render-module <module>` wrapper so its documented stem argument resolves to `modules/<module>.yaml`, matching the `render-modules` CLI contract.

Previously, `just render-module complex_i_assembly_factors` passed the bare stem and always failed with `Error: File not found: complex_i_assembly_factors`.

## Validation

- `just render-module complex_i_assembly_factors` — passed and rendered `pages/modules/complex_i_assembly_factors.html`
- `git diff --check` — passed

### DIFF
--- a/src/ai_gene_review/cli.py
+++ b/src/ai_gene_review/cli.py
@@ -3066,9 +3066,22 @@ def render_modules(
         typer.echo(f"\nRendered {len(output_paths)} module page(s) to {output_dir}")
     elif files:
         total_warnings = []
-        for module_file in files:
+        had_errors = False
+        for requested_file in files:
+            module_file = requested_file
+            if not module_file.exists() and module_file.parent == Path("."):
+                filename = (
+                    module_file.name
+                    if module_file.suffix == ".yaml"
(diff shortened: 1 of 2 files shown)


cmungall:
Addressed the review in a771eaa9c4.

The wrapper remains a pass-through, preserving the three documented full-path call sites. The `render-modules` CLI now also resolves a bare stem (and a bare `.yaml` filename) against `--modules-dir`, so both forms work:

- `just render-module complex_i_assembly_factors`
- `just render-module modules/complex_i_assembly_factors.yaml`

I also fixed the pre-existing silent failure: any missing requested file or render exception now makes the command exit 1 after processing the request list.

Regression coverage was added for stem resolution, full-path preservation, and missing-file failure. Validation:

- both wrapper forms above passed
- missing-file CLI test exits 1
- `just test` passed: 1799 tests + 154 doctests, mypy, and ruff
- `git diff --check` passed

@ai4c-reviewer please re-review.

PR #2733, opened 2026-08-29
Align UPB carrier-holdase decisions

cmungall:
## Summary
- align GO:0140309 terminology with its current label and acceptor/location semantics
- add an explicit carrier-holdase decision rule
- route SurA and SecB carrier-holdases consistently to GO:0140309
- regenerate the flagship project page

## Context
Separate project-policy dependency requested during review of #2732; no gene review is changed here.

## Validation
- `just render-project UNFOLDED_PROTEIN_BINDING`
- `git diff --check`

### DIFF
--- a/pages/projects/UNFOLDED_PROTEIN_BINDING.html
+++ b/pages/projects/UNFOLDED_PROTEIN_BINDING.html
@@ -458,10 +458,10 @@ Validate with: <code>just validate-all</code> (writes <code>reports/validation-a
 </tr>
 <tr>
 <td><strong>Carrier-holdase</strong></td>
-<td><a href="http://amigo.geneontology.org/amigo/term/GO:0140309">GO:0140309</a> unfolded protein carrier activity</td>
-<td>Binds unfolded protein and escorts it between cellular components, preventing aggregation in transit</td>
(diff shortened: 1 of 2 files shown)


cmungall:
Addressed the review at `6bc805902a`.

Definition verification: the current GO:0140309 record is **unfolded protein holdase activity** with definition “A protein carrier activity that binds to a protein in an unfolded state and escorts it to an acceptor molecule or to a specific location.” See [QuickGO](https://www.ebi.ac.uk/QuickGO/term/GO:0140309); the same current GO record is exposed by [ZFIN](https://zfin.org/ZDB-TERM-251121-3). The older “between two different cellular components” text in existing reviews/project history is stale.

The follow-up therefore:
- cites the 2026-08-29 QuickGO recheck and updates all live definition/label/synonym guidance;
- keeps the carrier-parent and defined-destination requirement as the discriminator for the general holdase NTR;
- aligns Skp, HdeA, HdeB, Spy, SlyD, and CpxP table rows with their current reviews, flagging HdeB for re-review where the acceptor/destination evidence is not established;
- marks SurA as the project decision with gene-review alignment explicitly tracked in separate one-gene PR #2732.

Historical changelog entries remain unchanged as historical records. `just render-project UNFOLDED_PROTEIN_BINDING` and `git diff --check` pass.

PR #2745, opened 2026-08-29
Refresh and harden PAINT family slices

cmungall:
## Summary
- delete stale family PAINT slices when current upstream mapping is empty, consistently in bulk and single-family refreshes
- report bulk stale-slice removals to callers and CLI users
- emit canonical LF-delimited TSV files with deterministic seed ordering
- add regression coverage for stale-slice pruning, reported deletions, single-family empty results, ordering, and line endings
- refresh all cached family slices from current PANTHER PAINT data

## PTHR13848 confirmation
- `just fetch-panther-paint PTHR13848` resolves 0 nodes and retains no slice
- `PTN002302768` is absent from the cached current upstream `IBD.gaf` fetched by the wrapper
- the removed historical `GO:0000151` row is therefore confirmed upstream churn, not a local join failure

## Verification
- `just fetch-panther-paint-all`
- `just fetch-panther-paint PTHR13848`
- `just verify-panther-paint`
- `just validate-modules`
- `just test`
- `git diff --check`


cmungall:
Addressed all requested changes in f17d7d50c8:

- bulk deletion is documented and returned as `removed`; CLI reports the count and family ids
- single-family and bulk paths now share the absence-based contract and remove obsolete/header-only slices
- tests cover deletion reporting, the single-family empty case, deterministic seed order, and LF endings
- seed identifiers are sorted to eliminate future with/from ordering churn
- stale PTHR36701 module prose now describes absence rather than an empty export
- `just fetch-panther-paint PTHR13848` reports 0 nodes/no annotations and leaves no slice; PTN002302768 is absent from the wrapper-fetched current IBD.gaf

Fresh local gates: `just validate-modules`, `just test`, `just verify-panther-paint`, and `git diff --check` all pass. Please re-review.

PR #2753, opened 2026-08-29
Respect citation provenance in SFT conflict checks

cmungall:
## Summary

- centralize exact AIGR action loading for the SFT auditor, HF converter, and GO-GPT refresh path
- preserve negative actions tied to MISCITED or WRONG_IDENTIFIER references as explicit provenance-limited evidence
- classify fresh provenance-limited cases as UNC rather than NPI or an unjustified GOA-presence CNN
- retain verified/unreviewed biological negatives and positive actions
- add cross-path parity, null-safety, mixed-evidence, and committed-cohort regression coverage

## Why

A REMOVE or MARK_AS_OVER_ANNOTATED action can reject a specific citation rather than the GO term itself. That should not automatically refute an independently assessed prediction, but it also must not be discarded and converted into affirmative evidence.

## Validation

- just test (1802 tests, 154 doctests, mypy, Ruff)
- documented --help entrypoints for all three scripts
- direct DnaJ check: GO:0051087 and GO:0032991 retain manual CNN assessments without deterministic conflicts

### DIFF
--- a/scripts/auto_review_sft_predictions.py
+++ b/scripts/auto_review_sft_predictions.py
@@ -17,7 +17,7 @@ import argparse
 import copy
 import csv
 import re
-from collections import Counter, defaultdict
+from collections import Counter
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Iterable, MutableMapping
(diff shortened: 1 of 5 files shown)


cmungall:
Addressed both IMPORTANT findings in 74b12ca16c:

- provenance-limited negatives are now preserved explicitly and fresh assessments resolve to UNC, not CNN
- the action loader is shared by the SFT auditor, HF converter, and GO-GPT refresh path
- verified biological negatives remain decisive when mixed with provenance-limited evidence
- added parity and null/missing-ID coverage
- moved the shared loader into the installed package after smoke-testing all documented direct script entrypoints

Full just test is green: 1802 tests, 154 doctests, mypy, and Ruff.

cmungall:
Addressed the residual MODIFY parity finding in b49892ae6a:

- auto_assess now compares biological_actions in the MODIFY branch, matching the HF converter
- added a mixed MODIFY + provenance-limited regression proving both paths return LSP/CS2
- restored strict GO colon seven-digit validation in the shared loader

Full just test is green: 1803 tests, 154 doctests, mypy, and Ruff.

PR #2764, opened 2026-08-29
Fix GO-GPT wrapper package imports

cmungall:
## Summary

- expose the repository `src` package to the external BioReason-Pro Python interpreter in every GO-GPT wrapper
- cover all five wrapper invocations with a regression test

## Bug

`just gogpt-compare-all` invoked `scripts/gogpt_predict.py` with the BioReason-Pro virtualenv but without the repository package on `PYTHONPATH`, so every gene failed immediately with `ModuleNotFoundError: No module named 'ai_gene_review'`.

## Validation

- red regression run before the fix: 1 failed, 1,804 passed
- `just test`: 1,805 tests passed; 154 doctests passed; mypy passed for 162 source files; Ruff passed
- `git diff --check`

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -2625,7 +2625,7 @@ ui-legacy port="5123":
 #   just gogpt-predict human TP53
 #   just gogpt-predict PSEPK rpoS
 gogpt-predict organism gene:
-    ~/repos/BioReason-Pro/.venv/bin/python scripts/gogpt_predict.py {{organism}} {{gene}} --output-dir .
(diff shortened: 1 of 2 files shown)


cmungall:
Addressed the two actionable review suggestions in `1b056fe6f5`:

- preserve any pre-existing `PYTHONPATH` after the repository `src` path
- scope the regression to external BioReason invocations and report offending commands without a magic count

I kept ownership in the public wrappers rather than adding a script-side `sys.path` bootstrap.

Revalidated with `just --list` and `just test`: 1,805 tests, 154 doctests, mypy across 162 source files, and Ruff all pass.

PR #2777, opened 2026-08-30
Remove shadowed Skp SFT override

cmungall:
## Summary
- remove the unreachable Skp GO:0061077 manual override that is shadowed by the ontology-pair adjudication table
- preserve the current CNN behavior from the committed pair decision
- add a behavioral regression through `repair_document`, confirming the pair decision is applied and the dead manual override does not return

## Validation
- `just test` (1909 tests passed, 156 doctests passed, mypy and Ruff passed)
- `git diff --check`

## Review follow-up
This supersedes the initial semantic reclassification. The reviewer correctly identified that the manual override was unreachable and conflicted with the committed adjudication table. The Skp curation change will update the adjudication row and generated sidecar together in the separate one-gene PR.

### DIFF
--- a/scripts/auto_review_sft_predictions.py
+++ b/scripts/auto_review_sft_predictions.py
@@ -349,15 +349,15 @@ MANUAL_OVERRIDES: dict[tuple[str, str, str], ManualOverride] = {
         ),
     ),
     ("ECOLI", "Skp", "GO:0061077"): ManualOverride(
-        assessment="CNN",
+        assessment="NPI",
         error_type=None,
         set_error_type=True,
         summary=(
-            "Correct but not novel under the curated reference. The replacement protein "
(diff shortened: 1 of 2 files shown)


cmungall:
Addressed the requested changes in 05ac54d2f0:

- removed the unreachable Skp GO:0061077 manual override instead of reclassifying it
- preserved the existing CNN result from the ontology-pair adjudication table
- replaced the constant-only assertion with a behavioral `repair_document` regression that verifies the pair decision wins, normalizes confidence, and adds the frozen-adjudication marker
- reran `just test`: 1909 tests, 156 doctests, mypy, and Ruff all pass

The semantic Skp reclassification is deferred to the separate one-gene curation PR, where the adjudication TSV and generated SFT sidecar will be updated together after the gene review changes are present.

PR #2779, opened 2026-08-30
Complete ECOLI Skp gene review

cmungall:
## Summary
- complete the 34-row qualifier-aware review of *E. coli* Skp with exact GOA parity
- replace obsolete/generic folding and binding claims with evidence-backed outer-membrane assembly, periplasmic localization, and unfolded-protein holdase terms
- retain homotrimerization as non-core and defer two abstract-only experimental rows as `UNDECIDED`
- add literature notes, manual reference assessments, and the fetched PTHR35089/PAINT slice used to evaluate the two IBA annotations
- refresh Skp SFT and GO-GPT prediction reviews, the canonical GO:0061077 adjudication, and all deterministic benchmark/publication outputs

## Curation decisions
- 14 `ACCEPT`, 15 `MODIFY`, 3 `KEEP_AS_NON_CORE`, 2 `UNDECIDED`
- GO:0051082 and generic protein-binding rows are narrowed to GO:0140309 (`unfolded protein holdase activity`)
- broad folding/maturation claims are replaced by GO:0043165 (`Gram-negative-bacterium-type cell outer membrane assembly`)
- broad periplasm is narrowed to GO:0030288 (`outer membrane-bounded periplasmic space`)
- PAINT self-provenance for UniProtKB:P0AEU7 is treated as valid descendant evidence, not circularity
- the abstract-only PMID:17908933 IPI and PMID:16858726 HDA claims remain undecided rather than overruling full-text curators

## Benchmark impact
- canonical Skp GO:0061077 adjudication changes from `CNN` to `NPI` / `CURATION_MISTAKE`
- GO-GPT three-level overlap decreases by one at post-review and core levels
- benchmark counts, manuscript, supplement, slides, rendered pages, and figure are regenerated through project wrappers

## Validation
- `just validate ECOLI Skp` (valid; one intentional mixed-action warning for curator deference)
- `just validate-goa ECOLI Skp`
- `just verify-panther-paint`
- PredictionReview LinkML validation for both prediction sidecars
- `just render ECOLI Skp`
- `just -f projects/BIOREASON_COMPARISON/justfile gogpt-overlap`
- `just -f projects/BIOREASON_COMPARISON/justfile slides`
- `just render-project BIOREASON_COMPARISON`
- `just test` — 1909 passed, 63 deselected

PR #2781, opened 2026-08-30
Add BioReason review refresh wrappers

cmungall:
## Summary
- expose read-only and applying SFT review refreshes through public `just` recipes
- expose GO-GPT web-export check and refresh modes through public `just` recipes
- route GO-GPT refreshes through the BioReason dependency bridge
- verify exact wrapper argument boundaries

## Validation
- `uv run pytest -q tests/test_bioreason_review_refresh_wrappers.py tests/test_gogpt_wrapper.py tests/test_bioreason_sidecar_wrapper.py`
- `just test` (1920 tests, 156 doctests, mypy, Ruff)
- `git diff --check`

This is an isolated wrapper-layer prerequisite found while responding to review feedback on #2779.

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -1481,6 +1481,22 @@ render-bioreason-eval:
 refresh-bioreason-benchmark-sidecars:
     uv run python projects/BIOREASON_COMPARISON/write_benchmark_sidecars.py
 
+# Audit SFT prediction reviews against the current GOA/AIGR snapshot without writing
+check-bioreason-sft-reviews:
+    uv run python scripts/auto_review_sft_predictions.py
+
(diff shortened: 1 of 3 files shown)


cmungall:
Addressed both requested changes in e13136a1bd:

- GO-GPT web-export check/refresh now use the repository interpreter directly; only inference recipes require the BioReason dependency bridge. The existing bridge invariant is narrowed accordingly.
- Wrapper tests now assert each target script exists and run each parser's `--help` to verify every exposed option is genuinely supported.

Focused validation: 16 passed. The earlier full `just test` run was also green (1,920 tests, 156 doctests, mypy, Ruff); these follow-up changes are covered by the focused wrapper suite.

PR #2785, opened 2026-08-30
Fail fitness fetch when no data source is reachable

cmungall:
## Summary

- distinguish an unavailable FEBA source from a reachable source with no record for the requested gene
- fail without writing a misleading fitness artifact when neither local nor remote data can be queried
- retain the existing placeholder artifact for genuine no-record results
- add regression coverage for both outcomes

## Verification

- `just format`
- `just pytest` (1925 passed, 63 deselected)
- `git diff --check`

### DIFF
--- a/scripts/fetch_fitness_data.py
+++ b/scripts/fetch_fitness_data.py
@@ -24,6 +24,31 @@ ORGANISM_TO_FEBA = {
 
 FEBA_REPO = Path.home() / "repos" / "feba"
 FEBA_DB = FEBA_REPO / "cgi_data" / "feba.db"
+REQUIRED_HTML_FILES = (
+    "fit_logratios_good.tab",
+    "fit_t.tab",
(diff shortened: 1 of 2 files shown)


cmungall:
Addressed all three requested changes:

- fixed the case-insensitive HTML challenge guard and now reject remote TSV responses unless both `fit` and `t` headers are present; added regression coverage for HTML and invalid headers
- probe the SQLite `Gene` table before treating the database as usable, guard SQLite queries against `sqlite3.Error`, and added corrupt-database coverage using a real minimal schema for the reachable case
- added direct coverage for a reachable remote source returning an empty result

Also consolidated BarSeqR required-file checks and expanded the recovery message to name the supported HTML-table source.

Verification after the update:

- `just format`
- `just pytest` — 1929 passed, 63 deselected
- `git diff --check`

PR #2797, opened 2026-08-30
Fix #2761: self-source rows are SUPPORTS_TRANSFER, not CIRCULAR_OR_REDUNDANT

### LINKED ISSUE #2761: 27 propagation_review rows mark the target's own accession CIRCULAR_OR_REDUNDANT
Found by a corpus audit of all 1,843 `review.propagation_review` blocks (366 files). Written up in `projects/IBA_REVIEW/propagation-review-audit.md`, linked from the IBA_REVIEW project page. Filing to track it, since the audit records the finding but nothing schedules a decision.

## The rule being broken

Both `CLAUDE.md` and `projects/IBA_REVIEW.md` state that a target appearing in its own `WITH/FROM` is **correct and expected** — the target's own experimental annotation is one of the descendant evidences the PAINT curator used to place the IBD — and that such a source must **never** be marked `CIRCULAR_OR_REDUNDANT`.

27 rows across 16 human genes do exactly that. Some state the inverted reading explicitly, e.g. AAK1:

> The IBA includes the human target itself among its sources, so that item is self-supporting and adds no independent propagation evidence.

## The data refutes the inverted reading in 26 of 27 cases

The target does carry its own experimental annotation for the term — which is precisely why it seeded the IBD:

| grounding on the target | rows |
|---|---|
| own experimental annotation at the **same** term (IDA/IMP/EXP/TAS) | 20 |
| own experimental annotation at a **descendant** term | 6 |
| none found in the cached set | 1 (LNX1, `GO:0005737`) |

The six descendant cases: LMTK2 `GO:0004672` → IDA `GO:0004674`; LMTK3 `GO:0004672` → EXP `GO:0106310`; AFF1 `GO:0006355` → IMP `GO:0032968`; LPGAT1 `GO:0005783` → IMP `GO:0005789`, `GO:0016746` → IDA `GO:0071617`, `GO:0012505` → the same ER evidence.

Affected genes (all human): A4GNT, AADAC, AAK1, AASDHPPT, ADAMTSL5, AFF1, LMTK2, LMTK3, LNX1, LPAR6, LPCAT4, LPGAT1, LPIN2, LRBA, LRCH1, LRCH3.

## The corpus already contains the correct handling

This is why it reads as a consistency fix rather than a judgement call — ADPRS marks its self-source `SUPPORTS_TRANSFER` with:

> self-reference: the target is its own IBD seed

## Suggested action

Retype the 27 sources to `SUPPORTS_TRANSFER` with an ADPRS-style comment, and fix the prose in the reviews that state the inverted reading (AAK1 is the clearest). LNX1 needs a look either way, since its grounding could not be confirmed from the cached set.

Deliberately **not** done as part of #2757: that PR adds the FamilyReview schema and residue-claim machinery, and rewriting existing curation across 16 gene reviews is a separate decision.

## Related

- Audit: `projects/IBA_REVIEW/propagation-review-audit.md`
- The same audit records a second, softer finding: the structured `propagation_review` enums drift from the prose beside them (ADPRS `SOURCE_BAD` with all sources `SUPPORTS_TRANSFER`; LPA tagged `PSEUDO_OR_SUBACTIVITY_LOSS` while its own text says the triad is retained). Worth a separate decision.

cmungall:
Closes #2761.

## The rule being restored

27 rows across 16 human gene reviews marked the **target's own accession** `CIRCULAR_OR_REDUNDANT`. Both `CLAUDE.md` and `projects/IBA_REVIEW.md` explicitly forbid this:

> The target appearing in its own `WITH/FROM` is correct and expected. When a gene has its own experimental annotation for the term, that annotation is one of the descendant evidences the curator used to place the IBD… **Do not** mark such a source `CIRCULAR_OR_REDUNDANT`.

The inverted reading treats a *marker of experimental grounding* as an *absence of evidence*.

## Re-deriving changed the picture

The audit reported 26 of 27 grounded, with LNX1 `GO:0005737` as the exception. Re-deriving with **GO ancestry closure** rather than literal term matching shows **all 27 are grounded** — LNX1's cytoplasm is grounded by its own IDA to `GO:0005829` cytosol, which is `part_of` cytoplasm. My earlier check only compared term ids as strings.

| grounding | rows |
|---|---|
| own experimental annotation at the same term | 20 |
| own experimental annotation at a descendant term | 7 |
| none | **0** |

## What changed

Each comment now names the grounding actually found on that gene rather than repeating boilerplate:

```yaml
      - source_id: UniProtKB:Q8IWU2
        source_label: human LMTK2 target
-       source_status: CIRCULAR_OR_REDUNDANT
-       comment: Self-token in WITH/FROM; direct human assays independently establish activity.
+       source_status: SUPPORTS_TRANSFER
+       comment: >-
+         Self-reference: the target is its own IBD seed, which is expected rather
+         than circular -- its own IDA annotation to GO:0004674 (protein
+         serine/threonine kinase activity), a descendant of this term, is one of
+         the descendant evidences behind the IBD. The IBA then asserts the
+         additional claim that the function is inherited rather than
+         lineage-specific.
```

AAK1's `reason` stated the inverted reading in prose — *"that item is self-supporting and adds no independent propagation evidence"* — and is corrected.

## Deliberately not changed

**AFF1's `UniProtKB-SubCell:SL-0191` stays `CIRCULAR_OR_REDUNDANT`.** It derives from the entry's own ECO:0000305 SUBCELLULAR LOCATION line, so it genuinely is not an independent witness — that is the case the enum exists for. Only sources naming the review's *own* UniProt accession were touched.

## Method note

Applied by text-level edit rather than round-trip YAML: `ruamel` re-joined wrapped scalars and produced a **1650-line diff for 27 field changes**, which would have been unreviewable. The surgical version touches only the `source_status` line and the `comment` block.

Key order varies across the corpus — AFF1 writes `comment` first, so `source_id` is not on the `- ` line — so entries are matched on block content rather than the leading key. My first attempt keyed on `source_id` and silently missed AFF1's two entries; the assertion that all expected rows were applied caught it.

## Verification

- all 16 files parse; re-deriving the query now returns **0 rows**
- `just validate` passes on 15 of 16 changed genes
- LRBA fails on `references[17].findings[1].supporting_text`, **pre-existing on main** and not in this diff (verified against a clean tree)

https://claude.ai/code/session_01Q2fW7K7A93LZvru2WdCicg

### DIFF
--- a/cache/go/terms.csv
+++ b/cache/go/terms.csv
@@ -1882,6 +1882,7 @@ GO:0008188,neuropeptide receptor activity,2026-05-01T15:49:15.420424
 GO:0008190,eukaryotic initiation factor 4E binding,2026-03-22T22:38:14.237952
 GO:0008191,metalloendopeptidase inhibitor activity,2026-05-08T22:12:02.942225
 GO:0008193,tRNA guanylyltransferase activity,2026-07-23T20:35:13.590483
+GO:0008195,phosphatidate phosphatase activity,2026-08-30T13:00:40.329676
 GO:0008198,ferrous iron binding,2026-05-08T22:12:02.942228
 GO:0008199,ferric iron binding,2026-05-08T22:12:02.942232
 GO:0008200,ion channel inhibitor activity,2026-05-08T22:12:02.942235
@@ -4056,6 +4057,7 @@ GO:0032153,cell division site,2026-05-08T22:12:02.947586
 GO:0032174,cellular bud neck septin collar,2026-05-08T22:12:02.947595
 GO:0032182,ubiquitin-like protein binding,2026-08-09T04:14:55.073218
 GO:0032183,SUMO binding,2026-05-08T22:12:02.947599
+GO:0032185,septin cytoskeleton organization,2026-08-30T13:01:15.136487
 GO:0032197,retrotransposition,2026-03-22T22:38:14.237952
 GO:0032200,telomere organization,2026-03-22T22:38:14.237952
 GO:0032204,regulation of telomere maintenance,2026-05-08T22:12:02.947602
@@ -7382,6 +7384,7 @@ GO:0070888,E-box binding,2026-05-08T22:12:02.956000
 GO:0070902,mitochondrial tRNA pseudouridine synthesis,2026-06-18T17:59:50.418226
 GO:0070911,global genome nucleotide-excision repair,2026-05-08T22:12:02.956003
 GO:0070914,UV-damage excision repair,2026-05-08T22:12:02.956007
+GO:0070915,lysophosphatidic acid receptor activity,2026-08-30T13:00:08.283626
 GO:0070922,RISC complex assembly,2026-05-08T22:12:02.956010
 GO:0070932,obsolete histone H3 deacetylation,2026-03-22T22:38:14.237952
 GO:0070935,3'-UTR-mediated mRNA stabilization,2026-05-08T22:12:02.956013
@@ -9244,6 +9247,7 @@ GO:1990597,AIP1-IRE1 complex,2026-05-08T22:12:02.960754
 GO:1990599,3' overhang single-stranded DNA endonuclease activity,2026-07-25T14:15:27.431102
 GO:1990604,IRE1-TRAF2-ASK1 complex,2026-05-08T22:12:02.960757
 GO:1990617,CHOP-ATF4 complex,2026-05-08T22:12:02.960760
+GO:1990624,guanyl nucleotide exchange factor inhibitor activity,2026-08-30T13:01:02.957774
 GO:1990627,mitochondrial inner membrane fusion,2026-05-08T22:12:02.960763
 GO:1990633,mutator focus,2026-05-08T22:12:02.960767
 GO:1990634,protein phosphatase 5 binding,2026-05-08T22:12:02.960770
@@ -9282,6 +9286,7 @@ GO:1990856,methionyl-initiator methionine tRNA binding,2026-05-08T22:12:02.96086
 GO:1990858,cellular response to lectin,2026-05-08T22:12:02.960865
 GO:1990861,Ubp3-Bre5 deubiquitination complex,2026-05-08T22:12:02.960868
 GO:1990863,acinar cell proliferation,2026-05-08T22:12:02.960872
+GO:1990869,cellular response to chemokine,2026-08-30T13:01:03.016071
 GO:1990871,Vma12-Vma22 assembly complex,2026-05-08T22:12:02.960875
 GO:1990874,vascular associated smooth muscle cell proliferation,2026-05-08T22:12:02.960878
 GO:1990880,cellular detoxification of copper ion,2026-03-22T22:38:14.237952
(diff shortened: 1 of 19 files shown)


cmungall:
## Completeness check on the fix

The fix matched sources containing the review's own `id`. That would miss a self-reference written a different way — a gene symbol, an isoform-suffixed accession, or a MOD id for the same gene — so I ran a deliberately over-broad heuristic across the corpus to look for survivors.

**61 `CIRCULAR_OR_REDUNDANT` sources remain. 21 were flagged as "plausibly self-referential"; on inspection none of them are.** They fall into one class: a **duplicate identifier for a source already listed**, and the enum is being used correctly for them.

AEBP2 `GO:0035098` is the clearest case:

```
SUPPORTS_TRANSFER      UniProtKB:Q6ZN18            AEBP2 itself (self-reference)     ← already correct
SUPPORTS_TRANSFER      UniProtKB:Q9Z248            mouse Aebp2 (reviewed, 504 aa)
CIRCULAR_OR_REDUNDANT  ensembl:ENSMUSP00000084896  mouse Aebp2 - the same entity as UniProtKB:Q9Z248
```

The Ensembl id is a second identifier for a protein already counted, so it genuinely "adds no independent support" — the enum's own wording. LRP11 is the same shape (`ensembl:ENSMUSP00000019931` beside `UniProtKB:Q8CB67`, both mouse Lrp11). And note AEBP2 already marked its *own* accession `SUPPORTS_TRANSFER` with the comment "AEBP2 itself (self-reference)" — that gene was doing it right before this PR.

My heuristic over-matched because it keyed on gene symbol, which collides across species: "mouse Aebp2" shares a symbol with human AEBP2 without being the same entity.

**Conclusion: the fix is complete for #2761's scope**, and the remaining 40 non-flagged sources (12 `UniProtKB-SubCell`, 9 `ensembl`, 7 `InterPro`, 7 `GO`, 5 `ARBA`) are unrelated to it.

Worth noting for anyone reading the enum later: `CIRCULAR_OR_REDUNDANT` is now carrying two distinct meanings in the corpus — genuine circularity (AFF1's SubCell keyword derived from the entry's own ECO:0000305 line) and mere duplicate identity (these Ensembl/UniProt pairs). Both fit the definition, but they are different situations and it may be worth splitting them if the distinction ever matters downstream.

cmungall:
## CI failure diagnosed: a citation marker, not a bad quote

`test (3.12)` failed on `Validate gene reviews (scoped)`. That step validates **changed** files, so a pre-existing broken quote in LRBA became this PR's problem the moment the PR touched that file. Fixed in `ac444d4533`.

**The quote was not wrong.** It is a literal substring of the cached publication — verified directly (`q in pub` → `True`), and `publications/PMID_33960403.md` is tracked and identical on main, so CI sees the same text I do.

The failure is a convention collision. `linkml_reference_validator` treats `[square brackets]` in a quote as **editorial notes to be stripped** (its own docstring: *"Editorial notes in [square brackets] that are ignored by default"*). So the paper's citation marker `[38, 39]` was removed and the two halves rejoined into:

```
...as LRBA KO mice showed limited pathology . Accordingly, we found that...
                                            ^^^ not a substring of anything
```

Its documented conventions are `[...]` for editorial notes and `...` for omitted text. I tested both:

| form | result |
|---|---|
| `[38, 39]` as stored | ✗ stripped, halves rejoined into one unmatched part |
| `[...]` elision | ✗ same — stripped and rejoined |
| `...` elision | ✓ splits into parts matched independently |

So the quote now reads `…limited pathology ... Accordingly, we found…`.

**Checked that this is faithful rather than merely passing:** both halves are verbatim substrings of PMID:33960403, and the elision stands in for exactly `"[38, 39]."` — a citation marker, not omitted substance. The finding's meaning is unchanged.

All 16 changed gene reviews now validate locally.

Worth noting independently of this PR: any quote spanning a bracketed citation will hit this, and the failure message is misleading — it prints the *post-stripping* text and says "not a verbatim substring", which points the reader at the quote rather than at the bracket convention. That cost me a couple of wrong turns before I read the validator source.

PR #2800, opened 2026-08-31
Load reference-validator config in build_supporting_text_validator

cmungall:
## Summary

- `build_supporting_text_validator` (the repo-local `findings` validation path used by `validate_reference_finding_supporting_text`) constructed a bare `ReferenceValidationConfig(cache_dir=…, fetch_full_text=False)` and **never loaded `conf/reference_validator_config.yaml`**. The CLI path (`cli.py`, `--config conf/reference_validator_config.yaml`) does load it, so the two paths disagreed: the config's `literal_bracket_patterns` — which preserve bracketed chemical notation like `[4Fe-4S]` and `[Na(+)]` instead of stripping them as citation markers — applied only to `supported_by` items, not to `references[].findings[].supporting_text`.
- Practical effect: a faithful, verbatim quotation of bracketed chemistry validated on the `supported_by` path but **failed on the findings path**, forcing curators to contort finding quotes to avoid the brackets (this surfaced in PR #2693, where two DANRE findings had to be re-quoted around `[4Fe-4S]` / `[Na(+)]`).
- Fix: `build_supporting_text_validator` now loads `conf/reference_validator_config.yaml` when present and merges the required overrides (`cache_dir` from the caller, `fetch_full_text` off, `reference_base_dir` resolved to an absolute path). Both validation paths now share the same bracket-preservation, `skip_prefixes`, and `unknown_prefix_severity` settings.

Closes #

## Validation

New `tests/test_supporting_text_config.py`:
- asserts a verbatim `[4Fe-4S](Cys)3` quote validates against `publications/PMID_34154323.md` on the findings path (fails before this change, passes after);
- asserts the built validator carries the repo config's `literal_bracket_patterns`.

```
$ uv run pytest tests/test_supporting_text_config.py tests/test_reference_validation.py \
    tests/test_finding_review_validation.py tests/test_goref_reference_validation.py \
    tests/test_file_reference_validation.py -q
28 passed, 1 deselected
```
`ruff format`, `ruff check`, and `mypy` on the changed files all pass.

## Test plan

- [x] New regression test fails before the fix and passes after
- [x] Existing reference/finding/file-reference validation tests still pass (28 passed)
- [x] `ruff check` + `ruff format --check` clean on changed files
- [x] `mypy` clean on `supporting_text.py`
- [x] `cache_dir` stays caller-controlled (absolute) and `fetch_full_text` stays off regardless of config contents

---
_Generated by [Claude Code](https://claude.ai/code/session_01DtkppRASTSuhnvKE8ngyeR)_

### DIFF
--- a/genes/DROME/Hsp26/Hsp26-ai-review.yaml
+++ b/genes/DROME/Hsp26/Hsp26-ai-review.yaml
@@ -855,9 +855,7 @@ references:
       member that exclusively assisted in HSP70-dependent refolding of 
       stress-denatured proteins. Furthermore, we report that HSP67BC, which has 
       no role in protein refolding, was the most effective small HSP preventing 
-      toxic protein aggregation in an HSP70-independent manner. ...[HSP26 
-      showed] overexpression of the classical small HSPs (HSP23, HSP26, and 
-      HSP27) increased luciferase refolding
+      toxic protein aggregation in an HSP70-independent manner.
 - id: PMID:32437379
   title: Small heat shock proteins determine synapse number and neuronal 
     activity during development.
--- a/genes/DROME/Hsp26/Hsp26-ai-review.yaml
+++ b/genes/DROME/Hsp26/Hsp26-ai-review.yaml
@@ -848,9 +848,9 @@ references:
     supporting_text: our results strongly suggest that the refolding capacity of
       D. melanogaster HSP27 and CG14207 is partially dependent on an intact 
       HSP70 machine
-  - statement: Hsp26 has intermediate activity in both refolding and 
-      anti-aggregation compared to specialized members CG14207 (strong refolder)
-      and HSP67BC (strong anti-aggregation).
+  - statement: Among Drosophila small HSPs, CG14207 is a specialized HSP70-dependent
+      refolder and HSP67BC a specialized HSP70-independent anti-aggregase, defining the
+      functional poles against which classical members such as Hsp26 sit.
     supporting_text: We identified CG14207 as a novel and potent small HSP 
       member that exclusively assisted in HSP70-dependent refolding of 
       stress-denatured proteins. Furthermore, we report that HSP67BC, which has 
--- a/genes/yeast/TIM10/TIM10-ai-review.yaml
+++ b/genes/yeast/TIM10/TIM10-ai-review.yaml
@@ -985,7 +985,7 @@ references:
       This publication is about colorectal surgery and has no relevance to TIM10 or
       mitochondria. It was likely cited by error in ComplexPortal (possibly confused
       with PMID:19037098).
-    supporting_text: "Anastomotic leakage is a feared complication following colorectal surgery [this paper has no relevance to TIM10]"
+    supporting_text: Anastomotic leakage is a feared complication following colorectal surgery and is associated with early and long-term morbidity and mortality.
 - id: PMID:23267104
   title: Proteome-wide protein interaction measurements of bacterial proteins of unknown
     function.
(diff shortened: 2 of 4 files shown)


cmungall:
Ran the repo-wide verification you asked for (an old-config-vs-new-config diff over **every** `references[].findings[].supporting_text`, which is exactly the population this change affects — more targeted than `validate-all`).

**Scanned 4,460 files, 15,804 literature findings quotes:**
- **115 improvements** (old FAIL → new OK) — real chemical/notation quotes that were previously unquotable: `[4Fe-4S]`, `[2Fe-2S]`, `[14C]`, `[R75Q]`, `[Ca(2+)]`, `[nic…]`, etc., across IBA57, ABCB7, HgcA/HgcB, SLC25A11, ACAP2, and many more.
- **4 findings affected the other way** — and, notably, **not** the numeric-citation-marker case (`[12]`) we were worried about. All four are **editorial-note brackets** (the CLAUDE.md "[editorial notes are ignored]" convention): `[NEK3 localization data is in the full text body…]`, `[mbiA/htgA was not among…]`, `[HSP26 showed]`, `[this paper has no relevance to TIM10]`. These would fail on the `supported_by` path too (it already loads this config), so the change surfaces pre-existing non-conforming quotes rather than creating a new class.

Under the **real** validator (which downgrades failures to WARNING when a finding is `full_text_unavailable` / abstract-only), only **2 of the 4 were blocking errors**: `DROME/Hsp26` and `yeast/TIM10`. Both are fixed in `cba7f57`:
- **Hsp26** — the finding embedded an editorial bridge `...[HSP26 showed]` joining two spans, and the trailing span duplicated a separate finding; truncated to the two contiguous verbatim sentences about CG14207/HSP67BC, which fully support the statement.
- **TIM10** — a deliberately mis-cited off-topic paper; replaced the `[this paper has no relevance to TIM10]` editorial quote with a clean verbatim off-topic sentence from its abstract (the mis-citation note stays in the finding `statement`).

`NEK3` and `mbiA` are `full_text_unavailable` findings, so they only warn (non-blocking) and are left as-is. **All four now validate at 0 errors**, so validate-all is clean.

On the 🔵 suggestions:
- **3, 4, 5, 6** — applied in `26a694f8` (added the `[Na(+)]` case, a non-verbatim negative control, exact set-equality on `literal_bracket_patterns`, and an `lru_cache`/`cache_clear` docstring note).
- **1 (config-loader DRY)** — checked: `linkml_reference_validator==0.0.0` here exposes no `from_yaml`/`load_config` helper (only a deprecated Pydantic-v1 `parse_file`), and `ReferenceValidationConfig.model_config['extra']` is `None`, so Pydantic **ignores** unknown keys — a future CLI-only key won't raise inside the `lru_cache`d builder. The hand-rolled `**config_data` construction is therefore safe; I kept it.
- **2** — wording nit noted; harmless.
- **7 (un-contort the DANRE quotes)** — that's the payoff and belongs to the curation PR #2693, not this infra PR.

---

cmungall:
Thanks for the thorough independent re-verification — glad the module-path coverage and severity-downgrade checks confirmed it.

Applied suggestion 1 in `ed359ab`: the Hsp26 truncation had left the statement ahead of its quote, so I reworded the `statement` to what the surviving quote actually shows (CG14207 = specialized refolder, HSP67BC = specialized anti-aggregase, defining the poles), with Hsp26's own activity carried by the sibling finding. Verbatim quote unchanged; file still validates at 0 errors.

Notes 2 (stale rendered HTML — regenerated by `generate-pages` on merge) and 3 (the editorial-bracket convention now needs a real home — an `editorial_note` field or a third `literal_bracket_patterns` exclusion) are follow-up material, not this PR. Note 3 is a good idea; I can open a separate issue for it if you'd like.

---

PR #2801, opened 2026-08-31
Complete E. coli CnoX annotation review

cmungall:
## Summary
- refresh CnoX UniProt/GOA data and reconcile the review to all 10 current annotations
- accept the PAINT-backed cell-redox-homeostasis IBA with an explicit propagation audit
- replace obsolete unfolded-protein-binding/core carrier claims with the shared holdase NTR and add the 2023 GroEL-CnoX primary evidence
- refresh the deterministic GO-GPT/BioReason sidecars, publication metrics, and rendered pages affected by CnoX completion

## Key curation decisions
- leave the two abstract-only PMID:15690043 mixed interaction rows `UNDECIDED`
- propose `protein-folding chaperone binding` only for the directly supported GroEL/DnaK pair assertions, with curator-facing row-splitting caveats
- retain protein refolding as a core process without claiming that CnoX is an ATP-dependent foldase

## Validation
- `just validate ECOLI CnoX`
- `just render ECOLI CnoX`
- `just render-project BIOREASON_COMPARISON`
- `just pytest` (1937 passed, 63 deselected)


cmungall:
Addressed all three requested changes in f1e1718dbe:

1. GO:0045454 is now `KEEP_AS_NON_CORE`, explicitly engaging the normal bulk protein-redox phenotype while retaining the client-level mixed-disulfide rationale and valid PAINT trace. It was removed from `core_functions`.
2. Fetched and cited PMID:21195694. The DnaN row now records the two-experiment IntAct support, beta-clamp refolding result, and replication phenotype; it remains `UNDECIDED` only because DnaN is a client and the current ontology lacks the proposed general holdase MF.
3. Restored the PMID:21498507 GroEL/GroES inhibition result and reconciled it with the 2023 GroES-release plugin model. The review deliberately declines a directional regulatory BP without in-vivo evidence.

Also added the requested standalone `NEW` GO:0051087 row from PMID:36764293, a `NEW` GO:0009408 row from the thermal-sensitivity phenotype, and the non-oxidoreductase clause in the description.

Revalidated with `just validate ECOLI CnoX`; targeted derived-artifact guards pass; full `just pytest` is green (1937 passed, 63 deselected). Please re-review the updated head.

PR #2803, opened 2026-08-31
Reuse existing review accession in fetch-gene

cmungall:
## Summary
- reuse the UniProt accession recorded in an existing gene review before attempting a symbol lookup
- preserve explicit `--uniprot-id` precedence and fall back to normal resolution for new reviews
- add hermetic regression coverage for both paths

This fixes refreshes for MOD-standard symbols absent from UniProt gene_exact indexing, including `SCHPO/dca7`, without bypassing the `just fetch-gene` wrapper.

## Validation
- `uv run pytest -q tests/test_gene_etl.py tests/test_fetch_gene_wrapper.py`
- `just fetch-gene SCHPO dca7` (resolved local `O74763`; no tracked gene files changed)
- `just test` (1939 passed, 63 deselected; 156 doctests passed, 37 skipped; mypy and ruff passed)

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -91,6 +91,7 @@ sync-codex-skills:
 # Fetch gene data from UniProt and GOA
 # Use --alias to specify a custom directory name and file prefix
 # Use --force to overwrite existing UniProt and GOA files
+# Accession precedence: --uniprot-id, existing review id, then symbol resolution
 # Example: just fetch-gene 9BACT F0JBF1 --alias HgcB
 # Example: just fetch-gene human TP53 --force
 [positional-arguments]
--- a/src/ai_gene_review/etl/gene.py
+++ b/src/ai_gene_review/etl/gene.py
@@ -22,6 +22,18 @@ import yaml
 import re
 
 
+# Official UniProtKB accession syntax (6 or 10 characters).
+# https://www.uniprot.org/help/accession_numbers
+_UNIPROT_ACCESSION_RE = re.compile(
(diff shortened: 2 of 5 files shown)


cmungall:
Addressed the requested changes in `f7cc469e55`:

- cross-check a reused review ID against the primary `AC` in the fetched UniProt flat file before calling QuickGO
- fall back to symbol resolution and fetch GOA only with the resolved primary accession when the review ID is stale or cannot be confirmed
- reject non-UniProt review IDs (including RNAcentral IDs) from the reuse path
- cover no review, missing/blank ID, ncRNA ID, malformed YAML, and stale/secondary accession behavior hermetically
- update the function and wrapper documentation with accession precedence

Validation: `just fetch-gene SCHPO dca7` leaves tracked gene data untouched; `just test` passes 1,945 tests (63 deselected), 156 doctests (37 skipped), mypy, and ruff.

cmungall:
Addressed the remaining DRY finding in `e9b13e03b7`:

- moved the single strict `is_uniprot_accession` implementation and its doctests to `etl.gene`
- imported that implementation from `deep_research_unified` and removed the duplicate regex/helper
- pinned the stricter trailing-newline behavior
- use the canonical primary `AC` already returned by UniProt for QuickGO instead of re-resolving the symbol
- recover from a deleted review accession via a narrow symbol-resolution fallback, with a hermetic test

Validation: 29 focused tests pass; live `just fetch-gene SCHPO dca7` leaves tracked gene data unchanged; `just test` passes 1,947 tests (63 deselected), 156 doctests (37 skipped), mypy, and ruff.

PR #2809, opened 2026-08-31
Avoid blank source metadata in hypothesis reports

cmungall:
## Summary
- render explicit `Not supplied.` placeholders for free-text hypothesis source metadata
- prevent trailing whitespace from propagating into saved OpenScientist/deep-research reports
- add a regression test for records without source files or selectors

## Validation
- `just test`
  - 1,948 tests passed / 63 deselected
  - 156 doctests passed / 37 skipped
  - mypy clean
  - Ruff clean
- `git diff --check`


cmungall:
Addressed the provenance/test suggestions in 2a64adb5c2: direct free-text records now retain the known review YAML as `source_file` with selector `free-text`; the regression suite now covers that real CLI path and asserts all template substitutions are non-empty. Re-ran the public `just test` wrapper: 1,949 tests passed (63 deselected), 156 doctests passed (37 skipped), mypy and Ruff clean.

PR #2814, opened 2026-08-31
Restore generic deep-research wrapper

cmungall:
Restores the documented public `just deep-research ORGANISM GENE --provider PROVIDER` entry point while preserving the legacy positional provider syntax and all provider-specific wrappers.

Adds validation for conflicting positional and option provider selections.

Verification:
- `just test`
- targeted wrapper tests
- `just --show deep-research`
- `git diff --check`


cmungall:
Addressed both requested changes in 4ee43b9e14:

- moved the generic recipe and its documentation above the OpenAI-specific block, preserving both entries in `just --list` and documenting supported options plus `--extra-args` ordering;
- moved provider tests to a dedicated file and added end-to-end coverage through `main()` for option and legacy positional syntax, conflict validation with `pytest.raises`, and a recording-`uv` test of the public `just deep-research` forwarding path.

Verification: focused 23-test suite, full `just test` (1,953 unit tests; 156 doctests; mypy; Ruff), `just --list`, and `git diff --check` all pass.

PR #2817, opened 2026-08-31
Validate deep-research provider names

cmungall:
Prevents misspelled generic or fallback provider names from reaching output-path construction and creating misleading `-deep-research-{provider}.md` artifacts.

The allow-list is derived from the repository’s supported wrappers and includes openai, perplexity, perplexity-lite, falcon, cyberian, codex, asta, and openscientist. Validation covers `--provider`, the legacy positional form, and `--fallback`.

Also makes the parsed UniProt context type explicit after targeted type checking exposed an existing heterogeneous-dictionary inference mismatch.

Verification:
- focused provider/wrapper suite: 28 passed
- full `just test`: 1,958 unit tests; 156 doctests; mypy; Ruff
- `git diff --check`

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -336,7 +336,7 @@ deep-research-asta organism gene_id *args="":
 #   just deep-research-codex human TP53
 #   just deep-research-codex METEA C5B1I4 --alias mllA
 deep-research-codex organism gene_id *args="":
-    uv run python scripts/deep_research_wrapper.py {{organism}} {{gene_id}} cyberian --extra-args --param agent_type=codex {{args}}
+    uv run python scripts/deep_research_wrapper.py {{organism}} {{gene_id}} codex {{args}}
 
 # Deep research on an InterPro entry (family/domain) behind InterPro2GO annotations.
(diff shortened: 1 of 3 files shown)


cmungall:
Addressed the blocking codex mismatch in 2ad36c6c90:

- gene `codex` now maps to client provider `cyberian` plus `agent_type=codex`, preserving the `-deep-research-codex.md` artifact name;
- `deep-research-codex` now passes the canonical `codex` wrapper value instead of disguising it as cyberian;
- added a direct command-mapping regression test;
- strengthened the provider/recipe test to parse the provider-specific recipes from `project.justfile` rather than asserting only a literal set;
- aligned the UniProt-context return annotation with its heterogeneous value type.

Verification: focused 29-test suite, full `just test` (1,959 unit tests; 156 doctests; mypy; Ruff), `just --show deep-research-codex`, and `git diff --check` all pass.

The separately identified pre-existing `perplexity-lite` fallback filename bug will be handled in a follow-up PR after this one merges.

PR #2842, opened 2026-09-01
Complete pmp20 propagation evidence

cmungall:
## Summary
- account for the fly donor in the mitochondrial PAINT propagation review
- add direct file-backed evidence for the mammalian transit peptide and target cytoplasmic localization
- align the homolog-panel finding with a complete verbatim results excerpt
- explain why fungal peroxisome evidence remains defensible while animal mitochondrial targeting does not transfer

## Validation
- just validate SCHPO pmp20
- just validate-families
- just render SCHPO pmp20
- git diff --check

### DIFF
--- a/genes/SCHPO/pmp20/pmp20-ai-review.html
+++ b/genes/SCHPO/pmp20/pmp20-ai-review.html
@@ -975,7 +975,7 @@
 
 
                         <div>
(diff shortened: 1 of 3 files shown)


cmungall:
Addressed both requested evidence fixes and both suggestions: replaced the activity quote with the direct CbPmp20 peroxisomal-membrane localization sentence; reframed fungal evidence as independent node grounding while keeping target transfer unverified because Pmp20 lacks an obvious PTS1; clarified that the PAINT node donors are mammalian; and added RGD:71007 to the mitochondrial source accounting. Revalidated gene and family reviews and rerendered.

PR #2862, opened 2026-09-01
Repair PSEPK glutathione fusion family evidence

cmungall:
## Summary
- represent the experimentally bifunctional Q9CM00 fusion in both glutathione reaction leaves using its exact PANTHER membership
- make clear that InterPro and PMID:16339152, not the narrow PANTHER label, establish its C-terminal glutathione-synthetase activity
- record and render the focused repair in the PSEPK batch page

## Validation
- ModuleReview LinkML validation
- module family and term validation
- `just validate PSEPK gshA`
- `just validate PSEPK gshB`
- `just validate PSEPK PP_3253`
- module and project rendering
- independent annotation-reviewer audit


cmungall:
Addressed the review: PTHR38761:SF1 is again asserted only on the first GshA/GCLC leaf. The second leaf retains Q9CM00 as an intentionally out-of-family bifunctional exemplar, grounded there by IPR040657 and PMID:16339152; module and batch prose now explain that the advisory is intentional rather than a validation requirement.

/review

PR #2913, opened 2026-09-01
Deduplicate projected GOA rows

cmungall:
## Summary
- deduplicate byte-identical rows after QuickGO annotations are projected into the repository TSV format
- preserve distinct annotations whenever an exported field differs
- cover extension-only variants and WITH/FROM distinctions with a regression test

## Validation
- `just format`
- `just pytest` (1974 passed, 63 deselected)


cmungall:
Addressed the two in-scope review suggestions in the latest commit:

- documented that extension-only distinctions collapse during TSV projection
- made the WITH/FROM regression assertion order-independent

Re-ran the wrapper checks after the changes: `just format` passed and `just pytest` passed (1974 passed, 63 deselected). The speculative pagination-overlap hardening remains out of scope because no overlap was observed and it is separate from the confirmed projection bug.

PR #2922, opened 2026-09-01
Review SCHPO scm3

cmungall:
## Summary
- refresh SCHPO/scm3 UniProt, QuickGO, and four primary publication caches
- remove three retired GOA protein-binding annotations and represent the two current IPI partners separately
- ground the core CENP-A histone-chaperone function in primary evidence and add manual reference reviews
- correct the overstatement that PMID:19217403 tested both H3 and H4 binding
- update notes, rendered HTML, and the related InterPro project row

## Research decision
No OpenScientist job was launched: accessible full-text studies directly cover the core function, centromeric localization, and zinc-dependent targeting mechanism.

## Validation
- `just fetch-gene SCHPO scm3 --force` (17 projected TSV rows from 19 QuickGO records after merged extension-only dedup fix)
- `just validate SCHPO scm3`
- `just validate-goa SCHPO scm3`
- `just validate-interpro-mappings`
- `just render SCHPO scm3`
- `git diff --check`


cmungall:
Addressed all review follow-ups in 993d41763a:

- added GO:0019237 centromeric DNA binding as an `existing_annotations` entry with `action: NEW` and direct PMID:38084929 IDA support (rather than `proposed_new_terms`, which is reserved for terms that do not yet exist)
- replaced the GO:0034080 IMP support with the direct newly synthesized CENP-A incorporation result
- refined the Cnp1 IPI replacement to GO:0042393 histone binding
- independently checked the other WITH/FROM identifier against PomBase and corrected our prose: SPCC970.12 is Mis18, not Mis16

Re-ran `just validate SCHPO scm3`, `just validate-goa SCHPO scm3`, `just render SCHPO scm3`, and `git diff --check`; all pass.

PR #2949, opened 2026-09-02
Add naked mole rat (HETGA) reviews for 8 landmark genes

cmungall:
Adds the first naked mole rat (*Heterocephalus glaber*, `HETGA`, taxon 10181) reviews to the
repository: eight landmark genes, plus a project page explaining why this species is worth
reviewing at all.

## Why this species

Across all of GOA, taxon 10181 has roughly 335,000 annotations and **exactly one** is
experimental (`Apoa1`, protein homodimerization, IDA). Every annotation in this PR is an
electronic projection — ISS from a named ortholog, Ensembl Compara, TreeGrafter/PANTHER,
ARBA/UniRule, InterPro2GO, or a UniProt keyword.

That normally makes a species un-reviewable. The naked mole rat is the exception, because it
is famous **precisely where its proteins diverged** from the mouse and human orthologs the
annotations were projected from. Each projection is a hypothesis with a literature that can
test it. `just fetch-gene-pmids` returns nothing for all eight genes, so the literature was
assembled by direct PubMed search and cached by hand (37 papers).

## Genes

**Hyaluronan axis** — the basis of the cancer-resistance phenotype.

| Gene | UniProt | Annotations |
|---|---|---|
| `Has2` | G5AY81 (Swiss-Prot) | 24 |
| `Hyal2` | A0A0P6J1Y4 | 45 |
| `Cd44` | A0AAX6R0R7 | 38 |

**Nociception** — the pain phenotype is distributed across the pathway, not localised to one channel.

| Gene | UniProt | Annotations |
|---|---|---|
| `Scn9a` | G9DCX3 | 12 |
| `Ntrk1` | A0AAX6QC09 | 24 |
| `Trpv1` | G9DCX1 | 8 |
| `Tac1` | A0A0P6JY17 | 5 |

**`Cgas`** (A0AAX6RS70), 13 annotations, all from one PANTHER ancestral node.

## Outcome

180 annotations reviewed, none left pending.

| Action | n |
|---|---|
| ACCEPT | 54 |
| KEEP_AS_NON_CORE | 58 |
| MARK_AS_OVER_ANNOTATED | 27 |
| MODIFY | 22 |
| NEW | 11 |
| REMOVE | 4 |
| UNDECIDED | 4 |

The shape of that distribution is the headline: only 4 of 180 were removed, while 85 were kept
but demoted to non-core or flagged as over-annotated. Ortholog projection into this species is
mostly **correct but unfocused** — the right molecular function arrives together with a long
tail of developmental and organ-physiology terms that no naked mole rat evidence supports.

## The findings that matter

**`Cgas` — a propagated annotation whose sign is inverted in this species.** `GO:2000042`
(negative regulation of double-strand break repair via homologous recombination) traces to an
IBD at a Euteleostomi-level node seeded by human and mouse cGAS, both of which genuinely do
suppress HR. The naked mole rat protein does the opposite. Removed, and replaced with a
positive-regulation term. The source annotation is sound; what fails is clade-wide extension.
A reproducible alignment (`Cgas-bioinformatics/`) confirms the catalytic Mg²⁺ triad, zinc
thumb and arginine anchor are all intact, and that the four divergent residues sit in the
C-terminal HhH domain, far from the active site — regulatory divergence with catalysis
untouched, which is what lets one term flip while the rest survive.

**`Hyal2` — a propagation that contradicts a curated negation in its own donor set.**
`GO:0001618` virus receptor activity is a `GO_REF:0000120` rule transfer whose donor set
includes mouse Hyal2 — and GOA separately records a curated `NOT|enables GO:0001618` IDA on
that same mouse protein. The pipeline propagated a positive assertion out of a donor set
holding an explicit experimental negation for the identical term. `GO:0046718` (symbiont entry
into host cell) exists only because the receptor annotation does, and falls with it.

**`Has2` — rat renal physiology projected onto an animal with different kidneys.**
`GO:0035810` (positive regulation of urine volume) and `GO:0070295` (renal water absorption)
come by ISS from rat, where the phenotype is a water-loading-responsive rise in medullary
interstitial hyaluronan. The naked mole rat renal medulla is the one tissue that runs against
the species' own trend, with *lower* hyaluronan than comparators. Marked over-annotated rather
than removed, with the resolving experiment named.

**`Ntrk1` — the annotation set has the evidence backwards.** ERK1/2 and p38 output through the
naked mole rat intracellular domain was measured and found reduced but present; PI3K/AKT was
never measured. GOA annotates the branch with no species data and omits the one with data. The
core receptor tyrosine kinase function is retained: the deficit is a right-shifted dose
response, not a lost function, and the receptor reaches rat-equivalent sensitisation at high
NGF. No `REMOVE` was warranted.

**`Tac1` — the "naked mole rats lack substance P" framing does not touch the gene product.**
The absence is tissue-restricted to cutaneous fibres; the peptide sequence is identical to
human and sits at the same coordinates. Substance P receptor binding kept, and the missing
neurokinin A arm added as a genuine gap.

**`Trpv1` — a well-argued negative.** The channel is unremarkable: naked mole rat cDNA
expressed in *Trpv1*-null mouse neurons gives normal heat- and pH-gated currents. The pain
phenotype is upstream. This is the outcome where the projection simply holds, confirmed
positively rather than by absence of evidence.

**`Scn9a` — the acid-insensitivity motif, verified rather than recalled.** A reproducible
script anchored on invariant flanking residues places the naked mole rat variant motif in the
domain IV P-loop against six other mammals. A late-surfacing paper qualified the mechanism, so
the original rationale was withdrawn and the motif is described as the sequence correlate of
enhanced proton block, mechanism open.

## Five proposed new terms

Each verified absent against QuickGO, including a `secondaryIds` check so a merged id is not
mistaken for a missing one:

| Gene | Proposed term |
|---|---|
| `Has2` | high molecular mass hyaluronan biosynthetic process |
| `Hyal2` | regulation of hyaluronan polymer size |
| `Scn9a` | proton-inhibited voltage-gated sodium channel activity |
| `Trpv1` | vanilloid-gated monoatomic cation channel activity |
| `Cd44` | positive regulation of basal ATF6-mediated signalling |

Three of them say something GO currently cannot: that a polymer's *length* is the biologically
important property, that a channel can be *closed* rather than opened by protons, and that a
sensor's resting set-point is distinct from the size of its stress response. On the sodium
channel: `GO:1905151` was obsoleted in April 2025 on the grounds that it "represents a
molecular function", but that molecular function term was never created.

## An expressivity gap worth discussing

For `Has2`, the naked mole rat coding sequence expressed in human HEK293 cells makes them
secrete high-molecular-mass hyaluronan (IDA), and shRNA knockdown in naked mole rat fibroblasts
reduces it (IMP). Both terms are currently ISS. There is no way to record that: the validator
rejects `action: NEW` for any term already in GOA, on the GO id alone, so *"the term is right
but the evidence code understates what is known in this species"* is inexpressible. For a
species where every annotation is electronic, that is arguably the most valuable recommendation
a review could make. Left in prose in `review.reason` for now.

## Note on deep research

Affinage is **human-only** and exits on any other species, so it was run on each **human
ortholog** and stored under a filename naming that scope
(`*-deep-research-affinage-human-ortholog.md`). It is a conserved-mechanism baseline, cited as
such in `references[].reference_review`, and no sentence of it is quoted as evidence about the
naked mole rat protein. Reviewers reported it contributed nothing to any species-specific call
— on `Tac1` about a third of its findings table describes *C. elegans*, *Candida* and rice
genes that share the symbol.

Falcon was run for the animal itself and produced a report for all eight genes. Note for anyone
scripting it: several tripped the wrapper's 600 s timeout and returned a non-zero exit code
while still writing a complete file, so check for the file rather than trusting the exit status.
Worth flagging one recall failure: on `Cgas`, falcon reported that no direct study of this
protein exists, when the *Science* paper the whole review turns on was already cached in this
repo.

## Validation

`just validate HETGA <gene>` passes for all eight, re-checked after merging `main`. Every `supporting_text` was checked
verbatim against its source, including `file:` quotes, which CI does not check. All reference
titles were checked against the cached publication metadata. `cache/go/terms.csv` adds 10 rows
and removes none.


cmungall:
The `test (3.12)` failure is **pre-existing on `main`, not caused by this PR**. Both failing tests compare a committed derived artifact against one regenerated from the current gene reviews, and both artifacts went stale earlier when gene reviews changed without the report being regenerated.

**`test_generated_sidecars_match_current_sources`** — I regenerated `projects/BIOREASON_COMPARISON/benchmark-quality.csv` from the current tree (which includes this PR) and diffed it against the committed copy. The entire difference is one hash field on a single **ARATH/COP1** row:

```
6c6
< ARATH,COP1,P43254,... ,3da735...,e1d94f8e76771a32e1755a5a0d8f0491c59c9d0de54d4cbeaf06f7381f7253c0,ec50b7...
---
> ARATH,COP1,P43254,... ,3da735...,d46ffe8b65704f5448cca410bb3d6f1527c8998dd615b4b10f6f062ba2ece057,ec50b7...
```

The file contains **0 HETGA rows** before and after regeneration.

**`test_committed_three_level_report_matches_current_reviews`** — `reports/gogpt-comparison-levels.json` covers a fixed set of 44 organisms (METEA 44, BACSU 35, ARATH 33, ECOLI 27, …). **HETGA is not among them**, and the reported diffs are all METEA genes (`mxcQ`, `pckA`, `pqqA`). It was last modified by #2801 (E. coli CnoX).

`git diff origin/main HEAD --name-only` from this branch touches **no** bioreason, gogpt, or benchmark path.

I have deliberately not regenerated either artifact here, since doing so would pull unrelated ARATH and METEA changes into a naked-mole-rat PR. They look like they want a separate housekeeping commit on `main`:

```
just refresh-bioreason-benchmark-sidecars
```

Happy to do that as its own PR if useful. Everything owned by this PR is green: `just validate HETGA <gene>` passes for all eight genes, re-checked after merging `main`.

cmungall:
Thanks — all four addressed, plus both ISS suggestions. Two of the items turned out to have a partly different underlying cause than the review assumed, so I want to be explicit about what changed and what did not.

**1. `Trpv1` cation terms — fixed as requested.** You are right, and the proposed new term contradicting its own parent was the tell. `GO:0005216` now modifies to `GO:0097604` and `GO:0099094`; both `core_functions.molecular_function` slots follow; the proposed term's parent is now `GO:0099094`. I confirmed against QuickGO that both are the cation-level children with the expected definitions. Since you flagged that `core_functions` is hard-validated, I re-ran `just validate HETGA Trpv1` locally after the change: ✓ Valid.

**2. `Cgas` — hedged, but the unsourced part was narrower than stated.** The TRIM41/p97 mechanism *is* in the `PMID:41066557` abstract, verbatim: "The changes enable cGAS to retain chromatin longer upon DNA damage by weakening TRIM41-mediated ubiquitination and interaction with the segregase P97." What the abstract never does is **name the four positions** — it says only "four amino acids". So I kept the mechanism (it is properly sourced) and hedged only the residue identities, which are now presented as a provisional alignment-derived assignment rather than established fact. I also removed the phrase "in this review" from my first attempt at the hedge, since `CLAUDE.md` bars curation framing from `description`.

**3. `Hyal2` size claim — harmonised.** It no longer asserts ">6 MDa"; it now records the dispute and the attribution to isolation procedure, matching the `Has2` treatment.

**4. `status: DRAFT` — the premise here is not right, and I have left four of the five as they are.** `status` is not a free-form field: `src/ai_gene_review/status_manager.py` computes it, and `COMPLETE` requires **no PENDING annotations *and* zero validation warnings**. `DRAFT` is precisely "no PENDING, but has warnings". Has2, Cd44, Ntrk1 and Trpv1 each carry exactly one warning (`no_deep_research_results`), so `DRAFT` is their correct computed value — I ran `just update-status` on all eight and it reports `Status mismatch: 0` for those four. Flipping them to `COMPLETE` would make the field disagree with the tool that maintains it.

`Hyal2` was a genuine mismatch and is now `COMPLETE`. Worth noting for the tool: `update-status` **reports** a mismatch and exits 1 but only **writes** when the field is absent (`needs_update = current_status is None`), so it could never have corrected Hyal2 on its own.

On the underlying warning: it fires because no annotation cites a deep-research file. The only naked-mole-rat one is the falcon report, and adding citations to it purely to clear a linter warning is the wrong incentive — especially given the `Cgas` case where falcon asserted no direct study exists. I would rather the four stay `DRAFT` and honest.

**🔵 Both ISS suggestions taken.** `Scn9a` `GO:0019233` now cites `PMID:17167479` (the human loss-of-function experiment) as `original_reference_id` with `UniProtKB:Q15858` in `supporting_entities`; `PMID:22174253` stays in `supported_by` where it belongs. Both `Tac1` NEW rows now carry `UniProtKB:P20366` as `supporting_entities`.

**On CI:** the `test (3.12)` failure is pre-existing on `main` and unrelated — evidence in my earlier comment. All eight genes pass `just validate` locally, re-checked after each change.

cmungall:
Confirming the validation point you asked about before merge.

**The `core_functions` change validates.** `just validate HETGA Trpv1` locally, after moving both `core_functions.molecular_function` ids to `GO:0097604` / `GO:0099094`:

```
Validating genes/HETGA/Trpv1/Trpv1-ai-review.yaml...
✓ Valid (with 1 warnings)
✓ All validations passed for HETGA/Trpv1
```

The one warning is the expected `no_deep_research_results`. All eight genes were re-validated after every change in this round.

**`test (3.12)` is still red, and it is still the same two pre-existing failures — not a regression.** The latest run (`33666206933`) is `2 failed, 1980 passed, 63 deselected`, and both failures are the identical stale-artifact assertions from before my fixes:

- `tests/test_bioreason_benchmark_policy.py::test_generated_sidecars_match_current_sources`
- `tests/test_gogpt_compare_levels.py::test_committed_three_level_report_matches_current_reviews`

Same two, same count, before and after this round. Neither touches naked mole rat data: regenerating `benchmark-quality.csv` from the current tree yields a one-hash diff on an **ARATH/COP1** row, and `gogpt-comparison-levels.json` covers a fixed 44-organism set whose reported diffs are all **METEA**. Zero HETGA rows in either, and `git diff origin/main HEAD --name-only` touches no bioreason, gogpt or benchmark path.

So there is no green `Build and test` to wait for on this branch — the job cannot go green until those two committed artifacts are refreshed on `main`, which is independent of this PR. I have deliberately left them alone rather than pull unrelated ARATH and METEA churn into a naked-mole-rat PR.

Happy to open that housekeeping PR against `main` (`just refresh-bioreason-benchmark-sidecars` plus the three-level report) if you would like it cleared — just say the word and I will do it as a separate change.

Agreed on `update_status_in_file` being a real tool gap rather than anything owed here; I have left it untouched.

cmungall:
Update on the red `test (3.12)`. I merged current `main` (122 commits) and re-ran everything.

**One of the two original failures is gone.** `test_committed_three_level_report_matches_current_reviews` now passes, because `reports/gogpt-comparison-levels.json` was refreshed on `main`.

**The two that remain are both inherited from `main` and are fixed by #2964:**

- `test_generated_sidecars_match_current_sources` — `benchmark-quality.csv` is stale; its two out-of-date rows are yeast `CDC37` and `IRE1`, from #2931 and #2936.
- `test_committed_argo95_has_no_deterministic_category_conflicts` — `('yeast', 'IRE1', 'GO:0005634', 'NPI_or_UNC_vs_positive_AIGR')`. #2936 changed the curated nucleus annotation to `KEEP_AS_NON_CORE` but left the SFT prediction review asserting that Ire1p is not nuclear.

Neither touches naked mole rat data, and `git diff origin/main HEAD --name-only` on this branch touches no yeast, bioreason or benchmark path.

I have put the fix in **#2964** rather than here, because resolving the second one correctly moves reported CNN/NPI statistics in the manuscript and supplement, and that should be reviewed on its own rather than buried in a naked-mole-rat PR that is already approved. With #2964 applied the full unit suite is **2045 passed, 0 failed**; on `main` it is 2 failed / 2044 passed. Once #2964 merges I will merge `main` here again and this should go green with no further changes.

**Also in this push:** `main` introduced the append-only `history/` mechanism while this PR was open, so I added the nine records `CLAUDE.md` now asks for — one per gene review plus one for the project page. All scaffolded with `just new-history` rather than hand-written, and each passes `just validate-history`. All eight gene reviews were re-validated after the merge and still pass.

Worth flagging for whoever owns CI: `.github/workflows/main.yaml` runs `just test` only `if: steps.changes.outputs.src == 'true'`, so data-only PRs never run the suite that guards their own consistency. That is why #2936 and #2931 went green while leaving these two tests broken.

PR #2960, opened 2026-09-05
Regenerate GO-GPT derived reports after the ARATH AT1G06680 review (unblocks main)

cmungall:
**`main` is red for every PR right now.** `tests/test_gogpt_compare_levels.py::test_committed_three_level_report_matches_current_reviews` and `tests/test_bioreason_benchmark_policy.py::test_generated_sidecars_match_current_sources` both fail on a **pristine `origin/main` checkout** — I verified that on a detached HEAD before assuming my own work was at fault.

## Cause

The AT1G06680 (PsbP) review merged in `6ac9ee91fa` without rerunning the derived reports it feeds.

This is the **second occurrence of the same shape** — the SlyD review did it a few days ago and was fixed by #2920 — so it looks like a recurring gap rather than a one-off. See the note at the bottom.

## What changed, and why

Regenerated with the project's own recipes (`just gogpt-overlap`, `just refresh-bioreason-benchmark-sidecars`) rather than hand-editing the numbers. Exactly one row moved, and I traced it before touching the pinned expectations:

| | before | after | why |
|---|---:|---:|---|
| `core_terms` | 0 | 4 | the completed review authors `core_functions` for the first time |
| `core_overlap` | 0 | 1 | `GO:0019684` becomes a predicted overlap |
| `post_review_overlap` | 9 | 8 | thylakoid annotation narrowed, dropping `GO:0009535` |

**GOA totals do not move** — which is the signal that this is a curation edit and not a GOA snapshot refresh. That distinction is what the surrounding audit comments in the test exist to record, so I added the corresponding note.

Three surfaces pin these numbers and all three are updated: the expected-stats block, the supplemental benchmark table (`2,761 | 848` and `1,234 | 351` at 4.0%), and the regenerated sidecars plus the overlap figure.

## Verification

Full suite: **2003 passed**, 65 deselected. Both previously-failing tests now pass.

## Worth considering separately

Two gene reviews in a week have merged without regenerating these artifacts, and in both cases the breakage surfaced as an unrelated PR's CI failure rather than on the PR that caused it. A paths-filter check on `genes/**` that runs `just gogpt-overlap` and diffs, or a note in the review workflow, would catch it at the source. I have not done that here — this PR is deliberately just the unblock.


cmungall:
Obsolete: main is green again. Upstream has since refreshed the benchmark sidecars, and independently reached the same NPI -> CNN conclusion for IRE1 GO:0005634 that I had queued in a follow-up — the curator-affirmed IDA from PMID:17035634 (Goffin et al. 2006, an 18-residue NLS recognized by importin alpha and multiple importin beta homologues) refutes the 'clear localization error' rationale.

Verified on a pristine origin/main checkout: tests/test_bioreason_benchmark_policy.py, tests/test_gogpt_compare_levels.py and tests/test_sft_prediction_review.py all pass (65 passed). Nothing here is still needed.

Standing observation, unchanged: three separate gene reviews in ~10 days (SlyD, ARATH AT1G06680, yeast IRE1/CDC37) landed without regenerating the derived artifacts they feed, and each time the breakage surfaced as an unrelated PR's CI failure rather than on the PR that caused it. A paths-filter check on genes/** that runs the regeneration recipes and diffs would catch it at source.

PR #2964, opened 2026-09-05
Reconcile yeast IRE1 SFT prediction review with its curated annotation

cmungall:
Fixes the two `test (3.12)` failures currently on `main`. Both trace to one thing: the yeast IRE1 curated annotation changed in #2936 and the BioReason benchmark artefacts were not updated with it.

## The inconsistency

#2936 changed the curated `GO:0005634` (nucleus) annotation on yeast IRE1 from `UNDECIDED` to `KEEP_AS_NON_CORE`, on the strength of an IDA from `PMID:17035634` reporting a validated nuclear localization signal whose disruption impairs HAC1 mRNA splicing.

The SFT prediction review was left behind. It still assessed the matching nucleus prediction as:

```yaml
assessment: NPI
confidence_score: 0
summary: Incorrect prediction. IRE1 is an ER transmembrane protein; it is not
  found in the nucleus. ... This is a clear localization error.
error_type: PATHWAY_CONTEXT_IGNORED
```

That premise is now contradicted by the repository's own curated annotation. The prediction was correct, and because the term is already in GOA the right category is `CNN` (correct but not novel), confidence 2, no `error_type`.

`test_committed_argo95_has_no_deterministic_category_conflicts` exists to catch exactly this, and it deliberately compares prediction reviews against the **working-tree** curated actions rather than the frozen baseline — so the design intent is that prediction reviews track curation changes.

## Everything else follows from that one reclassification

- **Regenerated the benchmark sidecars.** This also clears a second, independent failure: `benchmark-quality.csv` was stale, and its two out-of-date hash rows were yeast `CDC37` and `IRE1`, from #2931 and #2936.
- **Updated the reported CNN/NPI counts** in the project page, manuscript and supplement, because `test_publication_headlines_match_generated_metrics` requires the published numbers to equal the generated ones. The moves are all ±1: ARGO95 `681 (71.3%)` → `682 (71.4%)` CNN and `115 (12.0%)` → `114 (11.9%)` NPI, with the same ±1 propagating into the mixed-source, HF-catalogue and union tables.
- **Two hardcoded `cnn_exact_frozen_goa` expectations** moved `634` → `635`.

I have flagged the manuscript/supplement edits explicitly because they change reported statistics. They are mechanical consequences of the regenerated metrics rather than judgements of mine, but they are the part of this PR most worth a second pair of eyes.

## Verification

Full unit suite on this branch: **2045 passed, 0 failed.** On `main` the same suite is 2 failed / 2044 passed.

## Why this was invisible on main

`.github/workflows/main.yaml` runs `just test` only `if: steps.changes.outputs.src == true`. Data-only PRs such as #2936 therefore never execute the suite that guards their own consistency, and both runs went green. Worth considering whether the changed-paths filter should also trigger on `genes/**` and `projects/BIOREASON_COMPARISON/**`, since that is what these tests actually assert over.

Added a history record for the IRE1 edit per the new `history/` convention.


cmungall:
Closing as redundant — this was fixed on \`main\` in the meantime, independently and with the same resolution.

\`genes/yeast/IRE1/IRE1-sft-predictions.yaml\` on current \`main\` now has \`GO:0005634\` as \`assessment: CNN\`, \`confidence_score: 2\`, no \`error_type\`, with essentially the reasoning this PR proposed: the nucleus term is already in GOA as an IDA from PMID:17035634 and is retained by the gene review as \`KEEP_AS_NON_CORE\`.

Verified against current \`main\`: \`tests/test_bioreason_benchmark_policy.py\`, \`tests/test_sft_prediction_review.py\` and \`tests/test_gogpt_compare_levels.py\` are **65 passed, 0 failed**.

The one observation here that outlived the fix is the CI gap: \`.github/workflows/main.yaml\` runs \`just test\` only \`if: steps.changes.outputs.src == 'true'\`, so data-only PRs never run the suite that guards their own consistency. That is why #2936 and #2931 went green while leaving these tests broken for a while. Might be worth adding \`genes/**\` and \`projects/BIOREASON_COMPARISON/**\` to that filter, since those are what these tests actually assert over.

PR #2997, opened 2026-09-12
Review human AGTRAP

cmungall:
## Summary

AGTRAP's GO record is 199 rows, of which **188 are `GO:0005515 protein binding`** from nine CCSB two-hybrid releases. The interaction the gene is named for — AGTR1 — is in none of them. Two of the remaining rows are wrong for the same underlying reason, and it is not a propagation failure: it is a **gene-symbol confusion in the donor**.

**Row reconciliation:** GOA TSV 199 rows; the `fetch-gene` stub seeded 199 entries; the two key sets were compared as multisets of `(term, reference, evidence, with/from, gene-product)` before review began and are identical — nothing was collapsed here. Final file has **202** entries: 199 GOA + 3 `NEW` proposals.

## `GO:0004945 angiotensin type II receptor activity` → REMOVE

Two independent arguments, deliberately in this order so the second is not load-bearing.

**Biology.** `GO:0004945` is *"An angiotensin receptor activity that acts via Gi protein coupling and cGMP (NO) generation"*, and the qualifier is `enables` — the row asserts AGTRAP **is** an AT2 receptor. AGTRAP is 159 aa with three TM helices (UniProt FT 24-44, 56-76, 87-107), not a 7TM GPCR. The founding paper tested exactly this and excluded it:

> *"ATRAP interacts specifically with the carboxyl-terminal domain of the AT1a receptor but not with those of angiotensin II type 2 (AT2), m3 muscarinic acetylcholine, bradykinin B2, endothelin B, and beta2-adrenergic receptors."* — PMID:10358057

So the annotation names the wrong receptor subtype for a protein called "type-1", and is contradicted by direct experiment on the protein itself.

**Provenance.** The Compara donor `Q9WVK0` is mouse Agtrap — resolved, a true 1:1 orthologue (PTHR16521 has no paralogues). Going one level deeper than the donor, its `GO:0004945` is an **MGI IMP citing PMID:10425188**, *"Increased vasoconstrictor response of the mouse lacking angiotensin II type 2 receptor"*. Querying QuickGO **by reference** rather than by gene makes the mechanism explicit:

| entity | rows from PMID:10425188 |
|---|---|
| Agt (P11859) | 5 |
| **Agtr2** (P35374) | 5, incl. `GO:0004945` `enables` **IGI** ← correct holder |
| **Agtrap** (Q9WVK0) | 2, incl. `GO:0004945` `enables` **IMP** ← the confusion |

Two adjacent symbols, one of which is an angiotensin receptor. The error has since reached **three species**: rat Agtrap carries `GO:0004945` twice, by Ensembl IEA *and* by RGD ISO. One MGI edit clears four annotations — filed in `suggested_questions`.

## `GO:0008217 regulation of blood pressure` → KEEP_AS_NON_CORE

Right term, wrong evidence. `PTHR16521` has exactly **two** IBD annotations family-wide (`PTHR16521-paint.tsv`), both seeded from `MGI:MGI:1339977`, and mouse Agtrap's only experimental `GO:0008217` row is the same PMID:10425188 IMP. The conclusion survives on evidence nobody curated — *"Mean systolic BP was significantly higher in Atrap-/- mice compared with wild-type mice"* (PMID:20093357) — so this is a re-referencing request, not a deletion. Organism-level physiology downstream of AT1R restraint ⇒ non-core.

## `GO:0038166` → MODIFY to `GO:0110062`

An interpro2go mapping from `IPR009436`, i.e. from the family's **name**, asserting that AGTRAP is *part of* the angiotensin pathway. Every measurement runs the other way (PLC inhibition, reduced inositol lipids/c-fos/proliferation, elevated BP and raised surface AT1R on knockout). GO keeps regulation terms outside the process they regulate, so the direction-bearing term is `GO:0110062 negative regulation of angiotensin-activated signaling pathway`.

## `GO:0001666 response to hypoxia` → MARK_AS_OVER_ANNOTATED

Donor evidence queried, not just donor identity: rat Agtrap holds it by **one IEP** from PMID:17033689, a neonatal-brain hypoxic-preconditioning study that annotates **2 entities in all of GOA** (Agtrap, Angptl4), both IEP.

## The 188 binding rows, judged per partner

IntAct fully retrieved (`len(results) == totalElements` asserted): **617 evidences, 583 (94.5%) two-hybrid, only 23 of 213 pairs with any non-Y2H method.**

| publication | evidences | decomposition |
|---|---|---|
| PMID:32296183 (HuRI) | 276 | **92 + 92 + 92** across three sub-method terms |
| PMID:25416956 (HI-II-14) | 202 | **68 + 67 + 67**, same three |

That arithmetic is the `NbExp=3` artefact **measured** rather than assumed: UniProt's count is of sub-method vocabulary, not experiments. All 132 partner accessions resolved (121 Swiss-Prot / 11 TrEMBL — tested with `entryType.startswith("UniProtKB reviewed")`, since `"reviewed" in entryType` also matches *unreviewed*); several TrEMBL ones are partial ORFeome clones (`A0A0S2Z3E9` "FGA" **84 aa** vs canonical 866; `A0A0S2Z4D9` "GAD1" 111 vs 594; both PBX3 accessions non-canonical).

Searching **partners rather than the symbol** — the recall gap `gates_passed: True` cannot close — found orthogonal validation for exactly three, each titled for the partner:

- **PITPNC1** → `MODIFY` to **`GO:0043495 protein-membrane adaptor activity`**. PMID:21728994: the PITP domain of RdgBβ engages AGTRAP *"causing membrane recruitment"*. This is the one experiment yielding an informative MF for AGTRAP itself, which is the integral membrane protein doing the tethering. Deliberately **not** promoted to `core_functions`: one PMA-dependent experiment, no consequence of the recruitment measured for AGTRAP itself, no evidence it happens at endogenous levels — the rows now say so.
- **FAM114A1** → `KEEP_AS_NON_CORE`. PMID:35671117 starts from the published Y2H candidates and *"validated the interaction between AGTRAP and FAM114A1 in mouse heart lysates and cultured PMCFs by IP and Western blotting"*.
- **PBX3** → `KEEP_AS_NON_CORE`. PMID:35414770 co-IP, but over-expression in a cancer context and non-canonical accessions.

The remaining 178 rows + the isoform-2 self-interaction → `MARK_AS_OVER_ANNOTATED`, with per-partner reasons (28 partners are confined to the mitochondrial matrix, nucleoplasm or extracellular space and cannot reach AGTRAP's 52-residue cytosolic surface; reported alongside the 67 that *are* accessible, so the topology argument is supporting, not decisive). This matches merged **ADAMTSL5** and deliberately diverges from **ADAMTSL4**'s `REMOVE` on comparable rows — flagged rather than silent.

**The contrast is the finding:** the 23 pairs with AP-MS/co-IP/BioID support are almost all membrane proteins (CYB5D2, ENPP6, MBLAC2, TMEM47, TMEM106B, LAMP5, SLC15A3/4, TMEM63A, CFTR, PAFAH2) — a coherent neighbourhood for a 3-TM ER/Golgi protein — and **none of them is in GOA**.

## Three NEW rows

| term | evidence | why |
|---|---|---|
| `GO:0031702 type 1 angiotensin receptor binding` | IPI, PMID:12960423 | Measured, curated by UniProt on the human entry with `ECO:0000269`, given a `REGION 110..122` feature, and held by the mouse orthologue as IPI with P29754 — yet absent from human GOA. The term is already established for receptor **cytoplasmic-face** binders: of its 197 annotations over 188 gene products, 16 entities hold it by non-IEA evidence, including ARRB2, JAK2, TYK2 and ARAP1 by rat IPI. Human coverage is thin — 5 entities, only AGT and BDKRB2 by IPI. |
| `GO:0002092 positive regulation of receptor internalization` | ISS | The mechanism UniProt's FUNCTION line names. ISS not IMP because the assays are over-expression experiments whose species the abstracts do not state. |
| `GO:0141109 transporter activator activity` | ISS, PMID:27015675 | AGTRAP binds SERCA2a (MALDI-MS, co-IP, SPR) and raises its Ca²⁺ uptake — an AT1R-independent activity absent from GO entirely. Sibling of `GO:0141110`, which phospholamban holds by IDA. |

## The coverage gap is symmetric, and mouse shows it is fixable

Every finding above reads AGTRAP's own record. Reading each **partner's** record instead:

| paper | partner | annotations from that paper on the partner |
|---|---|---|
| `PMID:12960423` | human AGTR1 `P30556` | **0** of 79 |
| `PMID:11733189` | human RACK1 `P63244` | **0** of 166 |
| `PMID:21728994` | human PITPNC1 `Q9UKF7` | **0** of 28 |
| `PMID:27015675` | human ATP2A2 `P16615` | **0** of 84 |
| `PMID:35414770` | human PBX3 `P40426` | **0** of 23 |
| `PMID:35671117` | **mouse** Fam114a1 `Q9D281` | **10**, incl. reciprocal `GO:0005515` IPI with `Q9WVK0` |

Five targeted interaction papers, ten opportunities across both partners, zero human annotations. The counter-example is the informative one: MGI curated `PMID:35671117` **reciprocally** on both mouse partners, and also gave Fam114a1 `GO:0038166` by **IMP** — so human AGTRAP's only `GO:0038166` row is a family-name mapping while its own partner holds the same term experimentally, one species over.

The diagnosis is therefore a **curation** defect (the AFF4 distinction), not an over-annotation one. That coexists with 181 over-annotated rows because the screen rows and the missing targeted rows come from different places.

## Checks that came back negative (recorded, not skipped)

- **Retractions/errata/EoC**: all 21 relied-on PMIDs checked via `CommentsCorrections/RefType` on each cited article's own record *and* publication types — **none** flagged; only three benign `CommentIn`.
- **Paralogue transfer**: `PTHR16521` has no paralogues (all 5 Swiss-Prot members are AGTRAP orthologues, out of 1380 family proteins), so `WRONG_ORTHOLOG_OR_PARALOG` applies nowhere here.
- **IBA above its donor** (ACRV1 pattern): mouse Agtrap holds `GO:0005886` by IDA and the human IBA sits at the same term — no downward MODIFY warranted.
- **Dead accessions** (ACTR10 pattern): all 132 partner and 4 WITH/FROM accessions resolved to live entries with gene names.

## Verification

`just validate human AGTRAP` ✓ Valid · `just validate-changed --terms` ✓ · quote check **39/39, 0 problems** (raw `supporting_text` count == parsed, strict duplicate-key loader) · all 7 `file:` quotes `grep -F`-verified on single physical lines · `cache_lint` exit 0, **one** curie added (`GO:0141109`), zero deletions against the merge base, zero duplicate curies · no rendered HTML.

---

### Correction made after opening

The first version of this PR asserted that *all 197 holders of `GO:0031702` are angiotensinogen orthologues by IEA*. That was read off **page 1 of a paginated QuickGO result** and is false — paging the whole result (`len(results)` asserted against `numberOfHits`) shows 16 entities with non-IEA evidence, four of them cytoplasmic-face binders by rat IPI. The correction **strengthens** the proposal, and is recorded as a retraction in `AGTRAP-notes.md` rather than quietly edited away.

A second, smaller correction: `PMID:16189514`'s note said 3 binding rows where the TSV has 2 (the 3 was an IntAct evidence count including a self-pair GOA did not import), and chasing that gap showed the `GO:0042802` row's "detected once" framing was wrong — IntAct records the self-pair in three CCSB publications. Verdict unchanged; its grounds corrected to "recurrence on one platform, never tested by another method".


cmungall:
Thanks — the blocking item is a genuine inconsistency and I had seen the `109-161` numbering and failed to act on it, which is worse than not noticing. Pushed in `ea62926e`.

## 🟡 `GO:0031702` evidence code

Your premises check out: human `Q6RW13` is 159 aa with `TOPO_DOM 108..159`, mouse `Q9WVK0` is 161 aa, and the abstract's `109-161` matches the mouse length exactly. I also could not get the full text (`PMC284805`: XML restricted, HTML empty, no PDF link), so I could not settle it the direct way.

The tiebreaker neither of us used is **NLM's MeSH organism indexing**, which discriminates these papers cleanly:

| paper | MeSH organism terms | used for |
|---|---|---|
| `PMID:12960423` | **`Humans`** — no `Animals`, no `Mice` | `GO:0031702` **IPI** |
| `PMID:10358057` | `Animals`, `Mice` — no `Humans` | corroboration only |
| `PMID:11162453` | `Animals`, `Rats` — no `Humans` | `GO:0002092` **ISS** |
| `PMID:15757644` | `Animals`, `Mice` — no `Humans` | corroboration only |
| `PMID:27015675` | `Animals`, `Humans`, `HEK293 Cells` | `GO:0141109` **ISS** |

So the two codes **are** consistent once the indexing is consulted rather than the abstract alone — `12960423` is indexed as human work, `11162453` as rat work. NLM essentially always adds `Animals` when animal material is used, so its absence from `12960423` is meaningful, and it agrees independently with UniProt's `ECO:0000269` on the human entry.

That does **not** dissolve the numbering conflict, so I took your second option rather than pretending it away. `IPI` is retained on explicit deference to the curator, and the row now states the discrepancy, the likeliest reconciliation (the authors keeping the numbering of the sequence first published in `PMID:10358057` while using human constructs), that it cannot be confirmed without the full text, and the **condition for changing it**: if the full text shows rodent constructs the code becomes `ISS` with `Q9WVK0` — and only the code changes, not the term.

I also rewrote the `GO:0002092` justification, which previously said "the abstracts do not state the species". That was the weak form of the argument; it now cites the rat/mouse indexing directly.

## 🔵 1 — ISS from an ortholog holding nothing

Correct, and I did not want to fix it by quietly recoding. Both mouse-side annotations the transfers should descend from are now requested explicitly in `suggested_questions`: mouse Agtrap `GO:0141109` by IDA from `PMID:27015675`, and `GO:0002092` by IMP from `PMID:11162453`/`PMID:15757644`.

One datum in the other direction, now recorded on the `GO:0141109` row: `PMID:27015675` carries `Humans` **and** `HEK293 Cells` alongside `Animals`, so the SERCA2a *binding* measurements may have used human constructs even though the calcium-uptake and knockout work did not. `ISS` stays as the conservative reading, flagged for revisiting rather than silently upgraded.

## 🔵 2 — `GO:0043495` called the anchor MF but absent from `core_functions`

You are right that the two documents read inconsistently, and the omission *was* deliberate — the review just never said so. The four PITPNC1 rows now state it: one PMA-dependent recruitment experiment, no consequence of the recruitment measured for AGTRAP itself, no evidence it occurs at endogenous levels. Real molecular function, not a demonstrated core one, recorded where it was measured rather than promoted. I have also dropped "the anchor" framing from the PR body.

## 🔵 3 — ontology request rather than per-gene caveat

Agreed and filed. New `suggested_questions` entry asking GO to widen the `GO:0043495` differentia to cover an integral membrane protein acting as the tether, noting that `SUN1`, `SUN2` and `GPR179` already hold the term while satisfying neither clause of the current definition.

## 🔵 4 — `core_functions[1]` rests on one mouse paper

Noted, and kept, deliberately. It is the only *measured* molecular activity this protein has besides receptor binding — pull-down/MALDI-MS, co-IP and SPR for the interaction, a functional calcium-uptake increase, and a knockout relaxation phenotype — and the alternative is a gene whose core function set omits half of what has actually been assayed. The `knowledge_gaps` block states the limits (untested in human; site, stoichiometry and any phospholamban competition unknown) and the description says outright that it rests on a single mouse study, so nothing is claimed beyond that. Happy to demote it if you would rather a single-study activity never reached `core_functions`.

## Validation

`just validate human AGTRAP` ✓ · `just validate-changed --terms` ✓ · quotes **39/39**, raw `supporting_text` count == parsed under a duplicate-key-rejecting loader · all 7 `file:` quotes `grep -F`-verified on single physical lines · `cache_lint` exit 0, one curie added (`GO:0141109`), zero deletions against the merge base, zero duplicates · `existing_annotations` still 202 (199 GOA + 3 `NEW`), `suggested_questions` 6 → 8.

Every edit this round was applied through the YAML parser with an assert-anchor-present / re-verify-after guard, and checked against the **parsed** document — an earlier round's verifier compared against raw YAML and failed a correct edit because PyYAML's single-quoted style doubles apostrophes.

cmungall:
All three 🔵 items taken, in `8e5c668f`.

**1 — the four glued sentences.** Real, and self-inflicted in an instructive way: my `flat()` helper normalises whitespace, so `reason + flat(" This activity …")` silently drops the leading space, producing `").This"` in all four PITPNC1 reasons. Nothing in the repo can see that — schema validation, quote validation and reference validation all pass on prose damage. So the fix came with a guard: the verifier now scans `[.)][A-Z][a-z]` across every `reason`, `summary`, `suggested_questions[].question` and `core_functions[].description`, and the self-test mutates a real occurrence back to the broken form and requires the guard to fire. Detected 4, changed 4, asserted equal, re-checked 0 remaining.

**2 — the MeSH evidence is not reproducible from the repo.** Correct, and it is the sharper of the three points: the cached `publications/PMID_*.md` records carry title, authors and abstract but **no MeSH headings**, so a reader cannot check the table that decided an evidence code. The exact query is now recorded in the notes next to the table it produces:

```
curl -s "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed\
&id=12960423,10358057,11162453,15757644,27015675&retmode=xml" \
  | grep -o '<DescriptorName[^>]*>[^<]*</DescriptorName>'
```

The gap itself is stated there too. Capturing MeSH headings in the publication cache looks worth doing generally — organism indexing is exactly the kind of thing that settles a species question an abstract leaves open — but that is `fetch-pmid` scope rather than gene scope, so I have not touched the fetcher here.

**3 — the change-condition did not cover the partner accession.** Fixed; they move together now. The row reads: if the full text shows rodent constructs, the code becomes `ISS` with `UniProtKB:Q9WVK0` **and** the supporting entity becomes murine AT1a `UniProtKB:P29754` rather than human AGTR1 `UniProtKB:P30556` — and the term does not change either way.

`just validate human AGTRAP` ✓ · quotes **39/39** ✓ · `cache_lint` exit 0, zero deletions, zero duplicates · 202 `existing_annotations`, 8 `suggested_questions` unchanged.

cmungall:
🔵 1 taken — and it is the better catch of the two, because the command I recorded was not merely inconvenient, it was **structurally incapable of showing the claim**. The load-bearing entries in that table are per-paper *absences* (`PMID:12960423` carries `Humans` and neither `Animals` nor `Mice`), and a single efetch over five ids yields an unattributed union in which `Animals` appears — contributed by the other four papers. It read as verification and verified nothing. Same failure class as the "all 197 holders" retraction earlier in this PR, and recorded as such in the notes rather than quietly swapped.

Replaced with a per-PMID loop, **run before being recorded**, with its output committed beside it (`b44e24ac`):

```
PMID:12960423   Humans
PMID:10358057   Animals,Mice
PMID:11162453   Animals,Rats
PMID:15757644   Animals,Mice
PMID:27015675   Animals,HEK293 Cells,Humans
```

🔵 2 — agreed on both halves, and deliberately not done here. The prose guard belongs next to `src/ai_gene_review/validation/prose_panther_scan.py` with a `just` target, and MeSH capture belongs next to `etl/publication_type.py`; both are shared infrastructure whose blast radius is every gene, not this one. Flagging them as the two follow-ups rather than smuggling them into a gene PR.

`just validate human AGTRAP` ✓ · quotes 39/39 ✓ · `cache_lint` exit 0 · glued-sentence guard 0 hits · 202 `existing_annotations` / 8 `suggested_questions` / 28 `references` unchanged. Notes-only diff this round; no curation content touched.

PR #2998, opened 2026-09-12
Review human AFP: a biomarker whose measured ligands GO never recorded

cmungall:
Review of human **AFP** (alpha-fetoprotein, P02771), the fetal paralogue of serum albumin.

## The hazard, and what it turned out to be

AFP is one of the most-measured proteins in clinical medicine — second-trimester maternal serum screening, HCC surveillance, germ-cell tumour staging. Almost all of that literature measures **concentration as a readout**, which is neither an MF nor a BP. The expectation going in was over-annotation from that biomarker literature.

The opposite is true. **AFP's entire GO record is 8 annotations over 6 distinct terms, with no experimental molecular function beyond three bare `protein binding` IPI rows and no experimental biological process at all** — while two solution-binding studies on protein purified from human tissue, a 2024 cryo-EM structure with four fatty acids and a metal ion resolved, six ECO:0000269 FAM20C phosphosites, and two human null-allele reports have produced no GO annotation between them. The diagnosis is **under-annotation with exactly one over-annotation**.

## The ALB check the campaign keeps recommending — it fires here

Both IBA rows cite `PANTHER:PTN000147344` with **human serum albumin `UniProtKB:P02768` among the seeds**. All nine WITH/FROM tokens resolved: ALB from four species, human GC and AFM, rat Alb and rat Gc (via the RGD REST API; QuickGO rejects RGD ids in `geneProductId` with HTTP 400), plus the tree node. **Zero alpha-fetoproteins, from any species, are seeds on either row** — everything AFP inherits phylogenetically comes from its paralogues.

Family verified rather than assumed: **PTHR11385 "SERUM ALBUMIN-RELATED"**, cached `proteins: 2724`, of which the entries CSV holds **43 reviewed (Swiss-Prot)** members — i.e. 1.6% of the family, not the family. Four human members, one per subfamily: ALB/SF15, **AFP/SF7**, GC/SF11, AFM/SF14.

## `REMOVE GO:0031667 response to nutrient levels`

Following "go one level deeper than the donor". The four non-node donors yield exactly **three** experimental papers, and all three were read in full:

| donor evidence | what the paper actually measured |
|---|---|
| human ALB `GO:0009267` IDA, `PMID:16245148` | recombinant HSA expressed from the **rice αAmy3 sucrose-starvation promoter**. The starvation response is the *plant promoter's*; albumin is the cargo protein. |
| human ALB `GO:0072732` IDA, `PMID:16153637` | albumin at **29–49 mg/ml** added to serum-free medium against dotarizine/flunarizine/thapsigargin/cyclopiazonic acid, i.e. **drug-induced ER Ca²⁺ release**. Not a nutrient level. |
| rat Alb `GO:0007584` **IEP**, `PMID:20227002` | **serum albumin concentration** falling with fetal protein malnutrition and recovering — albumin as a nutritional-status biomarker. |

Every other donor row is an **ISS/ISO copy of the same two human ALB IDAs** (chicken ALB's only own IDA in the branch is `GO:0033189` response to vitamin A), so the apparent breadth of the donor set is one pair of experiments recounted. Human AFP is dispensable: two unrelated families with null alleles are asymptomatic with normal fetal development (`PMID:15280901`, `PMID:18854864`).

Note the shape: **the concentration-is-not-a-function fallacy is sitting inside the donor set**, not in AFP's own literature — and it has propagated to every member of PTHR11385. Filed as a node-level PAINT recommendation, stated once.

## `ACCEPT GO:0036094 small molecule binding` — general for the right reason

Checked against the AADACL4 corrective before reaching for a granularity call. The donors do **not** agree: ALB holds fatty acid (IDA ×2), pyridoxal phosphate, oxygen, enterobactin and copper; GC holds vitamin D; AFM holds vitamin E. `GO:0036094` is the genuine LCA. PAINT's own design confirms it — **general term at the family root, specific ligands on the three subfamily nodes** (vitamin E→afamin, vitamin D→GC, enterobactin→albumin).

(QuickGO ancestry checked rather than recalled: `GO:0005507` copper ion binding *is* a descendant of `GO:0036094`. I had that wrong from memory.)

## The reciprocal node question — and it finds something

Asking "which node's reach is exactly my gene, and what did it give it?": **`PTHR11385:SF7`, the alpha-fetoprotein subfamily, carries nothing at all.** AFP is the only one of the four human members that inherits no ligand term, from a subfamily node that exists and is empty. And `GO:0005504` **fatty acid binding sits at no node in the family**, though ALB holds it by two IDAs, rat Alb by IDA, mouse Afp by UniProt function, and human AFP by the structure below. That is the property that most clearly transfers family-wide, and it is the one PAINT has not placed anywhere.

## `MODIFY` both HCV rows → `GO:0140272 exogenous protein binding`

`UniProtKB:Q99IB8-PRO_0000045596` resolves to **Envelope glycoprotein E2, residues 384–750** of HCV genotype 2a JFH-1 — a *different species'* protein. `GO:0140272` is defined for exactly this, and the paralogue **ALB already holds it by IDA**, so this aligns AFP with how the family is curated. Applied to both rows, which are the **same** co-IP recorded by IntAct and by AgBase — kept separate precisely so the redundancy is documented rather than hidden, since two GOA rows for one experiment read as two supports in any tally.

## Three `NEW` molecular functions from measurements that already existed

- **`GO:0005504` fatty acid binding** (`PMID:38678117`, IDA) — cryo-EM 8X1N resolves four fatty acids in subdomains IIA, IIA/IIB, IIIA and IIIB, and the bound species were identified by **GC-MS**, not modelled from density shape: palmitate at 57.42% of total FA, stearate the other major peak. Retained through purification rather than soaked in.
- **`GO:0005507` copper ion binding** (`PMID:80265`, IDA) — 1 mol Cu(II)/mol on protein purified from **human cord serum and hepatoma ascites**, abolished by photooxidation of histidines, reconstituted by a synthetic N-terminal undecapeptide. **UniProt has carried this as `FT BINDING 22 /ligand="Cu(2+)" /evidence="ECO:0000269|PubMed:80265"` for decades with no GO annotation** — and the 2024 structure independently places a metal site on that same His22.
- **`GO:0046872` metal ion binding** (`PMID:38678117`, IDA) — tetrahedral site at His22/His264/His268/Asp280, metal acquired from the medium rather than added. **Revised in round 2 from `GO:0008270` zinc ion binding after review**: the paper's own ICP-MS assay on the `AFP-4mut` (H22A/H264A/H268A/D280A) mutant shows *magnesium, aluminium, nickel and zinc* falling together, and the authors conclude the four residues "possess the capacity to bind to different types of metal ions"; zinc entered the model only because they "selected Zn2+ fit into the density map". That is positive evidence for promiscuity, and weighing a deposited ligand identity above the paper's own measurement was the wrong way round. Copper keeps its specific term because copper is the one metal with a measured stoichiometry and a mapped site — **specific where a specific metal was measured, general where only "a metal" was.** The same ICP-MS result supplies the first independent support on the human protein for UniProt's otherwise unsourced **nickel** claim.

**`PMID:38678117` is absent from GOA *and* absent from the affinage report** — the campaign's recall lesson in concrete form.

## Proposed GO term: **bilirubin binding**, under `GO:0046906 tetrapyrrole binding`

GO has `GO:0015723` bilirubin transport and `GO:0015127` bilirubin transmembrane transporter activity but **no MF for binding bilirubin**. So a 1:1 measurement with Kd 2.6–7.4 × 10⁻⁷ M on human AFP by two independent methods (`PMID:89900`) is inexpressible, as is UniProt's statement that AFP binds bilirubin less well than albumin. The term would immediately serve ALB too.

## Checks that came back negative — reported, not skipped

- **The predicted rodent-estrogen leak has not happened.** Mouse *Afp* carries five knockout IMP rows (`GO:0001542` ovulation, `GO:0042448` progesterone metabolic process, and three others) and **none has been propagated to human AFP**; mouse Afp carries no estrogen-binding MF either, so there is nothing to transfer. Correct curation, given that the mouse female-infertility phenotype is estrogen-driven (`PMID:12297623`) and human AFP has essentially no estrogen binding.
- **GPC3 is not one screen counted three ways.** The three IntAct records are a HepG2 co-IP, an *in vitro* co-IP, and a **pull-down typed `direct interaction`** at MI-score 0.59 — the highest of AFP's 21 records — with purified Fc-AFP against GPC3-His and **MUC16-His as a negative control**.
- **Retraction/erratum/expression-of-concern sweep on all 29 candidate PMIDs: all clean**, reading both `PublicationTypeList` and `CommentsCorrections/RefType` (a Publisher Correction is invisible to a pubtype query).
- **GOA's partner curation is good and is said so.** IntAct holds **21 interaction records over 16 distinct partners**; GOA imported exactly the **2** whose own papers validated them outside the screen, and not the other **14**. Two Y2H screens supply eight of those: `PMID:21988832` contributes seven (AP4S1, PIDD1, MED27, EHD4, PHB2, PSMB7, GNB1) and `PMID:16169070` contributes SGSM2 — all cytosolic, nuclear or proteasomal, i.e. topologically inaccessible to a constitutively secreted signal-peptide protein, and all assayed in a yeast host. The remaining six are large-scale AP-MS hits with intracellular baits.
- **A number I could not reproduce was a defect, and it was mine.** I first wrote "12 distinct partners" and "the other nine", both counted by eye off a printed list; the true figures are **16** and **14**. Corrected in the second commit, recorded in the notes rather than silently patched, and the census is now **computed** by `audit_afp_claims.py --partners` so it cannot drift again. The review YAML was unaffected — it says "twenty-one interaction records" and "seven yeast two-hybrid partners", both correct.

## GPC3 → `MODIFY` to `GO:0043394 proteoglycan binding` (revised in round 2)

I first kept bare `GO:0005515` here, arguing that because a **glypican-3 mutant lacking heparan sulfate chains still precipitates AFP**, binding is to the core protein and `GO:0043394` would misdescribe it. **That confused two different objects.** QuickGO gives `GO:0043394` as *"Binding to a proteoglycan, any glycoprotein in which the carbohydrate units are glycosaminoglycans"* — the object is the glycoprotein — and GO has separate terms for the sugar moiety (`GO:0008201` heparin binding, `GO:1904399` heparan sulfate binding). So the ΔHS result rules out a *GAG*-binding term and leaves `GO:0043394` intact: the chains are **not required**, which is not the same as the partner not being a proteoglycan. GPC3 is a GPI-anchored HS proteoglycan, and both the HepG2 co-IP and the membrane colocalisation used native glycanated protein. Still non-core in substance, since no functional consequence is measured anywhere in the paper.

The generalisable error, recorded in the notes: **I let an experiment about *which part* of a partner is contacted decide a term about *what the partner is*.**

The cached record said `full_text_available: false`; **the flag was stale** — PMC11737099 had the full text all along, and this PR commits the refreshed copy (403 → 12,229 words, abstract preserved).

## `GO:0005737 cytoplasm` — right term, wrong provenance

GOA records it as an Ensembl Compara projection from mouse *Afp*, whose whole basis is one MGI IDA (`PMID:8607965`) — a paper whose own abstract says *"Only SA protein was detectable by immunostaining"* and that AFP "was not detected during this period". Only the abstract is cached, so this is recorded as **support that cannot be confirmed from the available record, not as a curator error**, and the mouse IDA is not second-guessed. The term survives on human evidence GOA does not cite (`PMID:33009373`, confocal colocalisation in HuH7/HepG2), so: `KEEP_AS_NON_CORE`, kept for a reason GOA does not give, with the recorded provenance flagged.

## Row reconciliation and mechanical gates

`AFP-goa.tsv` has **8** data rows; the `fetch-gene` stub seeded **7**, having collapsed the IntAct and AgBase records of one co-IP. Restored to 8, plus **3** `NEW` = **11** entries.

Committed as `genes/human/AFP/AFP-bioinformatics/audit_afp_claims.py`, which rebuilds `supporting_entities` **from the GOA WITH/FROM column** rather than by hand, asserts row-for-row coverage, verifies all **71** `supporting_text` quotes including the **5** `file:` quotes CI does not check, requires UniProt quotes to sit inside one physical line, rejects duplicate YAML keys, sweeps **13 retracted phrasings** across both the review and the notes, asserts **6 required corrections** with measured minimum counts, and reconciles the notes' own quote-count sentence against its derived value. Its **sixteen** named guards are each exercised by their own `--self-test` mutation, including negative controls that must *not* fire (the legitimate narration in the notes, and a run over the unmodified document); `RESULTS.md` is regenerated by the script, and a fresh run was diffed against the committed copy.

Two numbers that did not add up were investigated rather than explained away, per the campaign rule:
- 34 `reference_id:` lines against 63 parsed quote sites — **not** a bug: 34 in `supported_by` + 29 in `references[].findings[]`, which inherit the parent id. The checker was printing the wrong invariant and now asserts on `supporting_text` line count instead.
- 178 insertions / **28 deletions** in the refreshed publication — the deletions are re-wrapping of the same abstract, verified by grepping two distinctive sentences.

`just validate human AFP` is ✓ Valid on the tree merged with `origin/main`. `cache_lint` exits 0 and this branch touches `cache/` not at all. No rendered HTML committed.

### One warning left standing deliberately

`No annotations reference available deep research files`. The affinage record is added to `references[]` with a `reference_review` as CLAUDE.md requires, but **nothing in the review rests on it and no sentence from it is quoted**, so citing it from an annotation to silence the warning would be inventing content. Its recall was measured rather than asserted: **17 citations, 0 of which appear in AFP's GOA record and 1 of which is cited here**. Its gates passed and all 17 ids are numeric PubMed identifiers, so its precision is fine — the failure is recall and framing. Much of its narrative is the transcriptional regulation of the *AFP gene* (FOXM1, HBP1/HBx, gp96/NR5A2, the −119 HNF1 variant), which are functions of those regulators rather than of AFP, and the rest is HCC cell biology in which AFP concentration is the readout. **Ligand binding — what the protein actually does — is absent from it entirely.**


cmungall:
Both 🟡 items conceded and fixed in `4e15b1b`, plus all four 🔵 suggestions. One suggestion I checked and did not take, with the premise below. Everything re-verified against the sources before changing.

## 🟡 1 — zinc → `GO:0046872 metal ion binding`. You're right, and I should have found this.

Verified in the cached full text:

> We mutated four amino acids of AFP to alanine, obtaining the mutant AFP-4mut (H22A, H264A, H268A, and D280A). Following this, we assessed the binding of metal ions to both the wild-type AFP and the mutant AFP-4mut protein using inductively coupled plasma-mass spectrometry (ICP-MS). The findings indicated a reduction in the molar ratio of metal to protein in the mutant AFP-4mut, specifically for elements such as magnesium, aluminum, nickel and zinc […] This implies that the identified four amino acids in AFP possess the **capacity to bind to different types of metal ions**.

and `"After we selected Zn2+ fit into the density map"`.

Your framing is the part that actually changed my mind: this is **positive evidence for promiscuity**, not merely an absence of evidence for specificity, and my stated reason ("the deposited structure commits to zinc") weighed a modelling decision above the paper's own measurement. Changed to `GO:0046872`, with the ICP-MS and "selected Zn2+" sentences now in `supported_by`, and `core_functions[1]` updated.

**How I missed it is the uncomfortable part**: I had this full text, quoted from it five times, and never read the Results paragraph that tested the very site I was annotating. That's the ACBD3 lesson landing on me — the answer was inside a paper I was already citing.

Two things it **strengthens**, now said explicitly rather than left implicit:

- **`GO:0005507` copper is untouched and the asymmetry is now principled.** Copper rests on `PMID:80265` — equilibrium-dialysis stoichiometry on protein purified from human cord serum and hepatoma ascites, site mapped chemically by histidine photooxidation and reconstituted from a synthetic N-terminal peptide. *Specific where a specific metal was measured; general where only "a metal" was.*
- **Your 🔵 2 is right and I've folded it in.** Nickel is among the four metals lost in the 4mut, so my own `suggested_questions` entry about UniProt's unsourced nickel claim is partly answered by a paper I was already citing. The question now says so, and the annotation reason notes it as the first independent support on the human protein.

## 🟡 2 — `GO:0043394`: definition checked, you're right, now a `MODIFY`

QuickGO gives `GO:0043394` as *"Binding to a proteoglycan, any glycoprotein in which the carbohydrate units are glycosaminoglycans"* — the object is the **glycoprotein**, and the GAG-moiety terms exist separately exactly as you said (`GO:0008201` heparin binding, `GO:1904399` heparan sulfate binding, both confirmed non-obsolete).

So the ΔHS result rules out a *glycosaminoglycan*-binding term and leaves `GO:0043394` untouched: the chains are **not required**, which is not the same as the partner not being a proteoglycan. GPC3 is a GPI-anchored HS proteoglycan, and both the HepG2 co-IP and the membrane colocalisation used native glycanated protein. Now `MODIFY` → `GO:0043394`, still non-core in substance since no functional consequence is measured anywhere in the paper.

Named the generalisable error in the notes: **I let an experiment about *which part* of a partner is contacted decide a term that is about *what the partner is*.**

## 🔵 3 — `MISCITED` semantics: agreed, moved to `VERIFIED`

You're right that the field records the reference's status *in this document*. This review cites all three correctly, precisely to characterise what they measured; the miscitation is GOA's and PAINT's. All three are now `VERIFIED`, with `review_notes` opening on why they were reclassified so the distinction isn't lost.

## 🔵 4 — first REMOVE argument demoted to an observation

Agreed. It now reads as an observation explicitly flagged as not load-bearing ("worth noting but it is not itself a defect, because an IBA is a judgement about a node rather than a pairwise transfer, and AFP descends from `PTN000147344` whichever members happened to seed it"), and the removal is carried by the donor-evidence collapse and the two asymptomatic human null families. "Three independent lines converge" is gone.

## 🔵 1 — checked, and respectfully not taken

> The paper says only "This motif provides AFP with strong metal-ion binding capabilities" — no prior affinity comparison is cited anywhere in the review.

The paper states the comparison **twice**:

> Furthermore, AFP has a higher affinity for Zn2+ than HSA25.

> Previous studies have demonstrated that AFP exhibits a higher affinity for Zn2+ than HSA25, which can be attributed to structural differences, including the presence of an additional histidine residue involved in binding.

(`grep -cF` returns 2 for each against `publications/PMID_38678117.md`.) It is the paper's citation to its own ref 25 rather than a measurement made there, which is a real caveat and is now stated. So the fair half of your point — that I asserted it without quoting it — is taken: the quote is now in `core_functions[1].supported_by` and the knowledge gap says plainly that it is a citation to earlier work. The claim itself stands.

## On your two deferred verifications

Both were load-bearing, so here they are with sources, since you had no network:

- **`GO:0005507` under `GO:0036094`**: QuickGO `/ontology/go/terms/GO:0005507/ancestors?relations=is_a,part_of` returns `['GO:0003674', 'GO:0005488', 'GO:0005507', 'GO:0036094', 'GO:0043167', 'GO:0043169', 'GO:0046872', 'GO:0046914']` — `GO:0036094` is present. (`GO:0005504` likewise.) This is the memory error I self-corrected: I had ion binding filed as a *sibling* of small molecule binding, not a descendant.
- **ALB holding `GO:0140272` by IDA**: QuickGO annotation search for `UniProtKB:P02768` + `GO:0140272` returns exactly 1 hit — `IDA`, `PMID:16394536`, `assignedBy: UniProt`.

One more verification worth recording since it is the other half of the PAINT argument and I only asserted it from the cached file the first time: querying QuickGO directly confirms **AFM receives `GO:0008431` vitamin E binding by IBA from node `PTN000147369`, and GC receives `GO:0005499` vitamin D binding by IBA from `PTN002604574`**, while AFP's only two IBA rows both cite the family root `PTN000147344`. So "AFP alone inherits no subfamily ligand term" is confirmed against live annotations, not just the cached PAINT export.

## Also in this round, self-caught before your review

I had written "12 distinct IntAct partners" and "the other nine" — both counted by eye. The true figures are **16 partners over 21 records**, 2 in GOA and 14 not. Fixed in `d9c0418`, recorded in the notes rather than silently patched, and the census is now **computed** by `audit_afp_claims.py --partners` so it cannot drift again. It also completed the screen accounting: two Y2H screens, not one.

`just validate human AFP` is ✓ Valid (the one warning left standing deliberately, as you noted approvingly); the committed audit's four guards all still fire under `--self-test`; and all **25** inline `[PMID: "quote"]` citations in the notes were re-verified verbatim, since nothing in CI checks those.

Thank you — the zinc catch in particular is a genuine improvement to the annotation, not just to the prose.

cmungall:
Rounds 3–6 are in. One consolidated reply rather than four, since each round's item came from the previous round's fix.

## The round-5 clause was wrong, not just imprecise — thank you for catching it

> AFP-4mut is H22A/H264A/H268A/D280A and His22 is the residue this review has established is shared between the two sites, so the mutant ablates part of the amino-terminal site too.

Correct, and it is worse than a wording slip: I had *just* argued that His22 is the shared residue, then in the same paragraph proposed a mutant that removes it as the way to tell the two sites apart. The control I named would have destroyed the thing it was supposed to isolate.

Fixed to a **His264A/His268A/Asp280A triple retaining His22** — metal still bound is amino-terminal, metal lost is tetrahedral — in the knowledge gap, and rebuilt through the suggested experiment's `hypothesis`, `description` and `experiment_type`, since that is the place where the error would actually mislead someone at the bench.

Your second item followed from the same thing and is now stated rather than papered over: because AFP-4mut removes His22 along with the other three, **that experiment is silent on which site binds nickel.** The `suggested_questions` entry said "the amino-terminal site does engage nickel"; it now says a metal site does, and that which one is unresolved.

## On the pattern, since this is the third round of it

Rounds 3, 5 and 6 were each "one claim retracted at N−1 sites". The through-line: I kept verifying *the edits I had made* instead of sweeping *the document* for the claim I had retracted. Changed the method accordingly — round 6's check is a whole-document sweep over **seven** retired phrasings (reporting 0 remaining) plus a positive assertion that the corrected construct appears at all three sites that need it, rather than a check on the three hunks I touched.

Round 5 also produced a small object lesson in the same class: its own guard asserted `>=4` occurrences of the exact string `"separable"` and failed at 3, because the document legitimately also uses `"separability"`; and it guessed `4` for `"undecapeptide"` where the supporting quotes make it **10**. Both thresholds were hand-assigned rather than derived. Matched on the stem now, with counts printed rather than asserted blind.

## Rounds 3–4 recap

- **Parent/child redundancy** answered in both rows, and then answered *better* after your round-3 suggestion: the two are not really a parent/child pair in substance. `PMID:80265` reconstitutes Cu(II) binding with a synthetic amino-terminal **undecapeptide** — 11 residues, precursor 19–29 — which contains His22 and cannot contain His264/His268/Asp280, so copper binding does not require three of the four tetrahedral residues; and `PMID:38678117` independently infers "metal binding at multiple locations" from molar ratios above 1. Limit kept explicit: sufficiency of a peptide does not establish the sites are disjoint in the full-length protein, so the claim is **separability and a shared His22**, not independence.
- **`GO:0016151` nickel cation binding** named and declined, with the reason (a relative molar-ratio reduction in a supplementary panel is not a stoichiometry or an affinity).
- **Averaging-caveat quote** restored to `supported_by`; the fix script now asserts the reason still makes that argument before restoring the quote, so the two cannot drift apart again.

Your independent check of the residue arithmetic — `RTLHRNEYGIA`, and pinning the numbering convention through Asn251's N-X-T sequon so the paper and UniProt are demonstrably counting the same way — was a better verification of that argument than the one I gave for it. That sequon check is the part I would not have thought to do.

## State

`just validate human AFP` ✓ Valid on the merged tree (the one warning left standing deliberately). `cache_lint` exits 0 and this branch touches `cache/` not at all. The committed audit's four guards all fire under `--self-test`; all inline `[PMID: "quote"]` citations in the notes re-verified verbatim. No term id, action, qualifier, `supporting_text` or `core_functions` molecular function changed in round 6 — the annotation itself has been stable since round 4.

Not merging or enabling auto-merge; a coordinator gates that.

PR #2999, opened 2026-09-12
Review human AHNAK

cmungall:
## Summary

AHNAK's GO record is the opposite shape to its literature. The 66 GOA rows break down as
**31 cellular component** (16 distinct terms), **31 molecular function of which 26 are bare
`GO:0005515`**, and **4 biological process covering 2 terms** — one of which this review marks
over-annotated. So a ~700 kDa scaffold with hundreds of papers has effectively **no curated
biological process and no informative molecular function**. The diagnosis is a *coverage* defect,
not an over-annotation defect — though there are four specific over-annotations, and each one
turned up by resolving provenance rather than by reading labels.

**Coverage reconciliation:** GOA TSV 66 rows → `existing_annotations` 66 + 8 `NEW` = 74.
The `fetch-gene` under-seeding defect (collapsed `GO:0005515` rows) seen on ADAMTSL5 and ACTR5 is
**absent here** — all 26 protein-binding rows were seeded individually. Verified programmatically.

### 1. `GO:1905686` is a ComplexPortal projection sitting on top of a negative result

The row is NAS from `PMID:22940583` — a **3.5 Å crystal structure** that performs no
membrane-repair experiment; repair appears only in its framing sentence, which is why NAS is the
right code. Querying QuickGO *by reference* rather than by gene:

```
reference=PMID:22940583 -> 16 annotations, 10 distinct entities, complete
  GO:1905686 NAS -> 10 entities  (CPX-850/853/898/905 + ANXA2, S100A10, AHNAK in 2 species)
  GO:0045921 NAS ->  6 entities
```

All `assignedBy: ComplexPortal`, identical evidence code. The phenotype spreads with the complex —
the ACTR8 signature. The ACTRT3 discriminator ("does the functional term stay on the gene that was
perturbed?") cannot even apply, because nothing was perturbed.

**And the only experiment that ever tested it is negative:** *"A laser wounding assay with
AHNAK1-deficient fibers suggests that AHNAK1 is not involved in membrane repair."* (`PMID:20833135`).

→ `MARK_AS_OVER_ANNOTATED`, **not** `REMOVE`: one assay, one tissue, one species, hedged by its own
authors, and AHNAK's partners (dysferlin, ANXA2, S100A10) genuinely are repair proteins.

### 2. A plasma-membrane IDA whose reference predates the gene

`GO:0005886` IDA cites `PMID:3126079` — Osborn et al. **1988** on p11 and p36 (S100A10/ANXA2). It
contains no AHNAK experiment and could not: the AHNAK gene was first reported in **1992**
(`PMID:1608957`). Same projection mechanism, 8 entities.

The **term is correct** for AHNAK on three other rows, so this is a *reference* defect, not a term
defect → `ACCEPT` with the provenance recorded and `reference_review.correctness: MISCITED`, plus a
question for ComplexPortal.

### 3. One holdup screen supplies 8 of the 26 protein-binding rows

IntAct expansion of Q09666: **649 records, 608 distinct partners**. Of those, **415 come from
`PMID:36115835` alone**, every one `detectionMethod: holdup assay`, AHNAK always the **prey** — its
isolated ~80-residue PDZ domain against synthetic C-terminal 10-mers from a 448-peptide library.
The ACRV1 `NbExp=3` pattern, one order of magnitude up.

Judged **per partner**, not wholesale:

| partner | orthogonal record | action |
|---|---|---|
| CFTR | co-IP + BioID | KEEP_AS_NON_CORE |
| HSPA5, U2AF2, MCM7, ITK, CTBP1, COMMD3 | one each | KEEP_AS_NON_CORE |
| **RPS6KA1** | **none in IntAct**; MI-score 0.44, lowest of the eight | **MARK_AS_OVER_ANNOTATED** |

The flip side is a genuine positive: the screen is functional evidence that AHNAK's
`ECO:0000255`-predicted PDZ domain really folds and binds PDZ ligands.

### 4. `GO:0043484` is one experiment reaching 74 proteins by three routes

Donor-evidence query across the family: the **only** experimental annotation to this term anywhere
is a single mouse Ahnak IDA (`PMID:21940993`). Mouse/rat/human Periaxin and human AHNAK2 all hold
it by IBA alone. PANTHER node `PTN001156025` gives it to **74 gene products**, and human AHNAK
receives it **three times** — IBA, Compara IEA, CAFA ISS.

Two things make the PAINT call *more* defensible than it first looks, and both are recorded: the
paper tests **both** AHNAK and Periaxin, so the node has two members behind it; and the node
placement is clean (all three human members get exactly the same three terms from the same node —
no AADACL-style misplacement). What is wrong is the **level**: the activity belongs to the short
nuclear isoform **Q09666-2** (residues 1–149, essentially the PDZ domain), not to the 700 kDa
protein the canonical accession represents. All three rows → `KEEP_AS_NON_CORE`.

### 5. The CAFA ISS block cites one paper for five different terms

Five ISS rows all cite `PMID:21940993`, but the donor's own evidence for two of them is elsewhere:

```
E9Q616 -> GO:0043034 costamere  : IDA PMID:20833135   <- row cites 21940993
E9Q616 -> GO:0097493 elasticity : IMP PMID:20833135   <- row cites 21940993
E9Q616 -> GO:0005829, GO:0043484: IDA PMID:21940993   <- agree
```

`PMID:21940993` is about alternative splicing at the AHNAK locus and contains neither experiment.

### 6. `GO:0097493` is a molecular function inferred from organ mechanics

Its whole basis is one mouse IMP: AHNAK1⁻/⁻ fibres are stiffer by AFM. **Careful on direction** — I
initially wanted to call this a sign inversion and it is **not**; loss of AHNAK increasing stiffness
is *consistent* with AHNAK providing compliance. The objection is about **level**: a whole-fibre
knockout phenotype read as a molecular property, with no force spectroscopy, extensibility
measurement or reconstitution ever performed. Both rows → `MARK_AS_OVER_ANNOTATED`.

## Checks that came back negative — reported, not omitted

- **Stub under-seeding**: absent (66:66 exactly; all 26 binding rows individual).
- **Partner-accession substitution** (the ACRV1 truncated-ORFeome trap): all **15** distinct
  `GO:0005515` partner accessions resolve to **reviewed canonical Swiss-Prot entries at full
  length** (the HCV row correctly uses the mature-chain id `P27958-PRO_0000037576`). Scoped to
  *partners* deliberately — several WITH/FROM **donors** are TrEMBL, and conflating the two sets
  produced a false "19 of 19" in the first commit.
- **Paralogue-only support**: `GO:0005737` has AHNAK2 and mouse Ahnak2 in its WITH/FROM, but both
  carry their own experimental evidence, so it is not a sibling-only inference.
- **PAINT node misplacement**: none; node reach and term assignment are consistent across AHNAK,
  AHNAK2 and PRX.
- **Retractions**: none across the 58 PMIDs touched. **5 errata** found by reading
  `CommentsCorrections/RefType` on each *cited* article (a Publisher Correction is invisible to a
  pubtype search); all Author Corrections, all recorded, none load-bearing.

### The T-tubule row that looked like a paralogue artefact and was not

`GO:0030315` NAS from `PMID:17185750` ("dysferlin and AHNAK colocalize at the sarcolemmal membrane
and T-tubules") appears flatly contradicted by `PMID:20833135` ("no AHNAK expression was detected in
the T-tubule system"), which used AHNAK1- and AHNAK2-**specific** antibodies — a textbook
cross-reactivity story, and I was heading for `MARK_AS_OVER_ANNOTATED`.

Then a third paper: `PMID:12153988` localises the AHNAK C-terminal portion to the T-tubular system
of **human left ventricle** by confocal microscopy. The disagreement is confined to *skeletal*
muscle. It is a tissue difference, not a contradiction → **`ACCEPT`** with both observations
recorded. Logged in the notes as a case where the campaign's over-annotation prior nearly produced
a confident wrong answer.

## 9 `NEW` rows for what GOA does not record

| term | evidence |
|---|---|
| `GO:0005246` calcium channel regulator activity | Cavβ2 Kd 55–60 nM, PxxP motif, brake on I(CaL); human I5236T variant mimics PKA and occludes isoprenaline |
| `GO:1901385` regulation of voltage-gated calcium channel activity | fragments act in WT but **not** ahnak1-null cardiomyocytes; surface delivery in T cells, neurons, osteoblasts |
| `GO:0003779` actin binding | binds G-actin, cosediments with F-actin |
| `GO:0051017` actin filament bundle assembly | bundles purified filaments into paracrystalline arrays by EM |
| `GO:0030674` protein-macromolecule adaptor activity | one region binding Cavβ2 *and* actin; the A2t motif contacting both protomers at once — the child term, since its definition explicitly covers protein+protein and protein+lipid |
| `GO:0030866` cortical actin cytoskeleton organization | reciprocal siRNA separating the two arms |
| `GO:0032460` negative regulation of protein oligomerization | human AHNAK-KO U2OS: 53BP1 hyper-accumulation, enlarged NBs, enhanced phase separation |
| `GO:0005080` protein kinase C binding | the only annotation anywhere touching the **central repeat region** (~¾ of the protein) |
| `GO:0008047` enzyme activator activity | AHNAK stimulates LIG4-XRCC4 double-stranded ligation, purified proteins + in-cell co-IP |

The last of these was nearly missed, and how it surfaced is the point. `PMID:15177040` entered
this review only as the source of the "weak DNA-binding activity" caveat on the `GO:0003723` row —
**cited in prose, supporting no annotation**. Re-reading it for that caveat surfaced the actual
result. A cited reference that supports no annotation is a flag, exactly as an empty `findings:`
list on a full-text reference is. Term choice was constrained and both alternatives checked rather
than assumed: `GO:0051106` positive regulation of DNA ligation is **obsolete** at QuickGO, and
`involved_in GO:0006303` would over-reach with no cellular repair assay anywhere — so the activity
is annotated and the pathway role is filed as a knowledge gap. That gap is stark: AHNAK now has two
unconnected nuclear biochemical activities, and nobody has run an NHEJ assay on an AHNAK-null
background even though the U2OS knockout line exists.

Two `MODIFY`s move the S100A10 `GO:0005515` rows to `GO:0044548` — the one partner on this protein
with two crystal structures behind it. `GO:0005515` on DYSF/myoferlin was left alone **deliberately**:
it is the strongest IPI on the gene (direct, Ca²⁺-independent, domain-mapped both ways) but GO has
no term for binding a ferlin, and `GO:0019904` would be true and no more informative — filed as an
ontology question rather than forced.

One `proposed_new_terms` entry: an MF for *binding a self-association interface and capping oligomer
size*. `GO:0140311` protein sequestering activity fails on its differentia (53BP1 stays chromatin-
bound and active) and `GO:0060090` is the opposite direction; the BP parallel `GO:0032460` already
exists, so what is missing is the activity that carries it out.

## A number that refused to add up — twice

Per the campaign rule, neither was rationalised:

1. `raw supporting_text keys=68` vs `parsed quote entries=87`. Traced to **PyYAML anchors/aliases**
   from shared Python dicts. Accounting: 14 (two shared holdup quote dicts × 8 uses) + 2 + 2 + 1 =
   **exactly 19 = 87 − 68**. `copy.deepcopy` does **not** fix it (it preserves internal sharing by
   design); a dumper with `ignore_aliases` does.
2. Hand-counted aspect tallies of 39/23/4 rows and 17/4/2 terms — **wrong in five places** against
   the computed 31/31/4 and 16/5/2, and the protein-binding count was 21 (the IntAct subset) rather
   than 26 (the total). Corrected at every site.

Both led to committing `genes/human/AHNAK/audit_ahnak_claims.py` (`--self-test`), which recomputes
every asserted count from the GOA TSV, enforces one-review-per-row with no `PENDING` survivors,
rebuilds `source_entities` from the WITH/FROM field and fails on drift, rejects duplicate YAML keys
with a strict loader, and rejects the retracted first-pass numbers if they reappear. The claim check
was verified by breaking it (mutating the CC tally back to 39/17 → fires on both the missing correct
claim and the reappearing retracted one).

## Validation

- `just validate human AHNAK` → **✓ Valid**, 1 warning, left standing deliberately: no annotation
  cites the affinage deep-research file, because campaign policy forbids quoting a provider sentence
  as `supporting_text`. Every claim affinage contributed was re-grounded on the primary paper, and
  its curatorial assessment is recorded in `references[].reference_review` per CLAUDE.md.
  *(Worth recording for the recall metric: affinage supplied `PMID:20833135`, this review's decisive
  reference, which appears nowhere in AHNAK's GOA reference set.)*
- All **87** `supporting_text` entries verified — 85 literature via the repo's own
  `SupportingTextValidator`, and the 2 `file:` quotes by literal substring, since **CI does not check
  `file:` quotes**. The checker was self-tested by breaking it three ways (mutated literature quote,
  fabricated `file:` quote, and a `file:` quote spanning a UniProt `CC` continuation line) — all
  three fired.
- `cache/go/terms.csv`: 4 additions, **0 deletions**, 0 duplicate curies; `cache_lint` clean and the
  commit gated on its exit status.

## Second commit: four of my own counts were wrong

A self-audit found the same failure four more times — a count read off a printed list rather than
derived. `GO:0005634` "four of the five protein donors" (there are four proteins and one node);
`GO:0005737` "five of the six" (all six non-self donors carry their own experimental evidence);
the `PMID:3126079` entity list put `CPX-856` and `HLTF` in the IDA set when they carry NAS rows;
and the Swiss-Prot claim above conflated partners with donors.

Fixing them by line surgery on the builder then **spliced a clause** and left
*"...rather than a protein. They Human AHNAK also has its own record..."* in a `reason` — the
brief's "string replacement across wrapped YAML creates new garbage" failure, invisible to every
schema, quote and reference check.

So the audit gained a prose check, and **it took two attempts to make it real**: the first version
anchored on "lowercase word then capitalised word", which *cannot fire here* because the preceding
word keeps the full stop that survived the edit — a guard that reads as coverage while being
unreachable. The working version anchors on the pronoun and requires a lowercase letter after the
capital so acronyms do not match. Verified by reintroducing the exact defect (clean file: 0
problems; mutated file: fires), and both directions are in `--self-test`, with each mutation
asserting its anchor exists first so a drifted target cannot pass by mutating nothing.

## Review round 1 — three evidence codes fixed, all five suggestions taken

**Blocking issue conceded in full.** Three `NEW` rows carried experimental codes for experiments
done only in mouse or dog. My own working brief states the rule — *if the assay was on a mouse
ortholog, use ISS with supporting_entities* — and I broke it three times.

| row | was | now | donor |
|---|---|---|---|
| `GO:1901385` | IMP | **ISS** | `E9Q616` mouse Ahnak |
| `GO:0030866` | IMP | **ISS** | `A0A8I3NCQ4` dog AHNAK |
| `GO:0005080` | IPI + human `P17252` | **ISS** | `E9Q616`; partner is mouse Prkca `P20444` |

The `GO:0030866` reason previously read *"hence IMP"* about canine knockdowns, which inverts the
convention exactly backwards. Verified the premise directly rather than taking it: PMID:14699089
reads *"An siRNA designed against the highly conserved repeated central domains of AHNAK
efficiently and specifically down-regulates AHNAK levels in MDCK cells"*.

**Two premises I checked rather than accepted.** The review calls PMID:19261907 a *rat* knockdown —
MC3T3-E1 is a mouse line; it changes nothing, both are rodent. And I went looking for a Ku
precedent to *defend* `GO:0008047`: Ku80 does carry it, but by **TAS from a telomerase paper**
(PMID:19188702), so it is not precedent for end-bridging stimulation of ligation at all. That is
recorded as checked-and-discarded rather than quietly dropped.

**Suggestions, all five taken.**

1. **The Ku bridging alternative** — a direct hit, and the ACTR10 failure mode: I quoted
   PMID:15177040's DNA-binding sentence and stopped immediately before *"AHNAK is also able to link
   two DNA molecules to a similar extent to that previously reported for Ku."* That sentence is now
   quoted and the objection met. The term stands on two grounds generic bridging does not predict —
   the binding is selective for LIG4–XRCC4 over other ligases and other NHEJ components, and AHNAK
   sits in a stable complex with it *on* DNA — with the mechanism recorded as open.
2. **Proposed term** renamed to an activity shape (*protein oligomerization limiting activity*) and
   re-parented. The reviewer is right that `GO:0140678` presumed 53BP1 oligomerization is an MF; it
   is not (`GO:0051260` is BP). It now sits under `GO:0005515` with the reasoning spelled out,
   because **the absence of a fitting parent is the gap being reported**.
3. **CC consistency** — one rule now stated and applied: *refine when the same experiment identifies
   the more specific entity; keep when the assay cannot discriminate*. That is why the S100A10
   `GO:0005515` rows were MODIFYed and `GO:0016020`/`GO:0031982` were not — a bulk membrane prep
   names no membrane.
4. **Stale count** `notes:43` fixed, and the row totals are now **audited claims**, so this class of
   drift fails the gate instead of reaching a reviewer. Verified by reintroducing the exact drift.
5. **RPS6KA1 discriminator** stated as a reusable two-condition rule.

Applying fix 1 spliced a contradictory *"Evidence code IMP … Evidence code ISS"* into one reason —
caught by the prose check before commit, which is the second time that guard has paid for itself.


cmungall:
Thanks — the blocking issue is correct and I've conceded it in full. Pushed in `6ca1eac2fc`.

**The three evidence codes.** My own working brief states the rule (*"if the assay was on a mouse ortholog, use ISS with `supporting_entities`"*) and I broke it three times, which is worse than not having the rule.

| row | was | now | donor |
|---|---|---|---|
| `GO:1901385` | IMP | **ISS** | `E9Q616` mouse Ahnak |
| `GO:0030866` | IMP | **ISS** | `A0A8I3NCQ4` dog AHNAK (TrEMBL, 6134 aa) |
| `GO:0005080` | IPI + human `P17252` | **ISS** | `E9Q616`; the partner is mouse Prkca `P20444` |

I verified your MDCK premise directly rather than taking it — `PMID:14699089` reads *"An siRNA designed against the highly conserved repeated central domains of AHNAK efficiently and specifically down-regulates AHNAK levels in MDCK cells"*. The `"hence IMP"` sentence is gone, and the reason now names the inversion explicitly so it isn't repeated: a non-human perturbation is precisely the reason the code is *not* IMP.

**Two premises I checked rather than accepted.** `PMID:19261907` is MC3T3-E1, a *mouse* line, not rat — doesn't change anything, both are rodent. More usefully: I went looking for a **Ku precedent to defend `GO:0008047`** and it dissolved. Ku80 (`P13010`) does carry the term, but by **TAS from a telomerase paper** (`PMID:19188702`), so it is not precedent for end-bridging stimulation of ligation. That's recorded as checked-and-discarded rather than quietly dropped.

**On the Ku bridging alternative — a direct hit, and the ACTR10 failure mode.** I quoted the DNA-binding sentence and stopped one sentence short of *"AHNAK is also able to link two DNA molecules to a similar extent to that previously reported for Ku."* It is now quoted, and the objection is met rather than ignored. I kept `GO:0008047` on two grounds that generic DNA bridging does not predict: the binding is **selective** for LIG4–XRCC4 and not for other DNA ligases or other NHEJ components, and AHNAK forms a stable complex with it **on** DNA — a component of the ligating assembly, not a tether acting at a distance. The mechanism is now recorded as open rather than assumed.

**On the proposed term** — you're right that the parent didn't hold, and it's the more interesting half. `GO:0140678` and `GO:0098772` both presume the regulated thing is a molecular function, and 53BP1 oligomerization isn't one (`GO:0051260` is BP); `GO:0140313` fails the same differentia as `GO:0140311`. Renamed to *protein oligomerization limiting activity* and re-parented to `GO:0005515` — deliberately uninformative, because **the absence of a fitting parent is the gap**: GO's MF-regulator vocabulary is built to regulate molecular functions and has no way to say an activity modulates a partner's self-assembly.

**On the CC inconsistency** — fair, and the fix is a stated rule rather than a per-row judgement: *refine when the same experiment that produced the row identifies the more specific entity; keep when the term is general because the assay cannot discriminate*. The S100A10 rows qualify because those experiments name the partner; a bulk NK-cell membrane prep names no membrane, so MODIFYing to plasma membrane would assert something `PMID:19946888` did not show. Same rule applied to `GO:0031982`.

**On the stale count** — the loop is closed as you suggested: the row totals are now audited claims in `audit_ahnak_claims.py`, and I verified it by reintroducing the exact drift you found (`8 NEW` → fires on both the missing correct claim and the reappearing retracted one). The RPS6KA1 discriminator is now stated as a reusable two-condition rule.

One thing worth flagging for the repo generally: **`gh pr edit` exits 1 on the projects-classic GraphQL deprecation**, so my three earlier PR-description updates silently never landed — which is why you saw the table still at 8. `gh api repos/{owner}/{repo}/pulls/{n} -X PATCH -F body=@file` works. Same shape as the failure modes this review is about: a non-zero exit masked by a warning that looks benign.

Applying the Ku fix spliced a contradictory *"Evidence code IMP … Evidence code ISS"* into one reason; the prose check added last round caught it before commit.

Issue #2222, opened 2026-07-25
GO:0051082 obsoleted without the holdase NTR: 8 genes stranded, plus GO:0140309 misuse in in-situ holdases


cmungall:
Split out of #2219 review (SUGGESTION 5), scoped after resolving that PR's IMPORTANT 1.

## 1. The blocking request failed — 8 genes are stranded

`projects/UNFOLDED_PROTEIN_BINDING.md:134` asked that **GO:0051082 obsoletion be blocked** until a general "holdase chaperone activity" NTR exists. Verified against QuickGO/OLS on 2026-07-25:

- **GO:0051082 is obsolete** — `isObsolete: true`, name `obsolete unfolded protein binding`. The block was not honoured.
- **The holdase NTR was never created.** A GO search for "holdase chaperone activity" returns only GO:0140309 and GO:0044183.
- GO:0051082's obsoletion comment offers only those two as replacements, and **neither fits an in-situ holdase**: GO:0044183 requires assisting folding; GO:0140309 requires escorting the client to a destination.

**Stranded:** CRYAA, CRYAB, HSPB6, CLU, SCG5, DNAJB6, DNAJB8, HSPH1.

This needs escalating upstream on [go-ontology#30962](https://github.com/geneontology/go-ontology/issues/30962) / [#30552](https://github.com/geneontology/go-ontology/issues/30552) rather than resolving gene-by-gene.

## 2. GO:0140309 is used as a term id in genes that don't fit it

GO:0140309 was relabelled `unfolded protein carrier activity` → **`unfolded protein holdase activity`**, but its **definition is unchanged and still carrier-specific** ("escorts it to an acceptor molecule or to a specific location"), and it is still a child of GO:0140597 *protein carrier chaperone*. A relabel is not a redefinition — the new label makes it easy to misapply.

9 gene reviews still use GO:0140309 as a term id. They split two ways:

**(a) Wrong term — in-situ holdases, need the NTR (same fix as CRYAA in #2219):**
- **`human/CRYAB`** ⚠️ — one of the 8 stranded genes named on the project page, still pointing at GO:0140309 (4 sites). Highest priority; exact same bug just fixed in CRYAA.
- `SULAC/Hsp14` (2 sites) — an sHSP, same class.

**(b) Plausibly correct term — genuine carrier-holdases, only the *label* is stale:**
[...]

PR #3001, opened 2026-09-12
Review human AHSP

cmungall:
Review of human **AHSP** (Q9NZD4), alpha-hemoglobin-stabilizing protein.

AHSP is the inverse of the usual gene in this campaign. Its function is well
established, specific and structurally characterised — eight PDB entries that split
four/four between X-ray co-crystal structures with alpha-globin and solution-NMR
structures of the free protein. So the question is not "is this over-annotated" but
**can GO express what is actually known**, and largely it cannot.

## Row reconciliation

| | |
|---|---|
| GOA data rows (`AHSP-goa.tsv`, 25 lines − header) | **24** |
| `existing_annotations` entries | **26** |
| of which `NEW` proposals | **2** (`GO:0044183`, `GO:1903427`) |

24 + 2 = 26. Every GOA row was matched one-to-one against a review entry on the full
key `(term, evidence code, reference, WITH/FROM)` — not on counts, since the seeded
stub is known to collapse distinct `GO:0005515` partners. **The stub did not collapse
anything on this gene** (11 TSV `GO:0005515` rows → 11 stub entries); reported as a
negative because ADAMTSL5 and ACTR5 say to check.

Actions: **8 ACCEPT · 8 MARK_AS_OVER_ANNOTATED · 7 MODIFY · 1 REMOVE · 2 NEW** (was 10/8/5/1/2 before review — see the `GO:0031721` correction below).

## The one REMOVE, and why it happened

`GO:0005833 hemoglobin complex` / `part_of` / **NAS** / PMID:12066189 → **REMOVE**, on
three independent grounds:

1. **The cited paper says the opposite in its own abstract** — AHSP "forms a stable
   complex with free alpha-haemoglobin but not with beta-haemoglobin or haemoglobin A
   (alpha(2)beta(2))".
2. **UniProt says the opposite**, from two curated crystallographic papers
   (`ECO:0000269|PubMed:15550245, ECO:0000269|PubMed:15931225`).
3. **The term definition excludes it** — `GO:0005833` requires "two pairs of associated
   globin polypeptide chains"; AHSP·αHb is a 1:1 heterodimer of one *non-globin* chain
   with one globin chain, measured by AUC, gel filtration and amino acid analysis, and
   it is specifically **not** oxygen-carrying because binding drives the iron to an
   inert hexacoordinate ferric state.

NAS is a non-experimental code, so the "never REMOVE an experimental annotation you
cannot verify" rule does not engage.

**The cause is the more useful finding.** GO has no term for the complex AHSP actually
forms — an OLS search returns nothing. A curator with a crystallographically defined,
stoichiometrically measured heterodimer to record had only the tetramer term available.
So the removal is filed *together with* a proposed CC term rather than instead of it.

## Chaperone terms: definitions and parentage fetched before proposing anything

| term | verdict |
|---|---|
| `GO:0051082` unfolded protein binding | **obsolete**, no `replaced_by`; `consider` lists `GO:0044183`, `GO:0140309` |
| `GO:0140309` unfolded protein holdase activity | **excluded twice over** |
| `GO:0044183` protein folding chaperone | honest **partial** fit → proposed as a NEW IDA row |

`GO:0140309` fails on *both* clauses of its own definition, not just one:

- it is a **`protein carrier activity`** (ancestors `GO:0140597` → `GO:0140104`) that
  "escorts it to an acceptor molecule or to a specific location" — AHSP does not
  translocate anything; and
- its term comment says a holdase "binds an unfolded protein and **keeps it unfolded**"
  — AHSP's client is crystallographically ordered and fully helical in both the ferrous
  and ferric complexes.

`GO:0044183` is used because AHSP demonstrably does assist folding ("AHSP forms a
heterodimeric complex with αo that inhibits αo aggregation and promotes αo folding in
the absence of haem"), but the folding is *partial* and haem-dependent, so the residual
mismatch is filed as an `ONTOLOGY` KnowledgeGap with a proposed term rather than papered
over.

**Note this is a distinct shape from the other genes stranded by repo issue #2222.**
The sHSP cases hold genuinely unfolded clients and fail only the *carrier* clause; AHSP
fails the *unfolded* clause as well. A fix that only relaxes "escorts it to an acceptor
molecule" would still not reach this gene.

The proposal follows GO's own precedent: `GO:0140713 histone chaperone activity` already
sits under `GO:0140597` and describes binding a *folded* client, keeping it soluble, and
depositing it into its final assembly. `globin chaperone activity` is its sibling with
"nucleosome" replaced by "haemoglobin tetramer".

**Why not simply annotate `GO:0140597` and propose nothing?** Its parent `GO:0140104`
states in a comment that *"a carrier moves with its substrate/cargo"*, and AHSP moves
nowhere — the hand-off to β-globin is **competitive displacement of AHSP at the same
interface**, so a carrier annotation would assert the reverse of the measured mechanism.
And every existing child of `GO:0140597` is about getting a client somewhere, none about
changing its chemistry, so the carrier term alone would leave the redox half homeless in
both branches.

## Redox framing: tested, and `GO:0016209` rejected

Affinage's GO grounding proposes `GO:0016209 antioxidant activity`. **Wrong.** The
definition requires the antioxidant "can itself be more easily oxidized than the
substance protected" or "can trap free radicals". AHSP has no redox-active group at all
— it remodels its *client's* haem coordination sphere. Consistently, **all seven `is_a`
children of `GO:0016209` are redox enzymes** (peroxidase, SOD, the two disulfide
reductases, superoxide reductase, glutathione dehydrogenase, sulfiredoxin). Filed as a
second `ONTOLOGY` gap: GO has no MF for suppressing a *bound client's* cofactor
reactivity.

The process is captured instead as **NEW `GO:1903427`** (negative regulation of ROS
biosynthetic process), verified current and a descendant of `GO:2000378`, carried as
**IMP on PMID:18179859** — shRNA knockdown in human K562 and CD34+ erythroblasts. An
`involved_in` claim about a cellular process belongs on the cellular loss-of-function
datum; the purified-protein redox chemistry demonstrates the mechanism underneath it and
is cited as supporting evidence, not as the row.

**Wrong-ROS-species check: negative.** GOA carries no ROS term on AHSP at all, so there
was nothing to propagate wrongly. And the *generic* parent is the correct choice here
rather than a species-specific child, because the literature implicates several species
in sequence — O2 autoxidation, H2O2, ferryl haem, protein-based radicals.

## Structural evidence honoured over screen noise

IntAct (39 records read of 39 reported), counting **distinct experiments** rather than
`NbExp`:

| partner | methods | distinct pubs | MI | verdict |
|---|---|---|---|---|
| **P69905 HBA1/HBA2** | **x-ray diffraction · bio-layer interferometry · anti-tag coIP** | 3 | **0.71** | MODIFY → `GO:0031721` + `has_input` extension |
| Q5D1E8 ZC3H12A | Y2H only | 2 (same CCSB pipeline) | 0.67 | over-annotated |
| P62942, Q05086-2, P17024, P20618, Q8IUQ0, Q9Y2B5 | Y2H only | 1 each | 0.56 | over-annotated |

Three orthogonal methods in three papers converge on HBA and on nothing else. `protein
binding` cannot be the conclusion for the one partner a crystal structure resolves.

**Third confirmed instance of the `NbExp` trap** (after ACRV1 and ADAMTSL5): six of the
seven screen partners are logged as `two hybrid array` + `two hybrid prey pooling
approach` + `validated two hybrid` — three sub-methods of **one** screen, which is what
`NbExp=3` counts. ZC3H12A's `NbExp=4` is two screens, but both are CCSB Y2H against the
same ORFeome, so it is one method seen twice.

**Partner-accession discipline (the ACRV1 check): all clean** — every one of the eight
accessions resolves to a reviewed Swiss-Prot canonical entry of the expected length; the
only non-canonical token, `Q05086-2`, is a declared UniProt *isoform* id, not a
truncated ORFeome clone. **Expression-domain mismatch** is the strongest argument
against the screen rows: AHSP is erythroid-restricted, CLVS1 is neuron-restricted, ZNF20
is nuclear, VPS9D1 endosomal.

## Propagation: the cleanest IBA set in this campaign

All four IBA rows cite `MGI:MGI:2158492|PANTHER:PTN000408386`. Donor resolved to
`Q9CY02 AHSP_MOUSE`, Swiss-Prot, 102 aa — a one-to-one orthologue. Donor's own evidence
queried at QuickGO (13 annotations, `numberOfHits == len(results)`, not truncated):

| propagated term | donor's own evidence | same term? |
|---|---|---|
| `GO:0005737` | **IDA** PMID:12066189 | yes |
| `GO:0006457` | **IDA** PMID:12066189 | yes |
| `GO:0030218` | **IMP** PMID:12066189 | yes |
| `GO:0050821` | **IDA** PMID:12066189 | yes |

Every propagation lands at **exactly** its donor's term — no granularity drop (the ACRV1
defect), no paralogue transfer. So **no `propagation_review` is warranted on any row**
and all four are ACCEPT.

Negatives reported rather than skipped: `PTHR15914` has one subfamily and **170 proteins
/ 144 proteomes**, of which the InterPro reviewed endpoint returns 3 (the Swiss-Prot
subset, 1.8% — not the family). Querying *all* human members including unreviewed
returns three entries that are **all AHSP itself** (a duplicate entry and a 57-aa
fragment), so there is genuinely no human paralogue and the sibling-cross-check
technique is unavailable here. The node-placement question (AADACL/ACTG2) has no
purchase: one node, one orthogroup.

`IPR015317` is a **family-specific** signature ("Alpha-haemoglobin stabilising protein"),
not a bare fold, and the live `interpro2go` file confirms its four mappings are exactly
the four `GO_REF:0000002` rows here — ACCEPT, redundant but well-founded.

## Retraction / erratum sweep

27 PMIDs checked via `CommentsCorrections/RefType` on each article's own record (the only
way a Publisher Correction is discoverable). One flag: **PMID:40205054 carries an
unflagged `ErratumIn → PMID:41039152`**. Scope read rather than assumed — it replaces a
duplicated equation in the Methods "Loss functions" section and does not touch the
interaction data, so the AHSP–HBA row stands. Recorded on the row and as a reference.

## Two specificity MODIFYs on vague NAS rows

- `GO:0020027 hemoglobin metabolic process` → **`GO:0042541` hemoglobin biosynthetic
  process** (verified descendant), removing the unsupported implication of involvement in
  haemoglobin *catabolism*.
- `GO:0030097 hemopoiesis` → **`GO:0030218` erythrocyte differentiation** (verified
  descendant, already annotated twice), since AHSP has no part in myeloid or lymphoid
  development.

## Deliberately **not** annotated

`GO:0005634 nucleus`, despite a human primary-erythroblast imaging study reporting AHSP
becoming predominantly nuclear and being expelled at enucleation. One antibody-based
study from one laboratory; no NLS; at 11.8 kDa passive nuclear entry is expected; every
measured activity is on a cytoplasmic ligand. Filed as a `BIOLOGY`/`CURATION` knowledge
gap and a suggested experiment, with the reference marked `UNVERIFIED`.

## Verification

- `just validate human AHSP` → **✓ Valid**, on the tree after merging `origin/main`
  (merge was already-up-to-date). One standing warning: no annotation cites the affinage
  deep-research file. That is deliberate — the campaign rule forbids resting a
  mechanistic claim on a provider sentence, and every affinage lead here was re-anchored
  to the primary PMID.
- **66 `supporting_text` quotes** in the YAML and **17** in the notes verified with the
  repo's own `SupportingTextValidator` under a strict duplicate-key YAML loader; raw
  `supporting_text` key count **equals** the parsed count, so no duplicate key silently
  discarded provenance. **The checker was tested by breaking it** (one mutated word →
  exit 1 naming the row) before its clean run was believed.
- **Zero `file:` supporting_text entries** — the one unvalidated fabrication surface is
  not used at all here. (Also a practical finding: the repo's pre-write hook resolves
  `file:` references against the *main* checkout's project root, not the agent worktree,
  so a `file:` reference to a brand-new gene folder blocks the write.)
- Every GOA row reconciled against the review with a scripted key-match, also break-tested.
- `cache_lint` exit status **0**, checked directly rather than after a pipeline, both
  after my edits and after merging main. `terms.csv`: one row added (`GO:0030492`) in
  sorted position, **no deletions** (three-dot diff vs the merge base), **no duplicate
  curies**, and `origin/main` does not already carry the added curie.
- No rendered HTML committed.

## Affinage

`gates_passed: True` (precision only). 19 citations, all numeric PMIDs, none retracted.
**Missed** the discovery paper `PMID:12066189` entirely — the source of four GOA rows
and the paper whose abstract refutes the `GO:0005833` annotation — plus `PMID:12192002`
(the stoichiometry behind the proposed complex term), `PMID:15178680`, `PMID:19706593`
(see the correction below), and every reference behind the eleven `GO:0005515` rows. Its GO grounding proposed `GO:0016209`
(rejected above), `GO:0140313` molecular sequestering activity (rejected — AHSP's purpose
is the *opposite* of sequestration; β-globin displaces it) and `GO:0140110` transcription
regulator activity (rejected — AHSP is a *target* of STAT3 and NRF2/MAFG, not a
regulator).

## Correction made during review, recorded rather than quietly fixed

The first version of this PR asserted, in four places, that AHSP has **two** co-crystal
structures. That number came from reading UniProt's `RN` block, where exactly two entries
carry `X-RAY CRYSTALLOGRAPHY ... IN COMPLEX WITH HBA`. **Deriving** it instead — asking
PDBe which of the eight entries map *both* `Q9NZD4` and `P69905` — gives **four**:

| PDB | method | res. | chains | citation |
|---|---|---|---|---|
| 1Y01 | X-ray | 2.80 Å | AHSP + HBA | PMID:15550245 |
| 1Z8U | X-ray | 2.40 Å | AHSP + HBA | PMID:15931225 |
| 3IA3 | X-ray | 3.20 Å | AHSP + HBA | **PMID:19706593** |
| 3OVU | X-ray | 2.83 Å | AHSP + HBA + `Q6G8J7` IsdH (*S. aureus*) | unpublished |
| 1W09/1W0A/1W0B/1XZY | solution NMR | — | AHSP only | PMID:15178680, PMID:15550245 |

The count is load-bearing (it appears in the `ONTOLOGY` knowledge gap, the proposed
complex term's justification, and the ComplexPortal suggestion), so it was corrected with
a script asserting each anchor is present before replacing and re-grepping for the
retracted phrasing afterwards — **and it still landed in 4 of 5 sites.** A fifth
occurrence in `suggested_questions` read "GO:0030492 has two co-crystal structures", and
the post-edit grep missed it because the retracted-phrase list contained
`"two crystal structures"` but not `"two co-crystal structures"`.

That is the campaign's "fixed in N places, landed in N−1" recurrence happening *to a
script written specifically to prevent it*: **anchoring on enumerated phrasings inherits
whatever the author failed to imagine.** The fix was to lint for the *pattern* — a
cardinal number word within 40 characters of a structure noun, across the review, the
notes, the history record and this PR body — rather than for sentences. It found the miss
immediately; its own first run produced four false positives from digits inside `2.8 A`,
PMIDs and GO ids, which is why only count *words* are matched; and it is self-tested by
mutating `four` back to `two` and requiring the lint to fail.

**The recount surfaced a paper nothing else had.** `PMID:19706593` (Gell *et al.*, JBC
2009) is absent from UniProt's `RN` list, absent from GOA, and absent from the affinage
report — yet it is a crystal structure of the complex *plus* the Pro30 mutagenesis that
establishes how the conformational change is transmitted (cis-peptidyl Pro30 contacting
the N terminus of alpha-globin helix G; substituting it reduces conversion, and mutations
stabilising the cis conformation of free AHSP enhance it). Its abstract also states the
ROS claim outright. It is now cited on the `GO:1903427` row and in `core_functions`.

Generalisable: **a structure count taken from a reference list is a claim about the
reference list.** One API call fixes it — ask the structure database which entries contain
the gene *and its partner*.

## Round 2 — the reviewer found a real defect, and it was load-bearing

**`GO:0031721 hemoglobin alpha binding` exists.** Definition: *"Binding to a hemoglobin
alpha chain."* `is_a` child of `GO:0030492`, not obsolete. The review's claim that "no
alpha-globin-specific or free-globin-chain binding term exists in GO" was **false**, and
it was load-bearing in four places. Fully conceded and fixed.

**Why the first pass missed it.** I searched OLS for `"globin binding"` and
`"alpha-globin binding"`. Neither can match a label reading *hemoglobin alpha binding* —
OLS tokenises on words and **"globin" is not a token of "hemoglobin"**. The search was
incapable of returning the term, so its empty result was a fact about my query, not about
the ontology. Listing the `is_a` children of `GO:0030492` returns it in one call, which is
exactly the rule this repo already documents for term-absence claims. Note this is the
**second** instance of the same error in this one review, hours apart: the structure count
read off a reference list, and now an absence read off a keyword search. *A null result is
a claim about the method.*

**What changed:**

| | before | after |
|---|---|---|
| 3 × HBA `GO:0005515` MODIFY target | `GO:0030492` | **`GO:0031721`** |
| `GO:0030492` IEA (InterPro2GO) | ACCEPT + caveat | **MODIFY → `GO:0031721`** |
| `GO:0030492` NAS (PMID:12066189) | ACCEPT + caveat | **MODIFY → `GO:0031721`** |
| `core_functions[1].molecular_function` | `GO:0030492` | **`GO:0031721`** |
| action tally | 10/8/5/1/2 | **8/8/7/1/2** |

**This dissolves the reviewer's second point entirely.** The asymmetry was real: `GO:0005833`
was removed partly *because* its definition requires a tetramer, while a tetramer-defined
binding term was accepted two rows later on a broad-synonym mitigation. Both rows now get
the same standard.

**And it produced a better question.** The withdrawn `suggested_question` asked GO whether
to broaden `GO:0030492` or create a free-chain child — asking for a term that already
existed. It is replaced by one aimed at **InterPro**: `IPR015317` is specific to the AHSP
orthogroup, so retargeting one line of `interpro2go` from `GO:0030492` to `GO:0031721`
would fix **every AHSP orthologue at once** rather than gene by gene. Same shape as the
PAINT node-placement findings elsewhere in this campaign.

**The CC gap was then re-checked rather than assumed**, having just been caught asserting
absence from a single query: `GO:0005833` has **zero children**; seven independent QuickGO
query phrasings return no CC term for an AHSP or α-globin complex; and ComplexPortal holds
**nine** complexes containing AHSP and **none** containing α-globin (`P69905`) — verified by
reading each entry's participant accessions, since the nine names list AHSP alongside
MAPK8IP scaffolds and are not plausible erythroid complexes on their face. The CC proposal
stands, and the ComplexPortal asymmetry is now a `suggested_question` in its own right.

**That re-check also found the precedent that makes the CC proposal easy to accept.**
`GO:0031838 haptoglobin-hemoglobin complex` — *"A protein complex formed by the stable
binding of a haptoglobin to hemoglobin"*, a direct child of `GO:0032991` — is the same shape
of term, and its binding counterpart `GO:0031720 haptoglobin binding` sits **immediately
beside `GO:0031721`** in the id block the reviewer pointed at. So GO models the haptoglobin
case *completely* (binding function **and** complex) and the AHSP case only half. The
proposed term is renamed **`AHSP-hemoglobin alpha complex`** and its definition rewritten to
follow `GO:0031838`'s wording, using GO's own "hemoglobin alpha" phrasing rather than
"alpha-globin" for consistency with `GO:0031721`. A worthwhile lesson in itself: re-checking
a claim I had been caught getting wrong found something better than the original claim.

### The three 🔵 suggestions taken

- **`GO:1903427` evidence code** → **IMP on PMID:18179859**. Agreed: an `involved_in` claim
  about a cellular process belongs on the human-cell loss-of-function datum, not on a
  two-protein in vitro reconstitution, which demonstrates the chemistry underneath the
  process rather than the process. The in vitro work moves to supporting evidence.
- **"partial fit" vs "neither"** → softened. The justification now says `GO:0044183` "is
  annotated on this gene and is defensible … but it fits only the apo-globin regime, and
  only partially there", instead of "AHSP does neither".
- **`core_functions[1]`** → now carries `GO:0031721`, a different MF from
  `core_functions[0]`, and the description states why the two entries are kept separate:
  they are not one binding event with two consequences but two regimes with different
  client states — the folding arm works with no haem present at all, and the Pro30 mutants
  dissociate them, retaining α-globin binding while losing the conversion.

### The one 🔵 suggestion declined — with a measurement

**`GO:0140597`'s label.** The reviewer's premise is correct: the live primary label is
`protein carrier activity` (confirmed at QuickGO *and* OLS), so my `proposed_parent` is
right and `cache/go/terms.csv` is stale. But I tried the suggested refresh and measured the
effect:

```
just validate yeast RRB1   # with the refreshed cache label
⚠ WARN: Label mismatch for 'GO:0140597':
        expected 'protein carrier activity', got 'protein carrier chaperone'
```

The refresh puts a new warning on **three merged reviews** (`RRB1`, `SQT1`, `SHQ1`) that use
the synonym in a label-validated `core_functions.molecular_function` slot and that this PR
does not otherwise touch. A correct fix updates the cache *and* those three files in one
commit — shared-infrastructure work, not gene-review work, and the standing rule here is not
to regenerate shared caches inside a gene PR. Edit reverted; measurement and the three
affected filenames recorded in `AHSP-notes.md` so the refresh does not have to rediscover
them.

### Re-verified after the changes

`just validate human AHSP` ✓ Valid · **67** YAML quotes and **17** notes quotes verbatim,
raw = parsed · GOA reconciliation 24 + 2 = 26 still exact · `cache_lint` exit 0 · `terms.csv`
gains `GO:0031721` in sorted position, no deletions, no duplicates · claim lint (self-tested)
clean.


cmungall:
Thank you — 🟡 #1 was right, and it was the most valuable catch in this review.

## 🟡 #1 — confirmed and fixed in full

`GO:0031721` **is** `hemoglobin alpha binding`, definition *"Binding to a hemoglobin alpha chain"*, `is_a` child of `GO:0030492`, not obsolete. `GO:0031722` is `hemoglobin beta binding` and `GO:0031720` is `haptoglobin binding`, exactly as you inferred from the three consecutive enum ids. `GO:0030492` has precisely those two children and no others.

Your methodological point is the important part and I want to state it rather than just fix the symptom. My searches were `"globin binding"` and `"alpha-globin binding"`. **Neither can match a label reading *hemoglobin alpha binding*** — OLS tokenises on words, and "globin" is a *substring* of "hemoglobin", not a *token* of it. So the query was structurally incapable of returning the term, and the empty result was a fact about my query rather than about the ontology. One `children` call on `GO:0030492` finds it.

This is the **second** instance of the same error in this one review, about two hours apart: the structure count read off UniProt's reference list, and now an absence read off a keyword search. The generalisation I had already written for the first case covers the second exactly and I failed to apply it — *a null result is a claim about the method.*

**What changed:**

| | before | after |
|---|---|---|
| 3 × HBA `GO:0005515` MODIFY target | `GO:0030492` | **`GO:0031721`** |
| `GO:0030492` IEA (InterPro2GO) | ACCEPT + caveat | **MODIFY → `GO:0031721`** |
| `GO:0030492` NAS (PMID:12066189) | ACCEPT + caveat | **MODIFY → `GO:0031721`** |
| `core_functions[1].molecular_function` | `GO:0030492` | **`GO:0031721`** |
| action tally | 10/8/5/1/2 | **8/8/7/1/2** |

## 🟡 #2 — you were right, and it is now moot

The asymmetry was real, not just apparent: removing `GO:0005833` on "the definition requires a tetramer" while accepting a tetramer-defined binding term two rows later on a broad-synonym mitigation is not one standard. Both rows now get the same one, and the broad-synonym argument is no longer needed at all.

**It also produced a better question than the one it replaced.** The withdrawn `suggested_questions` #3 asked the GO editors to broaden `GO:0030492` or create a free-chain child — i.e. asked for a term that already existed. In its place is a question for **InterPro**: `IPR015317` matches the AHSP orthogroup and nothing else, so retargeting one `interpro2go` line from `GO:0030492` to `GO:0031721` would fix **every AHSP orthologue at once** rather than gene by gene. Same shape as the PAINT node-placement findings elsewhere in this campaign — right annotation, wrong level to attach it at.

**And it prompted me to re-check the CC gap** rather than leave it resting on the same kind of query that had just failed. It survives, three ways: `GO:0005833` has **zero children**, so there is no hemoglobin sub-complex term anywhere to refine; seven independent QuickGO query phrasings return no CC term for an AHSP or α-globin complex; and ComplexPortal holds **nine** complexes containing AHSP and **none** containing α-globin (`P69905`) — verified by reading each entry's participant accessions, because the nine names list AHSP alongside MAPK8IP scaffolds and are not plausible erythroid complexes on their face. That asymmetry is interesting enough to be its own `suggested_question`: the one AHSP assembly with four crystal structures behind it is the one ComplexPortal lacks.

## 🔵 2, 3, 4 — taken

- **`GO:1903427` evidence code.** Agreed, and for a sharper reason than "the IMP is also available": an `involved_in` claim about a *cellular* process should not rest on a two-protein reconstitution, which demonstrates the chemistry underneath the process rather than the process. Now **IMP / PMID:18179859** (shRNA in human K562 and CD34+ erythroblasts — human cells, so the code is not borrowed from the mouse knockout literature). The purified-protein redox work moves to supporting evidence, and the row says why.
- **"partial fit" vs "neither".** Conceded — as written the proposal argued against the annotation the same file makes. The justification now reads that `GO:0044183` "is annotated on this gene and is defensible, because AHSP does measurably assist alpha-globin folding — but it fits only the apo-globin regime, and only partially there".
- **`core_functions[1]`.** It now carries `GO:0031721`, so the two entries no longer share an MF. I kept them separate rather than folding them, and the description now says why: they are not one binding event with two consequences but two regimes with different client states — the folding arm is demonstrable with no haem present at all, and the Pro30 mutants dissociate them, retaining α-globin binding while losing the conversion.

## 🔵 1 — declined, with a measurement

Your premise is correct: the live primary label **is** `protein carrier activity` (confirmed independently at QuickGO and OLS; `protein carrier chaperone` is an exact synonym), so `terms.csv` is the stale one and my `proposed_parent` is right.

But I tried the refresh before deciding, rather than reasoning about it:

```
just validate yeast RRB1     # with the refreshed cache label
⚠ WARN: Label mismatch for 'GO:0140597':
        expected 'protein carrier activity', got 'protein carrier chaperone'
```

So it is not a free fix — it puts a new warning on **three merged reviews** (`RRB1`, `SQT1`, `SHQ1`) that use the synonym in a label-validated `core_functions.molecular_function` slot and that this PR does not otherwise touch. A correct fix updates the cache *and* those three files in one commit, which is shared-infrastructure work rather than gene-review work. Edit reverted; the measurement and the three filenames are recorded in `AHSP-notes.md` so whoever does the refresh does not have to rediscover them.

## Re-verified after the changes

`just validate human AHSP` ✓ Valid (one standing warning: no annotation cites the affinage file, which is deliberate — every provider lead was re-anchored to a primary PMID). **67** YAML `supporting_text` quotes and **17** notes quotes verbatim under a strict duplicate-key loader, raw count = parsed count. GOA reconciliation still exactly 24 + 2 = 26 on the full `(term, evidence, reference, WITH/FROM)` key. `cache_lint` exit **0** (checked directly, not after a pipeline). `terms.csv` gains `GO:0031721` in sorted position — no deletions against the merge base, no duplicate curies. Claim lint (self-tested by mutation) clean.

On your validation note: understood, and the runner's constraint is worth flagging upstream — but your `grep -F` quote pass and the enum-cache cross-reference both found things the toolchain would not have, so I would not read the missing `just` as having weakened this review.

PR #3002, opened 2026-09-12
Review human AHDC1

cmungall:
Review of human **AHDC1** (Q5TGY3, "Gibbin"), the Xia-Gibbs syndrome gene.

## Row reconciliation

| source | count |
|---|---|
| `AHDC1-goa.tsv` data rows (16 lines − 1 header) | 15 |
| `fetch-gene` stub `- term:` entries | 15 |
| QuickGO `geneProductId=UniProtKB:Q5TGY3` `numberOfHits` | 15 |
| `existing_annotations` in the review | **20** = 15 GOA + 5 `NEW` proposals |

> **Count correction.** At first submission this was **19 = 15 + 4**, and the PR body said 5/20 — an off-by-one the reviewer caught (`GO:0006357` was planned as a `NEW` row and implemented as a `MODIFY` replacement, and the narrative count was never re-derived). Acting on the reviewer's `GO:0000785` suggestion then added a fifth `NEW` row, so the number now reads correct again. It was not correct when written.

All three sources agree at 15, so the stub did not collapse rows on this gene (contrast ADAMTSL5 / ACTR5). The three `GO:0005515` rows were already split one-per-partner-per-publication and are judged separately.

## The predicted defect is not there — and the reason is checkable

The task brief predicted an InterPro2GO AT-hook DNA-binding MF with `GO_REF:0000002`. **There is no such row and there cannot be one.** InterPro (`/api/entry/InterPro/protein/UniProt/Q5TGY3/`, `count=2`, unpaginated) returns only `IPR032757` (DUF4683) and `IPR039225` (family), **both with empty `interpro2go`**, and `/entry/all/` returns four signatures total — no AT-hook signature (`PF02178` / `IPR017956` / the PROSITE profile) matches this protein anywhere. `PTHR15617` has `go_terms: null` and UniProt records `PAN-GO; Q5TGY3; 0 GO annotations based on evolutionary models`.

So AHDC1 has **no family-level or domain-level GO inference from any source**. Reported as a non-confirmation rather than dropped.

## …but the over-claim is present, by a different route

`GO:0003700 DNA-binding transcription factor activity` entered as a **manual `IDA` from UniProt** off PMID:35585237. The term's definition requires "selective and non-covalent binding to a **specific double-stranded genomic DNA sequence (sometimes referred to as a motif)**". The paper:

- assayed DNA binding **only** by ChIP-seq of a stably introduced, inducible, **tagged transgene** — the protein is never purified, and there is no gel shift, anisotropy, SELEX or de novo motif call anywhere;
- states outright that *"Gibbin DNA binding itself does not appear to be sufficient to drive gene expression, as ChIP-seq signal was not restricted to the mesoderm lineage or to Gibbin-regulated loci"*;
- puts the motif recognition in the **partners** — "enhancer- or promoter-bound Gibbin interacts with dozens of sequence-specific zinc-finger transcription factors" — with only ~10% of Gibbin sites overlapping GATA3 sites.

Sequence check: of the two UniProt AT-hooks, **only one is canonical**. `KRKRGRPPKNLLL` (544–556) carries the invariant `RGRP` core — the **sole occurrence of that string in all 1,603 residues** — while `RRKAGRGRKADAG` (396–408) is `GRGRK`, no proline. Neither `DNA_BIND` feature carries an `/evidence` tag.

→ **MODIFY `GO:0003700` → `GO:0003712` transcription coregulator activity**, a *sibling* under `GO:0140110` (verified), not a demotion to a parent. The DNA-side evidence is preserved as a new `GO:0003682 chromatin binding` row rather than deleted.

The genetics that *does* support the AT-hook mattering is recorded as such and not confused with a measurement: PMID:34950897 puts 4/10 Xia-Gibbs missense variants in 71 residues around AT-hook 2, two inside the 12-residue core (matching `VAR_086664` G548S and `VAR_086665` R549H) — while saying explicitly that AHDC1 "**likely**" acts through its AT-hooks and that there is no independent assay for AHDC1 protein function.

## Ensembl Compara round trip

Both `GO_REF:0000107` rows carry `UniProtKB:Q6PAL7|ensembl:ENSMUSP00000101535`. `Q6PAL7` = `AHDC1_MOUSE`, **reviewed Swiss-Prot, 1594 aa vs 1603** — a true 1:1 ortholog, not a paralog. Querying the *donor's own evidence* is what exposes the problem: QuickGO shows mouse Ahdc1 holds `GO:0001707` and `GO:0043589` **by IMP from PMID:35585237** — the same paper that already gives human AHDC1 direct annotations to both terms. The pipeline also runs the **reverse** direction, giving mouse `GO:0003700` and `GO:0140585` as `GO_REF:0000107` transfers from the human rows.

`root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT`, `failure_modes: [CIRCULAR_PROPAGATION]`, terms kept, `source_entities` built from the GOA WITH/FROM field.

**Same paper, same perturbation class, two different evidence codes.** UniProt coded the CRISPR loss-of-function experiments `IMP` on mouse and `IDA` on human. Flagged on each affected row and raised as a question; it changes no term.

## `NbExp` inflation — with a self-correction I want on the record

**My first draft of this section was wrong and I corrected it before the review landed** (commit `af3276a2`). I had written that "all twelve HTT records are the same single experiment logged twelve times", inferred from the records sharing one publication accession. Dumping the *full* IntAct records instead of the summary fields refuted it: the twelve rows carry **twelve distinct interaction ACs** and four distinct construct sets. The campaign's own rule is *derive the number, never find a story that makes the gap acceptable* — I had found the story.

Correct decomposition (73 records, 45 distinct partners, `totalElements` asserted against rows read; **2** of the 73 are RNA-level records where AHDC1 appears as `ENST00000374011`, reported rather than dropped):

| partner | records | pubs | what the records actually differ by | MI |
|---|---|---|---|---|
| ATXN1 `P54253` | 5 | 2 | PMID:16713569 → **2** records, one method, differing by **ATXN1 fragment**: 528-815 "c terminal" and 557-699 "axh region", both `sufficient to bind`. PMID:32814053 → **3** records differing **only by sub-method label**, one construct `p.Gln225[50]` | 0.67 |
| HTT `P42858` | 12 | **1** | **4 HTT constructs × 3 sub-method labels** — 1932-2642; 1-511 `Gln18[49]`; exon-1 `Gln18` at 17/20/23/49/51/79; 1-504 `Gln18[23]`/`Gln18[80]` | 0.56 |

So: the **sub-method triplication is real and is a threefold inflation** (`validated two hybrid` MI:1356 + `two hybrid array` MI:0397 + `two hybrid pooling` MI:0398 — third occurrence in this campaign after ACRV1 and ADAMTSL5), and the remaining fourfold variation on HTT is the screen's **standard huntingtin bait panel**, not a mapping done on this pair. But the records are not contentless: IntAct curates HTT 1932-2642 as `sufficient to bind` and curates some polyQ lengths as `mutation disrupting strength` and others as `mutation with no effect`, so the readout is graded.

**That changed the verdict, and the three rows now split:**

| row | action | why |
|---|---|---|
| ATXN1 (PMID:16713569) | `KEEP_AS_NON_CORE` | maps the binding region onto ATXN1's **AXH domain** (557-699); second lab replicates |
| ATXN1 (PMID:32814053) | `KEEP_AS_NON_CORE` | cross-laboratory replication; within this paper it is 3 sub-method labels on one construct |
| HTT (PMID:32814053) | `MARK_AS_OVER_ANNOTATED` | one laboratory, bait-panel design, **1,216**-partner hub, nothing tested functionally |

Neither is `REMOVE` — both are experimental IPI rows. No region is mapped on the **AHDC1** side of any ATXN1 or HTT record. Partner accessions resolve to **reviewed canonical** entries at full length (ACRV1's ORFeome-substitution check: run, negative).

## `GO:0140585` promoter-enhancer loop anchoring activity → over-annotated

The term asserts **bridging**. The evidence is a knockout (cohesin HiChIP: reduced number and size of contacts, lost contacts anchored at Gibbin-dependent promoters), the AHDC1/GATA3 spacing is offered by the authors only as a *suggestion*, and the paper's own mechanism is **indirect** — hypermethylation, elevated DNMT activity, and decreased CTCF binding at >2,700 sites, with the loop loss framed as downstream of that switch.

`GO:0140588 chromatin looping` was considered and **rejected on its definition** (loading of an SMC-family extrusion motor + chromatin extrusion) — label-reading would have shipped a wrong term. `GO:1902275 regulation of chromatin organization` proposed instead.

## `GO:0006355` → `GO:0006357`

This `GO_REF:0000108` row derives from `GO:0003700`. Verified rather than assumed: QuickGO ancestor closure over `is_a,part_of` puts `GO:0006355` among `GO:0003700`'s ancestors and **not** among `GO:0003712`'s (whose only ancestors are `GO:0140110` and the root). So the MF correction **voids this row's derivation** — but the process is independently real (~1,100 transcripts altered in the human knockout), and at the Pol II-specific level. MODIFY to `GO:0006357`, re-founded on PMID:35585237.

## Five `NEW` rows, four distinct assays

| term | code | assay |
|---|---|---|
| `GO:0003682` chromatin binding | IDA | ChIP-seq (hESC + ENCODE HepG2) |
| `GO:0140297` DNA-binding transcription factor binding (`P23771` GATA3) | IPI | BASU proximity proteomics + GATA3 epistasis |
| `GO:0040029` epigenetic regulation of gene expression | IMP | 850k methylation arrays + DNMT activity |
| `GO:1902275` regulation of chromatin organization | IMP | cohesin HiChIP |
| `GO:0000785` chromatin (`located_in`) | IDA | same ChIP-seq — added in the review round |
| (`GO:0006357` arrives via the MODIFY above, **not** as a separate NEW row) | | |

Every experiment is in **human** cells, so `IDA`/`IPI`/`IMP` are species-correct (the AHNAK trap).

**Declined, with reasons in the review:** `GO:0030216 keratinocyte differentiation` (the defect is **non-cell-autonomous** — WT mesoderm rescues GKO ectoderm and mature primary keratinocytes are unaffected); `GO:0003680 minor groove of AT-rich DNA binding` and `GO:0000987` (assert a preference nothing measured — the exact defect being reported); `GO:0090310` (excluded by the paper's own H3K9me3/HP1α negative); the **seven mouse metabolic IMP rows** from PMID:37819197 (mouse-only, would need `ISO`/`ISS`, mechanism unknown).

## Checks run that came back negative — reported as findings

- **Projection check** (ACTR8): QuickGO by reference gives PMID:35585237 → **7 annotations over exactly 2 entities** (human + mouse AHDC1), not paginated. No complex-level projection.
- **Retraction / erratum**: `CommentsCorrections` read from each cited article's own PubMed record for all 7 relied-on PMIDs. Clean; the only entry is a `CommentIn` Cell Preview on PMID:16713569.
- **Partner accessions**: all reviewed Swiss-Prot canonical, correct lengths. No TrEMBL/ORFeome substitution.
- **Sibling cross-check**: not applicable — `PTHR15617` holds **682** proteins of which the reviewed-member CSV has **2** (0.3%), human and mouse AHDC1. No human paralog exists.

## Diagnosis: coverage, not over-annotation

Derived (not hand-counted) by parsing the `RN`/`RP` blocks of the UniProt record and joining to the GOA TSV: **28** references, all with PubMed ids — **13 clinical genetics, 10 proteomics/PTM, 4 sequencing, 1 function**. Only **2 of 28** produced any GO annotation; the two remaining GOA publications are Y2H screens UniProt does not cite at all. This is AFF4's shape. Nine more case series will not produce a molecular function; one biochemical paper would.

**affinage**: the record has no `gates_passed` field (`faith_pct: 100.0`, `self_evaluation_pairwise: tie`). Precision is fine — all six citations are real numeric PMIDs that resolve to records matching their claims, none retracted. **Recall is zero for molecular function**: it never cites PMID:35585237 and asserts that "the direct molecular mechanism … has not been characterized in the available corpus", while that paper is the source of 5 of 15 GOA rows and of UniProt's entire `FUNCTION` block. It did surface one thing GO lacks on the human gene (PMID:37819197).

## Knowledge gaps

Five, including the headline `BIOLOGY`/`MF_DARK` gap the brief asked for — *no biochemical activity of any kind has been measured on this protein* — and one `ONTOLOGY` gap: `GO:0044030 regulation of DNA methylation` is **obsolete with `term_replaced_by: None`** (and `GO:0006306` likewise), so "restrains cytosine methylation at cis-regulatory elements" is expressible only at the much broader `GO:0040029`. Not filed as a `proposed_new_terms` entry, because AHDC1's own direct-vs-indirect question is unresolved and the term should be proposed from a case that exemplifies it cleanly.

## Review round — all five items from `ai4c-reviewer`

Approved with five non-blocking items. Four acted on, one answered in place; each premise checked first.

1. **Off-by-one — real, and conceded.** See the count correction above. The tell was that I *wrote* the number from a plan instead of deriving it with `grep -c '^- term:'` / `grep -o 'action: NEW'`. This is the campaign's most-confirmed lesson landing on its own author.
2. **Off-target quote — real, and the fix differs per row.** The `GO:0001707` Compara row was supported by perinatal lethality, which speaks to neither mesoderm formation nor to the circularity the row argues; replaced with the human scRNA-seq mesoderm result plus the day-3 onset, because what *this* row needs evidenced is that the target already holds the term from this publication. The `GO:0043589` Compara row carried the **identical** quote and got a **different** fix — the mouse skin data (KRT14/KRT10 loss, reduced differentiated layers) — because there the donor's evidence genuinely *is* a distinct experiment. One bad quote, two right answers.
3. **`GO:0000785 chromatin` — added, additively.** Ancestor closure checked: `GO:0000785` is a descendant of `GO:0005694` over `is_a,part_of`, so it specialises the existing row rather than competing with it. Added as a `NEW` `located_in` row on the same reference rather than re-graining the `EXP` row, which reflects UniProt's curated Chromosome location and is not wrong. Also added to both `core_functions.locations` — which cleared two "location term not reflected in existing_annotations" warnings that the core_functions-only version produced, so the additive route was the one the repo's own rules wanted.
4. **`core_functions` restructured.** The two entries now divide **by activity and by experiment** rather than overlapping: entry 1 is the partner-facing arm (`GO:0003712`, `GO:0006357`; proximity proteomics + epistasis), entry 2 the DNA-facing arm (`GO:0003682`, with `GO:0040029` **moved here from entry 1**, plus `GO:1902275` and the developmental outcomes; ChIP-seq, methylation arrays, HiChIP). Entry 2 now states the causal distance explicitly — mesoderm formation is the direct readout, skin morphogenesis is reached non-cell-autonomously, the same argument that declines `GO:0030216`.
5. **`IPI` for proximity labelling — not conceded.** Coding the row off the GATA3 epistasis was considered and rejected: the epistasis is a genetic result about GATA3's *dependence on* AHDC1, which is IMP-shaped and would support a *regulation* term, not a binding term. `IPI` on the proximity data with the labelling-radius caveat is what matches the measurement. The reasoning is now in the row, not only here.
6. **Tagged-transgene asymmetry (raised by the second, later-dismissed review) — the sharpest item, and answered rather than removed.** The same experimental limitation was weighted *against* `GO:0003700` and treated as a mere caveat *for* `GO:0003682`. The asymmetry survives being made explicit, so it is now argued on **both** rows: over-expression of a tagged chromatin protein distorts **which** sites are occupied far more than **whether** the protein reaches chromatin, and the two rows claim exactly those two different things. Two facts limit it further on the `GO:0003682` side and are cited only there, because neither rescues a specificity claim — the occupancy is **conditional** (absent day 0, present day 7, same cells, same construct) and an **independently generated** ENCODE dataset in another cell type reproduces it. The `GO:0003700` objection never rested on the construct anyway: all four of its arguments hold if the transgene were endogenous. The audit script gained an occurrence-count guard so the justification cannot silently drop from one of the two rows.

## Two methodological facts surfaced by review, both now in the review itself

**1,6-hexanediol.** The Gibbin ChIP-seq — *alone* among the ChIPs in that paper — pre-treated cells with 5% 1,6-hexanediol for 60 s before crosslinking. The word appeared **zero** times in my files before pass 5. It cuts both ways and both halves are on the `GO:0003682` row: it **strengthens** the binding call (condensate co-precipitation is the live alternative explanation for a ten-disordered-region protein appearing in a ChIP, and this occupancy survived condensate disruption) and it **filters the map** (a third independent reason not to read specificity off these peaks, alongside the ectopic promoter and the authors' own "not restricted to … Gibbin-regulated loci").

**ENCODE `ENCSR168AUX`, fetched not transcribed.** `AHDC1-bioinformatics/fetch_encode_ahdc1.py` + cached `ENCSR168AUX.json`. It is **not** an independent untagged replication: the ChIP is anti-3xFLAG. But the tag is a **CRISPR knock-in at the endogenous locus** (`ENCGM399CXU`, `category: insertion`, `method: CRISPR`, `3xFLAG (C-terminal)`), so that protein is expressed from its **own promoter at endogenous levels** — the axis the Stanford PiggyBac transgene cannot control, and the one that bears on occupancy. The script asserts both conclusions and warns if either flips. It also documents the trap that made a script necessary: an untagged-*looking* ENCODE target label is necessary but not sufficient, because the tag lives on the biosample's `genetic_modifications`.

Incidental, on-theme: ENCODE classifies AHDC1 as `investigated_as: ['transcription factor']` — a **third** database inheriting the classification from the gene's *name* rather than a measurement, after UniProt's `GO:0003700` IDA and PANTHER's `TRANSCRIPTION FACTOR GIBBIN`.

## Committed invariant check

`genes/human/AHDC1/AHDC1-bioinformatics/audit_ahdc1_claims.py` — this gene had a retracted claim *and* a narrative off-by-one, both of the same shape (a number or phrasing asserted in prose and never re-derived), so the campaign's guidance is to commit a lint. It checks GOA-row coverage by **presence** (a deleted entry fails rather than being skipped), the arithmetic `entries == GOA rows + NEW rows`, four retracted phrasings that must not reappear outside a retraction context, five required claims, and duplicate YAML mapping keys via a `SafeLoader` subclass that raises. Five self-test guards, all firing.

**It failed on its first run, and all three failures were real:** a digit-vs-word formatting difference (pattern widened); the "no InterPro2GO mapping exists" **non-confirmation** being present in the notes but **absent from the review** — so a curator reading only the YAML could not tell the absence had been checked, now added to the `GO:0003700` row naming `IPR032757`/`IPR039225`; and the claim that `GO:0003712` is a **sibling** of `GO:0003700` under `GO:0140110` — the crux of why the correction is lateral rather than a retreat to a parent — being stated in exactly one place. Written to stop a claim coming back, it caught a claim that was barely there.

## Verification

- `just validate human AHDC1` → `✓ Valid`, on the tree after merging `origin/main` (already up to date). One standing warning: no annotation cites the affinage file. That is honest — the record supported no decision — and the `file:` reference could not be added because the pre-edit hook resolves `file:` paths against `$CLAUDE_PROJECT_DIR` (the main checkout), where this new gene folder does not exist. The provider assessment is in `AHDC1-notes.md` §10 instead.
- Quote audit: **61 quotes, 0 problems**; strict duplicate-key loader clean; raw `supporting_text` lines = parsed quote sites = 56. Both self-test guards fire. Every quote was additionally pre-verified against a *live* fetch by the pre-edit hook, so they pass in both the cached and the fetched environment.
- `cache_lint` exit status `0` (tested directly, not after a pipeline), re-run after the merge. `cache/go/terms.csv`: two additions (`GO:0001707`, `GO:0043589`), **zero deletions**, **zero duplicate curies**.
- No rendered HTML committed.
- **Stale claim in a commit message:** the first commit's message still says the twelve HTT records are "one IntAct experiment ... logged twelve times". That claim is retracted; commit `af3276a2` states the correction, and the review YAML, the notes and this description all carry the corrected version. Flagging it because a squash-merge would otherwise leave the retracted wording on `main`.


cmungall:
Thanks — all five addressed in `4f732fbe`. Four acted on, one answered rather than changed. Detail is in the commit message and the updated PR body; the short version:

**1. Off-by-one — you were right, and I checked it rather than taking it.** `grep -c '^- term:'` gave 19 and `action: NEW` gave 4. Cause: `GO:0006357` was planned as a `NEW` row and implemented as a `MODIFY` replacement, and I wrote the narrative count from the plan instead of re-deriving it. Worth flagging the trap in the fix: acting on your item 3 **added** a fifth `NEW` row, so the file is now 20 = 15 + 5 and the original claim reads correct again. It was not correct when made, and the PR body and the history record both say so.

**2. Off-target quote — right, and the two rows needed different fixes.** `GO:0001707` `GO_REF:0000107` now cites the human scRNA-seq mesoderm result plus the day-3 onset, because what that row needs evidenced is that the *target already holds the term from this publication* — that is what makes the transfer redundant, and a mouse phenotype would not show it. The `GO:0043589` Compara row carried the identical perinatal-lethality quote and got the mouse skin data instead (KRT14/KRT10 loss, reduced differentiated layers), because there the donor's evidence genuinely *is* a distinct experiment. Same bad quote, two different right answers.

**3. `GO:0000785`.** Added as a `NEW` `located_in` row on `PMID:35585237` rather than re-graining the `EXP` row, with the ancestor closure checked first (`GO:0000785` is a descendant of `GO:0005694` over `is_a,part_of`, so it specialises rather than competes). Also added to both `core_functions.locations`. Useful side effect: the core_functions-only version raised two "location term not reflected in existing_annotations" warnings, so the additive route was the one the repo's own rules wanted.

**4. `core_functions`.** Restructured to divide **by activity and by experiment** instead of overlapping — entry 1 partner-facing (`GO:0003712`, `GO:0006357`; proximity proteomics + epistasis), entry 2 DNA-facing (`GO:0003682`, `GO:0040029` moved here from entry 1, `GO:1902275`, plus the developmental outcomes; ChIP-seq, methylation arrays, HiChIP). Entry 2 now states the causal distance explicitly: mesoderm formation is the direct readout, skin morphogenesis is reached non-cell-autonomously — the same argument that declines `GO:0030216`.

**5. `IPI` for proximity labelling — respectfully not conceded.** Coding the row off the GATA3 epistasis was considered and rejected: the epistasis is a genetic result about GATA3's *dependence on* AHDC1, which is IMP-shaped and would support a *regulation* term rather than a binding term. `IPI` on the proximity data, with the labelling-radius caveat stated, is what matches the measurement; the epistasis is what makes the partner worth naming, not what evidences the binding. That reasoning is now in the row itself.

---

**One correction of my own, made before your review landed** (commit `af3276a2`, so you may have read the pre-correction text): I had written that the twelve HTT `IntAct` records were "one experiment logged twelve times", inferred from a shared publication accession. Dumping the full records refuted it — twelve distinct interaction ACs, and the rows decompose as **4 HTT bait constructs × 3 sub-method labels**. So the sub-method inflation is threefold, not twelvefold, and the rest is the screen's standard huntingtin panel. That also flipped the ATXN1 verdict: `PMID:16713569` contributes two records mapping the binding region onto ATXN1's **AXH domain** (557-699, `sufficient to bind`), which with cross-laboratory replication makes `MARK_AS_OVER_ANNOTATED` the wrong call. The three `GO:0005515` rows now split **ATXN1 → `KEEP_AS_NON_CORE` ×2, HTT → `MARK_AS_OVER_ANNOTATED`**.

Noted on your validation caveat: `just validate human AHDC1` runs clean locally on the tree with `origin/main` merged, `cache_lint` exits 0, the quote audit is 60 quotes / 0 problems under a strict duplicate-key loader, and `cache/go/terms.csv` has two additions, no deletions and no duplicates.

cmungall:
All three taken, in `bdd07f1d`. Two were defects in the argument I had just written, and the ENCODE one turned a hedge into the strongest support on the row.

**1. "over-expressed" — you're right, and it was my inference.** The methods say only *"a doxycycline-inducible, HA-tagged Gibbin transgene"* on a PiggyBac vector, *"induced for 24 hours prior to crosslinking"*, and nothing compares the level to endogenous. I checked: the paper's only "over-expression" language is about **GATA3**, not Gibbin. Reworded everywhere it described the transgene → *ectopic, epitope-tagged expression from a heterologous promoter at an unmeasured level*. The argument is unchanged and slightly sharper for being stated precisely. (The two surviving uses of "over-expressed" are about the **Y2H** constructs, where it is IntAct's own `experimentalPreparations` value and therefore sourced.) You identified the exact move this review refuses everywhere else, and I made it while arguing against it.

**2. ENCSR168AUX — checked, and your worry was half right in a useful way.**

| field | value |
|---|---|
| target | `/targets/AHDC1-human/` — untagged target name, not an `eGFP-AHDC1` entry |
| genetic modification | `ENCGM399CXU`: `category: insertion`, `purpose: tagging`, `method:` **`CRISPR`**, `perturbation: False` |
| introduced tag | **C-terminal 3xFLAG** |
| antibody | `ENCAB697XQW`, targeting `3xFLAG-synthetic_tag` |
| lab / biosample | Richard Myers (HudsonAlpha) / HepG2 |

It **is** tagged, so it does not control for tagging — but the tag is a **CRISPR knock-in at the endogenous locus**, so that protein is expressed **from its own promoter at endogenous levels**. That is precisely the axis that bears on occupancy, and the one the PiggyBac transgene cannot control. So the replication is stronger than "different lab and lineage" on the axis I care about and weaker than "independent" on the axis you flagged. Both now stated, including the residual: **both datasets are epitope-tagged and no ChIP-seq of untagged endogenous AHDC1 with a validated antibody exists**, so tagging is the one axis neither controls.

Incidental but on-theme: ENCODE's own target record classifies AHDC1 as `investigated_as: ['transcription factor']`. That is a **third** database inheriting the classification from the gene's *name* rather than from a measurement — after UniProt's `GO:0003700` IDA and PANTHER's `TRANSCRIPTION FACTOR GIBBIN`. Recorded in the notes.

**3. The guard did not enforce its own docstring — correct, and it is the failure mode the script exists to catch.** Counting two matches anywhere in the file would have passed with both statements inside the same row. Replaced with a **paired-claim** check that parses the YAML, resolves `GO:0003700` and `GO:0003682` to their rows, and requires the justification in **each** row's `review.reason`. A missing row is an **error**, not a skip, so deleting the row cannot satisfy it. Two new self-test guards exercise it — `paired_claim_one_side_removed`, which mutates *through the parser* so exactly one side is thinned and asserts the mutation landed, and `paired_claim_row_deleted`. Six guards, all firing.

Thanks for the `core_functions` withdrawal and for checking the schema slot rather than leaving it as an open ask — and for re-deriving the counts rather than reading them, which is what caught the off-by-one in the first place.

Gates after these edits: `just validate human AHDC1` clean; audit `0 problems`; quote audit 60 quotes `0 problems`; `ruff` clean on the script; `cache_lint` exit 0; `cache/go/terms.csv` two additions, no deletions, no duplicates. This is my last push.

cmungall:
Both taken, in `ad1d0c3a`. The hexanediol catch is the best thing anyone has said about this gene and I had missed it entirely.

**1. Hexanediol.** Confirmed verbatim, and the scoping you flagged is right there in the string: *"**For Gibbin ChIP-seq**, freshly collected cells were treated with 5% 1,6-hexanediol in 5mL PBS in suspension for 60 seconds"* — applied to that ChIP and not to the others in the paper. "Hexanediol" appeared **zero** times anywhere in my files.

Both halves are now on the `GO:0003682` row, because it genuinely cuts both ways:

- **It strengthens the call.** A protein with ten annotated disordered regions showing up in a ChIP is precisely the case where condensate co-precipitation is a live alternative explanation for apparent chromatin binding. This occupancy survived a condensate-disrupting pre-treatment, so that explanation is excluded by the experiment rather than by assertion.
- **It also filters the map**, since any condensate-dependent occupancy was removed before crosslinking — a **third independent** reason not to read site specificity off these peaks, alongside the ectopic promoter and the authors' own "not restricted to … Gibbin-regulated loci".

The general lesson is in the notes: read the methods for the assay you are annotating, not only the results paragraph that reports it.

**2. ENCODE, now fetched rather than transcribed.** `AHDC1-bioinformatics/fetch_encode_ahdc1.py` + cached `ENCSR168AUX.json`; the RESULTS.md table is regenerated from it. The script **asserts** the two conclusions the review depends on instead of leaving them to be eyeballed, and warns explicitly if either flips — an untagged target would invalidate the "both datasets are tagged" caveat, and a tag introduced by transfection rather than knock-in would invalidate the "endogenous levels" claim that does the real work.

It also documents the trap in ENCODE's data model, which is why this needed a script: **an untagged-looking target label is necessary but not sufficient**, because the tag lives on the *biosample's* `genetic_modifications` — exactly this case (`ENCGM399CXU`, `category: insertion`, `method: CRISPR`, `3xFLAG (C-terminal)`).

Also correct on your point about provenance: ENCSR168AUX is the *paper's own* citation (Extended Data 2h–j), not something this review went looking for. That is now clear in the row.

**Gates on the merged tree:** `just validate` clean; audit `0 problems` with all six self-test guards firing; quote audit **61 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

Five passes, and every item you raised was real — including two that corrected arguments I had written that same round, and one you withdrew yourself after checking the schema. Thank you. I have nothing further to push.

cmungall:
All three taken, in `ba1f2582`. Item 3 is my own rule violated inside a script written to enforce rules, which is the most embarrassing kind.

**1. The filtering half was on the wrong row — and it is the same defect as last round.** You're right, and the tell was in my own sentence: I called filtering *"a third independent reason … alongside the ectopic promoter and the authors' own statement"* while **both of those live on the `GO:0003700` row**, where hexanediol appeared nowhere. A curator reading the MODIFY row would see four arguments and never learn of the fifth.

This is structurally identical to the tagged-transgene asymmetry you caught in pass 4: an argument that bears on two rows, stated on one. So rather than just move it, I added a **second `PAIRED_CLAIMS` entry** — both instances of this defect are now under the same guard, and the audit covers 2 paired claims over 4 rows.

**2. The reagent hedge.** Correct, and I've marked explicitly that it cannot be sourced from this paper, which says nothing about selectivity. The row now states that aliphatic alcohols also affect kinase/phosphatase activity and chromatin compaction directly, so "hexanediol-resistant" is not interchangeable with "not condensate-derived". Your scoping is exactly right and is preserved in the text: this cuts **only against the strengthening half**. The filtering half holds whatever the mechanism, because a map collected after a disruptive pre-treatment is a filtered map regardless of what the disruption acted on — so it is stated without the mechanistic hedge on both rows.

**3. `fetch_encode_ahdc1.py` reported but did not gate.** Caught cleanly — it printed the warnings and returned `0`, so a run where the tag turned out transfected would have exited clean. The verdict logic is now a separate function returning non-zero, with a `--self-test` exercising four synthetic records: the real record (must pass), an untagged modification, a transfected tag, and no modifications at all (all must fail). 4/4.

Thanks also for checking the scoping claim rather than accepting it — the *"GATA3 … CTCF … H3K9me3 … or HA for the Gibbin ChIP"* antibody line in the same paragraph is the right way to establish that the hexanediol step is Gibbin-specific, and it is better evidence than the `For Gibbin ChIP-seq` prefix I relied on.

**Gates on the merged tree:** `just validate` clean; audit `0 problems`, six self-test guards firing; ENCODE gate self-test 4/4; quote audit **61 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

cmungall:
All three taken, in `dcd1519b` — and acting on the first one exposed a guard that had been passing by luck.

**1. The hedge is now cited.** Your distinction is the one that mattered: it cannot come from *this* paper, but that is different from unsourceable. **`PMID:33814344`** (Düster et al., *J Biol Chem* 2021) — looked up in PubMed and cached, not recalled — reports kinases and phosphatases *"virtually inactive"* at the 5–10% concentrations used to dissolve condensates, and **the Gibbin ChIP used 5%**. The concentration match is what makes it a real citation rather than a gesture.

Scope discipline on it: I cite **only** the kinase/phosphatase result, because that is what the paper measured. My earlier draft also asserted that hexanediol alters chromatin compaction directly; no citation for that surfaced, so the claim was **dropped** rather than left standing unsourced. A `reference_review` records both the scope and the drop.

**2. Two ordinals for one fact.** Dropped on the `GO:0003682` side; it is counted on `GO:0003700`, where the other arguments live.

**3. Double computation.** Both the printed summary and the gate now come from a single `classify()` call.

---

**And then the self-test went red.** After those edits, `--self-test` failed on `retracted_phrasing` while the ordinary run stayed green.

The retraction-context window reached **backwards to the last blank line**. In markdown that is a paragraph; in a **YAML** file, where blank lines are rare, it can reach the start of the document — so a retraction marker *anywhere above* silently excused every later match. The guard had been firing only because no marker happened to sit above the mutation point, and item 1's edit put one there. Fixed with a bounded ±400-character window.

The generalisation is in RESULTS.md: **a scoping heuristic tuned on one file format behaves differently on another, and the unbounded version is the one that fails silently** — because it makes the guard *more permissive*, so nothing goes red. That is the one class of defect a self-test finds that reading does not, and it only surfaced because acting on your suggestion happened to move a marker.

**Gates:** `just validate` clean; audit `0 problems`, six self-test guards firing; ENCODE gate self-test 4/4; quote audit **62 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

cmungall:
Both taken in `e4294b67`, and the first is a **withdrawal rather than a hedge** — you read further into the paper I added than I did.

**1. The citation I brought in to qualify the claim refutes it.** Both statements verify verbatim in the cached full text:

> *"We found that at least 7.5% 1,6-hexanediol is required to dissolve phase-separated GST-CTD"*

> *"This holds already at concentrations of 1,6-hexanediol where the agent inhibits kinase activity while condensates are not yet dissolved"*

**The Gibbin ChIP used 5%** — below the only dissolution threshold that paper reports, and inside the range where it finds kinases and phosphatases virtually inactive. So the pre-treatment cannot be relied on to have dissolved anything, and "survived a condensate-disrupting pre-treatment" may partly mean the condensates were never disrupted.

So the condensate-control reading is **withdrawn**, not softened. The `GO:0003682` row no longer rests on it; the binding call rests on the conditional occupancy (day 0 vs day 7, same cells, same construct) and the endogenous-locus ENCODE replication, neither of which involves the reagent. The filtering half is untouched exactly as you scoped it: whatever occupancy did depend on the disrupted interactions is still missing from the map, and a pre-treatment that inactivates kinases and phosphatases at 5% is not inert either way.

Worth naming the shape, because it is the second time on this gene: **I brought in a citation to qualify a claim and did not read past the fact I wanted from it.** The threshold that overturns the claim was in the same paper's Discussion. That is this repo's own ACRBP lesson — *if you cite a PMID more than once, read its full text* — arriving a round late.

**2. `bool(run())` proved a problem appeared, not the intended one.** Correct, and the shared-round-trip point is the real risk: three of the six mutations go through a YAML round-trip, so had the round-trip itself perturbed the document, all three would have "passed" while testing nothing. Each case now asserts **which** problem appeared, via `fired_with(run(), "<expected text>")`.

Incidentally the patch script that made that change caught its own false positive — an unanchored `"bool(run()) survived"` residual check matched the new docstring *naming* the defeated idiom. Anchored to assignments. Same substring-needs-an-anchor bug the file exists to catch, one level up.

**Gates:** `just validate` clean; audit `0 problems` with six guards firing on matched text; ENCODE gate self-test 4/4; quote audit **64 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

cmungall:
All three in `86eb181a`, and your prediction about item 1 was exactly right.

**1. The withdrawn claim is now linted — and adding the lint failed immediately, as you said it would.** You were right that it is the likelier one to come back: the `GO:0003682` row still states the *intent* of the pre-treatment right before withdrawing it, so trimming that paragraph would leave the claim standing. Three phrasings added to `RETRACTED`.

The live test you identified fired: `AHDC1-notes.md` still carried *"It strengthens the `GO:0003682` call"* unqualified, ~1,700 characters from the withdrawal and far outside the ±400 `RETRACTION_WINDOW`. Fixed by **striking the bullet in place** with an explicit *superseded by §15* pointer rather than deleting it — a journal should record what was thought and why it changed — and `withdrawn`/`superseded` added to `RETRACTION_MARKERS` so an in-place strike is a recognised context. That is the resolution you suggested and it is better than either deleting the bullet or widening the window.

**2. Misread antecedent — confirmed and fixed.** In the paper *"This"* refers to the preceding sentence about impairing the phospho-mark, not to kinase inhibition, so my lead-in made it self-referential. Dropped; the subordinate clause asserts the needed fact directly and the inference is unchanged.

**3. Both unused facts added, both cited verbatim.** They do cut the same way:

- the 7.5% figure is specific to **one construct in a cell-free system**, and the paper itself says *"the concentration required to dissolve LLPS probably varies with respect to protein and experimental condition"* — a floor established in vitro, not a value that transfers;
- **duration**, which was absent from the argument entirely: the Gibbin pre-treatment ran **60 seconds**, against *"1.5% 1,6-hexanediol for 30 min"* and 1% for 2 h in the cellular studies that paper cites. Two to three orders of magnitude shorter, and independent of concentration.

That is the same lesson as last round arriving twice: this paper held the threshold, the variability caveat *and* the duration comparison, and all of it surfaced one reviewer pass at a time rather than when I first cited it.

**Gates:** `just validate` clean; audit `0 problems` — seven retracted phrasings scanned, six self-test guards firing on matched text; quote audit **66 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

cmungall:
Fixed in `b88489eb`. You're right on both counts, and the second half is the more valuable finding.

**The blocking item.** The `reference_review` for `PMID:32814053` was still carrying the pre-correction reading. Your provenance check is exactly what happened: `af3276a2` rewrote the `GO:0005515` row and the notes and never touched the references block, and no later pass re-read it. Rewritten to the corrected decomposition — four bait constructs × three sub-method labels, threefold not twelvefold — with `relevance: LOW` / `correctness: VERIFIED` left alone, as you said they should be.

The failure mode is worth naming: I corrected the *row* and treated the claim as corrected. `reference_review` is a second, independent surface asserting the same thing, and nothing in my process re-read it. That is the ACBD3 "one claim asserted at N sites with no generation relationship between them" pattern, at N=2.

**The lint could not have caught it — and that is the real finding.** Both diagnoses check out:

1. **`RETRACTED` was seeded with the wordings the fix deleted.** Such a list can only confirm that a fix happened; it cannot find the claim restated in different words, which is the failure mode that actually occurs. Your count is right: of the seven phrasings, **zero** could match anything in the review YAML.
2. **`re.escape(phrase)` against raw text made every wrapped claim invisible by construction** — and everything in the review file lives in block scalars.

Both fixed. `RETRACTED` is now **eight regexes anchored on the numbers and the semantic core** (`single\s+AHDC1-HTT\s+experiment`, `twelve-?fold\s+overstatement`), matched against **whitespace-normalised** text — which is what `REQUIRED` already did and I had not carried across.

**Proof rather than assertion:** splicing your quoted historical sentence back into a copy now fails on **two independent patterns**, and that splice is committed as the `regression_stale_reference_review` self-test guard — the only case in the file taken from a real defect rather than a mutation I invented. Seven guards, all firing.

The generalisation is in RESULTS.md: **a regression lint seeded from the diff of a fix inherits the fix's vocabulary, not the claim's meaning.** Seed it from what the claim *asserts* — preferably a number — and it still fires after someone rewrites the sentence.

You also predicted the specific consequence of the `"survived a condensate-disrupting pre-treatment would"` pattern matching nothing: deleting the withdrawal would have left the audit green. That pattern is replaced with one anchored on the assertion itself.

**🔵 The arithmetic.** You derived it and I had not: 60 s vs 30 min is 30×, vs 2 h is 120× — 1.5 and 2.1 orders, not "two to three". Corrected in place with the arithmetic shown. Your observation about *why* it stayed harmless is the useful part: the review YAML states the raw durations and draws no multiplier, so the error never reached the curation.

**Gates:** `just validate` clean; audit `0 problems` — 8 retracted patterns, 7 self-test guards firing on matched text; quote audit **66 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

cmungall:
Both in `eed26f24`. The second one is the better catch: my pass-10 fix repeated the mistake it was fixing, and you proved it from the same commit.

**1. `binary interaction` — right, and it contradicted two rows I keep.** "Binary" is the IMEx/IntAct term of art and MI:0018 is its canonical binary detection method, so the gap statement denied exactly the evidence the two `GO:0005515` rows rest on — including the sufficient-to-bind fragment mapping onto the ATXN1 AXH domain. Your point about the `boundary` field is the part that made it invisible: it listed ChIP-seq, the proximity interactome and the knockout phenotypes and omitted the two-hybrid data, so nothing inside the entry surfaced the contradiction.

Reworded to the narrower claim the same sentence already made twice (no assay on purified components), Y2H and the AXH mapping added to `boundary`, and the term-of-art collision recorded so it does not come back.

You also named the structure correctly: **this is the pass-10 finding one section over.** The IntAct correction rewrote the row and nothing re-read the *other* surfaces characterising that same evidence. Second instance of one claim living on multiple unlinked surfaces.

**2. The fix repeated the mistake — and the proof was in the same commit.** Two of the five replacement patterns were verbatim fragments of the one stale sentence, seeded from the diff of *that* fix exactly as the original four were seeded from `af3276a2`'s. And my replacement note restates the retracted reading as *"one experiment, twelvefold"* while **none of the eight patterns matched it**. Added, together with `pre-correction` / `rewritten` / `earlier version` as markers — your paired-half point is right, a note may legitimately *quote* the reading it replaced, and the pattern without the marker would have red-flagged that quotation.

Rather than assert the list is now well-founded, I made the seeding basis **auditable**:

```
uv run python genes/human/AHDC1/AHDC1-bioinformatics/audit_ahdc1_claims.py --provenance
```

It replays every past revision of the review and the notes on this branch and classifies each pattern as **HISTORIC** (matches text that really existed — fires on a literal revert) or **INVENTED** (matches nothing, ever), exiting non-zero on any INVENTED.

Running it immediately caught **one invented pattern of my own** — `twelve\s+records\s+are\s+…\s+one\s+experiment`, a paraphrase I had guessed at — which is removed. All **nine** surviving patterns are HISTORIC over **24** replayed revisions. A lint guarding imagined phrasings while presenting itself as evidence-based is this file's own failure mode one level up, so it is now checked by machine rather than claimed in prose.

**Gates:** `just validate` clean; audit `0 problems`; self-test 7/7 on matched text; provenance 9/9 HISTORIC; quote audit **66 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

cmungall:
Taken in `83a91632`, and acting on it found that my first attempt at the fix used the wrong algorithm.

**"The list only ever widens" — correct, so it now has a shrinking force.** `--markers` computes a **greedy minimal cover** (remove a marker only if the audit still passes with it already removed) and exits non-zero if anything can be dropped. It took the list from **eight markers to three**: `refuted`, `superseded`, `pre-correction`.

**And the obvious version of that test is wrong — I wrote it that way first.** Dropping each marker *in turn* and calling any whose removal leaves the audit green "inert" reports inert for markers that are **jointly** necessary: two markers covering the same window each test inert while together being the only cover.

Measured rather than reasoned, because the numbers disagreed with the story: **six of eight looked individually inert, and removing all six turned the audit red on three counts.** `withdrawn` is droppable only because `superseded` covers the same windows; drop both and the struck bullet in the notes is uncovered. Acting on the naive report would have broken the guard while appearing to tidy it — which is a neater version of the same defect you have been finding all along.

Greedy gives a *minimal* set, not necessarily the *minimum* one. That is the honest claim and it is enough: it certifies nothing can be deleted one at a time from what is committed.

Both derived properties of this lint are now machine-checked rather than asserted in prose:

| mode | property | result |
|---|---|---|
| `--provenance` | every retraction pattern matches text that really existed on this branch | 9/9 `HISTORIC`, over 24 replayed revisions |
| `--markers` | the context-marker list is minimal | 3/3 required |
| `--self-test` | each guard fires, on the *intended* problem | 7/7 |

**Gates:** `just validate` clean; audit `0 problems`; quote audit **66 quotes, 0 problems**; `ruff` clean; `cache_lint` exit 0; `terms.csv` two additions, no deletions, no duplicates.

Twelve passes, and the curation itself has not moved since pass 2 — `GO:0003700` → `GO:0003712` argued from the definition, `GO:0140585` over-annotated rather than removed, the `GO_REF:0000107` rows discounted for provenance with terms kept, five `NEW` rows, five knowledge gaps. Everything since has been you finding claims that were asserted rather than derived, in the review, in my notes, and finally in the lint I wrote to catch exactly that. Thank you — it is a markedly better review than the one I opened.

PR #3003, opened 2026-09-12
Review human AHI1

cmungall:
Full review of human **AHI1** (Jouberin, Q8N157) — a ciliary transition-zone protein and Joubert syndrome gene.

## Coverage

| | |
|---|---|
| `AHI1-goa.tsv` data rows | **37** |
| QuickGO `geneProductId=UniProtKB:Q8N157` | **37** |
| `existing_annotations` reviewed | **37** |
| `existing_annotations` proposed (`NEW`) | **3** |

Computed evidence-code breakdown of the 37 rows: **ISS 13, IPI 6, IDA 6, TAS 3, NAS 3, IEA 2, IBA 2, IMP 1, IC 1**. All 13 ISS rows transfer from a single accession, `UniProtKB:Q8K3E5` (mouse Ahi1); each donor annotation was checked individually against QuickGO and all 13 have genuine experimental evidence on the mouse entry.

The `fetch-gene` stub did **not** under-seed this gene: all five `GO:0005515` rows arrived one-per-partner, and the three Reactome `GO:0005829` rows and every duplicate-term pair kept their own entries. Coverage is asserted programmatically, keyed on (term, evidence code, reference, WITH/FROM set) — not by eye.

**Actions:** 9 ACCEPT · 21 KEEP_AS_NON_CORE · 1 MODIFY · 6 MARK_AS_OVER_ANNOTATED · 3 NEW · **0 REMOVE**.

---

## 1. Experimental evidence codes on non-human experiments — in the shipped GOA

`GO:0060271 cilium assembly` carries **IMP** from `PMID:21959375`. That paper's only experimental systems are zebrafish morpholinos and an siRNA pool directed at ***mouse*** Ahi1 in **mouse IMCD3** cells. There is no human cell, tissue or reagent in it.

The tell that makes this a defect rather than a judgement call is internal to the curation. QuickGO **by reference**:

```
reference=PMID:21959375   9 annotations, 4 entities
  GO:0034329 cell junction assembly   IMP  ->  Q8K3E5 (mouse) ONLY
  GO:0060271 cilium assembly          IMP  ->  Q8K3E5 (mouse) AND Q8N157 (human)
  GO:0005929 cilium                   IC   ->  both
```

Both readouts come from the same siRNA pool in the same mouse cells; one was restricted to mouse and the other was not. GOA **already carries** the species-correct encoding of this claim — the `GO:0060271` **ISS** row from `UniProtKB:Q8K3E5` — so the IMP row duplicates the mouse evidence under a code that asserts a human experiment.

The term is right and core, so the row is **ACCEPTed** with the defect documented. Human evidence exists in a different paper: Joubert-patient dermal fibroblasts show ~50% fewer primary cilia (`PMID:23532844`), which would make this a genuine human IMP on re-reference.

Same shape in `PMID:18633336`, which supplies **five** human IDA rows. Its abstract names the systems explicitly — cell-cell junctions / primary cilia / basal body in **mouse mIMCD3**, centrosomes from an overexpressed GFP fusion in **canine MDCK** — which is the one situation where an organism caveat is safe from an abstract-only record. None removed (four have independent human support; HPA's human-cell-line IF is the real evidence for centrosome), all demoted, mismatch reported.

## 2. A complex-level process projected onto every subunit

QuickGO by reference on `PMID:22179047` (fully paginated, 211 annotations / 36 entities):

Annotation counts and entity counts are reported separately throughout, because conflating them is exactly what inflates a projection signal:

| term | annotations | entities | experimental |
|---|---|---|---|
| `GO:0036038` MKS complex | 33 | 24 | **18 IDA across 10 entities** — one of them mouse *Ahi1* itself |
| `GO:0035869` ciliary transition zone | 20 | 19 | **5 IDA across 4 entities** (all mouse accessions) |
| `GO:1904491` protein localization to ciliary TZ | 15 | 15 | **zero — no IDA/IMP/IGI in any organism** |

The 15 entities holding `GO:1904491` are exactly `ComplexPortal:CPX-2531` plus its 14 subunits. The paper's own knockouts were *B9d1* and *Tmem231* — *Ahi1* was never perturbed — and mouse Ahi1 does not carry the term at all.

The discriminator (does the *phenotype* spread with the complex, or stay on the perturbed gene?) separates them cleanly: membership was **measured**, the process term was **distributed**. So `GO:1904491` → `MARK_AS_OVER_ANNOTATED` with a `propagation_review`, while `GO:0036038` and `GO:0035869` are kept and accepted.

**Cross-checked against a sibling rather than diverging silently.** AHI1 has no human paralogue, but it has 13 complex subunits, and `genes/human/CC2D2A` on `main` carries the **byte-identical** row — `GO:1904491 / NAS / PMID:22179047` — with `action: ACCEPT`. Both verdicts are right, and the evidence says so rather than the judgement:

| | annotations for `GO:1904491` | reaches an experimental annotation? |
|---|---|---|
| human **CC2D2A** `Q9P2K1` | **3** — IBA, IEA/Compara, NAS | **Yes** — mouse `Cc2d2a` `Q8CFW7` holds it by IMP |
| human **AHI1** `Q8N157` | **1** — NAS only | **No** — mouse and rat `Ahi1` carry it not at all |

## 3. Not over-annotation — a coverage gap

The jouberin/β-catenin Wnt axis is established by three Nature-family papers and cited by UniProt's own FUNCTION and SUBUNIT blocks at `ECO:0000269` on the human entry. GO has **nothing**:

```
QuickGO Q8N157 (human) 37 annotations -> 0 in {GO:0016055, GO:0060070, GO:0090263, GO:0008013, ...}
QuickGO Q8K3E5 (mouse) 75 annotations -> 0
QuickGO Q6DTM3 (rat)   63 annotations -> 0
```

The affinage record never mentions Wnt, β-catenin or CTNNB1 across all 31 findings — and the papers it missed are titled for *cystic kidney disease*, for *the primary cilium*, and for *a mouse model*. None for AHI1.

Two rows proposed, with the evidence codes chosen deliberately:

- **`GO:0008013` beta-catenin binding — IPI.** The co-IP was in human 293T against *endogenous* β-catenin, and UniProt marks this `ECO:0000269` on the human entry while marking RAB8A/CEND1 `ECO:0000250|UniProtKB:Q8K3E5` **in the same SUBUNIT block**. That within-block discrimination is corroborated independently: the RAB8A paper states its mouse antibody *cannot see human AHI1*, which is why HEK293 was chosen. Caveat recorded — the founding paper's Methods are online-supplement only, so the construct's species is not in the accessible text.
- **`GO:0090263` positive regulation of canonical Wnt signaling pathway — ISS, not IMP.** The *in vivo* work is entirely `Ahi1−/−` mouse (BATgal reporter, lithium rescue). The claim is also bounded by its neighbouring sentences: jouberin **is not a pathway activator** — expressing it alone gives no activation — it amplifies an already-initiated response 1.6-fold. `GO:0090263`'s definition accommodates an amplifier; a term implying initiation would not.

## Other resolutions

- **`GO:0044458` motile cilium assembly (IBA)** → MARK_AS_OVER_ANNOTATED. Sole non-node donor is zebrafish `ahi1` (**TrEMBL only** — no Swiss-Prot entry), holding the term by **one** morpholino IMP where the cilia lost were **Kupffer's vesicle** cilia. Human AHI1 has no motile-cilium data, already carries `GO:0097730 non-motile cilium`, and already carries the parent `GO:0060271`.
- **Three `GO:0005829` cytosol TAS rows** → non-core. Querying Reactome participants shows AHI1 appears in all three reactions **only inside** `Tectonic-like complex [cytosol]` / `basal body:transition zone proteins [cytosol]` — Reactome has no transition-zone compartment. `R-HSA-5626681` alone supplies cytosol to **87 distinct entities**.
- **`GO:0008104` → `GO:0061512` protein localization to cilium** (MODIFY). The donor's own evidence is opsin mislocalisation from the photoreceptor outer segment, and the replacement term has direct *human* support (ARL13B in 39% of patient-fibroblast cilia vs 85%) — the rare case where the replacement is better evidenced than the term replaced.
- **`GO:0043066`, `GO:0045944`, `GO:0050795`** → over-annotated. The apoptosis one is refuted by its own source: photoreceptor death is *delayed by halving opsin dosage*, so it is two steps downstream of the structural defect.
- **`GO:0030674` protein-macromolecule adaptor activity — NEW, ISS.** "Scaffold" is asserted everywhere from domain content alone (in one paper it is an Introduction citation to someone else's review). Exactly one study tests bridging with the correct reciprocal control — OTUD1–TYK2 association is abolished in AHI1-knockout macrophages, while OTUD1 loss does not affect AHI1–TYK2 — and that assay is in **mouse** macrophages, hence ISS. The BCR-ABL/JAK2 ternary complex was checked and **rejected** as adaptor evidence: no dependency test exists.

## Checks run, including the ones that came back negative

- **Retraction / erratum / expression of concern**: 21 PMIDs checked against `CommentsCorrections/RefType` on each cited article's own record (the only way a Publisher Correction is discoverable). **All clean.**
- **Partner identity** (the ACRV1 check): 5/5 resolve to reviewed canonical Swiss-Prot entries with matching lengths. No TrEMBL or ORFeome substitutions.
- **Paralogue transfer**: impossible — `PTHR44499` has a single subfamily whose reviewed members are exactly human, mouse and rat AHI1. (3 rows is the **Swiss-Prot subset**; cached family total is `proteins: 1838`.)
- **Sibling-review consistency**: run and it *found* something — see the CC2D2A comparison above. My first pass reported this check as "not applicable, no paralogues", which was the wrong frame: a complex subunit has siblings even when a gene has no paralogue.
- **Self-audit of my own numbers**: found one wrong count (I wrote "14 of 37 rows are ISS"; it is 13) and two annotation-vs-entity conflations, all fixed in the second commit. The defect in each case was a number that was never computed, not a number computed wrongly.
- **`NbExp` counts sub-methods, not experiments**: HAP1 `NbExp=3` is **one** study logged as *2 hybrid* + *anti tag coip* + *molecular sieving*. Third instance in this campaign.
- **Downward-MODIFY check on `GO:0042802`**: not warranted — a direct human IPI at the correct term, not a propagation.
- **Stub under-seeding**: absent here.

## Verification

- `just validate human AHI1` → **✓ Valid**, 1 warning left standing (no annotation cites the affinage record — deliberate; it is assessed in `references[].reference_review` and used as evidence nowhere).
- **72/72** `supporting_text` quotes verbatim, `raw == parsed`. **Zero `file:` quotes in the review** — the one unvalidated fabrication surface is simply not used.
- `cache_lint` exit **0**, no `cache/` changes, no duplicate curies.
- Committed invariant check `AHI1-bioinformatics/audit_ahi1_claims.py` (strict duplicate-key loader, GOA-row bijection, WITH/FROM-derived `source_entities`, no `PENDING`, reference declaration, raw-vs-parsed arithmetic) with a 5-mutation `--self-test`. **The self-test found a real bug in the check itself**: a `KeyError` that aborted every later check while the harness still printed as though it had run.


cmungall:
Thanks — all five addressed in `d5a279c`. Four adopted; suggestion 1 I checked rather than conceded, and it does not hold.

## 1. NPHP1 `NbExp=4` — the original claim is correct

Your premise was checkable, so I re-queried IntAct **per record**, resolving each to its PubMed id instead of the internal `EBI-*` experiment id my first pass had printed:

```
NPHP1 (O15259): 4 records
  EBI-9662903  molecular sieving  pubmed=23532844
  EBI-9662916  molecular sieving  pubmed=23532844
  EBI-9662962  anti tag coip      pubmed=23532844
  EBI-9661010  anti tag coip      pubmed=23532844
  distinct PubMed ids: ['23532844']
```

4 records, 2 methods, **1 paper**. Every one of the five partner pairs resolves the same way.

But you were pointing at something real, and the reconciliation is now in the row:

- **`NbExp` counts records, not methods.** That is exactly why NPHP1's 4 does not equal its method count while HAP1's 3 happens to — I had left that coincidence unexplained.
- **`PMID:18633336`'s NPHP1 evidence is missing from IntAct because that GOA row is `assignedBy: UniProt`** — curated straight from the paper without being lodged in IntAct — while the other is `assignedBy: IntAct`. So the two GOA rows *are* independent evidence even though IntAct's count is not. The apparent arithmetic problem is an assigner artefact, not a counting error.

Also recorded in the notes: reading `publicationIdentifiers[0]` returned an `EBI-*` token that says nothing about how many papers are involved — the `size=1` trap in another guise, a lookup answering confidently about the wrong thing.

## 2. `GO:0090263` — now an ordered two-step ask

You're right that it was unimplementable as written. Both the row's `reason` and `suggested_questions` now spell out the order: mouse `Ahi1` (`Q8K3E5`, `MGI:87971`) gets `GO:0090263` **IMP** from `PMID:21623382` first (BATgal cross, reduced midline proliferation, lithium rescue); the human ISS follows from a donor that actually holds the term. Noted that `GO:0008013` is independent of the ordering and can go straight onto the human entry, since UniProt already records it there at `ECO:0000269`.

## 3. Adaptor MF marked as inferred-by-transfer

`core_functions` entry 1 now states it in the `description` rather than only in `knowledge_gaps`: the only bridging assay in the AHI1 literature is OTUD1–TYK2 in mouse macrophages, and what is established *at the ciliary base* is the position, the partner binding and the requirement — not the mechanism.

## 4. `GO:0042802` and `GO:0097730` now in `core_functions`

`GO:0097730` added to entry 1's `locations`. Self-association got its **own** entry rather than being tucked into an existing one, and I think that is the right call on the evidence rather than just a relocation: **V443D destroys the AHI1:NPHP1 heterotetramer while leaving self-association intact**, so they are separable properties, and it is one of only two molecular activities measured on the human protein in human cells.

## 5. `GO:0007169` — the line it sits on

Fair hit; the reason read like an over-annotation argument. The test I am applying is not *how far downstream* but **whether the annotated process is one the protein's measured activity acts on**.

AHI1's established activity is controlling where membrane proteins end up (ARL13B and MCHR1 to the cilium, RAB8A retained at the basal body). Endocytosed TrkB being degraded instead of sorted onward is that same activity one step out, so the pathway it feeds is a legitimate if generic `involved_in`. `GO:0045944` and `GO:0050795` differ in **kind**, not degree: RNA Pol II transcription is not a process AHI1 touches by this evidence chain (no DNA-binding or chromatin-associated domain), and regulation of behavior is an organismal readout several causal layers up. `GO:0002092` is held for a third reason — possible direction mismatch between the term and what the abstract describes, on an abstract-only cache — so it is deferred to the curator rather than decided.

---

On your deferred checklist item: `just validate human AHI1` is **✓ Valid** in my worktree (1 warning left standing — no annotation cites the affinage record, which is deliberate). Also re-run after every edit: the committed audit and its `--self-test` (exit 0), 72/72 `supporting_text` quotes verbatim with `raw == parsed = 42`, notes quotes 34/34, and `cache_lint` exit 0 with no `cache/` changes and no duplicate curies.

cmungall:
All four fixed in `7f4b87d`. Every one was real, and all four sat inside the `GO:0042802` entry I wrote *last round* — the newest prose was the weakest part of the file.

They share one shape: **I reported the authors' interpretation as their result.**

**1. The oligomeric state is hedged, and I asserted it.** You're right, and the paper argues against my reading twice over:

> "which **could correspond to** an AHI1 homodimer. **However, the molecular weight of the peak is smaller than the expected weight of the dimer**"

A hedge followed by an argument against itself — and the "tetramers and dimers of AHI1" phrase I leaned on is in the **V443D mutant lane**, not the wild-type one. This is the quote-to-the-end-of-the-interpreting-clause rule: stopping before "However" converts a hedged non-result into a finding. The entry now separates the **activity** (established — co-IP of two differently tagged copies in human cells) from the **stoichiometry** (not established), and the `knowledge_gap` says which is which.

**2. A verbatim quote attached to the wrong claim.** The AHI1–NPHP1 co-migration sentence evidences the heterocomplex, not self-association. Worth naming the class: **every mechanical check in this repo validates a quote against its *source*; none validates it against the *claim* it is attached to**, so a correctly-quoted, correctly-attributed sentence supporting the neighbouring assertion passes every gate. Replaced with the sentence that carries both the point and the separability argument:

> "These results demonstrate that AHI1 is capable of self-association; however, the V443D mutation in AHI1 did not disrupt the self-association of AHI1"

**3. "Retained" asserted a mechanism the data cannot distinguish.** PMID:19625297 reports Rab8a levels *fall* and it *cannot be detected* at the basal body — which separates neither failed recruitment from failed retention, nor localisation from stability. Now "accumulate", with the ambiguity named rather than smoothed over.

**4. UniProt's hedge dropped in transit.** UniProt writes "**probably** as heterodimers and/or AHI1(2):NPHP1(2) heterotetramers"; my `description` had "assemble into a **defined** heterotetramer and heterodimer". Corrected at all four sites where a stoichiometry appears, applied with a script that asserts each anchor is present before replacing **and absent afterwards** — the "fixed in N places, landed in N−1" guard.

---

I've recorded the generalisable version in `AHI1-notes.md`: this review's oldest and most-checked prose was fine, and every defect this round was in text added one round earlier which had been gated only by *"is the quote verbatim?"*. **A block written to satisfy a reviewer deserves the same scepticism as the original draft, not less** — and I had given it less.

Gates re-run after the edits: `just validate` **✓ Valid** (1 warning left standing), audit + `--self-test` exit 0, **74/74** quotes verbatim with `raw == parsed`, notes quotes 36/36, `cache_lint` exit 0 with no `cache/` changes and no duplicate curies.

cmungall:
Fixed in `48d678c`. You're right, and this is the most useful catch of the three rounds — the diagnosis is more important than the edit.

## The IMPORTANT item

`AHI1-ai-review.yaml:971-972` did still carry the retracted phrase, and your account of *why* is exact: the round-2 script **did** assert each anchor present-before and absent-after, and it **did** report four edits applied. It verified the four sites it was handed. **The four sites had been enumerated from memory rather than from a search.**

So the failure mode the guard exists to prevent moved one level up, into the guard's *input*. A guard over a hand-supplied site list inherits the incompleteness of that list — the same shape as a guard inheriting the scope of the claim it checks.

And you're right that line 978 was the one that mattered: the rejection of a MODIFY to `GO:0042803` rested **entirely** on the retracted claim. The conclusion survives and is stronger — an unresolved copy number is a reason not to narrow to a dimer-specific term — but restated on the ground that now holds, and both `supporting_text` quotes on the row now carry the hedge.

## The fix is structural, not another careful pass

`audit_ahi1_claims.py` gains two checks:

- **check 7** — scans for every retracted phrasing across **both** the review YAML and the notes. It **takes no site list as input**, so it cannot be defeated by an incomplete one. The notes are scanned with double-quoted spans stripped, so the write-up can quote the retracted phrases while documenting them.
- **check 8** — the **replacement** claims must be present. A retraction with nothing in its place is a deletion, and check 7 alone passes on that silently.

Check 8's first version was itself defective and the self-test caught it: it took an **OR-list of alternatives**, so deleting one site still passed because a sibling phrase elsewhere satisfied the list — *"a guard defeatable by deleting the thing it guards"*, reproduced verbatim, one commit after writing the fix for a related failure. Rewritten as one required string per site with its own minimum count. Self-test is now 7/7, and **two of those seven guards were found broken by trying to break them, neither by reading them.**

## Your two non-blocking items — both verified, both right

**1. My RAB8A caveat over-corrected.** I had written that the abundance drop means "some of the effect may be stability rather than localisation at all". The paper controls for exactly that:

> "Diffuse expression of HA-Rab8a was found throughout much of the cytoplasm in control, scramble and Ahi1-knockdown cells; however, unlike control and scramble cells, HA-Rab8a **was never found to localize to the basal body in Ahi1-knockdown cells**"

Overexpression restores abundance and localisation still fails, so the stability alternative is **excluded**, not open. Removed; the recruitment-versus-retention ambiguity is real and kept. Worth naming the general point: **a hedge can be wrong in the cautious direction, and that is still wrong** — I had traded one unsupported assertion for another.

**2. HAP1 hedged harder than NPHP1** although both rest on the same kind of evidence and the same authorial hedging. Levelled.

**3.** `AHI1-notes.md` NbExp table carried the same retracted "defined stoichiometry" framing — corrected, and it is now inside check 7's scan surface rather than outside it.

---

Gates: `just validate` **✓ Valid** (1 warning left standing), audit exit 0 with checks 7 and 8 green, `--self-test` **7/7**, **75/75** quotes verbatim with `raw == parsed = 43`, notes quotes 36/36, `cache_lint` exit 0, no `cache/` changes, no duplicate curies.

cmungall:
Both points are correct, and I'm **declining to push for them** — recording the reasoning rather than spending a round.

## Why this one stops here

Campaign data from the AFP review that just finished: it ran to **13 rounds**, and its own assessment of rounds 5–13 was that they changed nothing a curator would act on — fixes-to-fixes in prose and in its audit script — while each push reset the quiet window and kept the PR unmergeable throughout.

The test I'm now applying before pushing: *would a curator reading the merged YAML act differently?* Concretely — does it change an `action`, a `term`, an `evidence_type`, a `supporting_text`, or a `core_functions` entry?

For this suggestion: **none of the five.** The `action` stays `KEEP_AS_NON_CORE`, and as you say, "nothing rests on the word; the non-redundancy conclusion survives on the hedged version." The row's own `summary` two lines above already carries the correct hedge — *"co-migrate as 430 kDa and 210 kDa species, which the authors read as a heterotetramer and a heterodimer"* — so a curator reading that row gets the hedged version first and the conclusion is unaffected either way.

So I'm leaving `reason:` line 170 as-is and recording here that **"supplies the defined stoichiometry" is a residual instance of the retracted class**, hedged correctly in the same row's summary. If a blocking item ever reopens this PR, it goes in that push.

## Your methodological point is the better half, and it's right

> moving a guard's input from sites to phrases narrows the incompleteness without closing it; a regex over the class would close it

That is the round-3 diagnosis generalised one step further than my fix went, and it is correct. `RETRACTED_PHRASES` is a literal list enumerated from the four known retractions, so `"the defined stoichiometry"` matches none of its four stoichiometry entries — the same incompleteness, one level down, exactly as the site-list was one level up from the anchor guard.

And your second observation is the sharper one: **check 7's self-test reintroduces a string that is already in the list, so it proves the lookup works but cannot detect that the list is short.** That is the "a self-test proves the guards you thought of fire; it cannot tell you which guard you failed to write" limit, instantiated. A regex over the class — `\bdefined\b` within N tokens of `stoichiometr|heterotetramer|heterodimer|dimer|tetramer` — would close it, and a self-test mutation using a phrase deliberately *not* in the list would be the one that could detect shortness.

I'm not pushing that either, for the same reason: it is tooling, and it changes nothing a curator acts on. But it belongs in the campaign brief rather than only in this thread, so I'm carrying it back to the coordinator as a generalisable finding — **a guard's input can be narrowed indefinitely without ever being closed; only a check over the *class* rather than over *instances* terminates that regress**, and a self-test drawn from the same list cannot measure the list's completeness.

---

Unless something blocking appears, this is my last push on #3003. Final state: `just validate` **✓ Valid** (1 warning left standing), audit exit 0 with checks 7 and 8 green, `--self-test` 7/7, **75/75** quotes verbatim with `raw == parsed = 43`, notes 36/36, `cache_lint` exit 0, no `cache/` changes, no duplicate curies.

PR #3005, opened 2026-09-12
Review human AGGF1

cmungall:
## AGGF1 — the gene name makes three claims, and they do not all survive

"Angiogenic factor with G-patch and FHA domains 1" asserts an angiogenic
function plus two domains. Tested separately, they come out differently.

| claim | verdict |
|---|---|
| **angiogenic factor** | supported — but the activity maps to residues 604–613, a motif in **neither** named domain, acting through integrin α5β1 |
| **G-patch domain** | intact and canonical (8/8 conserved columns). The GO term derived from it is nonetheless a domain-name mapping no G-patch protein anywhere holds experimentally |
| **FHA domain** | every phosphothreonine-anchoring residue retained — and **not one phosphopeptide has ever been measured**. Untested, not refuted |

**Reconciliation:** GOA has 24 rows; the `fetch-gene` stub seeded 24 entries with
per-partner WITH/FROM intact (no collapse on this gene). The review has **33**
entries = 24 + 9 `NEW`. Asserted programmatically by
`AGGF1-bioinformatics/reconcile_goa.py` on the full
`(term, evidence, reference, WITH/FROM)` key, not on row counts.

**Actions:** 10 ACCEPT · 10 KEEP_AS_NON_CORE · 3 MODIFY · 1 MARK_AS_OVER_ANNOTATED · 9 NEW.

---

### 1. `GO:0003676` is the InterPro2GO output for the whole family and never a measurement

`domain_residues.py` pulls every reviewed human protein with an annotated G-patch
(`IPR000467`) — **34 domains across 33 proteins** — aligns each to AGGF1's
domain-local window and computes per-column agreement. AGGF1 matches **8/8** of
the columns conserved at ≥80%, and all 8 are identical to the corresponding
residues of **NKRF**, whose G-patch is the one resolved in complex with DHX15
(PMID:32179686). The domain is real.

The term derived from it is not:

```
G-PATCH REFERENCE CLASS SUMMARY (controls DHX15/DHX16 excluded): 12 queried, 11 reported;
  11 carry GO:0003676 nucleic acid binding;
  11 carry it by IEA and NOTHING ELSE;
  0 carry it with any experimental code;
  6 carry GO:0003723 RNA binding (or a child) with an experimental code.
```

Across NKRF, GPKOW, RBM17, SUGP1, CMTR1, GPATCH1, GPATCH4, RBM5, PINX1, ZGPAT and
AGGF1, not one instance of `GO:0003676` rests on an experiment. Where anyone *has*
measured, the answer lands on the child term. That matches what the domain is now
understood to be — a helicase-activation tether, not an autonomous
nucleic-acid-binding module.

**MODIFY → `GO:0003723 RNA binding`**, which AGGF1 has in its own right by RNA-IP
against NEAT1 (PMID:35608889). Not REMOVE: the conclusion survives on better
evidence at a better term. A **`NEW` `GO:0003723` IDA row from PMID:35608889** is
added alongside, so the experimental result is not left behind an electronic code
whose InterPro2GO provenance could not have produced the replacement term.

### 2. The FHA is intact at every anchor — and nothing has been measured

Anchors from CHEK2 and Rad53 FHA1, verified against the live sequences and
transferred by **domain-local** alignment:

| anchor | → AGGF1 | identical | inside AGGF1 FHA (434–487) |
|---|---|---|---|
| CHEK2 G116 / R117 / S140 / N166 | **G437 / R438 / S454 / N478** | yes | yes |
| Rad53 FHA1 S85 / N107 | **S454 / N478** | yes | yes |
| Rad53 FHA1 G69 / R70 | gap | — | — |

6/8 strict, and the two independent anchor sets **agree exactly wherever both
resolve**. The tumour data agree on the same surface: of six COSMIC somatic FHA
variants in PMID:33069768, **G437E** — the conserved Gly of the pThr-anchoring G-R
pair — impairs function, while **R447Q** (not the conserved Arg; R438 is) and
**V497I** (outside the 434–487 core) do not.

And yet: **no phosphopeptide, phosphothreonine or phospho-dependence experiment on
AGGF1's FHA exists.** Both published FHA functions — p53 (PMID:33069768) and
14-3-3α/β (PMID:33471274) — are plain co-IP with deletion constructs, no
phosphatase, no phospho-site mutant, no competition. No PDB entry for any part of
human AGGF1. So this is reported as **untested, not refuted**, and no
phosphoprotein-binding term is proposed.

*(A first pass globally aligned full-length AGGF1 to CHEK2 and mapped every anchor
into AGGF1 residues 112–169 — a clean-looking "0/8, the FHA is degenerate" result
that was pure alignment noise. The script now refuses to align anything but domain
windows, and the docstring records why.)*

### 3. The human genetics that named the gene does not hold up

AGGF1 is "the Klippel-Trenaunay gene" because of E133K. That is no longer true:

- **PMID:16443853** — *"Nine (3.3%) of the controls carried the E133K sequence change, indicating that this variant has reached polymorphic frequency in population controls"*, and the one affected carrier inherited it from **a clinically normal parent**.
- **PMID:17103452** — reported as a polymorphism in a second population.
- **PMID:18564129** — the **originating laboratory's own** larger study, conceding the AGGF1–KTS relationship was not firmly established.
- **ClinVar** classifies `NM_018046.5(AGGF1):c.397G>A` **Benign/Likely benign**, multiple submitters, no conflicts.

**UniProt still presents it as functional** (`FT VARIANT 133 /note="E -> K (displays
a stronger angiogenic activity…)"`), filed as a correction to report. None of the
GO rows depend on it, which is why they are all accepted — the biochemistry of the
wild-type protein was measured separately and stands.

**None of these three papers is in the affinage record or in GOA.** Nor is
PMID:33471274, which is *titled* "FHA domain of AGGF1 is essential for its
nucleocytoplasmic transport and angiogenesis" — i.e. directly on one of the three
questions. Consistent with the campaign's finding that the trust gates measure
precision, not recall.

### 4. The defect here is coverage, not over-annotation

Querying QuickGO **by reference** for all 35 relevant papers:

```
SUMMARY: 35 papers; 22 produced NO GO annotation anywhere;
         3 annotate other genes but nothing on AGGF1.
```

Exactly **two** papers ever produced an AGGF1 annotation: the 2004 discovery paper
and a 2005 review. Invisible to GO: the receptor (α5β1), the FAK–Src–AKT pathway,
paraspeckles, splicing of 436 genes, FHA-dependent p53 stabilisation,
nucleocytoplasmic transport, autophagy.

AGGF1 has **no nucleus annotation at all**, against three independent lines from
three laboratories — and against its own interaction record, in which **three of the four HuRI partners — DHX15, MCRS1, FBXO28 — carry `GO:0005634`
nucleus with experimental evidence** (the fourth, MAB21L3, has no
cellular-component annotation at all).

9 `NEW` rows: `GO:0005634` nucleus, `GO:0042382` paraspeckles, `GO:0003723` RNA
binding, `GO:0000381` regulation of alternative mRNA splicing, `GO:0005178`
integrin binding, `GO:0048018` receptor ligand activity, `GO:0051897` positive
regulation of PI3K/AKT, `GO:0010508` positive regulation of autophagy,
`GO:0010595` positive regulation of endothelial cell migration.

### 5. Almost all of this literature is one laboratory

Added after review. `lab_independence.py` reads the senior author out of each
cached record rather than taking independence on impression:

`lab_independence.py` **derives** its paper set (GOA references + affinage
citations + the review's references, minus high-throughput screens and
corrections) rather than using a hand-written list — the first version did use
one, omitted PMID:29885663, and therefore reported "exactly one independent
group" about its own list rather than about the literature. Over the resulting
**27 cached AGGF1 primary studies**:

| | count | |
|---|---|---|
| dominant group | **18** | Wang Q / Wang QK |
| trainee/collaborator lineage | **5** | Tian XL ×2, Xu Y, Zhou B, Lu Q — each an author on a Wang-group paper |
| senior author on **no** Wang-group paper | **4** | Qi J, **Zhang JH (29885663)**, Liao S, **Chen L (39905000)** |

Two of those four are substantive. An earlier draft also hedged the paraspeckle
paper by anchoring the nucleus proposal on PMID:33069768 "instead" — which has the
**same senior author**, so the hedge was empty. Withdrawn. The review leans on
PMID:39905000 for TNFSF12 and the extracellular pool, and cites PMID:29885663 on
the PI3K/AKT row; the paraspeckle and splicing results remain entirely
single-group.

### 6. Per-partner judgement on the binding rows

`intact_partners.py` paginates IntAct to completion (**742 records, 695 distinct
partners, 22 publications**) and expands `NbExp` into independent experiments:

| partner | records | independent experiments |
|---|---|---|
| **O43143 DHX15** | 3 | **1** — one HuRI Y2H as array + prey-pooling + "validated two hybrid" |
| **Q96EZ8 MCRS1** / **Q9NVF7 FBXO28** | 3 each | **1** each — the same screen |
| **Q8N8X9 MAB21L3** | 5 | **2** — Y2H + anti-tag co-IP in human U2OS |
| **Q9UL45 BLOC1S6** | 3 | **2** — Y2H + BioPlex co-IP in HEK293T |
| **Q8N302 self** | 3 | **2** Y2H — the third is a BioID artefact (below) |
| **O43508 TNFSF12** | **0 in IntAct** | Y2H + GST pull-down + co-IP (PMID:14961121), re-confirmed by co-IP in human retinal EC (PMID:39905000) |

- **DHX15 → MODIFY to `GO:0017151` DEAD/H-box RNA helicase binding.** The same
  measured fact stated informatively, with **no** claim that AGGF1 activates the
  helicase — a 2026 review still lists AGGF1's interacting helicase as *Unknown*
  (PMID:42052570), and the evidence is one screen.
- **TNFSF12 → MODIFY to `GO:0019955` cytokine binding.** `GO:0043120 tumor necrosis
  factor binding` was considered and **rejected on its definition** ("a
  proinflammatory cytokine produced by monocytes and macrophages"), even though GO
  places `GO:0038057 TNFSF11 binding` beneath it — that inconsistency is raised
  separately.
- **`GO:0042802` from PMID:39251607 → MARK_AS_OVER_ANNOTATED.** IntAct gives AGGF1
  `experimentalRoleA=bait` in **all 635** proximity records, and a BirA\* fusion
  biotinylates its own polypeptide, so recovering the bait in its own streptavidin
  pull-down carries no information about self-association. The two Y2H
  self-interaction rows are kept — there, both fusions have to associate.
  That one screen is also what makes AGGF1 look like a 695-partner hub: **634 of
  742 records** come from it.

### 7. Proposed new term

**`TNFSF12 binding`**, modelled verbatim on the existing `GO:0038057 TNFSF11
binding`, with a note that `GO:0043120`'s definition is narrower than its own child
placement and should be settled at the same time.

### 8. Checks that came back negative, recorded because they ran

- **IBA donors both sound.** `ZFIN:ZDB-GENE-060929-836` → zebrafish **aggf1**
  (Q08CK9) — a true orthologue carrying its own `GO:0001525` **IMP**. The
  `GO:0005576` IBA is self-referential (`NO_FAILURE_CORE`). No paralogue transfer,
  no propagation landing above its donor.
- **All seven GOA partners** resolve to **reviewed Swiss-Prot canonical** entries
  at canonical lengths — no TrEMBL/ORFeome substitution.
- **Errata**: 22 PMIDs checked via `CommentsCorrections/RefType` (a Publisher
  Correction is not findable by pubtype search). Two flagged — PMID:39468017 (a
  missing author affiliation) and PMID:41039152 (a subscript typo in a Methods
  equation). Scope established by reading both; neither touches any evidence used.
- **Projection check**: no projection pattern behind any `NEW` BP row —
  PMID:27513923 and PMID:34551592 produced zero GO annotations anywhere. It did
  sharpen the coverage point: PMID:40035560's ten annotations are **all
  `GO:0048514` blood vessel morphogenesis in zebrafish**, so a paper whose finding
  is that AGGF1 is a general splicing factor produced ten annotations, none about
  splicing in any species.
- **Duplicate-term check**: before proposing `TNFSF12 binding`, all **90**
  children of `GO:0019955`, `GO:0043120` and `GO:0005102` were **enumerated, not
  searched** — ontology search is token-based and could not match a term spelling
  the ligand out in full. None names or defines TNFSF12/TWEAK.
- **Predicted `PSEUDOENZYME_OVERANNOTATION`: not confirmed, and could not be.**
  AGGF1 carries no catalytic MF term in GOA at all. The name-derived over-reach on
  this gene is a *binding* term.
- **Isoform caveat, computed:** 24 IntAct records are logged against `Q8N302-2`.
  Fetching each isoform gives 714 / **109** / **176** aa — so isoforms 2 and 3
  contain neither domain nor the 604–613 motif. None of those records reached GOA.

### Verification

`just validate human AGGF1` clean, no warnings. All **77** `supporting_text`
quotes checked with a strict duplicate-key loader (raw 77 = parsed 77), including
the **6 `file:` quotes CI does not check** — one was copied from console output
rather than `RESULTS.md` and was fixed. All 13 PMID quotes in `AGGF1-notes.md`
verified verbatim. All 28 reference titles synced from the cached records and
re-verified from disk. `cache_lint` clean (exit status tested directly, not after
a pipeline); one `terms.csv` addition, zero deletions, zero duplicates.
`audit_claims.py` guards every reported number against drift — 11 guards, each
exercised by `--self-test`, which found and fixed one no-op mutation of its own.


cmungall:
Both blocking items conceded in full; both were right, and #2 is sharper than stated. Pushed in `30bd2d3` (with `d7251fef`, which was written before the review landed and already covers one of the 🔵 notes).

### 🟡 #1 — the fifth phrasing

Fixed. The summary now reads "against three of the four partners its HuRI two-hybrid screen returned, which carry `GO:0005634` experimentally themselves", which is the verified statement.

You also identified the real defect, which is the guard's design: **enumerating phrasings of a retracted claim is a losing game**. I matched four literal strings and you found a fifth. `audit_claims.py` now carries regex patterns for the *claim* rather than for spellings of it —

```
(?:all four|every)[^.]{0,80}partner[^.]{0,80}(?:is|are)\s+nuclear
(?:three|four|several)\s+(?:independent\s+)?(?:later\s+)?laborator
```

— applied to all three files with whitespace normalised (so YAML wrapping cannot hide one), backing up the literal strings rather than replacing them. Both are exercised by `--self-test` with wordings no literal string covers.

One consequence worth flagging: the notes now **paraphrase** the retracted wording instead of quoting it, because a guard that has to distinguish a retraction from a report of one is a guard that will eventually let the claim back in.

### 🟡 #2 — lab independence: I measured it, and it is worse than "two groups"

`lab_independence.py` (new) reads the senior author out of each cached record. Of the 13 papers this review relies on:

| senior author | papers |
|---|---|
| **Wang Q / Wang QK** | **11** — 14961121, 33069768, 35608889, 34551592, 27513923, 27522498, 40035560, 23197652, 24277077, 35202649, 37081014 |
| Tian XL | 1 — 33471274; first author of the discovery paper, so a separate group but the same lineage |
| **Chen L** | **1 — PMID:39905000** |

So **exactly one independent group**, and your diagnosis of the hedge is exact: the `reference_review` for PMID:35608889 pointed at a paper with the same senior author. Withdrawn and replaced with the honest version, which — as you say — is the more interesting finding. The nucleus proposal still anchors on PMID:33069768, but now for the stated reason that its imaging is in a named human cell line with full text cached, **not** because it is independent.

`:482-484` and `:119-122` are corrected the same way, naming which results are same-lab continuations. The script fails if the review text reintroduces a laboratory count the author lists contradict, so this stays a measurement rather than an assertion — which is the standard the rest of the review was already held to.

### 🔵 The `GO:0003676` MODIFY left the experimental evidence behind an electronic code

Agreed, and fixed as you suggested. A **`NEW` `GO:0003723` IDA row from PMID:35608889** now sits alongside the MODIFY: same term, correct code, correct reference. You are right that InterPro2GO does not map IPR000467 to GO:0003723 and could not have produced it, so the MODIFY alone would have recorded the NEAT1 RIP with false provenance. A curator who prefers to leave the electronic row at its generic parent can now do so and still keep the specific term.

On the related point — RIP cannot separate direct binding from recruitment via NONO/PSF/HNRNPK — I have put that caveat into the new row's `reason` so it travels with the annotation, pointing at the existing `knowledge_gap` rather than silently resolving it. I kept `GO:0003723` as `core_functions[0].molecular_function`: it is the best available term, the gap is declared, and demoting the core MF while proposing the row would be inconsistent. Counts move to **33 entries / 9 NEW**, asserted by `reconcile_goa.py`.

### 🔵 The notes recommended the term the YAML rejected

Already fixed in `d7251fef7`, pushed with this round — it was written before your review landed. That commit also adds a cross-file guard comparing every `proposed_replacement_terms` id against the prose in RESULTS.md and the notes. Writing the guard exposed a bug in the guard: it keyed MODIFY targets on the *source* term id, and two of the three MODIFY rows are both `GO:0005515`, so a dict silently dropped one and it reported a false failure against correct data.

### 🔵 `GO:0001570` kept while its reference is `MISCITED`

Fair, and the reason now says what a curator should actually do rather than leaving the tension implicit: keep the term, but **re-source the row** to PMID:23197652 with ISO (or an IBA from the zebrafish orthologue), because TAS to a review discussing angiogenesis cannot support a vasculogenesis annotation on the human protein. The term survives; this `(term, evidence, reference)` triple does not. I kept `KEEP_AS_NON_CORE` rather than `MARK_AS_OVER_ANNOTATED` because the term is not an over-reach — the citation is the problem — but I have no objection if you would rather see the action changed.

### Also in this round, found before the review landed

`term_choice_checks.py` defends both MODIFY targets by GO's **usage** rather than by reading the definition:

- `GO:0019955 cytokine binding` — asking each holder's own record whether it carries `GO:0038023` (rather than guessing from the symbol, which put ACVRL1, BMPR1A and IFNAR1 in the wrong column) gives **83 receptors and 9 non-receptors**; **BGN**, a secreted proteoglycan that binds secreted cytokines, is the directly analogous precedent.
- `GO:0017151` — an ontology search for *DEAH* returns **zero** terms, and its ten human annotations name DDX5 and, via POT1, the RecQ **DNA** helicases BLM and WRN. GO already applies it well outside a strict DEAD-box reading.

`audit_claims.py` is now 16 guards, each exercised by `--self-test`. All gates re-run on the tree with `origin/main` merged.

Thank you — the fifth-phrasing catch in particular was the one my own tooling was built to prevent and did not.

cmungall:
Thanks — all three optional notes taken in `c8c915b`, because the first one is a factual error in the most-read field in the file and the other two are cheap.

**🔵 `1.4%` was an allele frequency doing a carrier frequency's work.** You are right, and the arithmetic lands exactly where you said: at AF = 0.0142 the carrier frequency is 2·AF·(1−AF) = **2.80%**, against Barker's **9/275 = 3.27%** of controls. The three numbers agree once the right one is named. The `description` now reads "its gnomAD v4 allele frequency is about 1.4% overall, i.e. roughly 2.8% of people carry a copy, and it was found in 3.3% of healthy population controls", and `audit_claims.py` rejects the conflated form.

Your related point was the sharper one: the gnomAD and ClinVar numbers were the only quantities in this review with no script and no cached quote behind them, which is not the standard the rest of it is held to. `AGGF1-notes.md` now records the provenance of each (gnomAD v4 GraphQL for `5-77035624-G-A` GRCh38, cross-checked against Ensembl REST `/variation/human/rs34203073?pops=1`; NCBI eutils `esummary` for `VCV000402345`) and states plainly that neither is reproduced by a committed script rather than leaving that implicit.

**🔵 `laborator` but not `group`.** Caught the guard failing in the direction the prose had already moved — the corrected text says "independent group", so "three independent groups" would have passed the pattern built to stop that exact claim. Widened to `(?:laborator|group|lab\b|team)`, with a self-test mutation using the wording the prose actually uses, not the wording the guard was written against.

**🔵 Stray word.** Fixed.

**On your Xu C observation** — I measured it rather than taking it on report, since "one independent replication" is load-bearing here. `lab_independence.py` now computes the overlap between the independent paper and the dominant group:

```
CAVEAT: 39905000 shares author name(s) with the dominant group:
    Xu C         also on 8 of its papers: 24277077, 27513923, 33069768,
                 34551592, 35202649, 35608889, 37081014, 40035560
```

8 of 11, which is more than I would have guessed. I have kept PMID:39905000 counted as independent — different senior author, different institution, and a shared surname-plus-initial is not proof of the same person — but the overlap is now printed by the script and stated in both the notes and RESULTS.md rather than omitted. If the independence claim is ever leaned on harder than it is here, that is the number someone will want.

All gates re-run on the tree with `origin/main` merged; `audit_claims.py` is now 18 guards, each exercised by `--self-test`.

Two rounds, and in both of them the thing you found was a claim my own tooling was built to catch and didn't — a fifth phrasing the literal strings missed, then a guard that had drifted behind its own prose. That is a useful pattern to have had pointed out twice.

cmungall:
Fixed in `363cf65`. You are right that this is the third instance of the same pattern, and right that the guard was the problem rather than my attention.

**The two summaries.** Both now say what their own `reason` says. `GO:0001938` → "extended and given a mechanism by later work from the same group"; `GO:0005576` → "confirmed by an independent group's detection of AGGF1 in human vitreous fluid". I found a third while grepping for the class: `GO:0051897`'s summary said the AKT result is the most consistently reported consequence "across laboratories" — corrected to "across cell types, species and disease models — though almost all of those reports come from one group".

**The guard failed in both the ways you identified.** `lab\b` cannot match `labs`, and requiring a count word let the uncounted "later labs" through. Rewritten to forbid a vague plural attribution *whatever the quantifier*:

```
\b(?:three|four|five|several|multiple|many|various|numerous|other|later|across)
\s+(?:\w+\s+){0,2}(?:labs?|laborator\w*|groups?|teams?)\b
```

Both wordings that reached you are now self-test mutations, so those two spellings provably fire rather than being trusted to.

**But a blanket ban on the plural noun would have been wrong**, and that is the part I would not have got to without checking. Two surviving occurrences are true and checkable, and I verified both against the cached author lists before deciding the pattern:

- `"Two methods from two laboratories"` (BLOC1S6) — PMID:16189514 senior author **Vidal M**, PMID:33961781 senior author **Gygi SP**. Genuinely two labs.
- `"Two laboratories report opposite effects"` (TWEAK–Fn14) — PMID:36696895 **Wang QK** versus PMID:39905000 **Chen L**. Genuinely two labs.

So the rule the pattern encodes is: a **counted, checkable** claim that names its papers is fine; a vague one is not. Both of those are deliberately not matched.

One false positive worth recording, since it is the same shape as the `lab\b` bug: unanchored `other\s+group` matched "an**other group**" inside a faithful paraphrase of PMID:33069768's own sentence. Anchored with `\b`.

**🔵 The stray word.** Fixed properly this time. Last round I changed `gives`→`gave` and left the actual stray `read` — which is, pointedly, the same miss-the-real-target failure as the finding above.

`audit_claims.py` is 20 guards now, each exercised. All gates re-run.

Three rounds, and each time what you found was the guard being narrower than the claim it was written to stop — literal strings missing a fifth phrasing, then `laborator` missing `group`, then `lab\b` missing `labs`. The pattern in my errors is enumerating spellings instead of describing the class, which the current version finally does.

cmungall:
Your non-blocking finding turned out to be a correctness problem, so it is fixed rather than deferred — `41d977e`. This is my last push unless something blocking appears.

**PMID:29885663 was not just uncited, it broke the measurement.** You are right that it is cached with full text, is AGGF1 primary research, and has a senior author (Zhang JH) on none of the dominant group's papers. It was also missing from `lab_independence.py`'s hand-written paper list — so the script was reporting "exactly one independent group" about *its own list* rather than about the literature. That is the defect, not the missing citation.

The script now **derives** its paper set (GOA references + affinage citations + the review's references, minus the high-throughput screens and corrections, filtered to what is cached), and the lineage test is computed instead of judged: a senior author who also appears in the author list of a dominant-group paper is a trainee or collaborator. The old version checked only the discovery paper's first author, which is why it caught Tian XL and missed four others. Over the resulting **27 cached AGGF1 primary studies**:

| | count | |
|---|---|---|
| dominant group | **18** | Wang Q / Wang QK |
| lineage | **5** | Tian XL ×2, Xu Y, Zhou B, Lu Q — each an author on a Wang-group paper |
| senior author on no Wang-group paper | **4** | Qi J (17884784), **Zhang JH (29885663)**, Liao S (33168501), **Chen L (39905000)** |

PMID:29885663 is now cited on the `GO:0051897` row, which is exactly where you said it belongs — it makes "almost all of those reports come from one group" checkable from the review's own citations. It shows recombinant AGGF1 raising PI3K and phospho-Akt with the effect abolished by LY294002, in **Sprague-Dawley rats**, so it corroborates the pathway without bearing on the human IDA code, and the row says so explicitly.

**Two checks run at the coordinator's suggestion; both paid for themselves.**

*Projection.* No projection pattern behind any `NEW` BP row — PMID:27513923 and PMID:34551592 produced zero GO annotations anywhere. But it sharpened the coverage finding and caught an imprecision of mine: PMID:40035560's ten annotations are **all `GO:0048514` blood vessel morphogenesis in zebrafish** (4 × `aggf1` IMP, 6 × `srsf6a`/`srsf6b` IGI). So a paper whose entire finding is that AGGF1 is a general splicing factor regulating 436 genes produced ten GO annotations, every one about vessel morphogenesis in fish and none about splicing in any species. My `reference_review` said "none of them on AGGF1" — four are, on the zebrafish orthologue. Corrected.

*Enumerate, do not search.* Before the `TNFSF12 binding` proposal, `proposed_term_check.py` lists **all 90 children** of `GO:0019955`, `GO:0043120` and `GO:0005102` rather than searching, since a query for "TNFSF12 binding" cannot match a term spelling the ligand out in full. None names or defines TNFSF12/TWEAK, and the script exits non-zero if one ever does.

**One more thing, in my own tooling.** My terms.csv gate used the two-dot `git diff origin/main`, and it reported two deletions I had not made — `GO:0030492` and `GO:0031721`, which a sibling branch added after my branch point. Three-dot confirms zero deletions. Corrected, with the reason recorded in the script.

`audit_claims.py` is 21 guards, each exercised, now including the three superseded phrasings of the one-independent-group claim. All gates re-run on the tree with `origin/main` merged; `just validate` and `just validate-changed` both clean.

cmungall:
Taken — `7ec2e26`. Your note is the mirror image of the previous fix and it is right: I derived the count end to end and then hand-applied a significance filter on top of it, which put the judgement straight back in, in a place that looked like it had already been measured. Same failure as the hand-written paper list two rounds ago, one level up.

**PMID:33168501 is now a `supported_by` on the `NEW GO:0005634` row.** Verified before adding: human HCT116, immunofluorescence, `"Immunofluorescence assay showed the co-localization of AGGF1 and γH2AX."` Since γH2AX marks nuclear chromatin at double-strand breaks, that is independent human-cell corroboration that AGGF1 is nuclear — and the row, which until this commit rested entirely on Wang-group and lineage papers, now has one line of support that does not. Cited for **localisation only**; the review makes no DNA-repair claim and both the row and the `reference_review` say so.

So three of the four independent papers are used — PMID:39905000 (TNFSF12, extracellular pool), PMID:29885663 (PI3K/AKT, rats), PMID:33168501 (nuclear localisation, human cells) — and only PMID:17884784 is not. The "two of the four are substantive" sentence is gone from both the notes and RESULTS.md, with the reason it was wrong recorded in each.

**On PMID:33069768's "confirmed by another group [28]"** — agreed, and I am not resolving it either. The cached text omits the bibliography, so [28] cannot be dereferenced. The review quotes that clause as reported speech and attributes nothing to it; it is not load-bearing anywhere, and I would rather leave it unresolved than guess that it is PMID:33168501 when the citation genuinely could be something else.

One thing the gate caught on the way out, worth recording: a `--self-test` mutation still targeted the superseded wording and so reported "mutation target absent" rather than passing. A stale mutation is a guard that silently stops testing anything, which is why `--self-test` treats an absent target as a failure rather than a skip.

All gates re-run on the tree with `origin/main` merged. That is everything from this round; I have nothing further queued.

cmungall:
Fixed in `086accd`. Conceded in full — and your framing of it as round 1's finding inverted is exactly right: then the `reason` was corrected and the `summary` above it missed, here the notes and RESULTS were corrected and the YAML missed. Same shape, opposite direction, and I still did not check all three files.

**The YAML sentence is gone**, replaced by the three papers the review actually uses. The secondary contradiction went with it: the notes opened "All four are used" and closed seven lines later saying PMID:17884784 is not. Now "Three of the four" in both files.

**Your diagnosis of why the guard could not cover it was the useful part.** Adding the literal was blocked because the notes and RESULTS *quoted* the phrase while retracting it — the quote-vs-report problem I removed two rounds ago by paraphrasing, reintroduced in a new place. Both self-reports now paraphrase, so one literal covers the class, and three phrasings are guarded (`two of the four are substantive`, `two are substantive`, `All four are used`) with a self-test mutation that re-inserts the sentence into the YAML specifically.

**And the guard's own rationale string was asserting the retracted claim** — "two of them substantive (Zhang JH, Chen L)". That one is worth naming: a guard that explains itself in the words of the thing it forbids is one careless edit away from being cited as authority for it. Corrected.

**🔵 Both optional notes taken.** The `reference_review` for PMID:33168501 now records the Chinese-language/English-abstract status and the affinage Low rating — two further reasons the MEDIUM grade is right rather than leaving it a bare judgement.

**🔵 The title.** Agreed it is a fetcher issue, and agreed the YAML must keep matching the cache. `[ole of AGGF1` has lost the R of `[Role` to what looks like a bracket-stripping step, which would hit every translated PubMed title of that form — so it is recorded in `AGGF1-notes.md` as a report against `src/ai_gene_review/etl/publication.py` rather than patched locally, where it would just be masked.

One more caught by the gate on the way out: a `--self-test` mutation targeted a span the YAML wraps at width=100, so it no-opped and reported as an absent target rather than as a pass. Shortened to one physical line. That is the third time wrapping has broken a check in this PR, which is itself worth knowing.

No term, action, evidence code, count or quote changed; 33 entries and 10/10/3/1/9 unchanged. All gates re-run on the tree with `origin/main` merged.

cmungall:
Fixed in `9d16373`, and your diagnosis *is* the fix — I stopped adding literals.

**Both instances corrected.** The PMID:29885663 `reference_review` no longer ranks the independent set; it says `lab_independence.py` counts the paper among that set and that the review cites it. RESULTS.md no longer asserts the superseded count inside the sentence explaining why the superseded count was wrong.

**The structural change.** `lab_independence.py` now exposes `measure()`, and `audit_claims.py` imports it and checks the prose against the computation rather than against a blacklist:

- any counted `N independent group/lab/laboratory/team` claim must match the measurement — 4 independent senior authors, 3 of them cited — or it is a problem;
- any significance-ranking applied to the independent set (`second`, `only`, `single`, `main`, `most`, `substantive independent`) is a problem outright.

**Switching it on immediately found two more that every literal had missed**, both in files I had already "fixed" this round:

1. `AGGF1-notes.md` quoted the retracted claim again while retracting it — the quote-vs-report trap, third occurrence.
2. `PMID:39905000`'s `reference_review` still called it *the only independent replication*. That was true when written and was superseded the moment PMID:29885663 and PMID:33168501 joined the set. No literal I would have written covered it, because I did not know it was there.

That second one is the strongest argument for your position: the wrong claim was not a spelling variant of anything I had retracted, it was a **different sentence that became false when the underlying number changed**. Only a derived check can catch that.

One more thing worth reporting: writing the RESULTS.md paragraph that *explains* all this tripped the new guard, because I spelled the retracted wordings out as examples. Paraphrased, per the rule from two rounds ago. A guard that fires on its own documentation is mildly annoying and exactly right.

`audit_claims.py` is 24 guards, including two `--self-test` mutations that invent counts and rankings no literal in the file contains — which is the point of deriving.

Nothing else queued. No term, action, evidence code, count or quote changed; 33 entries and 10/10/3/1/9 unchanged.

cmungall:
Both actionable notes taken in `bc7e79a`. The third needs nothing and you were right to say so.

**The noun-list mismatch was a real gap with a live instance behind it.** `COUNT_RE` omitted `replicat` while `RANK_RE` included it, so "only independent replication" was caught and "one independent replication" was not — and one of the latter was still in RESULTS.md, superseded since PMID:29885663 and PMID:33168501 joined the set. Removed, and the two patterns now share a single `NOUN` list (`group|laborator|lab|team|replicat|contribution`).

Worth naming the shape: two checks that disagree about what their own subject is each look correct in isolation, and the gap between them is exactly where the claim survives. That is the detector/mutator scope divergence again, one level down.

**`measure()` duplicating `main()` was the more dangerous one**, and you are right that the drift would have been invisible — a human reads the printed table, the guard checks the other copy. `main()` now calls `measure()` and **asserts** agreement on `total`, `dominant`, `lineage` and the independent set, so a divergence is a hard failure rather than two plausible-looking outputs.

New `--self-test` mutation on exactly the noun the two patterns used to disagree about. 25 guards.

**On your third note** — the guard keying on the literal word "independent", which the corrected PMID:29885663 wording no longer uses — agreed, nothing to fix, and it is the right thing to have flagged rather than pre-emptively patched. The honest statement is that the derived check covers claims phrased *around the word*, and a future sentence making the same claim without it would still get through. I have not tried to guess that phrasing, because the last four rounds are a demonstration of what guessing phrasings achieves.

Nothing further queued from me.

cmungall:
All three taken in `1065d88`.

**`lab\b` → `labs?\b`.** Worth acting on for exactly the reason you give: merging the two lists removed the *divergence* but kept the *broken member*, and because `COUNT_RE` and `retracted_patterns` are complementary by design — the latter omits one/two/six precisely because those are `COUNT_RE`'s job — both were blind on the same noun. Latent, not live, but this is the second time this one element has been the hole, so it is closed with a `--self-test` mutation on the plural specifically.

**`m` rebinding.** Renamed to `meas`. It was correct only because the dict happened not to be read after the rebind, which is the kind of correctness that stops being true the moment someone adds a line between them. The patch asserts no read survived on the old name.

**`assert` under `-O`.** Converted to explicit raises naming both values. A verification that can be compiled away is not a verification, and these are the checks that keep the printed table and the guard's inputs from diverging — the one place where the failure would be silent.

27 guards, all exercised. Nothing further from me.

Thank you for the seven rounds. The shape they traced is worth stating once: every finding was a claim my own tooling was built to catch and did not, and the progression was literal strings → wider literals → regex over spellings → a derived check against the measurement. Only the last one caught a sentence that had become false by drift rather than by being written wrong, which is the failure mode I had no way to see.

cmungall:
Both taken in `f0bbbcb`. The first was worth doing properly.

**Per-mutation markers.** Each mutation now declares a substring the intended guard's message must contain, and the newly-appearing problems (`set(check(mutated)) - baseline`) must include it. My first attempt used a file-name predicate instead; it needed so many escape hatches for guards that do not name a file that it was close to no check at all, so I threw it away and did what you actually suggested.

**Installing the markers immediately exposed three mutations whose labels described the wrong guard.** Two of them fire the vague-plural pattern rather than the derived count check — because `three` is a *legal* value in the ok-set, equalling `independent_cited_n`. So your second note turned up as a concrete case while fixing the first, which is a fair demonstration that it is not hypothetical.

**`break_selftest.py` (new) tests the tester**, on a copy, restoring the real file byte-for-byte:

| breakage | required | result |
|---|---|---|
| marker pointing at a string no guard emits | exactly that mutation fails | 1 failure, the right one |
| one guard disabled outright | exactly its own mutation fails | 1 failure, the right one |
| control: unmodified | clean | exit 0 |

**On the ok-set**, I have documented the limit rather than closed it, and I want to be explicit about why. The check constrains a count to be one of two measured values and cannot tell which sense a sentence means; `four independent groups are cited` and `three senior authors are independent` would both pass. Inferring the sense from surrounding words is a heuristic over prose, which is precisely what failed for four rounds in this PR — and a wrong heuristic is worse than a stated limit, because it looks like coverage. The check still rejects every count that is neither value, which is what caught the live instances.

That is everything. 28 guards, each with a marker, and a script that proves the harness can fail for the right reason.

PR #3013, opened 2026-09-12
Fix deep research recipes on systems with older default Python

### LINKED ISSUE #2977: deep-research-* recipes fail on systems whose default Python is <3.12 (uvx resolves wrong interpreter)
## Summary

`just deep-research-falcon`, and presumably the other `deep-research-*` recipes that shell out to `deep-research-client`, fail outright on an environment whose default/first-resolved Python is older than 3.12, even though a suitable interpreter (3.12+) is installed and available.

## Reproduction

On a fresh container where `python3 --version` → `3.11.15` (with `/usr/bin/python3.12` also present but not first on PATH):

```
$ just deep-research-falcon human TAS2R38 --fallback perplexity-lite
...
Running: uvx --from deep-research-client[cyberian]==0.2.7rc1 deep-research-client research ...
  × No solution found when resolving tool dependencies:
  ╰─▶ Because the current Python version (3.11.15) does
      not satisfy Python>=3.12,<4.0 and you require
      deep-research-client[cyberian]==0.2.7rc1, we can conclude that your
      requirements are unsatisfiable.
Provider falcon exited with code 1
...
error: recipe `deep-research-falcon` failed on line 325 with exit code 1
```

## Root cause

`deep-research-client[cyberian]==0.2.7rc1` requires `Python>=3.12,<4.0`. The recipe invokes it via bare `uvx --from ...`, which lets `uv` pick whatever Python it resolves as the ambient default — on this environment that's 3.11.15, so resolution fails before any research even starts. A 3.12 interpreter was available the whole time; `uv` just wasn't told to use it.

## Workaround found

Setting `UV_PYTHON=3.12` in the environment (or installing one via `uv python install 3.12` first, if not already present) fixes it immediately:

```
UV_PYTHON=3.12 just deep-research-falcon human TAS2R38 --fallback perplexity-lite
```

## Suggested fix

Pin the interpreter explicitly at the call site rather than relying on ambient `PATH`/`python3` resolution, e.g. either:
- add `--python 3.12` to the `uvx --from deep-research-client[cyberian]==... ` invocations in `scripts/deep_research_wrapper.py` (or wherever the recipe builds the `uvx` command), or
- set `UV_PYTHON=3.12` (or whatever floor `deep-research-client` currently requires) in the `justfile`/`ai.just` recipe environment for the `deep-research-*` family.

This would make the recipes robust to whatever Python happens to be first on `PATH` in a given dev/CI/session environment, rather than depending on it happening to already be 3.12+.

## Secondary observation (same session)

While retrying with a fallback provider, `perplexity-lite` also failed with `ERROR - Provider 'perplexity' not available. Available: falcon, asta, openscientist` in this environment — likely just a missing API key/credential locally rather than a bug, but flagging in case it's useful context; the primary issue here is the Python-version resolution failure above.

cmungall:
## Summary

Request Python `>=3.12,<4.0` explicitly in the gene and module wrappers' `uvx` invocations so deep-research-client starts when the ambient Python is older than its requirement. The range also allows newer compatible interpreters when Python downloads are disabled. Fallback providers and the pathway wrapper inherit the fix; package and custom-command overrides remain supported and are documented.

Closes #2977

## Validation

- After rebasing onto current main: 77 tests passed across client-command, provider-selection, module-wrapper, timeout-propagation, module-logic, and cache-order suites.
- Ruff checks on all changed Python files and `git diff --check` passed.
- Real client startup previously passed with `--python '>=3.12,<4.0'`, `UV_PYTHON=3.11`, Python 3.13 on PATH, `UV_PYTHON_PREFERENCE=only-system`, and `UV_PYTHON_DOWNLOADS=never`.

## Test plan

- [x] Check explicit interpreter requests, package overrides, and custom-command overrides across all three wrappers; ensure the interpreter flag precedes the client command.
- [x] Verify compatibility with main's explicit-gene precedence logic and cache fixes.


cmungall:
Addressed the latest blocking review by rebasing onto current main and dropping the superseded CI-repair commit entirely. The PR no longer changes the methionine module, its gap-candidate expectations, the GO cache, or the associated unmerged history record. This preserves main's explicit-gene precedence fix and resolves the history rationale/actor concerns without retaining unrelated curation changes.

Also documented the interpreter request and both existing overrides in the wrapper docstrings. The optional helper consolidation and additional interpreter environment variable are left out to keep this fix scoped; the existing custom-command override still controls interpreter selection.

Validation after rebase: 77 tests passed (interpreter/provider/module-wrapper/timeout/module-logic/cache suites), Ruff passed on all changed Python files, and git diff --check passed. The PR description now reflects the final scope and current validation. Ready for re-review.

PR #3017, opened 2026-09-12
Review human AP1S1: sigma1A supplies half the dileucine cargo-signal site; remove receptor-mediated endocytosis

cmungall:
## Headline

**sigma1A is not a structural filler — it supplies half of the binding site for dileucine cargo sorting signals — and the one GOA row that is actually wrong is `receptor-mediated endocytosis`.**

AP1S1 (P61966, 158 aa, PANTHER PTHR11753) had 53 GOA rows, one IBA, and **zero experimental annotations on the human gene**. The molecular-function story is entirely absent from GOA and was missed wholesale by the deep-research record. Dileucine `[DE]XXXL[LI]` signals bind *no single AP subunit* — they bind the γ1‑σ1A hemicomplex, with the hydrophobic pockets on σ1A itself. That is why `contributes_to` is the correct reading, and it is what makes the MEDNIK L90P allele comprehensible.

## Actions (scripted counts, 54 rows = 53 GOA + 1 NEW)

| Action | n |
|---|---|
| ACCEPT | 32 |
| KEEP_AS_NON_CORE | 19 |
| REMOVE | 1 |
| MARK_AS_OVER_ANNOTATED | 1 |
| NEW | 1 |

**REMOVE — GO:0006898 receptor-mediated endocytosis (TAS, PMID:9733768).** Decided from the definition, which is explicit that a *cell-surface* receptor–ligand complex undergoes endocytosis from the plasma membrane. AP-1 does not act there: recruitment is Arf1-GTP/PI4P-dependent at the TGN [PMID:15377783, PMID:23415225], and its remit is TGN↔endosome plus basolateral/somatodendritic sorting [PMID:39269494]. The endocytic adaptor is AP-2 (paralog AP2S1). The cited paper has no endocytosis experiment — it is a γ2-adaptin identification study whose σ1A content is a Y2H interaction and whose localisation result is paranuclear. TAS is a traceable author statement, not an experimental annotation, so no curator judgment is being overruled; the argument is positive and biological. The *same paper's* AP-1-complex row is ACCEPTed.

**The mirror case, deliberately not removed — GO:0005905 clathrin-coated pit.** Looks like the same plasma-membrane error. It is not: the GO definition extends coated pits to "the trans-Golgi network, and on some endosomes". ACCEPT.

**MARK_AS_OVER_ANNOTATED — GO:0009615 response to virus (IEP).** One microarray, one cell line, 191 transcripts moved >2-fold, AP1S1 not named in the abstract, reference projects to 20 gene products. (AP-1 *does* have real virus biology — Nef and Vpu hijack the γ1‑σ1A dileucine site — but that is a different claim from different papers.)

## IBA propagation verdict

One IBA: **GO:0016192 vesicle-mediated transport → `NO_FAILURE_CORE`.**

Node **PANTHER:PTN000204281** (PTHR11753). I fetched the PAINT slice and resolved **all 11 gene-level donors individually** — they are *not* a set of AP-1 orthologs:

- **AP-1 sigmas:** mouse Ap1s1 (P61967, the 1:1 ortholog), mouse Ap1s2, fly AP-1sigma, yeast Aps1, pombe vas2
- **AP-2 sigmas:** human AP2S1, rat Ap2s1, fly AP-2sigma, worm **aps-2** (WBGene00000157 → F02E8.3/Q19123, resolved via the GO API after UniProt xref search returned nothing)
- **AP-3 sigmas:** yeast Aps3, *C. albicans* Aps3

The tempting review — "vesicle-mediated transport is vague, MODIFY to a TGN-sorting term" — would be **wrong**. `GRANULARITY_MISMATCH` requires donors that *agree* plus an available specific term; here the donors work in three different compartments (TGN/endosome, plasma-membrane endocytosis, lysosome/LRO), so the general term is precisely their LCA and a specific one would over-propagate AP-1 biology onto every AP-2 and AP-3 sigma. The target is inside the inheriting clade (its own mouse ortholog is a donor) with no evidence of loss.

Also resolved: the node's *other* IBD (GO:0043231) is seeded by P61966 itself, but no GO:0043231 IBA row exists in this gene's GOA, so there is nothing to review for it.

**13 `propagation_review` blocks** (the IBA + every IEA row with `supporting_entities`), **27 `source_entities`**, all *generated* from each row's `supporting_entities` by script — a build-time assertion fails if the two sets differ in either direction, so they cannot drift or gain an invented entry.

Notable non-IBA traces: `UniProtKB:P61967` is **mouse Ap1s1, not AP1S2** — checked rather than assumed, since the donor's evidence (PMID:20203623) is titled for σ1B. Abstract-only cache + MGI curator read the full text ⇒ `KEEP_AS_NON_CORE`, not REMOVE. ARBA00033972 was fetched and read: 10 FunFam+taxon condition sets, and AP1S1 genuinely satisfies set 8 (FunFam `3.30.450.60:FF:000005` AND Catarrhini).

## NEW term

**GO:0005198 structural molecule activity (IMP, PMID:39269494)**, both perturbation directions in human cells: WT σ1A co-IPs with endogenous γ1, restores γ1 levels and γ1 TGN/endosomal staining in triple-σ1-KO HAP1 cells; L90P drops γ1 co-IP to 0–13% and rescues nothing. IMP not IDA (transfected constructs in a knockout background).

## Proposed new GO term

**`dileucine sorting signal binding`.** GO has GO:0089710 `endocytic targeting sequence binding` for the *tyrosine* motif but **no counterpart for the dileucine motif**. Verified by searching QuickGO for "dileucine" (0 hits), "sorting signal binding", "targeting sequence binding", "leucine motif binding", and enumerating the MF children of GO:0140312 (only GO:0035615). Filed with an ONTOLOGY knowledge gap attached to the affected core function.

Also raised as a suggested question, not acted on: GO:0035615's definition says "responsible for the formation of **endocytic** vesicles", which does not describe AP-1 — yet GO_Central applies it to AP-1 sigmas (pombe vas2 carries it as IC `contributes_to` GO:0030121).

## Bioinformatics — including an honest negative

`genes/human/AP1S1/AP1S1-bioinformatics/` fetches nine AP sigma subunits **live** from UniProt, asserts every expected length, and maps the structurally defined σ2 dileucine pocket onto each.

1. All five literature-named σ1A positions verify against P61966: R15, A63, V88, I103, and L90 (the MEDNIK variant).
2. The alignment **independently reproduces every homology the literature states** (σ2 88→σ1A 88, σ2 103→σ1A 103, σ2 88→σ3A 94, σ2 103→σ3A 109) without being told them.
3. **The verified negative control defeats the conservation argument.** AP4S1 (σ4) is in the same family and the ε‑σ4 hemicomplex demonstrably does *not* bind these signals — yet it scores **4/5 identical at the pocket, the same as σ1A, σ1B and σ1C**.

So "AP1S1 retains the dileucine pocket" is true but carries no weight alone; the claim rests on direct σ1A mutagenesis and on PDB 4P6Z (which contains this protein as chain S). This is written up as a limitation, encoded as 4 `residue_claims` (RETAINED ×3, SUBSTITUTED ×1 for the conservative L103I), and turned into a `RESIDUAL_SUBGAP` knowledge gap.

## What affinage missed

Trust gate tripped (`self_evaluation_pairwise: tie`) → `LOW_QUALITY`, every claim re-verified. It found the disease and cell-biology literature well, but **missed the entire molecular-function story**: PMID:17360967, PMID:14691137, PMID:21097499 (the three dileucine papers), PMID:24843023 (PDB 4P6Z), PMID:15377783, PMID:23415225. All are titled for the complex, a partner subunit, or a virus — never for AP1S1. Found via Europe PMC on "dileucine signal AP-1 gamma sigma1 hemicomplex" and by resolving 4P6Z through the RCSB API. Two factual errors corrected: it presents PMID:2040623 as human (the paper used **rat and mouse** brain libraries), and says σ1A nucleates a Rabex-5 complex when PMID:27411398 states *"Unexpectedly, Rabex-5 binds σ1B, not σ1A"*.

## Validation evidence

```
Validating genes/human/AP1S1/AP1S1-ai-review.yaml...
✓ Valid
✓ All validations passed for human/AP1S1          # no ❌, no ⚠️

checked 175 quotes: 0 failures, 0 skipped          # checkquotes.py
residue-claim validation ...: 91 pass, 0 fail, 0 unresolved
no duplicate keys
terms.csv: 0 deletions, no duplicate ids

GOA rows: 53   seeded YAML rows: 53   NEW rows: 1
GOA rows with WITH/FROM: 13   YAML rows with supporting_entities: 13
propagation_review blocks: 13
RECONCILIATION PASS
```

53 references, **every one with a `reference_review`** — including the 8 donor-side papers cited inside `source_entities` comments, which were fetched and title-checked rather than left as bare PMIDs.


cmungall:
Thanks — all five suggestions addressed in `12a2fc3`. Four applied; **one declined with evidence**, which I want to flag explicitly rather than bury.

### 1. `proposed_parent` — declined, and here is why

The suggestion was to move `dileucine sorting signal binding` off `GO:0005515 protein binding` onto `GO:0005048 signal sequence binding`. I checked both against QuickGO before changing anything, and the suggestion does not survive the check:

- **`GO:0089710`** — the tyrosine counterpart this proposal is explicitly modelled on — **is itself `is_a` GO:0005515.** Its `is_a` ancestors are exactly `GO:0005515, GO:0005488, GO:0003674`. So the parent I chose *mirrors the sibling term*, which is the behaviour the suggestion was asking for.
- **`GO:0005048` is not named "signal sequence binding".** Its actual label is **"signal sequence receptor activity"**, its `is_a` ancestors are just `GO:0005048, GO:0003674` (a root-level MF), and its definition is about "a signal for its proper localization in the cell" — i.e. ER/organelle targeting. It is in the **receptor** branch, not the binding branch, and is not a genus for binding a cytosolic sorting motif.

You were right that `GO:0005515` is the least informative option and that it deserved scrutiny — CLAUDE.md says to avoid that term. But here it is the correct genus, and the fix was to *document the check* rather than change the id. The `justification` now records both findings, plus the note that if GO ever introduces a sorting-signal-binding grouping term, GO:0089710 and this term should move under it together.

### 2–5. Applied

- **`RESULTS.md` now carries the R15 caveat.** Good catch, and it cut deeper than flagged: σ1A R15E is among the substitutions that *did not* abolish binding, and γ1 R15E "caused a much greater reduction … than mutation of σ1A Arg 15". Position 15 is now marked as the weakest of the five scored anchor sites. I kept it in the scoring rather than dropping it — removing an inconvenient site from a conservation score is exactly how such a score gets quietly biased.
- **GO:0035615 qualifier.** The `enables` → `contributes_to` recommendation is now stated explicitly and actionably in `review.reason` instead of being implicit in `core_functions`, with the precedents named: pombe vas2 carries GO:0035615 as IC `contributes_to` GO:0030121, and human AP2S1 carries it as TAS `contributes_to`. So `enables` on AP1S1 is the outlier among sigma subunits, not a considered exception.
- **GO:0005198 framing.** The NEW row now pre-empts the tension with the "not a structural filler" headline: both hold at once — the subunit does hold the complex together (that is the part GO can express), and it also supplies the cargo pockets (the part with no term).
- **affinage in `core_functions[1]`.** Kept, because removing it re-triggers the validator's "no annotation references the deep research file" warning, but the coherence objection is now answered in the reference's own `review_notes`: it is cited exactly once, last of six supports, for the uncontested identity statement that AP1S1 is the AP-1 small subunit — which the five primary supports beside it already establish — and it carries no mechanistic weight anywhere. That is the condition under which a `LOW_QUALITY` source can legitimately be cited at all.

### Revalidated after the changes

```
just validate human AP1S1        →  ✓ Valid  /  ✓ All validations passed   (no ❌, no ⚠️)
checkquotes.py                   →  175 quotes, 0 failures
gene_residue_claims              →  91 pass, 0 fail, 0 unresolved
GOA↔YAML reconciliation          →  PASS (53↔53, 13 WITH/FROM ↔ 13 supporting_entities ↔ 13 propagation_review)
duplicate-key check              →  none
cache/go/terms.csv               →  0 deletions, no duplicate ids
```

No curation decision changed in this round — counts are still ACCEPT 32 / KEEP_AS_NON_CORE 19 / REMOVE 1 / MARK_AS_OVER_ANNOTATED 1 / NEW 1.

On your validation caveat: `uv`/`just` being unavailable in the runner is a real gap in what the bot can confirm, and worth noting independently of this PR — the structural checks you did by hand (row counts, `grep`-based quote verification, cached-enum GO id resolution, `results.json` vs `RESULTS.md`) covered the things most likely to be wrong, and they agree with the local output above.

cmungall:
Good catch on the method, and thank you for stating the falsifier explicitly ("if a live QuickGO pull returns the receptor name today, the caches are stale and the finding dissolves") — that made this resolvable in one step instead of turning into a disagreement.

### The label check, run live against two independent sources

| source | GO:0005048 label |
|---|---|
| **live QuickGO** (`/ontology/go/terms/GO:0005048/complete`, today) | `signal sequence receptor activity` |
| **live OLS4** (`/api/ontologies/go/terms?iri=…GO_0005048`, today) | `signal sequence receptor activity` |
| `cache/ontologies/go.tsv:747` (2026-03-21) | `signal sequence binding` |
| `cache/go/terms.csv:1063` (2026-05-08) | `signal sequence binding` |

Both live sources also list **`signal sequence binding` as a `narrow` synonym** (alongside `leader sequence binding`, `protein signal sequence binding`, `signal sequence receptor`). So the label claim was correct and **the two in-repo snapshots are stale** — and they agree with each other only because they are two readings of the same stale state, which is the failure mode worth naming: two caches agreeing looks like corroboration and isn't.

### But your underlying point was right, and that's what I changed

The sentence should never have leaned on a label at all. I've rewritten it to rest on the **`is_a` closure**, which is immune to relabelling:

```
GO:0005048  is_a ancestors: ['GO:0005048', 'GO:0003674']      ← no GO:0005488, no GO:0005515
GO:0089710  is_a ancestors: ['GO:0005515', 'GO:0089710', 'GO:0005488', 'GO:0003674']
```

GO:0005048 hangs directly off the `molecular_function` root with **no binding ancestor whatsoever**, so it cannot be the genus of a binding activity regardless of what it is named. That is the real disqualification; "receptor branch, not binding branch" was a label-shaped restatement of it and deserved the challenge.

The term is genuinely a trap and the justification now says so: its **name** says receptor activity, its **definition** opens *"Binding to a signal sequence…"*, and its **narrow synonym** says binding. Three artefacts, three different stories — which is exactly why the decision rests on the closure and the definition rather than any of them. The mirror half you couldn't verify in-runner is re-confirmed live: GO:0089710 really is `is_a` GO:0005515, so the proposal does sit in the same genus as its sibling. Parent unchanged.

The full table, including the stale-cache row, is recorded in `AP1S1-notes.md` so the next person doesn't re-litigate it from the caches.

### Your other two observations

- **The vas2 detail is better than what I had.** You found that `genes/SCHPO/vas2/vas2-goa.tsv` carries *both* the IC `contributes_to` row (line 19, with GO:0030121 as the with-field) *and* an InterPro IEA row with `enables` (line 12) for the same term. That is a stronger form of the argument than I made: the qualifier tracks the **evidence pipeline**, not the biology, and `enables` on AP1S1 is just what the InterPro2GO route emits by default. Folded into the GO:0035615 reason.
- **AP2S1 precedent provenance.** Correct that there's no `genes/human/AP2S1/` to check against — a sibling agent is reviewing it concurrently, so the directory doesn't exist yet on main. The claim came from a live QuickGO annotation query on `UniProtKB:P53680` (GO:0035615, TAS, `contributes_to`, PMID:12121421), and the reason now states that provenance inline so it's checkable without the local directory.

### Possible repo-level follow-up (not for this PR)

`cache/ontologies/go.tsv` and `cache/go/terms.csv` both carry a label for GO:0005048 that GO has since demoted to a narrow synonym. Since `just validate` checks author-supplied labels against these caches, any review that legitimately uses the current name would be flagged — and, as here, a reviewer reasoning from them can reach a confidently wrong conclusion. Worth a cache refresh independent of this gene.

### Revalidated on `db40b74`

```
just validate human AP1S1   →  ✓ Valid  /  ✓ All validations passed   (no ❌, no ⚠️)
checkquotes.py              →  175 quotes, 0 failures
gene_residue_claims         →  91 pass, 0 fail, 0 unresolved
GOA↔YAML reconciliation     →  PASS (53↔53, 13↔13↔13)
```

No curation decision changed in this round either — still ACCEPT 32 / KEEP_AS_NON_CORE 19 / REMOVE 1 / MARK_AS_OVER_ANNOTATED 1 / NEW 1, 13 `propagation_review`, 53 `reference_review`, and no `supporting_text` added or altered.

cmungall:
Fixed in `cce7302` — notes-only, no YAML touched.

You were right, and the location is exactly as you said: `grep -c GO:0140309 projects/IBA_REVIEW.md` → **0**, while `projects/UNFOLDED_PROTEIN_BINDING.md` carries it at the lines you cite, including the quotable form *"A relabel is not a redefinition"*. The link now points there, the relative path is verified to resolve from `genes/human/AP1S1/`, and the quoted line is confirmed present in the target rather than assumed.

The irony is noted and deserved: the broken pointer was sitting inside the paragraph arguing against trusting unchecked pointers. The analogy is also better than the one I had — UNFOLDED_PROTEIN_BINDING hit the same failure from the **opposite** direction (a synonym *promoted* to primary label, rather than a primary label demoted to a synonym as with GO:0005048) and resolved it the same way, on the definition and the hierarchy rather than the name.

On your provenance note about PMID:12121421: agreed on all counts, and no change made. It is prose provenance for *another gene's* annotation, so it correctly stays out of `references:` — adding it would assert a relevance to AP1S1 that it does not have. Your in-repo enumeration of every GO:0035615 row (P61966 `enables`, Q9P7N2 `enables` via GO_REF:0000002, Q9P7N2 `contributes_to` via GO_REF:0000111) is a cleaner demonstration than my external one anyway — same protein, two rows, split by pipeline rather than by biology — and that is what the reason now leads with.

```
just validate human AP1S1   →  ✓ All validations passed   (no ❌, no ⚠️)
checkquotes.py              →  175 quotes, 0 failures
```

Counts unchanged for the fourth round running: 54 rows, ACCEPT 32 / KEEP_AS_NON_CORE 19 / REMOVE 1 / MARK_AS_OVER_ANNOTATED 1 / NEW 1; 13 `propagation_review`; 53 `reference_review`. No `supporting_text` added or altered since round 1.

PR #3026, opened 2026-09-12
Review human AP4S1: sigma4's molecular function is holding the AP-4 core together, not reading cargo signals

cmungall:
## Headline

**Sigma4 has a molecular function, and it is not the one its paralogues have.** GOA gives AP4S1 no molecular function at all. The literature gives it one that is specific to this subunit and reproducible from three directions: epsilon-adaptin and beta4-adaptin associate with each other **only in the presence of sigma4** ([PMID:11409905](https://pubmed.ncbi.nlm.nih.gov/11409905/)), loss of sigma4 in patient fibroblasts destabilises all four subunits and abolishes complex assembly and tepsin recruitment ([PMID:25552650](https://pubmed.ncbi.nlm.nih.gov/25552650/)), and the 2026 cryo-EM structure shows the epsilon trunk wrapping around sigma4 to build the central core ([PMID:41565640](https://pubmed.ncbi.nlm.nih.gov/41565640/)). That is `GO:0005198 structural molecule activity`, proposed as a NEW row.

Conversely, and against the pattern set by AP1S1/AP1S3/AP2S1, **sigma4 contributes nothing to cargo-signal recognition** — in AP-4 that job is mu4's, whether the canonical YXXØ signal, APP's non-canonical YKFFE, or the ATG9A peptide. No cargo-binding MF is proposed.

## Actions — 14 GOA rows + 2 NEW

| Action | n | Rows |
|---|---|---|
| ACCEPT | 12 | GO:0005737 (IEA), GO:0005794 (IEA), GO:0005802 (NAS), GO:0006605 (IC), GO:0008104 (IC), GO:0015031 (IEA), GO:0016192 (IBA), GO:0016192 (NAS), GO:0030124 (IDA), GO:0030124 (NAS), GO:0032588 (TAS ×2) |
| MODIFY | 1 | GO:0012505 endomembrane system → **GO:0032588 trans-Golgi network membrane** |
| REMOVE | 1 | GO:0031904 endosome lumen |
| NEW | 2 | GO:0005198 structural molecule activity (IDA); GO:0006892 post-Golgi vesicle-mediated transport (IMP) |

### The REMOVE, with a positive argument

`GO:0031904 endosome lumen` (TAS, Reactome:R-HSA-5229111) is a compartment-inheritance artefact, and the Reactome ContentService shows the mechanism exactly: the reaction is a `BlackBoxEvent` whose **output entity is the complex `AP4:APP` placed in compartment "endosome lumen"**, so every member inherits it. The reference-projection test quantifies it — QuickGO by `reference=Reactome:R-HSA-5229111` returns **11 annotations over 5 entities** (the four AP-4 subunits plus APP), with GO:0031904 on five of them. One compartment assignment, five proteins.

AP-4 is a cytosolic peripheral coat and cannot be in a lumen: UniProt records "Peripheral membrane protein"; [PMID:11707398](https://pubmed.ncbi.nlm.nih.gov/11707398/) states the four-subunit complex "localizes to the cytoplasmic face of the trans-Golgi network (TGN)"; and the cryo-EM structure puts both ARF1 sites on epsilon and mu4-CTD facing the membrane. The `GO:0032588` rows from the *same* projection are correct and are accepted — which is the point: the mechanism is not evidence either way, so each term has to be judged on the biology.

*Note:* `genes/human/AP4B1` marks the identical row `KEEP_AS_NON_CORE`. I think that is too generous and say so in the notes, but I did not edit that gene — AP4B1 is not a donor for any AP4S1 row.

### The two NEW rows

- **GO:0005198 structural molecule activity, IDA, PMID:11409905.** The three-hybrid experiment is a direct assay of sigma4's *effect* (whether two other proteins can find each other), not a binary interaction test — hence IDA rather than IPI. The same paper shows the identical dependence for AP-1 (gamma+beta1 need sigma1) and COPI-F (gamma-COP+beta-COP need zeta-COP), so this is the conserved architectural role of the small subunit, not an assay artefact. Supported additionally by patient-cell loss of function with an internal AP-1 control, by the structure, and by AAGAB chaperoning epsilon and sigma4 specifically ([PMID:35976721](https://pubmed.ncbi.nlm.nih.gov/35976721/)). Bounded by [PMID:39865903](https://pubmed.ncbi.nlm.nih.gov/39865903/): 28 heterozygous carriers of a nonsense allele are asymptomatic, so the requirement is absolute but not dosage-limiting.
- **GO:0006892 post-Golgi vesicle-mediated transport, IMP, PMID:31915823.** Four AP4S1 patient fibroblast lines among fifteen; ATG9A retained in the TGN in all patient lines without exception; rescued by re-expressing an AP-4 subunit. Term choice checked both ways against QuickGO's closure: GO:0006892 **is** under GO:0016192 (so it sharpens the IBA) and **is not** under GO:0015031 (so it does not duplicate the InterPro row). `GO:0006895 Golgi to endosome transport` was deliberately rejected — its definition names early sorting endosomes and clathrin vesicles, and AP-4's route is non-clathrin.

## Propagation review — 10 blocks

Required on 7 rows (1 IBA + 2 IC + 4 IEA with `supporting_entities`); added on 3 more (the Reactome TAS rows, where the projection *is* the analytic content). `source_entities` are built programmatically from each row's seeded `supporting_entities`, and a reconciliation script asserts they match exactly.

**GO:0016192 IBA — `NO_FAILURE_CORE`.** `just fetch-panther-paint PTHR11753` returns exactly **one** annotated node, PTN000204281, with two IBD rows and **no IRD or IKR anywhere in the family**. Yes, this is the same node the merged AP1S1/AP1S3/AP2S1 reviews inherit from — it is the only one there is. All 11 gene donors resolved (WB:WBGene00000157 needed the GO API bioentity endpoint, not a UniProt xref); **10 of 11 carry a direct experimental annotation** to GO:0016192 or a term beneath it (counted by script, not asserted). The one exception, FB:FBgn0039132, is recorded as a limit of my tracing rather than a donor defect.

The temptation worth naming: this looks like a `GRANULARITY_MISMATCH` candidate and is not one. The donors *disagree* on sub-process — AP-2 sigmas → endocytosis, AP-3 sigmas → Golgi-to-vacuole, AP-1 sigmas → retrograde/endosomal — so GO:0016192 is the genuine LCA of a heterogeneous set. The AP4S1-specific refinement therefore goes in a separate NEW row rather than replacing the IBA. Q9Y587 does **not** appear in its own WITH/FROM.

| Row | root_cause | note |
|---|---|---|
| GO:0016192 IBA | `NO_FAILURE_CORE` | correct LCA; no IRD/IKR in family |
| GO:0005737 IEA (ARBA) | `NO_FAILURE_NON_CORE` | true, generic, entailed by GO:0030124 |
| GO:0012505 IEA (ARBA) | `TERM_SCOPING_PROBLEM` / `GRANULARITY_MISMATCH` | → GO:0032588 |
| GO:0005794 IEA (SubCell) | `NO_FAILURE_CORE` | faithful coarse restatement of SL-0132 |
| GO:0015031 IEA (InterPro) | `NO_FAILURE_CORE` | AP4S1 really matches IPR016635 |
| GO:0006605 IC, GO:0008104 IC | `NO_FAILURE_CORE` | source is GO:0030124 on the same protein |
| GO:0031904 TAS | `PROPAGATION_BAD` / `COMPARTMENT_OR_COMPLEX_MISMATCH` | see above |
| GO:0032588 TAS ×2 | `NO_FAILURE_CORE` | same projection, right answer |

## Bioinformatics — `AP4S1-bioinformatics/`

The campaign prompt asked me to take a position on the merged **AP1S1** bioinformatics, which uses sigma4 as a "verified non-binder" negative control that nonetheless scores 4/5 at the sigma-side dileucine pocket.

**Position: the analysis is right and its design is the right design, with one citation-weighting correction and one extension.**

*Correction.* AP1S1 cites the sigma4 negative to `[PMID:14691137, PMID:21097499]` jointly; those are not of equal weight. PMID:14691137 (2003) tested two signals and scored **both** alphaC-sigma2 and epsilon-sigma4 negative, and its GST pull-down arm used a prep containing "traces of AP-4". The alphaC-sigma2 half did not hold — by 2011 the same laboratory reported all three signals binding alphaC-sigma2. **PMID:21097499 is the load-bearing citation**: three signals, working positive controls, only epsilon-sigma4 negative. That makes sigma4 a *stronger* negative control than the AP1S1 write-up claims for it. I did not edit the AP1S1 review (not a donor for any row here); flagged for the coordinator.

*Extension.* AP1S1 left the next question open: the difference "must lie elsewhere — in the partner large subunit...". The site is a **two-part** basic patch, one Arg on the sigma and one on the large subunit (gamma1 R15, alphaC R21, delta R26), with the large-subunit one dominant. `basic_patch.py` transfers those three anchors onto epsilon, gated by a **6/6 method control** (the three anchors must map onto each other, which they do exactly):

| anchor | → epsilon | residue |
|---|---|---|
| gamma1 R15 | 44 | **T** |
| alphaC R21 | 44 | **T** |
| delta R26 | 44 | **T** |

All three converge on epsilon Thr44, and epsilon/delta share an identical `LVRGI` immediately upstream so the register is visible without trusting the aligner. Sigma4 retains the sigma-side Arg15. **The basic patch is half-complete in AP-4: present on sigma4, absent on epsilon.** Every complex that binds `(D/E)XXXL(L/I)` has both halves; the one that does not is the one missing the large-subunit half.

Stated as a hypothesis consistent with the measured negative, **not a test of it** — the test is an epsilon T44R substitution, which is in `suggested_experiments`. It supports no annotation either way. All eight sequences fetched live with lengths asserted; script exits 2 rather than answering if the control fails.

## Knowledge gaps — 5

1. **MF_DARK / NARROWING** — sigma4 has no assigned activity outside the AP-4 core, and the cargo-signal question has never been tested by mutating sigma4 itself.
2. **RESIDUAL_SUBGAP** — the causal step from ATG9A missorting to neurodegeneration is missing, and the obvious candidate is excluded where the missorting is measured: **bulk autophagic flux is intact in AP-4-deficient patient fibroblasts**. This is why no autophagy BP term is proposed for AP4S1.
3. **CURATION** — *neither ARBA row can be reproduced from its own rule's condition sets.* Fetched both rules; ARBA00026971 has 2388 condition sets, the only one naming an AP4S1 signature also requires IPR027156; ARBA00028953 has 533, its candidate also requires IPR010908 and IPR044565. AP4S1 matches exactly three InterPro entries (checked live): IPR011012, IPR016635, IPR022775. Both terms are biologically right, so this is provenance, not biology — but a reviewer cannot tell correct-by-luck from correct-by-derivation. `GO_REF:0000117` marked `LOW_QUALITY` on this specific basis.
4. **CURATION/BIOLOGY** — three incompatible isoform frames. Clinical variant nomenclature uses NM_007077 (= Q9Y587-**2**); MANE Select and UniProt canonical are NM_001128126.3 (= Q9Y587-**1**); and PMID:31660686 calls NM_001128126.3 "the canonical isoform **2**". No GOA row carries an isoform qualifier.
5. **ONTOLOGY** — `GO:0140312 cargo adaptor activity` is defined as binding coat scaffolding "such as clathrin or COPII". AP-4 uses neither and **no scaffold partner for AP-4 is known**, so the complex-level activity can only be expressed with the generic parent `GO:0060090`. ComplexPortal nonetheless annotates CPX-5151 to GO:0140312 by NAS. Raised as an observation, not a proposed term, because the right repair depends on whether AP-4 has a scaffold.

## `core_functions`

Two entries, per the complex-subunit pattern: (1) subunit-level `molecular_function: GO:0005198`; (2) complex-level `contributes_to_molecular_function: GO:0060090` — deliberately the generic parent rather than GO:0140312, for the reason in gap 5. No "protein binding" anywhere.

## What affinage missed

Trust gates clear, `self_evaluation_pairwise: win`, `faith_pct: 100`, all eight citations numeric PMIDs, right protein, and all eight findings check out. Its problem is recall, exactly as the brief predicts — **it missed every paper bearing on sigma4's own molecular function**, because each is titled for the complex, the partner or the family:

`PMID:11409905` (the sigma4-dependence result, the basis of the MF row) · `PMID:41565640` (2026 cryo-EM — found via the EMDB cross-refs in the UniProt entry) · `PMID:21097499` + `PMID:14691137` (the dileucine negative) · `PMID:35976721` (AAGAB holds epsilon and sigma4) · `PMID:11707398` (cytoplasmic face of the TGN) · `PMID:39865903` (the 2025 carrier negative) · `PMID:20230749` + `PMID:24498434` (mu4 cargo sites) · `PMID:32979048`, `PMID:41032520`, `PMID:30262884`, `PMID:11802162`, `PMID:31142229`.

Its `mechanism_profile` GO ids were **not** imported; both MF terms were re-grounded from primary literature and land in different slots than a direct import would have given.

## Tooling

- **Europe PMC was down for the entire session** — `www.ebi.ac.uk/europepmc/webservices/rest/search` returned HTTP 503 on every attempt over several hours and `www.europepmc.org` returned 530, while UniProt REST, QuickGO, InterPro, Reactome and NCBI eutils were all fine. The brief's Europe PMC searches were run against NCBI eutils instead, covering the symbol, `sigma4 adaptin`, `SPG52`, `AP-4 sigma`, `AP-4 deficiency`, each partner, the family and the paralogues.
- `interpro/panther/PTHR11753/` was already current; re-fetching the PAINT slice produced no diff, so nothing to commit there.

## Validation

```
$ just validate human AP4S1
✓ Valid (with 1 warnings)
  ⚠ [WARNING] No annotations reference available deep research files (AP4S1-deep-research-affinage.md)
✓ All validations passed for human/AP4S1

$ uv run python checkquotes.py genes/human/AP4S1/AP4S1-ai-review.yaml
checked 84 quotes: 0 failures, 2 skipped (non PMID/file refs)

$ python3 .scratch/reconcile.py
GOA rows: 14 (14 distinct keys) | seeded YAML entries: 14 | NEW entries: 2
rows requiring propagation_review (IBA/ISS/ISO/IEA/IC with supporting_entities): 7; present: 7
propagation_review blocks in total: 10
actions: {'ACCEPT': 12, 'MODIFY': 1, 'NEW': 2, 'REMOVE': 1}
references: 32; with reference_review: 32
PASS
```

The one warning is deliberate: the campaign brief forbids quoting an affinage sentence as `supporting_text`, so the affinage record is listed and adjudicated in `references[]` but never cited — the same shape as the merged `genes/human/AGT` review. No duplicate YAML keys; `cache/go/terms.csv` unchanged (no deletions, no duplicate ids); `gene_residue_claims` passes 79/79 repo-wide (this review asserts `residue_claims_not_applicable`, not residue claims).


cmungall:
## Round 1 response — head now `0a899bd`

Thanks for the review. Two of the five suggestions were acted on, one was acted on in a different way than suggested, and two are declined with reasons. The branch as it now stands:

### 1. `source_status` vs its own comment — **fixed, with a different enum than suggested**

You were right that the machine-readable field contradicted the prose, and this is exactly the drift class this campaign exists to catch. I re-checked the donor directly before changing it rather than taking either suggested value:

```
$ QuickGO geneProductId=Q9VCF4 goId=GO:0016192 goUsage=descendants
GO:0035615 IEA GO_REF:0000002  InterPro:IPR044733
GO:0016192 IEA GO_REF:0000002  InterPro:IPR000804, InterPro:IPR044733
GO:0016192 ISS GO_REF:0000024  SGD:S000004160
```

So the record is neither missing nor untraceable — which rules out `SOURCE_STALE_OR_MISSING` ("current source record no longer carries the transferred term, or tracing cannot recover it") and `UNRESOLVED` ("could not be classified confidently"). It carries the term, just never experimentally. `SOURCE_WEAK_OR_INFERRED` ("source exists but is only inferred") is the accurate value and is what the row now uses.

One thing the re-check turned up that is worth having on the record: the ISS strand traces to **`SGD:S000004160`, which is itself another donor in this same `WITH/FROM`**. So that strand adds nothing independent, and the honest count of independently grounded donors is unchanged at 10 of 11. Both facts are now in the source comment and in `AP4S1-notes.md` §6.

### 2. Coarse-parent policy — **fixed, both halves**

The discriminator is now stated in `GO:0005737`'s reason rather than left implicit: *coarse-but-independently-informative is kept; coarse-and-strictly-subsumed is sharpened.* `GO:0005737` carries content nothing else on the gene carries — it is the topological fact the `GO:0031904` removal rests on — whereas `GO:0012505` is strictly subsumed by terms already present and adds nothing.

The `MODIFY` no-op point is also now stated in `GO:0012505`'s reason: `GO:0032588` is already on the gene twice by TAS, so the practical effect is removal of a redundant parent, not creation of a third duplicate row. Good catch — a consumer reading the replacement term alone could easily have got that wrong.

### 3. `KEEP_AS_NON_CORE` on `GO:0008104` — **declined, with the reason now written into the row**

I don't think the enum supports it. `KEEP_AS_NON_CORE` is defined for annotations that are *real but peripheral* — "for pleiotropic genes, this may be the developmental processes, or other processes that are not the core function". `GO:0008104` is not a different, secondary process; it is the core function stated coarsely. Marking it non-core would misreport what kind of annotation it is, and the repo's usage agrees: `genes/human/AP4B1` reserves `KEEP_AS_NON_CORE` for `GO:0061938` (somatodendritic localization) and `GO:0031904`, while accepting `GO:0008104` and `GO:0006605`.

What the suggestion is right about is that six co-equal "core" BP rows under-express the hierarchy. Rather than misuse the enum, I've added a clause to `GO:0008104`'s reason making the nesting explicit and pointing at the IC propagation block, which already records that the three rows from `PMID:10066790` are one observation projected across the complex's subunits. If the project wants generic-parent rows distinguished systematically, that is an enum gap worth raising separately — `ActionEnum` currently has no value for "true, entailed, retained for completeness".

### 4. `IDA` on the three-hybrid row — no change

Noted as the contested point; we agree on the call, and it is now on the record in both directions.

### 5. Cross-gene follow-ups — confirmed out of scope here, and tracked

Neither `AP4B1` nor `AP1S1` is a donor for any AP4S1 row, so the campaign brief confines me to flagging them. Both are written up in `AP4S1-notes.md` (§7 for the AP4B1 `GO:0031904` row, §4 for the AP1S1 citation weighting) and both are in my return report to the coordinator, so they will not be lost with this PR.

### Re-validation on `0a899bd`

```
$ just validate human AP4S1
✓ Valid (with 1 warnings)          # the deliberate uncited-affinage warning
✓ All validations passed for human/AP4S1

$ uv run python checkquotes.py genes/human/AP4S1/AP4S1-ai-review.yaml
checked 84 quotes: 0 failures, 2 skipped (non PMID/file refs)

$ python3 .scratch/reconcile.py
GOA rows: 14 (14 distinct keys) | seeded YAML entries: 14 | NEW entries: 2
rows requiring propagation_review: 7; present: 7
propagation_review blocks in total: 10
actions: {'ACCEPT': 12, 'MODIFY': 1, 'NEW': 2, 'REMOVE': 1}
references: 32; with reference_review: 32
PASS
```

No duplicate YAML keys; `cache/go/terms.csv` unchanged (no deletions, no duplicate ids). Action counts, term choices and all 84 quotes are unchanged from the approved commit — this round touches four `review.reason` strings, one `source_status` enum, one source comment, and one paragraph of the notes.

cmungall:
## Round 2 response — head now `148707e`

You were right, and the fix is structural rather than a one-line enum change, because the root cause was in how I built the file.

### The defect

`source_status` is a judgment about a source **with respect to one propagated annotation** — the schema says so explicitly ("Mechanical status of a source entity with respect to a propagated target annotation"). My builder keyed source judgments by `source_id` alone, so `Reactome:R-HSA-5229111`, which sources two rows on this gene, was forced to carry one verdict on both. `SOURCE_BAD` is right for `GO:0031904` and wrong for `GO:0032588`, and the shared comment made the contradiction visible. Same drift class as the `FBgn0039132` fix, one field over, and a better illustration of it — there the prose and the enum disagreed by accident, here they disagreed because the data structure could not represent the distinction.

The builder now takes per-row `source_overrides`, and errors out if an override names a source the row does not actually have.

### The value, and why not `SUPPORTS_TRANSFER`

`SOURCE_WEAK_OR_INFERRED`, of the two you offered. `SUPPORTS_TRANSFER` would overclaim in the other direction: Reactome does place the AP4:APP complex at the TGN membrane and that *is* correct for AP-4, but it is correct in a way the mechanism cannot take credit for. Whole-complex compartment inheritance from a `BlackBoxEvent` assigns the same compartment to every member regardless of whether it is physically possible for them — which is exactly how the sibling row got an impossible one. A mechanism with no discriminating power is "otherwise weak", not supporting. The row's `reason` already said the projection "is not evidence either way", so `SOURCE_WEAK_OR_INFERRED` is what that sentence was asserting all along.

Final state across the three Reactome-sourced rows, which now say three different things because they mean three different things:

| row | source | `source_status` |
|---|---|---|
| `GO:0031904` | `R-HSA-5229111` | `SOURCE_BAD` |
| `GO:0032588` | `R-HSA-5229111` | `SOURCE_WEAK_OR_INFERRED` |
| `GO:0032588` | `R-HSA-5229132` | `SUPPORTS_TRANSFER` |

Each carries its own comment now rather than sharing one; the `R-HSA-5229111` default comment also states that its status is term-specific, so a reader hitting either row alone is not misled. The general point is written up in `AP4S1-notes.md` §7, including the admission that the first build could not have got this right.

### Re-validation on `148707e`

```
$ just validate human AP4S1            → ✓ Valid (1 deliberate warning); all validations passed
$ checkquotes.py …AP4S1-ai-review.yaml → checked 84 quotes: 0 failures, 2 skipped
$ .scratch/reconcile.py                → PASS
    GOA rows: 14 (14 distinct keys) | seeded YAML entries: 14 | NEW entries: 2
    rows requiring propagation_review: 7; present: 7 | blocks in total: 10
    actions: {'ACCEPT': 12, 'MODIFY': 1, 'NEW': 2, 'REMOVE': 1}
    references: 32; with reference_review: 32
```

No duplicate YAML keys; `cache/go/terms.csv` unchanged. No term id, label, action, `proposed_replacement_terms`, `supporting_text` or reference was touched this round — the diff is one `source_status`, two source comments, the builder's override mechanism, and one notes paragraph.

Thanks for both catches. They were the same bug at two depths, and the second one is the more useful finding.

PR #3029, opened 2026-09-12
Review human AP5B1: a beta adaptin that cannot recruit clathrin, working in a complex GO has no term for

cmungall:
## Headline

**AP5B1 is a beta adaptin that cannot recruit clathrin, and the complex it actually works in has no GO term.**

Two findings drive this review. First, beta-5 has the beta-adaptin fold and the beta-mu
contact, but not the clathrin apparatus — re-tested rather than repeated, with the four
classical beta adaptins as controls. Second, AP-5 acts *in vivo* as a stoichiometric hexamer
with spatacsin and spastizin; ComplexPortal curates that assembly as CPX-20045, GO does not
have a term for it, and the cost is visible in this gene's annotations as five bare
`GO:0005515 protein binding` rows that have nowhere accurate to land.

## Actions

24 seeded GOA rows, all reviewed; 1 NEW row added. Counted by script, not by eye.

| Action | n | Rows |
|---|---|---|
| MODIFY | 12 | 9 intra-complex `GO:0005515` IPI → `GO:0005198`; `GO:0030119` IDA → `GO:0044599`; `GO:0016197` IMP and `GO:0016192` NAS → `GO:0042147` |
| ACCEPT | 9 | 2 IBA, 1 IEA, 5 ComplexPortal NAS (CC/BP), 1 HDA |
| MARK_AS_OVER_ANNOTATED | 3 | CAMK2B ×2, NIF3L1 ×1 |
| NEW | 1 | `GO:0001895 retina homeostasis`, IMP, PMID:40081374 |

No `REMOVE` and no `UNDECIDED`: nothing in GOA for this gene is contradicted by the
literature, and every cited reference was readable at least in abstract.

**Why the nine IPI rows become `GO:0005198 structural molecule activity`, not "protein
binding".** The claim is not "it binds something". Mu-5 bait recovers AP5B1 residues 16-229
in a two-hybrid screen — the same N-terminal quarter that makes the beta-mu contact in the
AP-1 and AP-2 structures — and knocking AP5B1 down lowers mu-5 protein. That is a
structural-integrity contribution, which is exactly GO:0005198's definition, and it is the
most specific MF GO offers for an adaptor-complex subunit. `GO:0140312 cargo adaptor
activity` was considered and **rejected**: its definition requires bridging membrane, cargo
receptor and deformation machinery, and PMID:29381698 states in its own abstract that
nothing was known about AP-5's cargo. That is recorded as an MF knowledge gap instead of
being asserted.

**Why CAMK2B and NIF3L1 are downgraded.** Reference projection test, paginated fully and
counting entities: PMID:25416956 → 24,599 annotations / **937 entities**; PMID:33961781 →
9,514 / **2,196**. Neither partner appears anywhere in the AP-5 literature. Note the trap
that was avoided: CAMK2B is recovered by both screens and carries `NbExp=4` in IntAct, but
two high-throughput methods agreeing is still two high-throughput methods, and NbExp counts
experiments, not studies.

## IBA propagation verdicts

`just fetch-panther-paint PTHR34033` → the committed slice has **exactly two IBD rows, one
node, one seed**:

```
PTHR34033  PTN002145545  GO:0030119  C  IBD  false  UniProtKB:Q2VPB7  taxon:2759
PTHR34033  PTN002145545  GO:0016197  P  IBD  false  UniProtKB:Q2VPB7  taxon:2759
```

| Row | root_cause | Node / seed | Verdict |
|---|---|---|---|
| `GO:0016197` IBA | `NO_FAILURE_CORE` | PTN002145545 at taxon:2759, sole seed UniProtKB:Q2VPB7 | ACCEPT |
| `GO:0030119` IBA | `NO_FAILURE_CORE` | same node, same sole seed | ACCEPT |
| `GO:0016197` IEA | `NO_FAILURE_CORE` | InterPro:IPR038741 | ACCEPT |

The seed is **AP5B1 itself**, and the two node terms are precisely the two experimental
annotations AP5B1 held in 2017 (the IDA and IMP from PMID:22022230). This is the textbook
self-referential IBA: the target in its own WITH/FROM is the expected marker that the
family's experimental grounding sits on the target, not circularity, and a one-donor list is
not weak evidence. What the IBA adds over the IMP is the phylogenetic judgement that the
function is ancestral to the whole eukaryotic clade — which the same paper's comparative
genomics supports (AP-5 subunits in all five eukaryotic supergroups).

The obvious leak risk was checked and does not exist: **AP1B1, AP2B1, AP3B1 and AP4B1 are all
PTHR11134** (ADAPTOR COMPLEX SUBUNIT BETA FAMILY MEMBER, 17,626 proteins) while **AP5B1 is the
sole occupant of PTHR34033** (1,227 proteins, one subfamily, five reviewed members all being
AP5B1 orthologues at 876-883 aa). No clathrin-binding term from the classical beta adaptins
can reach this gene by descent.

The IEA was verified at source rather than assumed: the InterPro API returns IPR038741 as a
family-type entry whose only GO mapping is GO:0016197 and whose description reads "AP5B1 is
part of the AP-5 adaptor protein complex, which may be involved in endosomal transport" —
AP5B1-specific, so not a promiscuous mapping.

## NEW term

`GO:0001895 retina homeostasis`, IMP, PMID:40081374 (+ PMID:42568187, PMID:41830174).
AP5B1 became a macular dystrophy gene after the GOA snapshot and none of it is annotated:
two families with truncating alleles removing >2/3 of the protein in the 19-family 2025 AP-5
cohort, five probands in an independent 2026 replication, and 22 affected individuals from 20
families carrying the p.Leu785Pro founder allele (OR 11.9 vs gnomAD NFE). The same 2025 paper
supplies the first **subunit-specific** localisation of AP5B1 protein, by immunofluorescence
in human RPE explants and iPSC-RPE, showing puncta overlapping Rab7 and partially TGN46. The
row's own `reason` flags it as a tissue-level, non-core consequence, not a molecular function.

## Bioinformatics

`genes/human/AP5B1/AP5B1-bioinformatics/` (uv project, live fetch, asserted lengths, no
hardcoded results). Re-tests the four 2011 claims about clathrin binding rather than
repeating them:

| Claim (PMID:22022230) | Verdict |
|---|---|
| classical clathrin box absent | **confirmed in the form that matters** — the one L(Φ)x(Φ)[DE] match, `LLRLE`@841, is inside folded PF21590, whereas AP1B1 `LLNLD`@632 and AP2B1 `LLNLD`@631 sit in the hinge |
| `LLDLL` and `YQW` absent | **confirmed**, verbatim |
| `WDW` unusable because it is "in the middle of the α-helical solenoid" | **confirmed** — PDBe coverage shows residues 223-229 modelled in 8YAB chain B (modelled span 7-631), between `Helix 182-203` and the disordered `Region 234-260`. A short ordered island in an inter-helical loop of the trunk, not an exposed hinge. Independently: `PWxxW` is absent, and beta-5 is undetectable in the CCV fraction |
| no long solenoid-to-appendage linker | **confirmed quantitatively** — 62 aa in AP5B1 vs 187 (AP1B1), 178 (AP2B1), 99 (AP4B1), 90 (AP3B1) |

The document states its own caveat: PF21587-PF21590 are family-specific models while PF01602
is pan-family, so the boundary calls are not derived identically in subject and controls. The
62-vs-178/187 gap survives that, and two independent sources corroborate the AP5B1 trunk
boundary at PF21588's end (627) — the UniProt secondary-structure span (10-629) and the 8YAB
chain B modelled range (ending 631).

*Revised after review round 1: the `WDW` row originally read "qualified", on the grounds that
neither Pfam nor UniProt annotates residue 223. That was an over-reading of an absence —
neither database annotates inter-helical loops. A PDBe polymer-coverage check was added to the
script and settles it directly; the 2011 claim is confirmed. The correction is recorded in
`RESULTS.md` §3 and in the notes rather than quietly overwritten.*

## What affinage missed

Gates clear (`self_evaluation_pairwise: win`, faith 100%), right protein, no symbol
collision, all 7 citations numeric and resolving to the papers described. Recall was the
weak point — found by independent Europe PMC searching:

1. **PMID:22022230** — the paper that defines AP-5 and contains *every* AP5B1-specific
   experiment in GOA. Not cited (it cites the 2013 follow-up).
2. **PMID:20613862** — supplies 4 of the 12 IPI rows, and the study in which AP5B1 was itself
   a bait. Its title is about a DNA repair screen: the "titled for something else" failure
   mode. Not cited. It also carries a **measured negative** for this gene — AP5B1 knockdown
   had no effect in the HR assay while three complex partners did — which is why no DNA
   repair term belongs here.
3. **PMID:42568187** and **PMID:41830174** — both 2026 disease cohorts, published after the
   affinage run.
4. **PMID:17897319** — source of the HDA row.

Its `mechanism_profile` GO grounding was not imported, per the brief, and would have been
wrong: it proposes `GO:0060090 molecular adaptor activity`, which overstates what has been
shown.

## Knowledge gaps recorded

Three at gene level plus two attached to core functions: no demonstrated molecular activity
for AP-5 (MF_DARK); no GO CC term for the hexamer despite ComplexPortal CPX-20045
(ONTOLOGY/CURATION, CC_DARK, with a `proposed_new_terms` entry and a `skos:exactMatch` to
CPX-20045); no ligand known for the beta-5 appendage, which uniquely retains both subdomains
and hosts the recurrent disease allele; the unexplained retinal-vs-spastic-paraplegia split
between subunits of an obligate complex; and the unattributed curvature-sensing surface.

Deliberately **no** `contributes_to_molecular_function` on either core function. `GO:0140090
membrane curvature sensor activity` was the candidate and is rejected: the activity was
measured on the hexamer while GO's only complex term here is the tetramer, the same study
solved SPG11-SPG15 without AP-5 so AP-5 may be dispensable, and the term's definition names an
untested helix-insertion mechanism. An omitted slot says "not established"; a filled one says
something downstream tooling will believe.

## Validation

```
$ just validate human AP5B1
Validating genes/human/AP5B1/AP5B1-ai-review.yaml...
✓ Valid (with 3 warnings)
  ⚠ Inconsistent review actions for term GO:0016197: ACCEPT (IBA, IEA); MODIFY (IMP)
  ⚠ Inconsistent review actions for term GO:0030119: ACCEPT (IBA); MODIFY (IDA)
  ⚠ No annotations reference available deep research files (AP5B1-deep-research-affinage.md)
✓ All validations passed for human/AP5B1

$ uv run python checkquotes.py genes/human/AP5B1/AP5B1-ai-review.yaml
checked 95 quotes: 0 failures, 0 skipped (non PMID/file refs)

$ reconcile_AP5B1.py
GOA rows: 24   YAML existing_annotations: 25 (24 seeded + 1 NEW)
RECONCILIATION: PASS
```

Plus: no duplicate YAML keys; `cache/go/terms.csv` diff is one addition (`GO:0044599`) and
zero deletions; no duplicate ids in terms.csv.

**All three warnings are deliberate and each is justified in the artifact itself.** The two
action-consistency warnings are the point rather than an oversight: the direction of transport
and the four-subunit composition are *human* results, and pushing either down a level on a
eukaryote-root PAINT node or on an InterPro family signature would propagate a human-specific
finding to 1,227 proteins across 2,973 taxa. All four `review.reason` fields state the
divergence explicitly. The third warning stands because the campaign rule is that affinage is
a lead and not evidence, so no `supporting_text` cites it — recorded in its `reference_review`.

`source_entities` were built mechanically from each row's seeded `supporting_entities`, never
hand-typed, and every supporting quote was verified against the cache before being written.


cmungall:
## Round 1 response — both blocking items fixed, both verified first

Head is now `6c4cb3ba03c926ef2d728bd899d326f78a3365e3`. I checked each finding against the artifact before agreeing; both were correct, and the second one was correct in a way that made the analysis stronger rather than just more cautious.

### 🟡 1 — `contributes_to_molecular_function: GO:0140090` — **removed**

Agreed without reservation. The slot asserted exactly what the adjacent `knowledge_gaps` entry declares undetermined, and the `in_complex: GO:0044599` mismatch you spotted makes it worse: the curvature sensing was measured on the hexamer. Your point that PMID:40175557 also solved SPG11-SPG15 *without* AP-5 is the one I should have drawn myself — with an abstract-only cache I cannot rule out that AP-5 is dispensable for the activity, so there was no version of that assertion I could defend.

The second core function now stands on `GO:0005198` + `GO:0007040`. The description says why the slot is empty, and the knowledge gap now carries the SPG11-SPG15-without-AP-5 point explicitly.

### 🟡 2 — the `WDW` claim — **confirmed, not qualified**

You were right, and checking it properly produced a better answer than rewording would have. I added a **PDBe polymer-coverage fetch** to `clathrin_box_scan.py` so the question is settled by data rather than by reading absences:

```
WDW at 223-225, outside any Pfam domain; UniProt features here: no UniProt feature covers this residue
  nearest feature before: Helix:182-203
  nearest feature after:  Region:234-260 (Disordered)
  8YAB chain B: modelled=True segment=[223, 229] chain span=[7, 631] within span=True
  nearby unmodelled gaps: [[213, 222], [230, 260]]
```

Residues **223-229 are modelled**, inside a chain span of 7-631. The tripeptide is a short ordered island in a poorly ordered inter-helical loop of the trunk — not an exposed hinge of the AP-1/AP-2 kind, and not outside the solenoid. My error was exactly the one you named: treating "no Pfam domain and no UniProt feature covers 223" as positive evidence, when neither database annotates inter-helical loops at all. Hirst et al. 2011 is confirmed; the apparent inter-domain position is a Pfam family-model coverage gap and nothing else.

`RESULTS.md` §3 is rewritten to say so, and it records the correction rather than overwriting it quietly, since the wrong reading is the more instructive half. The notes and the PR body table are updated to match.

### 🔵 3-7 — all five taken

- **3 (reproducibility framing).** Fixed at the source rather than by disclaimer: the script now emits, for every motif hit, the nearest flanking UniProt features, the structure coverage, and the protein's secondary-structure span. Every figure in `RESULTS.md` §3 and §4 now comes from `run-output.txt`. The one paragraph that genuinely is not from the run — the PANTHER family comparison in the conclusion — is now labelled as such, with its sources named (the `DR PANTHER` lines and the committed `PTHR34033-metadata.yaml`/`-entries.csv`).
- **4 (docstring).** `AP1B1 LLDLD` → `LLNLD`. Fixed.
- **5 (HDA row).** The trailing "retained as non-core" clause is gone; the row is `ACCEPT` and now says so cleanly.
- **6 (nine rows → one term).** Added to the lead IPI row: the nine replacements should collapse into a single `GO:0005198` annotation, and that collapse implies a re-code, since what supports the term is the mapped beta-mu interface (IDA) and the loss of mu-5 on beta-5 depletion (IMP), not the IPI evidence line. Good catch — the intent would not otherwise have survived the hand-off.
- **7 (NEW row IMP basis).** Made explicit in the `reason`: human genetic association across three cohorts plus subunit-specific localisation in the affected tissue, not an engineered mutant.

Also, unprompted: the generator was emitting YAML anchors/aliases (`&id001`/`*id001`) for the shared replacement-term lists. Disabled, so the file is now plain.

### Validation on the new head

```
just validate human AP5B1      → ✓ Valid (with 3 warnings)   [same three, all deliberate]
checkquotes.py                 → checked 95 quotes: 0 failures, 0 skipped
reconcile_AP5B1.py             → 24 GOA rows / 25 YAML rows (24 seeded + 1 NEW): PASS
duplicate YAML keys            → none
cache/go/terms.csv             → 1 addition (GO:0044599), 0 deletions, no duplicate ids
```

Thank you for the runner caveat about `just`/`uv` — noted, and the results above are from the worktree.

cmungall:
## Round 2 response — blocker fixed, all four suggestions taken

Head is now `310235b5ad21528f974087b7bc95003a9ec11704`.

### 🟡 8 — the stranded retraction — **fixed**

Correct, and it is the textbook version of the campaign brief's own "fixed in N places, landed in N−1" lesson. I carried the WDW retraction through `RESULTS.md`, the notes and the PR body and missed the one place where the review file assesses its own artifact — `reference_review.review_notes` on `file:human/AP5B1/AP5B1-bioinformatics/RESULTS.md`, marked `relevance: HIGH / correctness: VERIFIED`, still saying "qualifies the fourth". That block now says all four claims are confirmed, states the PDBe coverage result that confirms the fourth, and records that an earlier revision read it the other way and why that was wrong.

I then grepped the whole gene folder for the retracted phrasing. The only surviving occurrences are the deliberate records of the correction in `RESULTS.md` §3 and the notes.

### 🔵 9 — the coverage test applied to `LLRLE@841` — **taken, and it is the inconvenient answer**

You are right that applying the discriminator to one hit and not the other is uneven, and right about which way it cuts. The run now prints it:

```
LLRLE@841: inside PF21590; 8YAB chain B modelled=False within chain span [7, 631]=False
```

841 is C-terminal to the modelled span, so the appendage is unresolved in 8YAB and "buried in a folded domain" rests on the Pfam assignment alone — and unmodelled often means flexible, which is the direction that weakens my own §1 claim. §1 now states this, and the notes carry the same caveat. Two things keep the conclusion standing and are named there: the AP1B1/AP2B1 boxes are not merely unstructured but sit in a linker three times longer, and AP-5's clathrin independence is a biochemical result, not a motif one.

### 🔵 10 — the hardcoded chain letter — **asserted now**

`assert_chain_is()` checks the SIFTS UniProt mapping before any coverage is read, so a re-lettered chain fails loudly rather than handing §3 another subunit's numbers. It also prints the entry composition, which turned out to be worth having on the record:

```
Structure 8YAB chain B (SIFTS-verified); entry composition:
  A=Q3U829 (AP5Z1_MOUSE), B=Q2VPB7 (AP5B1_HUMAN), C=Q9NUS5 (AP5S1_HUMAN),
  D=Q96JI7 (SPTCS_HUMAN), E=Q8BJ63 (AP5M1_MOUSE)
```

The deposited complex is a human/mouse hybrid — which is the same cross-species reconstitution the IntAct `Xeno` flag records for the AP5B1–Ap5m1 IPI row, so the two artifacts now corroborate each other.

### 🔵 11 — the `AP5B1_C` short name — **emitted**

`fetch_domains()` keeps the Pfam names, and the run prints them. One note: InterPro returns the long name, so the document now reads `PF21590 "AP5B1, C-terminal"` rather than `AP5B1_C` — the short form is not in the payload, and inventing it back would have broken the same rule twice.

### 🔵 12 — CPX-20045 provenance — **committed**

This was the right flag and I have closed it rather than noting it. New `complexportal_provenance.py`, with `complexportal.json` and `complexportal-output.txt` committed:

```
ComplexPortal complexes containing Q2VPB7 (AP5B1):
  CPX-5181     AP-5 Adaptor complex
  CPX-20045    AP5-Spastizin-spatacsin complex
  CPX-13475    CDC123:AP5B1
  CPX-19795    AP5Z1:CDC123:AP5B1:ZFYVE26:SPG11:AP5M1:AP5S1

CPX-20045: AP5-Spastizin-spatacsin complex
  participant: CPX-5181     AP-5 Adaptor complex  (1:1)
  participant: CPX-26503    Spastizin-spatacsin complex  (1:1)
  PubMed: 40175557, 25365221, 23825025
  GO cross-references: NONE
```

The script asserts CPX-20045 still contains Q2VPB7, and prints a retirement notice if a GO cross-reference ever appears — so re-running it is how a later curator discovers the proposed term is no longer needed. New `RESULTS.md` §5. The header's cross-check rule now names all four output files and flags the two paragraphs that are not from a run.

Incidental, recorded in the notes but not built on: ComplexPortal puts CDC123 (O75794) with AP5B1 in two complexes. CDC123 appears in no AP-5 paper I read and in no GOA row for this gene, so nothing in the review depends on it.

### Validation on the new head

```
just validate human AP5B1      → ✓ Valid (with 3 warnings)   [unchanged, all deliberate]
checkquotes.py                 → checked 95 quotes: 0 failures, 0 skipped
reconcile_AP5B1.py             → 24 GOA rows / 25 YAML rows: PASS
duplicate YAML keys            → none
cache/go/terms.csv             → 1 addition (GO:0044599), 0 deletions, no duplicate ids
```

Action tally unchanged: ACCEPT 9 / MODIFY 12 / MARK_AS_OVER_ANNOTATED 3 / NEW 1.

cmungall:
## Round 3 follow-up — the three optional items are closed

Head is now `c7094372506cf90e3f566f6ac997272e7ef205d5`. You approved on the previous head and marked 13-15 as not needing doing before merge; I took 13 anyway because it is the same internal-contradiction class as the round-2 blocker, sitting in the one paragraph the review YAML quotes, and 14-15 were one-line changes once I was in the file.

**13 — Conclusion aligned.** It read "buried in a folded C-terminal domain" while §1 had just said that call rests on the Pfam assignment alone. It now reads "assigned to", carries the §1 pointer, and adds the thing that was missing from the summary rather than merely hedged: of the three observations in that sentence, the linker-length one is load-bearing and is the only one resting on measured boundaries rather than a model assignment. The three passages quoted from this file in the review YAML are untouched — checkquotes re-run at 95/0 to confirm nothing was stranded.

**14 — CPX-26503 fetched.** `complexportal_provenance.py` now pulls detail for it too, so both halves of the notes' sentence are under the same assertion discipline:

```
CPX-20045: AP5-Spastizin-spatacsin complex   GO cross-references: NONE
CPX-5181:  AP-5 Adaptor complex              GO cross-references: NONE
CPX-26503: Spastizin-spatacsin complex       GO cross-references: NONE
```

**15 — the `GO:0044599` aside sourced.** You were right that the header rule is stated strongly enough that the one exception should be handled. Rather than exempting it I rewrote the bullet to say explicitly that the asymmetry is not from these runs, and to point at where it is sourced: GO:0044599 is on this gene in GOA and in the review's `core_functions`, and its definition names beta5. The document now has exactly three passages labelled as not-from-a-run, and all three say so.

### Validation on the new head

```
just validate human AP5B1      → ✓ Valid (with 3 warnings)   [unchanged, all deliberate]
checkquotes.py                 → checked 95 quotes: 0 failures, 0 skipped
reconcile_AP5B1.py             → 24 GOA rows / 25 YAML rows: PASS
cache/go/terms.csv             → 0 deletions
```

The diff is four files, all under `AP5B1-bioinformatics/`; no annotation content moved and the action tally is unchanged at ACCEPT 9 / MODIFY 12 / MARK_AS_OVER_ANNOTATED 3 / NEW 1.

Thanks for three careful rounds — the WDW item in particular turned a wrong conclusion into a better-evidenced right one rather than just a softer sentence.

cmungall:
Merged origin/main to clear a stale/real conflict on shared caches. Resolved: (no add/add conflicts; only untracked uncited caches moved aside). Preference rule: keep the copy with `full_text_available: true`, else the longer one. No review content changed; quotes re-verified: checked 95 quotes: 0 failures, 0 skipped (non PMID/file refs).

cmungall:
Fixed the 🔴: ran `uv run python -m ai_gene_review.tools.cache_lint --fix`, which removed the duplicated `GO:0044599` row that git's line-based merge had kept twice; `cache/go/terms.csv` now has zero duplicate CURIEs and its diff against origin/main is empty. No review content changed. The conflict-resolution helper used for this campaign now runs the cache lint after every merge, so this will not recur on later sibling PRs.

PR #3032, opened 2026-09-12
Review human AP5S1: sigma-5 is a real sigma-adaptin, but its dileucine site is chemically gone and plugged by the zeta N-terminus

cmungall:
## Headline

**Sigma-5 is a real sigma-adaptin that sits where sigma1/sigma2/sigma3 sit — but the one experimental AP-5 structure argues positively against it carrying a cargo-signal pocket, and there is no GO-expressible cargo function to give it.** The review annotates what is demonstrable (an architectural role in the AP-5 tetramer, a contact with the spatacsin WD40-hairpin, late-endosomal/lysosomal retrieval) and declines the adaptor MF that the family name invites.

AP5S1 / Q9NUS5 / 200 aa / PANTHER **PTHR16120:SF0** — *not* PTHR11753, which is why no clathrin-related term leaks onto this non-clathrin complex.

## Actions (30 rows: 28 seeded + 2 NEW) — revised after review round 1

| Action | n | What |
|---|---|---|
| ACCEPT | 15 | lysosome, late endosome and their membranes; AP-5 complex membership; endosomal transport (IBA, IEA, IMP); vesicle-mediated transport |
| MARK_AS_OVER_ANNOTATED | 6 | the five bare `GO:0005515` IPI rows, and the InterPro-derived `GO:0000724` IEA |
| KEEP_AS_NON_CORE | 6 | three cytosol rows (the unrecruited pool), `GO:0016192` vesicle-mediated transport (true grandparent, superseded by the child), `GO:0007040` lysosome organization, the `GO:0000724` IMP |
| MODIFY | 1 | `GO:0030119` **IDA** → `GO:0044599` AP-5 adaptor complex (the co-IP identifies a specific complex, and the parent's definition is written around linking clathrin) |
| NEW | 2 | `GO:0005198` structural molecule activity (IDA); `GO:0034499` late endosome to Golgi transport (**IC**, inference basis GO:0016197 + GO:0044599) |

Counts are produced by `.scratch/reconcile_AP5S1.py`, not asserted: it re-keys every GOA row on (term, evidence, reference, normalized WITH/FROM), requires a 1:1 match with the YAML, and checks that every `propagation_review.source_entities` id set equals its row's `supporting_entities`.

## The five IBAs

All five trace to **PTN000413531**, the only PTN node in PTHR16120. The local PAINT slice has exactly five IBD rows — `GO:0005764`, `GO:0005770`, `GO:0005829`, `GO:0030119`, `GO:0016197` — each seeded by **`UniProtKB:Q9NUS5` alone**, i.e. by the target itself. That is the valid self-referential pattern: the seed's evidence is IDA/IMP, not another propagation, and AP5S1 is the only human gene in the family, so it is inside the inheriting clade by construction. No IRD or IKR anywhere in the slice.

| term | root_cause |
|---|---|
| GO:0005764 lysosome | `NO_FAILURE_CORE` |
| GO:0005770 late endosome | `NO_FAILURE_CORE` |
| GO:0005829 cytosol | `NO_FAILURE_NON_CORE` (the unrecruited pool of a cycling coat) |
| GO:0016197 endosomal transport | `NO_FAILURE_CORE` |
| GO:0030119 AP-type membrane coat adaptor complex | `NO_FAILURE_CORE`, carrying the residue claims |

`GRANULARITY_MISMATCH` is explicitly **not** used on the `GO:0016197` IBA: its finer destination (late endosome to Golgi transport) was established seven years after the IBD was placed, and by a knockout of a different subunit, so the parent was the right call on the evidence the curator had. The finer term is added as a NEW row instead.

The NEW `GO:0034499` IC row and six IEA rows also carry `propagation_review` (three UniProt SubCell mappings, three InterPro2GO). Each SubCell identifier was resolved at the UniProt locations endpoint and each maps to exactly the GO term used (SL-0157→GO:0005765, SL-0091→GO:0005829, SL-0151→GO:0031902).

## The one propagation defect found

`InterPro:IPR029392 → GO:0000724` (homologous recombination repair) is marked `EVIDENCE_CIRCULAR_OR_REDUNDANT`. The InterPro entry was fetched from the API: its description says the family "is required for efficient homologous recombination DNA double-strand break repair" and cites **PMID:20613862** — the same single HeLa DR-GFP esiRNA experiment that already gives this gene its IMP. So the electronic row adds no evidence about AP5S1 while doubling the apparent support for its weakest process claim, and it carries a human cell-line phenotype family-wide (790 proteins / 2,294 taxa). Within that original experiment the phenotype does not behave like a property of the complex either: knocking down SPG11 or beta5 had no effect. Raised with InterPro in `suggested_questions`.

## Bioinformatics — `AP5S1-bioinformatics/sigma5_structure.py`

Fetches **8YAB** (AP-5 core + SPG11 WD40-hairpin, 3.26 Å) and **2JKR** (AP-2 core + CD4 dileucine peptide) live from RCSB and all sequences from UniProt; asserts expected lengths and verifies deposited-versus-UniProt numbering per chain, which caught a one-residue offset in the zeta chain that would otherwise have mislabelled every zeta residue reported.

1. **Interfaces.** Sigma-5 buries **2,621.8 Å²** against the zeta trunk over 60 residues and contacts **neither beta5 nor mu5** — the canonical sigma position. It additionally buries **755.4 Å²** against the SPG11 WD40-hairpin over 20 residues, a contact no other AP sigma subunit has.
2. **The dileucine site, defined empirically.** The site is taken as the 19 sigma2 residues within 4.5 Å of the bound CD4 peptide in 2JKR (no residue list typed in), then carried onto sigma-5 by sequence-independent CE superposition. **18/19 positions have a structural equivalent** (RMSD 3.66 Å; positive control, the second sigma2 copy, 0.03 Å; negative control, the beta5 solenoid, 6.82 Å) but **only 3/19 carry the same residue**: sigma2 V88→T143 and V98→R151 replace pocket-lining hydrophobics with polar and basic side chains, and R15 — the FHH3 disease residue that reads the acidic position of the motif — is Y25. Three of the equivalent positions are in contact with the **zeta N-terminus** (zeta residues 1–10) in the assembled core. The fold is retained; the chemistry is not; and the surface is not free.
3. **Sequence control.** Every pairwise alignment of sigma-5 to a human sigma paralog scores negative and within a few SD of shuffled sequence (best z = 3.09), so this mapping could not have been made from pairwise sequence alignment — consistent with Hirst et al. having needed HHpred rather than BLAST. This is why the comparison is done on structures.

Encoded as five `residue_claims` (2 RETAINED, 3 SUBSTITUTED, method STRUCTURE) on the `GO:0030119` IBA; `uv run python -m ai_gene_review.validation.gene_residue_claims` → 94 pass, 0 fail, 0 unresolved.

## GO representation gaps recorded

* **No term for the hexamer.** ComplexPortal curates CPX-20045 (AP5Z1:AP5B1:ZFYVE26:SPG11:AP5M1:AP5S1) with **no GO cross-reference**, while the tetramer CPX-5181 maps to GO:0044599. The consequence is visible in this gene's own rows: the three ComplexPortal rows sourced from the structural paper reach AP5S1 as *location* and *process* statements because there is nothing to be `part_of`. Proposed as a new term.
* **No fitting MF for a non-clathrin coat adaptor.** GO:0140312 requires binding clathrin- or COPII-like scaffolding and bridging a cargo receptor; ComplexPortal asserts it on CPX-5181 anyway. Not adopted here.
* **No cargo, no signal.** Recorded as a BIOLOGY gap with `dark_aspect: MF_DARK`, with the field's own admissions as provenance.

## What affinage missed

Everything. `n_discoveries: 0`, `citation_count: 0`, "No mechanistic discoveries found in literature", with trust gates clear — a clean illustration that the gates measure precision, not recall. All eight AP-5 papers used here (20613862, 22022230, 23825025, 26085577, 29381698, 33464297, 40175557, 40081374) were found by independent Europe PMC searches on the symbol, the complex and the partner proteins, plus two AP-2 comparator papers (19140243, 23222959). The affinage reference is kept with `relevance: NONE`.

## Validation

```
just validate human AP5S1        → ✓ Valid (with 3 warnings)
checkquotes.py                   → checked 127 quotes: 0 failures, 0 skipped
gene_residue_claims              → 94 pass, 0 fail, 0 unresolved
duplicate-key check              → no duplicate keys
cache/go/terms.csv               → no deletions, no duplicate ids
reconcile_AP5S1.py               → GOA rows 28 / YAML reviewed rows 28 / NEW 2; 12 propagation_review blocks; reconciliation OK
reference titles vs cache        → 11/11 exact, 0 mismatches
```

The three warnings are deliberate and are explained in the rows themselves: two "inconsistent actions for the same term" (the `GO:0000724` IEA is redundant while the IMP is a real if peripheral observation; the `GO:0030119` IDA is modified to the specific child while the IBA and IEA are kept as correct general parents), and one "no annotation references the deep research file" — which is because that file is empty, and citing it as support for anything would be false.

## Self-audit after opening (commits 97bde83..4dbb672)

Five corrections made by re-checking my own claims rather than in response to review:

* "AP5S1 is the sole human member of PTHR16120" → **the only human *gene*** in it; a UniProt query for the signature in taxon 9606 returns Q9NUS5 plus two unreviewed TrEMBL records of the same gene.
* "AP-5 is the only coat complex known to work on this compartment" → attributed as the statement made **on discovery in 2011**, not asserted as current fact.
* Two `supporting_text` quotes were verbatim and from the right paper but did not bear on their claims ("The AP5 complex is in a super-open conformation" was supporting a structural-subunit claim); replaced with quotes that do.
* The proposed-term justification said spatacsin is "required for the AP-5 tetramer to form at all"; the cited abstract says its N-terminal region is required for "AP5 complex interaction and assembly", and the text now says that.
* The notes' decisions summary said the two SubCell membrane IEAs were kept non-core; they are ACCEPT, and cytosol is the non-core case.
* The five protein-binding rows still said the subunit-level molecular statement "has no GO molecular function to modify to" after GO:0005198 had been added as a NEW row; they now say why an IPI row is nevertheless not the route to it.
* The mouse-Ap5z1 row described the co-expression protocol, which is only in the preprint; it now states what is checkable from the deposition (8YAB chain C = human Q9NUS5, chain A = mouse Q3U829, each verified against its UniProt sequence).


cmungall:
## Round 1 addressed — all eight points, at head `48c450f`

**🟡 #1 (blocking) — `GO:0034499` evidence code.** You are right, and I checked the paper again rather than taking the finding on trust: PMID:22022230 reports the CIMPR/Vps26 redistribution and concludes "a role for the C14orf108-DKFZp761E198 complex in endosomal trafficking", and its introduction does say of AP-1 that "there is still some uncertainty about the direction". Coding IMP there asserted an experiment that paper did not do.

Re-encoded as **`IC`** on `PMID:29381698`, with `supporting_entities: [GO:0016197, GO:0044599]` — the two annotations the inference actually rests on, both held by this gene — and a `propagation_review` whose `source_entities` name each of them with what it contributes (GO:0016197 gives the pathway from sigma-5's own IMP; GO:0044599 is what carries the complex-level direction to the subunit, licensed by the all-six-knockdowns phenocopy). The `reason` now states explicitly why IMP was wrong. I took the IC branch rather than the "IMP on PMID:29381698" branch because the perturbation there is AP5Z1, and IMP on this gene would have the same defect in a different place.

**🔵 #2 — the AP-4 precedent.** Checked and you are right twice over: AP4S1 (Q9Y587) has **no** GO:0005198 in GOA (QuickGO, molecular_function aspect, returns nothing), so "carries" was false, and the AP4S1 review that proposes it merged to main as #3026 and is not in this branch's base. Clause dropped from the YAML; the notes now say what is actually true, flagged as not verifiable from this branch. The justification never depended on it.

**🔵 #3** — noted, and the bounding language is kept prominent (it is also the third knowledge gap).

**🔵 #4 — `GO:0016192`.** Agreed, and the enum semantics are the deciding argument: changed to `KEEP_AS_NON_CORE` with a reason saying it is a true grandparent superseded by the child already on the gene.

**🔵 #5 — `GO:0005765`.** Omission, not intent. Added to `core_functions[0].locations`.

**🔵 #6 — the description over-reach.** Fixed; it now reads "an interface whose spatacsin side is required for the complex to assemble", which is what the abstract supports. Fair hit that this survived the self-audit that caught the same class of error elsewhere.

**🔵 #7 — mouse anchor vs human disease evidence.** Added to the R15 claim's comment, and measured rather than asserted: human P53680 and mouse P62743 were fetched and compared, and are identical over all 142 residues, so position 15 is R in both and nothing numeric moves.

**🔵 #8 — affinage `correctness`.** Changed to `UNVERIFIED`; you are right that the enum is about whether the record supports the claim, not whether the fetch was faithful.

### Validation at this head

```
just validate human AP5S1   → ✓ Valid (3 warnings, the same three, each explained in-row)
checkquotes.py              → checked 127 quotes: 0 failures, 0 skipped
gene_residue_claims         → 94 pass, 0 fail, 0 unresolved
reconcile_AP5S1.py          → GOA 28 / YAML 28 / NEW 2; 12 propagation_review blocks; OK
reference titles vs cache   → 11/11 exact
duplicate keys / terms.csv  → none / no deletions, no duplicate ids
```

Action counts are now 15 ACCEPT · 6 KEEP_AS_NON_CORE · 6 MARK_AS_OVER_ANNOTATED · 1 MODIFY · 2 NEW.

Thank you for verifying the PAINT slice and the residue table against the repo data independently — that is exactly the check I most wanted a second pair of eyes on.

cmungall:
## Round 2 leftovers closed at `190af4a`

Both were real defects, so I fixed them rather than leaving them.

**R15 anchor comment — the claim now lives where it says it does.** You were right that "were fetched and compared here" had no referent in the code: I had done that comparison in a throwaway shell command, and `sigma5_structure.py` only fetched P53680 as one of the paralog-alignment controls. The comparison is now a section of the script in its own right — it takes the peptide-contacting positions it already computed, fetches P53680, and reports length equality, full-chain identity, the list of differing positions (empty), and whether every contact position is identical. `RESULTS.md` gains:

> Same length: True. Identical over the full chain: True. All peptide-contacting positions identical between the two: True.

The residue-claim comment now points at that section instead of asserting the result. The `RESULTS.md` diff is purely additive (+6 lines), so no previously quoted line moved and `checkquotes.py` is unchanged at 127/0.

**History record.** Rather than rewriting the CREATE record — whose 16/6/5 and 123-quote figures were accurate at the time it was written — I added a `REVIEW` record documenting the round-1 revision, the reason the IMP→IC re-encoding was necessary, all seven polish items, and the final counts, with a note that it supersedes the earlier figures. `just validate-history` passes on it.

Validation at this head is unchanged from the last comment: `just validate` ✓ (same 3 explained warnings), `checkquotes.py` 127/0, `gene_residue_claims` 94 pass, reconciliation OK (28/28, 2 NEW, 12 propagation_review blocks).

cmungall:
Merged origin/main to clear a stale/real conflict on shared caches. Resolved: (no add/add conflicts; only untracked uncited caches moved aside). Preference rule: keep the copy with `full_text_available: true`, else the longer one. No review content changed; quotes re-verified: checked 127 quotes: 0 failures, 0 skipped (non PMID/file refs).

cmungall:
Post-merge cache lint: removed duplicate CURIE row(s) (GO:0044599) that the line-based merge of origin/main had kept twice in `cache/go/terms.csv`. No review content changed.

PR #3036, opened 2026-09-12
Review human APOC4: three of four IBAs rest on the target's own overexpression data, and the family's only lipid-binding measurement is uncurated

cmungall:
## Headline

**Three of APOC4's four IBAs are seeded by exactly one gene — human APOC4 itself — and for the newest of them that seed is a single 2008 transfection of a hepatoma line.** The PANTHER node is sound and the family is clean (`PTHR32288` is apoC-IV only, so none of the feared apoC-II/apoC-III biology leaks in), but the evidence underneath the node is much thinner than four IBA rows across eleven species suggests. Separately, the Ensembl-Compara VLDL IEA turns out to re-derive, via the mouse, a claim the human gene already holds by direct experiment **from the same publication** — and its two `WITH/FROM` identifiers are one donor, not two.

Gene: **APOC4** (apolipoprotein C-IV), UniProtKB **P55056**, PANTHER **PTHR32288:SF0**.

## Actions

41 seeded GOA rows, all reviewed; 1 NEW row added. Every count below is produced by `.scratch/APOC4_reconcile.py`, not asserted.

| action | n | rows |
|---|---|---|
| `ACCEPT` | 11 | GO:0005319 lipid carrier activity (TAS); GO:0005576 extracellular region ×5 (1 IEA + 4 Reactome TAS); GO:0034361 VLDL particle ×5 (IBA, 2×IDA, IEA, IMP) |
| `KEEP_AS_NON_CORE` | 7 | GO:0006629 lipid metabolic process; GO:0019915 lipid storage (IBA, IMP); GO:0034364 HDL particle (IBA, IDA); GO:0070328 triglyceride homeostasis (IBA, IMP) |
| `MARK_AS_OVER_ANNOTATED` | 23 | GO:0005515 protein binding, all from PMID:32296183 (HuRI) |
| `NEW` | 1 | GO:0005543 phospholipid binding, ISS from rabbit P55057 |
| `REMOVE` | 0 | — |

No `REMOVE`: nothing here is contradicted. The 23 two-hybrid rows are experimental and project policy is not to remove such rows on compartment grounds, so they are marked over-annotated instead.

## Per-IBA propagation verdicts

All four IBAs descend from one node, `PANTHER:PTN000792756`, at `taxon:9347` (Eutheria), read from the committed PAINT slice `interpro/panther/PTHR32288/PTHR32288-paint.tsv`. The node propagates all four terms to exactly **11** eutherian gene products (human, mouse, rat, chimp, gorilla, macaque, cat, dog, horse, pig, cow), counted from the cached PAINT GAF; there is no IRD or IKR in the slice.

| term | seeds (from the IBD row) | `root_cause` | action |
|---|---|---|---|
| GO:0034361 VLDL particle | `MGI:MGI:87878` + `UniProtKB:P55056` | `NO_FAILURE_CORE` | ACCEPT |
| GO:0034364 HDL particle | `UniProtKB:P55056` | `NO_FAILURE_NON_CORE` | KEEP_AS_NON_CORE |
| GO:0070328 triglyceride homeostasis | `UniProtKB:P55056` | `NO_FAILURE_NON_CORE` | KEEP_AS_NON_CORE |
| GO:0019915 lipid storage | `UniProtKB:P55056` | `SOURCE_WEAK_OR_INFERRED` (+`SOURCE_EVIDENCE_WEAK`) | KEEP_AS_NON_CORE |

Three points the blocks make in detail:

- **The self-seed is correct, not circular.** `UniProtKB:P55056` appearing in its own `WITH/FROM` is a marker that experimental grounding exists on the target; it is recorded `SUPPORTS_TRANSFER` in every block. What varies between the four rows is the *strength* of that descendant evidence, and that is what the `root_cause` values track.
- **The VLDL row's "two donors" are two species in one study.** `MGI:MGI:87878` resolves (via `rest.uniprot.org/uniprotkb/search?query=xref:mgi-87878`) to `Q61268`, APOC4_MOUSE — a true orthologue. Its **only** experimental GO annotation in QuickGO is `GO:0034361 IDA` from **PMID:12700345**, which is the mouse-plasma clause of the same paper that gives the human target its IDA. The claim is still solid (direct isolation, independently replicated on human VLDL by PMID:17154273) but it is not the extra independence a second donor usually implies.
- **The lipid-storage IBD is dated 2025-06-24**, seven months after the BHF-UCL IMP it rests on (2024-11-01), and that IMP is one transient transfection of Huh-7 cells (~2.7-fold triglyceride rise, PMID:18809223) in a paper whose main subject is what *regulates* APOC4. One cell-culture result now asserts intracellular lipid storage as an ancestral eutherian property of a secreted plasma apolipoprotein, in eleven species.

Two further `propagation_review` blocks on non-IBA rows (7 blocks total):

- **GO:0034361 IEA, GO_REF:0000107** → `EVIDENCE_CIRCULAR_OR_REDUNDANT`. The donor `UniProtKB:Q61268` is the right orthologue, but its own GO:0034361 IDA is PMID:12700345 — the human target's own paper. The reciprocal transfer also runs the other way (mouse carries the term by GO_REF:0000107 from human ENSP00000468236). Both `WITH/FROM` entries (`UniProtKB:Q61268`, `ensembl:ENSMUSP00000003071`) are the **same mouse gene product under two identifiers**, and both are marked `CIRCULAR_OR_REDUNDANT`. Kept as ACCEPT — the term is right and duplicates are not a defect — with the redundancy recorded rather than acted on.
- **GO:0005576 IEA, GO_REF:0000044** → `NO_FAILURE_CORE`. `UniProtKB-SubCell:SL-0243` resolved to "Secreted" against the UniProt locations service; keyword and mapping both correct.

`source_entities` are built **from** each row's seeded `supporting_entities` by script, in the seeded order; the builder hard-fails on an authored source the row does not carry, or a seeded source with no judgment.

## NEW term

**GO:0005543 phospholipid binding, ISS, PMID:8576182, `supporting_entities: [UniProtKB:P55057]`.**

After review, APOC4's only informative molecular function is GO:0005319 (lipid carrier activity), which states the role but not the binding step beneath it. That step *has* been measured — on the rabbit orthologue, which "contains two potential amphipathic helices characteristic of plasma apolipoproteins and forms discoidal micelles with phosphatidylcholine". Coded **ISS, not IDA**, because the measurement is on P55057 and the supporting sequence analysis here is a statistic, not an assay. GO:0031210 (phosphatidylcholine binding) would over-specify from one reconstitution lipid; GO:0008289 (lipid binding) would lose the monolayer. APOC3 and APOA1 both carry GO:0005543 by IDA and IBA, so this puts apoC-IV at the same resolution as its characterised relatives.

## Checked, and cleanly negative

- **No lipase-regulation row exists on APOC4, and none should.** The mechanism proposed for the transgenic phenotype is apoE displacement, *explicitly not* a lipase effect (PMID:37226733). LPL activation (GO:0060230) and phospholipase activation (GO:0016004) are on APOC2, not here.
- **No chylomicron row, and no evidence for one.** Every direct measurement puts apoC-IV on VLDL and HDL.
- **The APOC4-APOC2 readthrough contributes nothing here.** It is a separate UniProt accession — `K7ER74`, unreviewed, 178 aa, which UniProt names *Apolipoprotein C-II*. It carries its own seven IEA rows including exactly the terms one would worry about leaking (GO:0008047 enzyme activator activity, GO:0042627 chylomicron, GO:0006869 lipid transport). **None of them appears on P55056.** The GWAS literature's "APOC4-APOC2" is a locus label, not this transcript.
- **Reference-projection test on PMID:32296183 (HuRI).** Full QuickGO export: **85,343 annotations across 6,769 distinct gene products, 84,937 of them GO:0005515.** Resolving all 23 APOC4 partners in UniProt by script: **0** have a curated Secreted/Extracellular location, **10** are curated mitochondrial, **10** carry a transmembrane segment, 1 (THBD) has a cleaved signal peptide, 1 (SYT16) has no curated location. APOC4 is secreted with a cleaved signal peptide and never occupies the nuclear compartment where Gal4 fusions are tested. THBD is flagged in its own row as the single partial exception (it does face blood) with the caveat that a nuclear two-hybrid cannot test that topology.

## Bioinformatics

`genes/human/APOC4/APOC4-bioinformatics/` (uv project, two scripts, live fetches, assertions on inputs, no hardcoded results). One script rejected a demerged accession (`P62988`) rather than silently scoring it — the guard is retained.

1. **Family taxonomic span** — `family_taxonomy.py` pulls all PTHR32288 members from InterPro and resolves each source organism's lineage from UniProt taxonomy. **185/189 members and 146/149 organisms are inside Eutheria**; the only non-eutherian apoC-IVs are three marsupial sequences (wombat, koala ×2) plus one uncharacterised *Tetrahymena* protein. This independently corroborates the node's `taxon:9347` placement and the published phylogeny ("ApoC-IV first arose in Eutheria", PMID:31722659) — the node sits at or just inside the family's own origin, so the target is inside the inheriting clade by construction.
2. **Amphipathic-helix scan** — `amphipathic_helix.py`, Eisenberg hydrophobic moment over 18-residue windows at 100°/residue on mature CHAINs. The single strongest window does **not** discriminate (both globular controls reach μH ≈ 0.46–0.48). The discriminating statistic is the density of non-overlapping μH ≥ 0.35 segments: seven exchangeable apolipoproteins fall at **0.0206–0.0351** per residue, two globular controls at **0.0096–0.0101**, no overlap (2.04× to 3.66× separation, derived by the script). Human apoC-IV is **0.0300** — 0.0009 from rabbit apoC-IV and 0.0021 from APOA1.

## What affinage missed

Trust gates cleared (`APOC4: trust gates clear`), `faith_pct: 100.0`, empty `self_evaluation_pairwise`, correct protein, no symbol collision — so no `LOW_QUALITY` flag. The problem is recall. Its four citations are PMID:8530039, PMID:10996355, PMID:29580721, PMID:33705959. APOC4 has 29 experimental GOA rows, but 23 are the single HuRI screen; the other **six** trace to exactly four references — PMID:12700345, PMID:8827523, PMID:18809223, PMID:17154273 — and affinage cited **none of them**. It also missed PMID:8576182 (the rabbit paper carrying the family's only direct lipid-binding measurement), PMID:37226733 (the review with the apoE-displacement mechanism and the not-a-lipase negative), and PMID:31722659 (the Eutheria origin). Its `mechanism_profile` is empty, so no GO grounding was available to import. It is cited once, for its own admission of ignorance, and carries a `reference_review`.

## Curation gap worth flagging to GOA/UniProt

**Rabbit APOC4 (P55057) carries five GO annotations and all five are IEA** (four from the GO_REF:0000118 PANTHER ancestral-node pipeline, one from the subcellular-location mapping). Yet rabbit is the species where apoC-IV was purified, shown to form discoidal micelles with phosphatidylcholine, localised to VLDL and HDL, shown to be liver-restricted and shown to be secreted (PMID:8576182). PAINT therefore **could not** use the best-characterised member of the family as an IBD seed, which is why three of four IBDs fall back on a single human seed, two of which are overexpression experiments. Curating PMID:8576182 to P55057 would give PTN000792756 its only non-human, non-overexpression descendant evidence. Recorded as a `CURATION` knowledge gap and as a suggested question.

## Synthesis

- 2 `core_functions`: phospholipid-surface binding via amphipathic helices (MF GO:0005543, `in_complex` GO:0034361), and exchangeable apolipoprotein constituent of plasma VLDL (MF GO:0005319, `in_complex` GO:0034361). Lipoprotein particles are GO:0032991 descendants, so they are recorded in `in_complex`, not `locations`.
- 4 `knowledge_gaps` (3 BIOLOGY, 1 CURATION, 1 with both), each with verbatim provenance: what endogenous apoC-IV does at native concentration (no knockout exists in any species); whether the apoE-displacement mechanism is real for apoC-IV or borrowed from APOC1/C3; the uncurated rabbit biochemistry; and the unexplained rabbit-vs-human abundance difference.
- 5 `suggested_questions` (to PAINT, GOC/MGI, UniProt/GOA, IntAct, and the ontology editors) and 5 `suggested_experiments`.
- 22 references, **every one** with a `reference_review` (relevance + correctness + notes), including the affinage record, the UniProt flat file and the bioinformatics RESULTS.md.

## Validation evidence

Run after merging `origin/main` into the branch:

```
$ just validate human APOC4
Validating genes/human/APOC4/APOC4-ai-review.yaml...
✓ Valid

Checking PMID references in markdown files...
No pathway markdown files found in genes/human/APOC4
✓ All validations passed for human/APOC4

$ uv run python checkquotes.py genes/human/APOC4/APOC4-ai-review.yaml
checked 148 quotes: 0 failures, 0 skipped (non PMID/file refs)

$ uv run python .scratch/APOC4_reconcile.py
GOA lines: 41 (41 distinct keys)
YAML entries: 42 = 41 seeded + 1 NEW
references: 22 declared, 22 used
actions: {'ACCEPT': 11, 'MARK_AS_OVER_ANNOTATED': 23, 'KEEP_AS_NON_CORE': 7, 'NEW': 1}
propagation_review blocks: 7

reconciliation OK: every GOA row maps to exactly one YAML entry and back
```

Also clean: `linkml-validate` (schema) "No issues found"; `linkml-term-validator --labels` "✅ Validation passed"; duplicate-YAML-key check "no duplicate keys"; no `cache/go/terms.csv` deletions and no duplicate ids in it. No residue claims were made, so `gene_residue_claims` does not apply — the two places a residue argument might have been expected (the amphipathic helices, the particle partitioning) are distributed surface properties, and each `propagation_review` says so via `residue_claims_not_applicable`.


cmungall:
Thanks — all four suggestions acted on in `07cbf19d48`. Three adopted as written, one adopted in the alternative form you offered, with the reason below.

**🔵 #2 — the partner script is not in the diff.** Agreed, and this was the one that mattered most: those counts carry 23 annotation decisions. The analysis is now a committed, re-runnable one — `APOC4-bioinformatics/huri_partner_topology.py`, with `huri_partners.tsv` (per-partner rows) and a new RESULTS.md §3. It asserts the target resolves to APOC4_HUMAN, is curated Secreted and carries a signal peptide before it scores anything, and rejects inactive accessions rather than silently scoring them. The review YAML text is now *generated from* that script's JSON output, so the 0/10/10/1/1 counts in the YAML, the notes and RESULTS.md all derive from one artifact that is in the diff. Reproduces exactly: 0 secreted, 10 mitochondrial, 10 transmembrane, 1 signal peptide (THBD), 1 with no curated location (SYT16).

`.scratch/APOC4_reconcile.py` stays out of the tree deliberately — it is a QA harness that checks the deliverable rather than producing any of it, and the campaign brief asks for scratch files to be kept out of the repo. Its output is quoted in the PR body so the numbers are still checkable.

**🔵 #4a — RSAD2.** You were right and the fix generalised. The bespoke mitochondrial clause now fires only for partners curated *exclusively* to the mitochondrion — MICOS10, MICOS13, MAIP1, BCL2L2, TIMMDC1 — so RSAD2, whose curated list leads with ER membrane / Golgi / lipid droplet, falls back to the general argument. And the general argument is now the one you pointed at: not "intracellular", which proves nothing for a protein that transits the ER and Golgi, but **which face** of the compartment. APOC4 is lumenal in transit and extracellular after, and never faces the cytosol, the mitochondrial matrix or the nucleus — which is where all 23 partners work, and which is exactly what a nuclear Gal4 two-hybrid tests.

**🔵 #4b — the `description` hedge.** Caught a real miss: I tightened that claim in `core_functions` in `ee7e2dad2` and left the same claim unhedged one field away. Now reads "No experimental structure has been determined for apoC-IV in any species, but the sequence has the architecture of an exchangeable apolipoprotein — predicted amphipathic helices that would lie on the phospholipid monolayer …".

**🔵 #5 (your note) — NbExp=3.** Added, since a reader will hit it: each row now says the NbExp=3 in the UniProt INTERACTION block is HuRI's own screen-and-retest count from this one publication, not three independent studies.

**🔵 #1 — the ISS has no annotated source yet.** Adopted. The `review.reason` for GO:0005543 now states the ordering plainly: under GOA's ISS convention the WITH/FROM source is expected to carry the transferred term experimentally, P55057 carries no experimental GO annotation of any kind, so the row is contingent on PMID:8576182 first being curated to P55057 — which is the curation request already recorded as a CURATION knowledge gap and as suggested question #3. The measurement exists; the GO record of it does not.

**🔵 #3 — lipid-storage IBA action.** Adopted as the sentence, not the action change, and I want to be explicit about why rather than quietly declining. I checked the action change first: `src/ai_gene_review/validation/validator.py` warns `inconsistent_review_actions` when annotations to the same term carry different actions (the only exemption is a REMOVE split with disjoint references). Setting the IBA to `MARK_AS_OVER_ANNOTATED` while the IMP stays `KEEP_AS_NON_CORE` would introduce a validation warning into a file that currently has none. So the `reason` now does what you suggested as the alternative — separates the two rows explicitly, says the IMP is non-core because one ectopic overexpression does not establish a physiological role while the IBA is a propagation whose eleven-species reach is what over-extends, concedes that `MARK_AS_OVER_ANNOTATED` is arguable on the propagation alone, and points at where the over-reach is actually recorded: `propagation_review.root_cause: SOURCE_WEAK_OR_INFERRED`. If the project would rather the two rows diverge and the consistency check is not meant to bind here, I will make that change.

**Validation, re-run on `07cbf19d48` after merging `origin/main`:**

```
just validate human APOC4          ✓ Valid   (no errors, no warnings)
checkquotes.py                     checked 148 quotes: 0 failures, 0 skipped
APOC4_reconcile.py                 41 GOA lines -> 42 YAML entries (41 + 1 NEW); 22 refs declared, 22 used;
                                   7 propagation_review blocks; reconciliation OK
linkml-validate (schema)           No issues found
linkml-term-validator --labels     ✅ Validation passed
duplicate-YAML-key check           no duplicate keys
cache/go/terms.csv                 no deletions, no duplicate ids
```

On your stated limitation — the live-lookup claims you could only check for internal consistency were all re-verified here against the live services: the GO:0005319 rename (QuickGO `history`, `Updated TERM` 2026-02-09, narrow synonyms `apolipoprotein`/`lipophorin`, `replaces` GO:0005321/GO:0005323 and `consider` GO:0005320); APOC3 and APOA1 both carrying GO:0005543 by IDA and IBA; the QuickGO export totals for PMID:32296183 (85,343 annotations / 6,769 gene products / 84,937 GO:0005515); and the reciprocal mouse GO_REF:0000107 row on Q61268 pointing back at human ENSP00000468236. I also re-confirmed from the cached PAINT GAF that the node's 44 rows are all IBA across 11 eutherian gene products, with no IRD, IKR or NOT qualifier anywhere in the slice.

cmungall:
Fixed in `b42440d9b3` — you were right that the paragraph overstated its own exceptions. RESULTS.md §3 now reads "Twenty-one of the 23 partners do their work on the cytosolic side or inside an organelle APOC4 never enters; the two that are not in that class are handled below", and the claim that genuinely does cover all 23 — that every pair was scored by a Gal4 two-hybrid in the yeast nucleus, a compartment the target cannot occupy at all — is now its own clause rather than being folded into the weaker one. The 21/2 split is derived from `huri_partner_topology_result.json` (THBD has a cleaved signal peptide, SYT16 has no curated location), not counted by eye.

Nothing else changed: no action, term, supporting_text or reference value is touched by this commit. `just validate` still ✓ Valid with no warnings and `checkquotes.py` still 148/148.

cmungall:
Both findings acted on in `53eb9de0f8`; I verified each against the ontology/data before changing anything.

**#1 — the GO:0006629 justification asserted a relationship that does not hold.** You were right, and I checked it rather than taking it on trust. QuickGO's `ancestors` endpoint: `GO:0070328` resolves to `[GO:0008150, GO:0055088, GO:0042592, GO:0048878, GO:0055090]` — lipid homeostasis, not lipid metabolic process — and `GO:0019915` to `[GO:0008150, GO:0170062, GO:0009987]`, cellular process. `GO:0006629` is in neither ancestry, under `is_a` or under all relations. And the two particle terms are cellular components, so they could not have been descendants of a BP term in any case. So *none* of the four terms I named was a child of GO:0006629 and the "broad parent alongside specific children" framing was simply false.

The reason now says what is actually true: the row is **not** redundant with anything on the gene — it is the most general true statement available about APOC4 and the only metabolic-process term it carries, which is why it is uninformative rather than duplicative. The action is unchanged; `KEEP_AS_NON_CORE` is what a true, uninformative term should get, and that was never what the bad justification was load-bearing for.

**#2 — PORCN.** Taken, and generalised rather than patched. I could not settle MBOAT active-site topology either, and the right response to an unsettled point is not to pick a side — it is to stop resting a count on it. RESULTS.md §3 no longer claims any exact number of topologically impossible partners. It now separates the firm claim from the soft one:

- **Firm:** all 23 pairs were scored by a Gal4 two-hybrid in the yeast nucleus, a compartment APOC4 cannot occupy at all, so the screen could not have observed any of them where it would matter. This covers every row regardless of partner topology.
- **Soft, and now labelled as such:** the per-partner argument is strongest for the ten curated mitochondrial partners and sharpest for the five curated exclusively there; it is weakest for ER/Golgi-membrane partners, where APOC4 transits the same organelle and everything turns on which face the partner works on — with PORCN named as the example where that face is contested.
- **What the table shows without qualification:** none of the 23 has a curated secreted or extracellular location, so not one is a plasma protein of the kind apoC-IV actually encounters.

No action, term, supporting_text or reference value changed in this commit. `just validate` ✓ Valid with no warnings; `checkquotes.py` 148/148; reconciliation still 41 GOA lines ↔ 42 YAML entries with 7 propagation_review blocks.

Both of these were the same class of error — a justification stronger than what I had actually established — and both were worth the round. Thanks for catching them.

cmungall:
Fixed in `9523dcfe81`. This was the right catch and it is the failure mode the campaign brief names explicitly — fixed in N places, landed in N−1. I hedged the analysis and left the deliverable asserting the version I had just retracted, in 23 rows plus the notes.

All 23 `MARK_AS_OVER_ANNOTATED` rows now carry the same two-tier structure RESULTS.md §3 uses:

- **Firm, no partner topology required:** not one of the 23 is a plasma protein of the kind apoC-IV encounters, and every pair was scored by a Gal4 two-hybrid — reconstituted in the yeast nucleus, a compartment the mature chain cannot occupy at all.
- **Softer, and now explicitly graded:** each partner's own compartment — strongest for the ten curated mitochondrial partners, weakest for the ER/Golgi-membrane ones, since APOC4 does transit those organelles and the argument then turns on a membrane face that is not settled for every partner. Where the per-partner point is strong it is made in the row itself rather than in the shared paragraph.

The THBD row is the one you singled out and it no longer contradicts itself: the generic paragraph makes no blanket topology claim, and the bespoke clause now says the encounter is not topologically absurd "the way it is for the mitochondrial partners" rather than "the way the others are" — which is what the table actually supports. `APOC4-notes.md` got the same treatment, with PORCN named there too as the case that makes an exact count unclaimable.

No action, term, supporting_text or reference value changed; `just validate` ✓ Valid with no warnings, `checkquotes.py` 148/148, reconciliation unchanged at 41 GOA lines ↔ 42 YAML entries with 7 propagation_review blocks.

Three of the four rounds on this PR have been the same class of defect — a sentence claiming more than the data behind it — and each time the fix was to weaken the claim rather than defend it. Worth noting for whoever picks up the next gene in this campaign.

cmungall:
Fixed in `4a70564b5f` — and you were right that the fix belonged in the topic sentence rather than the retraction. `APOC4-notes.md` now opens the paragraph on what the table establishes without qualification ("gives a partner set with no plasma protein in it at all — nothing of the kind a secreted apolipoprotein actually encounters") instead of on a blanket impossibility claim it then withdraws.

Grepping the whole gene folder, the phrase "topologically impossible" now survives in exactly two places, both of them the explicit retractions: `RESULTS.md:178` and `APOC4-notes.md:241`. Nothing asserts it any more.

Notes-only commit; no YAML change at all. `just validate` ✓ Valid, `checkquotes.py` 148/148.

That closes the last instance of the defect this PR kept hitting: a sentence claiming more than the data behind it. The count that survives everywhere is the one the TSV licenses — 0 of 23 partners with a curated secreted or extracellular location.

cmungall:
Fixed in `945e7e3021`. You were right that the `9523dcfe8` sweep stopped at the 23 rows and the notes and never reached `knowledge_gaps`.

`knowledge_gaps[1].boundary` now reads: the 23 GO:0005515 rows all come from a single yeast two-hybrid screen, scored in the yeast nucleus, and not one of its 23 partners is a plasma protein of the kind apoC-IV encounters. That is what `huri_partners.tsv` licenses flatly (`secreted=False` on all 23), and it is the same firm half the rows and RESULTS.md now use. The substantive point of the boundary — that no partner has been demonstrated in a physiologically relevant assay — is unchanged.

I grepped `intracellular` across the YAML, the notes and RESULTS.md to confirm nothing else survives. The three remaining hits are all legitimate: a verbatim `supporting_text` from PMID:18809223, the GO:0019915 term sense ("propagates intracellular lipid storage to eleven eutherian orthologues"), and a term-class reference in a suggested question ("intracellular lipid-transfer proteins"). No blanket claim remains.

`just validate` ✓ Valid with no warnings; `checkquotes.py` 148/148; reconciliation unchanged at 41 GOA lines ↔ 42 YAML entries, 22 references, 7 propagation_review blocks, actions 11/7/23/1.

Six passes, and five of the findings were one defect wearing different clothes: a claim pitched above the evidence, then only partly retracted. Worth recording for the campaign — the retraction sweep has to cover `knowledge_gaps`, `suggested_questions` and the `description`, not just `existing_annotations` and the notes.

cmungall:
Both fixed in `ef3a3b509d`, and both were worth checking rather than waving through.

**#1 — the affinage self-description.** Correct, and it was self-contradictory: the `reference_review` said "no claim in this review rests on it" while the GO:0006629 row lists the file in `additional_reference_ids` and quotes it in `supported_by`, which is the schema's evidence slot. It now says what is actually the case — cited once, in that row, where its independent arrival at the same conclusion corroborates a `KEEP_AS_NON_CORE` that already stands on the primary literature; no mechanistic or grounding claim rests on it, and none of its GO ids was imported.

**#2 — "intracellular lipid storage".** You flagged it as worth checking rather than asserting, so I checked the definition rather than the label, which is the rule that applies here. QuickGO: GO:0019915 is "The accumulation and maintenance **in cells or tissues** of lipids…" — no intracellular restriction anywhere in the definition. So the word was mine, not the term's, and it was doing rhetorical work the term does not license. The question now says the IBD propagates "lipid storage" and puts the real basis of the objection where it belongs: cytoplasmic lipid droplets in one transfected Huh-7 line — which is exactly what the row's own `reason` and `propagation_review` already argue, so the objection loses nothing.

`just validate` ✓ Valid with no warnings; `checkquotes.py` 148/148; reconciliation unchanged at 41 GOA lines ↔ 42 YAML entries, 22 references, 7 propagation_review blocks, actions 11/7/23/1.

Both findings were the same shape as the previous rounds' — a sentence claiming slightly more than what was established — and #2 in particular is the "decide from the definition, never the label" rule catching me using a label-flavoured word the definition does not carry. Good catch.

Issue #2789, opened 2026-08-30
linkml-reference-validator: expose a public API for materializing a FullTextLocation (download/sniff/extract) without persisting


dragon-ai-agent:
*Filed here for transfer to `linkml-reference-validator` (LRV).*

## Use case

ai-gene-review's publication-cache warm sweep (`src/ai_gene_review/etl/publication_warm.py`, added in 85db6a9ad) drives LRV ≥ 0.2.1's full-text provider chain (pmc → epmc_preprint → unpaywall → openalex) over a cache that is **not** in LRV's native cache format — the repo's own `publications/PMID_*.md` layout (`pmid`/`full_text_available` frontmatter, `## Abstract` / `## Full Text` sections). monarch-initiative/dismech's `warm_reference_cache.py` has the easy case: its cache *is* LRV-format, so it can just call `ReferenceFetcher.fetch()` and let LRV persist. A downstream consumer with its own cache format can't use `fetch()`/`apply_full_text_location()` because those persist in LRV's format (and here would overwrite the repo-format files in `cache_dir`).

## What's public today vs. what isn't

The locate half of the chain is cleanly public:

- `FullTextProviderRegistry.get(name)` / `provider.locate(ids, config)`
- `ReferenceIdentifiers`, `FullTextLocation`
- `ReferenceFetcher.locate_full_text(content, provider_name)`

But turning a `FullTextLocation` into text — download, byte-sniffing (`sniff_format`), format resolution, extractor dispatch, and the crucial transient-vs-durable error classification — lives only in the private `ReferenceFetcher._materialize(location) -> (text, fmt, pdf_bytes, transient_error)`. The warm sweep currently calls it directly (see the comment at `publication_warm.py` in `warm_publication()`), which is a version-pin hazard: any LRV refactor of that internal silently breaks downstream.

The `transient_error` flag matters as much as the text: it's what makes dismech-style durable `full_text_attempted` tagging correct (clean miss → tag durably; transient download/extract failure → leave retryable). Reimplementing the routine downstream would drift from LRV's sniffing/extraction fixes; that's why we call the private one.

## Request

Promote the materialize step to public API in LRV, e.g. either:

1. a public `ReferenceFetcher.materialize_location(location) -> MaterializedFullText` (text, format, pdf_bytes, transient_error), or
[...]

PR #3048, opened 2026-09-15
Update LTC/LRV to latest and add LRV-driven publication full-text warm sweep

dragon-ai-agent:
## Summary

Updates the two validators to latest and adds a new **full-text warm sweep** for the publications cache, modeled on monarch-initiative/dismech's `warm-reference-cache` workflow:

- **Dependency updates**: `linkml-term-validator` 0.4.4 → 0.4.5; `linkml-reference-validator` 0.2.1rc1 → 0.2.1 (stable, replacing the rc pin).
- **New `ai-gene-review warm-publications` command** (+ `just warm-publications` / `warm-publications-preview` targets and `src/ai_gene_review/etl/publication_warm.py`): drives LRV 0.2.1's full-text provider chain (PMC → Europe PMC preprints → Unpaywall → OpenAlex) over `publications/PMID_*.md`. Unlike `refresh-publications` (PMC-id-gated), this also covers DOI-only records, which were previously never retried.
- **dismech-style durable tagging**: every cleanly concluded attempt is recorded as `full_text_attempted: true` (never re-queried); successes gain provenance tags (`full_text_provider`, `full_text_extraction_method`, `oa_status`, `license`, `full_text_url`). Transient errors are left untagged so later runs retry. Bounded `--limit` sweeps are therefore idempotent and resumable. Only open-access locations are merged, mirroring LRV's own policy.
- **First sweep results (included here)**: 305 of the 18,301-record backlog attempted — **42 upgraded to full text** (41 via OpenAlex, 1 via PMC direct), 257 durably tagged as having no open full text, 6 left retryable.
- **Docs**: new `docs/publication_fulltext_warming.md` (in mkdocs nav).
- The sweep currently uses LRV's private `_materialize` download/extract routine; a public-API request is tracked in #2789 (for transfer to linkml-reference-validator).

## Validation

- New unit tests `tests/test_publication_warm.py` (8 tests; candidate selection, tagging, rewriting, provider-chain behavior via LRV's public registry hook), no network required.
- Full non-integration suite green with the updated validators (1832 passed).
- `just validate ECOLI CpxP` passes with the new LTC/LRV versions.
- ruff + mypy clean on the new module.

## Test plan

- [x] Unit tests for warm sweep pass
- [x] Full non-integration test suite passes with updated validators
- [x] Live 300-record sweep completed; upgraded files carry valid frontmatter and `## Full Text` sections

https://claude.ai/code/session_01RTcRTT7K2QeanY8rTe36A1


dragon-ai-agent:
All review findings addressed in ece3f85…head.

**🔴 1 + 2 (false full texts, `html_abstract_only` overwrites):** confirmed and fixed at the mechanism. New `is_usable_full_text` guard rejects paywall/preview and PMC-stub marker text, and any text adding < `MIN_FULL_TEXT_CHARS` beyond the cached abstract — using an alphanumeric-only comparison, since HTML scrapes drop spacing around italics (that formatting drift is exactly why PMID_10193316 slipped past a whitespace-normalized check). Rejected text is a clean miss, never an upgrade. All 13 bad records (your 12 + the bronze Springer one) were reverted to their pre-sweep state — restoring the 4 `html_abstract_only` markers — and re-attempted with the fixed chain: every one now honestly carries `full_text_attempted: true` with `full_text_available: false`.

**🟡 3–6:**
- Per-record error containment: a failing record now counts as `transient_error` instead of aborting the bounded sweep (fault-injection test included).
- `_rewrite` no longer touches untouched bodies. The whole first sweep was repaired: all 299 files had their bodies restored verbatim from the pre-sweep base and the tags re-applied through the fixed writer, eliminating the ~5,300 lines of whitespace churn — untouched abstracts are now byte-identical to `main`, so the `supporting_text` risk is gone by construction.
- `--retry-attempted` flag added for re-targeting durably attempted records after chain/guard improvements (it is how the 13 reverts were re-attempted).
- Bronze OA is now gated: no explicit license → clean miss; any reported license is persisted. The unlicensed Springer PDF text (PMID_10447589) is reverted. Result: all 29 surviving upgrades are `oa_status: green` PMC-hosted text.

**🟡 7 (PR description):** was already rewritten before your review posted (race) — current body describes the code + data change; counts updated for this round (29 upgrades kept).

**🔵 8–12:** PMC URLs normalized to the canonical `PMC`-prefixed form on all kept records and in `apply_full_text` (spot-checked resolving, HTTP 200 → `pmc.ncbi.nlm.nih.gov`); frontmatter parsing is now line-boundary; tests added for `transient_error`, the `MIN_FULL_TEXT_CHARS` rejection, the `access_type` filter, the bronze gate, and both stub-rejection paths (22 tests total); `license` parameter renamed; dry-run output fixed.

Also re-sorted `cache/go/terms.csv` (dirtied by a `just validate` run in this session, caught by `test_cache_sorted`). Full publication-related test suites pass locally; `test (3.12)` will confirm on this head.

---

dragon-ai-agent:
**`test (3.12)` failure root-caused and fixed.** The failing check on the previous head was `just validate-all`: one invalid file, `genes/human/LONP1/LONP1-ai-review.yaml` (2 errors) — not this PR's change (the file is byte-identical to `main`, so the base is red on it too). The 2025-10-27 GOA refresh dropped the WITH/FROM value on LONP1's `GO:0004252` EXP `PMID:14739292` row, so the review's stale `supporting_entities: [UniProtKB:P36776]` made the annotation matcher report the same annotation as simultaneously "missing from GOA" and "not in GOA". Fix ported into this PR (drop the stale supporting entity to match the current GOA row); `just validate human LONP1` now passes locally. CI is re-running on the new head.

---

dragon-ai-agent:
Re-review addressed in the latest push.

**A (blocking):** `linkml-reference-validator` is now bounded `>=0.2.1,<0.3` with a comment pointing at #2789; lock refreshed.

**Non-blocking suggestions carried:**
- **B:** `_materialize` is now wrapped per provider, symmetric with the `locate` guard — a failed download counts as a transient error for that provider and the chain continues to the next one.
- **C:** docs now state the sweep targets `PMID_*.md` records only; `DOI_*.md`-keyed records are not candidates.
- **D:** the abstract-echo check switched from a single midpoint probe to adaptive overlapping shingles, so a short abstract whose cached copy starts with a PubMed citation header is still caught (new test). Regression-checked against all 29 kept upgrades: none newly rejected.
- **H:** backfilled the canonical `PMC` prefix on the 8 pre-existing bare-numeric `full_text_url` values (`PMID_10339610` included).

**Left as-is, with reasons:**
- **E (license on green records):** OpenAlex returned no license string for these PMC locations; writing an explicit `license: null` on every record adds frontmatter noise without information. Any license a provider does report is persisted — happy to add the explicit null if you prefer auditable absence.
- **F (openalex HTML beating pmc XML):** agreed worth investigating, but it's a fetch-quality improvement, not a correctness issue — the 29 texts passed the guards. Note the suggested `--retry-attempted --providers pmc` wouldn't re-target them: upgraded records aren't candidates (they have full text). Filed mentally as follow-up work alongside #2789.
- **G (monkeypatched sweep test):** a provider raising from `locate` doesn't exercise the sweep-level containment — `warm_publication` catches provider errors itself and returns `transient_error` without raising. The monkeypatch is the only seam that makes `warm_publication` itself raise, which is exactly the failure mode the sweep guard exists for; it's fault injection at the boundary under test, not a mock of the behavior being verified.

Separately, the `test (3.12)` failure on the earlier head was root-caused as a base-branch issue (LONP1 GOA drift after the 2025-10-27 GOA refresh) and fixed in ef25460 — see the previous comment. CI is running on the current head.

---

PR #3049, opened 2026-09-15
Add generic Embden-Meyerhof-Parnas glycolysis module

cmungall:
## Summary

- Creates `modules/emp_glycolysis.yaml`, a taxon-neutral template for Embden-Meyerhof-Parnas (EMP) glycolysis grounded in GO:0061615
- Defines nine ordered reaction steps from hexose entry through pyruvate production
- Models three axes of cross-taxon variation as explicit variant sets: hexose entry mechanism (ATP kinase, ADP kinase, PTS, or glucan-derived), phosphofructokinase phosphoryl donor (ATP, diphosphate, or ADP), and glyceraldehyde-3-phosphate oxidation mode (phosphorylating vs. non-phosphorylating)
- Provides the generic EMP counterpart to the existing taxon-neutral gluconeogenesis module and complements the concrete human glycolysis investment/payoff phase modules
- Includes comprehensive evidence grounding in GO terms, MetaCyc pathways, and related modules

## Validation

N/A — This is a new module creation, not a gene review. The YAML structure follows the existing module schema and is consistent with related modules (gluconeogenesis.yaml, entner_doudoroff_and_gluconeogenesis.yaml).

## Test plan

- [ ] Validate module structure with schema validation tools
- [ ] Verify cross-references to GO terms (GO:0061615, GO:0061621, GO:0061633, GO:0061622, etc.) are accurate
- [ ] Confirm variant sets and annotons follow the same pattern as existing modules
- [ ] Review evidence statements for accuracy and completeness

https://claude.ai/code/session_01FcdRwtVons6Q1NixiJ51Da


cmungall:
All nine findings addressed in c7a55027b. Each was checked against a primary source before acting; all nine were accepted, though **#4 is resolved differently from what was proposed** — details below.

**Blocking**

**1. PFK `EXACTLY_ONE` → `ONE_OR_MORE`.** Correct, and the inconsistency you point to was the real tell: the same argument was already applied to the other two axes. Notes now name the plant ATP-PFK/PPi-PFP co-expression case.

**2. Hexokinase + glucokinase as conjunctive annotons.** Confirmed against the local GO build that `GO:0004340` is a **direct child** of `GO:0004396`, so subsumption, not exclusion. Restructured as a nested `variant_set` on the specificity axis with `ONE_OR_MORE` (vertebrates carry both), which is the idiom used elsewhere in the file.

**3. F6P commitment claim.** Withdrawn. F6P is shared — non-oxidative PPP output, phosphoketolase intermediate, hexosamine precursor — and step 3 already named PFK as committed. GPI is now described as *routing* carbon into EMP rather than committing it, which is also what `GO:0061615` actually names.

**Non-blocking**

**4. Fructose-1-phosphate route — scoped out rather than added.** Checking the hierarchy before adding the variant changed the answer: `GO:0061625 glycolytic process through fructose-1-phosphate` is a **sibling** of `GO:0061615` under `GO:0006096`, not a child. GO therefore places the FruK route *outside* the F6P pathway this module models, so adding it as a fourth entry variant would have contradicted the module's own core term. It is now explicitly scoped out in `notes` with `GO:0061625` recorded as evidence, cross-referencing `bacterial_fructose_pts_catabolism.yaml` and `fructose_pts_uptake_and_catabolism.yaml`. Your underlying point stood — the node description did quietly promise coverage it did not deliver — so thank you for it; only the remedy differs. The retained glucose PTS variant is unaffected: it yields G6P and rejoins at the isomerase.

**5.** Both slips fixed — four payoff ATP per hexose, and seven of ten *reactions*.
**6.** Edge retargeted to `pts_entry_variant`; the top-level graph is now acyclic unless PTS entry is selected.
**7.** `EC:1.2.7.6` added, citing the `NCBIFAM:NF040818` equivalog mapping you located.
**8.** `plasma membrane` on `pts_activity`; `cytosol` on both ADP-dependent archaeal annotons.
**9.** `GO:0061621` added as a concept; the three entry variants are now symmetric.

**On the validation checkbox you left unconfirmed:** the four labels you could not check locally (`GO:0061622`, `GO:0061633`, `GO:0043844`, `GO:0043797`) are label-checked by `module_validator`, which reports **0 warnings**. Full chain after the edits — `linkml-validate -C ModuleReview` clean, `module_validator` term labels OK, `collect_module_qc` 100% recommended-field compliance with no leaf-grounding, conformance or chaining findings, 163 module tests passing, page re-rendered.

Two things remain open and are recorded in the module's `notes` as prerequisites for leaving `DRAFT`: there is no module deep-research report, and the template has not been checked for satisfiability against a bacterial or archaeal genome — the archaeal ADP-kinase and ferredoxin-oxidation variants are the weakest claims in the file.

---

cmungall:
Both residual nits cleared in f880d3bf8.

**A — "defining commitment".** Fixed; my own inconsistency. Withdrawing the step-2 claim left the same word in the two most prominent places in the file. Now "defining feature" / "defining criterion", and the concept description states the reservation explicitly. Every remaining use of *committed*/*commitment* in the file refers to the PFK step.

**B — this one was a modelling gap, not just wording, and you were right to separate it from #4.** Fructose via broad-specificity hexokinase and mannose via mannose-6-phosphate isomerase both land on F6P, so unlike the F1P route they are squarely *inside* `GO:0061615` — the F1P scope-out did not cover them and could not. Reworded the node description to describe only what the four variants do, and added a scope paragraph recording direct F6P entry as a **known gap for a future revision**, explicitly distinguished from the F1P exclusion that follows it. I did not add the variant in this PR: it needs GO terms I would have to verify for fructokinase and MPI, and widening an already-approved PR to do it seemed worse than recording the gap honestly. Happy to add it here instead if you'd prefer.

**On your parentage caveat — a fair catch, and now recorded.** You're right that `module_validator` checks term *labels*, not parentage, so the `GO:0061625` sibling claim is not covered by CI either way. The evidence statement now names what makes it auditable: the asserted superclasses of `GO:0061625` are `GO:0006001` fructose catabolic process and `GO:0006096` glycolytic process, with `GO:0061615` absent. It rests on that check against the GO release in use, and the statement now says so rather than asserting the relationship bare.

Re-validated after the edits: `linkml-validate -C ModuleReview` clean, `module_validator` 0 warnings, `collect_module_qc` 100% recommended-field compliance with no leaf-grounding, conformance or chaining findings, 163 module tests passing, `git diff --check` clean, page re-rendered.

---

cmungall:
Fixed in 9f481d3f8 — worth spinning the render for, since as you note the `role` renders ahead of the description that contradicts it.

`role: aldose-ketose isomerisation routing carbon into the EMP route`.

**Why my previous sweep missed it:** I grepped for `committed` and `commitment`. The survivor was the participle `committing`, which matches neither — the one morphological form I didn't cover, in the one field that renders most prominently. Re-swept on the bare stem `commit` this time: eight remaining hits, all the PFK step. I agree on `:215` too — "Commits hexose carbon to intracellular metabolism" is the hexose-trapping sense, and its next sentence already disclaims EMP-exclusivity for G6P, so it stays.

Verified the render rather than assuming it: the new phrasing appears in both rendered positions (edge label and part header) and the old phrasing appears zero times.

`linkml-validate -C ModuleReview` clean, `module_validator` 0 warnings, 113 module tests passing, `git diff --check` clean.

That closes every item from all three rounds. The two items in the module's `notes` remain open by design and are the prerequisites for leaving `DRAFT`: no deep-research report, and no genome satisfiability check for the archaeal ADP-kinase and ferredoxin-oxidation variants. The direct-F6P entry variant is recorded there as a known gap for a future revision.

---

cmungall:
Fixed in a00d3546a. You read it correctly: 113 was a narrower three-file selection I ran during that edit, not a coverage drop — but as committed provenance the lower number reads as fifty tests having disappeared, which is exactly the wrong thing for a history record to imply.

Re-ran the full five-file suite (`test_module_qc`, `test_module_validator`, `test_module_conformance`, `test_render_modules`, `test_module_logic`) on the current head: **163 pass**. The clause now names the selector and the count, and states plainly what the earlier number was and why it differed, rather than quietly overwriting it — the point of the record is that someone can audit it later, so silently swapping 113 for 163 would have been worse than the original imprecision.

**On the renderer observation — thank you, and I'm deliberately not fixing it here.** You're right that `concepts[].description` and `context[].cellular_components[].description` are dropped by the module renderer, so the reservation sentence and the cytosol rationale are invisible on the page. That's a renderer gap affecting *every* module in the repo, not something about this file, and fixing `render_modules.py` inside a module-content PR that has already been approved four times over is exactly the widening I declined on the direct-F6P variant. Flagging it to the repo owner as a separate issue instead.

That closes every item across four rounds. Still open by design, recorded in the module's `notes` as prerequisites for leaving `DRAFT`: no deep-research report, no genome satisfiability check for the archaeal ADP-kinase and ferredoxin-oxidation variants, and direct-F6P entry (fructose via hexokinase, mannose via MPI) unmodelled.

---

cmungall:
Agreed on the residual imprecision, and taking your advice not to spend a commit on it — but it shouldn't go unanswered either, so putting the answer in the thread instead.

You're right about the distinction: lines 133 and 181 record the *number* 163, not the selector, so "the other two events cite the same full 163-test selection" is something I know rather than something the record establishes. It is true — I ran the same five-file suite for both of those events — but that's my attestation as the author, not a derivation a later reader can make from the file. Your static collection arithmetic (136 functions across the five files, all parametrised, expanding to 163; and 110 across the three-file subset landing at 113) independently corroborates it, which is a better check than either of us asserting it.

Since the sentence is accurate and the clarification now lives in the thread alongside the record, I'd rather not spin another render and CI cycle to reword it. Happy to tighten it if @cmungall would prefer the record fully self-contained.

On the append-only precedent: your reading matches mine. A fifth event whose only content was "the previous event's test count was imprecise" would have put indirection in front of a fact the corrected clause now states directly, and `sections` would have had nothing honest in it. The correction is self-documenting precisely so the edit doesn't destroy what a reader would want to audit.

---

PR #3059, opened 2026-09-17
Review human ARHGAP21: a contested GAP substrate that GO cannot record, and seven uncurated papers

cmungall:
Complete GO annotation review of human **ARHGAP21** (`Q5T5U3`, 1958 aa).

**Row reconciliation:** the GOA TSV has **47 data rows**; `existing_annotations`
has **51 = 47 + 4 NEW**. The `fetch-gene` stub did *not* collapse rows here
(checked by multiset comparison on `(term, evidence, reference, with/from)`, not
by eye). Actions: 19 ACCEPT, 3 KEEP_AS_NON_CORE, 8 MODIFY, 17 REMOVE, 4 NEW.

---

## 1. GO cannot record which GTPase a GAP acts on

The task flagged ARHGAP21's substrate specificity as contested and asked that the
disagreement not be collapsed. It is worse than contested — it is **inexpressible**.

Resolving each historical substrate-specific GAP term through QuickGO's
`/complete` endpoint (the only endpoint that distinguishes a **merged** id from
an **absent** one; OLS reports both identically):

| requested | resolves to |
|---|---|
| `GO:0005099` Ras GTPase activator activity | **MERGED** → `GO:0005096` |
| `GO:0005100` Rho GTPase activator activity | **MERGED** → `GO:0005096` |
| `GO:0008060` ARF GTPase activator activity | **MERGED** → `GO:0005096` |
| `GO:0005097` Rab GTPase activator activity | **MERGED** → `GO:0005096` |

and `GO:0005096` has **zero `is_a` children**. So it is already maximal:
`ACCEPT`, no replacement proposed, and **no `proposed_new_terms` entry** to
re-create what GO deliberately merged. Substrate identity is recorded
machine-readably instead — `core_functions[].substrates` plus `has_input`
(`RO:0002233`) extensions.

**What the papers actually say** (recorded in the review, not adjudicated):

| paper | substrate(s) |
|---|---|
| Dubois 2005 `PMID:15793564` | Cdc42, "preferentially" |
| Sousa 2005 `PMID:16184169` | **RhoA and Cdc42** |
| Anthony 2011 `PMID:21173159` | **RhoA** |
| Lazarini 2013 `PMID:23200924` | **RhoA and RhoC** (PC3, not LNCaP) |
| Hehnly 2010, Barcellos 2013 | Cdc42 |

The dispute is narrower than "RhoA vs Cdc42": both 2005 papers agree on Cdc42
activity and differ on whether RhoA is *also* a substrate. And the catalysis
itself is not in dispute — **both** groups independently built the R1184A
arginine-finger mutant and both saw loss of function (UniProt carries
`ECO:0000269` from both PMIDs on that one feature).

## 2. Coverage, not over-annotation, is this gene's real defect

QuickGO queried **by reference**. None of the ten primary-paper result sets was
paginated, so these counts are complete:

| PMID | annotations in all GOA | on ARHGAP21 |
|---|---|---|
| 15793564 Dubois 2005 | 1 | `GO:0005096` IDA |
| **16184169 Sousa 2005** | **2** | **0** (both on CTNNA1) |
| 16527809 Klein 2006 | 0 | 0 |
| 17347647 Ménétrey 2007 | 2 | `GO:0005515` IPI |
| 20525016 Hehnly 2010 | 12 | 5 rows |
| 21173159, 22922005, 23200924, 23235160, 29212046 | 0 | 0 |

**Seven of ten primary papers produced no GO annotation anywhere.** Every paper
reporting RhoA or RhoC is in that set — so GO is silent on that half of the
literature for a reason *independent* of the ontology gap. A 2.1 Å crystal
structure of the ARF1–ARHGAP21 complex is recorded as bare `protein binding`.

The largest single gap is `PMID:16184169`, a *Nature Cell Biology* paper titled
for ARHGAP21's requirement in α-catenin recruitment, which produced two
annotations — **both on α-catenin, none on ARHGAP21** — despite UniProt carrying
the interaction in FUNCTION, in SUBUNIT, and as a mapped region (1592–1861).
Hence the three NEW rows:

- `GO:0045294` **alpha-catenin binding** (IPI) — term verified active, defined
  *"Binding to catenin complex alpha subunit."*
- `GO:0034333` **adherens junction assembly** (IMP) — two independent knockdowns;
  loss and gain of function move the phenotype opposite ways. **Species caveat
  recorded on the row:** both supporting papers are abstract-only, neither names a
  cell line, and `PMID:16184169` is absent from PMC/Europe PMC full text with no
  text-mined annotations, so the cell system could not be confirmed. IMP is kept
  (human α-catenin in the Y2H, human target) but flagged for a curator to confirm.
- `GO:0032956` **regulation of actin cytoskeleton organization** (IMP) — the
  founding paper's title finding had no BP term on the gene at all
- `GO:0031005` **filamin binding** (IPI, added in review round 2) — see §4

## 3. ARHGAP21 is an ARF **effector**, not an ArfGAP

The task brief described an "ArfGAP-like region"; the review corrects this. The
930–1097 region is an ARF-**binding** domain (PH domain + C-terminal helix), and
`PMID:16527809` classifies ARHGAP21 as an effector while explicitly contrasting
it with the two Arf GAPs assayed in the same experiment. There is no ArfGAP
domain and no ARF-GAP annotation — GOA has correctly avoided the
name-implied-catalysis error here.

## 4. The PDZ rows: a second ontology gap

Ten of the 24 `GO:0005515` rows come from one PDZ/PBM affinity survey
(`PMID:36115835`). Its full text **never names ARHGAP21**, so rather than assume,
the claim was checked against IntAct participant ranges: all ten map to residues
**50–159**, the PDZ domain. But GO's only PDZ term, `GO:0030165`, is defined
*"Binding to a PDZ domain of a protein"* — the **ligand** side — and has no
descendants. There is no term for ARHGAP21's side, so those rows are `REMOVE`d as
uninformative with the gap recorded.

**Round-2 correction.** The first draft also tagged this gap `MF_DARK` and called the
PDZ "the one module with no assigned cellular role". That was refuted by
`PMID:41957357` — a file this same PR adds — which reports the PDZ domain mediates
binding to filamin A. The darkness framing is gone; the *ontology* gap survives, since
finding a ligand does not create a term for the PDZ-domain side. A fourth NEW row,
`GO:0031005` **filamin binding**, is added — and the fact that GO can express this one
interaction *only* because a partner-specific term happens to exist is itself recorded
as evidence the gap needs a general fix.

By contrast **14-3-3 binding is genuinely replicated** — 4 distinct publications,
3 orthogonal methods in IntAct — so those 6 rows are `MODIFY` → `GO:0071889`
rather than removed. Per-partner judgement, not per-gene.

## 5. Negative results (recorded because they were checked)

- **Symbol collision: clean.** Both 2005 papers call the protein *ARHGAP10*,
  which is this gene's former symbol and now belongs to a **different** gene
  (`A1A4S6`, absent from PANTHER `PTHR23175`). No annotation from any primary
  paper landed on `A1A4S6`. For the seven paginated screens the check is reported
  as *not possible*, not as clean.
- **The one `Xeno` WITH/FROM is harmless.** ARF1 is recorded as mouse `P84078`;
  fetching both FASTAs shows it is **sequence-identical** to human `P84077`
  (181 aa). Bookkeeping, not a biological defect.
- **No retractions** among 30 cited PMIDs. One erratum (`PMID:36115835` →
  `PMID:36477203`) is **cosmetic** — missing figure labels and legends, no data
  changed. Found by reading `CommentsCorrections/RefType` off each cited record,
  which is the only way a Publisher Correction is discoverable.
- **All 7 IBA rows are self-referential** (`PANTHER:PTN…|UniProtKB:Q5T5U3`) —
  the expected marker of a PAINT curator judging the function core.
  `NO_FAILURE_CORE`, never `CIRCULAR_OR_REDUNDANT`.
- **ARHGAP23** (`Q9P227`, SF5) is the genuine close paralog; ARHGAP21 is SF16.
  `PTHR23175` has 5182 proteins of which **7 are reviewed** — the entries CSV is
  the Swiss-Prot subset, not the family.

## 6. On affinage

`gates_passed: true`, 15 citations, no bioRxiv-DOIs-in-PMID-fields, no
retractions. But it returned **neither founding 2005 paper nor `PMID:20525016`**
— i.e. it missed *every* paper that carries a GOA row on this gene, plus the
α-catenin paper driving the largest gap. Consistent with the gates certifying
precision and saying nothing about recall. It is cited exactly once, at
`core_functions` level, for a corpus-level statement; no mechanistic claim rests
on it.

## Committed analyses

Six rerunnable scripts under `genes/human/ARHGAP21/ARHGAP21-bioinformatics/`
(`resolve_partners`, `check_corrections`, `check_gap_terms`,
`reference_coverage`, `intact_methods`, `audit_review`) with `RESULTS.md` and
committed JSON outputs. All derive the repo root rather than hardcoding a
worktree path.

`audit_review.py --self-test` reconciles the review against the GOA TSV, rejects
duplicate YAML keys via a strict loader (PyYAML silently discards the earlier of
a repeated key, and every other gate walks the *parsed* document), and verifies
all 48 `supporting_text` quotes — including `file:` ones, which CI does not
check. Its self-test asserts each mutation anchor exists before mutating.

### Two defects I found in my own work

- `intact_methods.py` first parsed IntAct ids as `db:ACC` and reported **0
  partners from 502 rows** — every GOA partner "ABSENT", a silent zero that reads
  as a finding. The real format is `ACC (db)`. The script now aborts rather than
  reporting if that recurs.
- Two rows initially carried a **verbatim, correctly-attributed** quote from
  `PMID:15161933` describing *that* paper's 14-3-3 affinity-column method, sitting
  under summaries about BioPlex AP-MS. Every mechanical gate passed — no gate
  checks a quote against the claim it is attached to. Fixed by quoting each row's
  own paper.
- `core_functions[1]` listed **ARF1/ARF6 under `substrates`** while its MF is
  `GO:0031267 small GTPase binding`. For a GAP, "substrate" means the GTPase whose
  hydrolysis it accelerates — so that entry asserted ARHGAP21 *is* an ArfGAP, the
  exact claim §3 refutes. Removed; the ARF identity survives as the `RO:0002233`
  extension on the annotation row, and the fix asserts that extension exists before
  deleting the list.

## Unresolved

Rank order of GAP preference (no study has assayed Cdc42/RhoA/RhoC side by side
with one construct and one assay); whether the RhoA/RhoC activity is
context-restricted; and which PDZ ligand ARHGAP21 engages in cells. All recorded
in `knowledge_gaps` / `suggested_questions` with experiments proposed.


cmungall:
All six addressed in `abad3a4`. Every item was verified against the cached files before acting — two came back differently than stated, noted at the end. No term or action changed, and nothing you validated was touched.

### 1 + 2 — `GO:0032956`

**Confirmed, and worse than reported.** `grep` for `stress fib`, `spreading` and `spread` in `publications/PMID_15793564.md` returns **nothing**. The claim is UniProt's `MISCELLANEOUS` line — and that line spans a `CC` continuation (lines 274–275), so it **cannot be quoted verbatim on one physical line at all**. Re-sourcing was not available; the clause is removed.

The row now records something more useful than the clause did: the **sign of the effect is not fixed**. UniProt says depletion *accumulates* stress fibres; `PMID:25744409` says antisense knockdown *reduces* F-actin ("F-actin was reduced in AS-islets, as judged by lower phalloidin intensity") in neonatal mouse islets. That is exactly why the sign-agnostic `GO:0032956` is right and a directional term would not be. The mouse paper is flagged in its `reference_review` as context only, not as evidence for a human code.

**On the copy-paste: you're right, and it was self-inflicted.** The offending caveat was something I *added* in round 1 to improve honesty about species, then pasted onto a row it doesn't describe. Replaced with a row-appropriate caveat about `PMID:15793564` being abstract-only.

### 3 — `GO:0051684`

Confirmed; your sentence is verbatim in the cached file and is now the row's only quote. The `reason` now states the distinction explicitly, because it is the whole point of the row: **nocodazole washout measures re-establishment after dispersal**, which is `GO:0051683`. Maintenance is the steady-state observation that the Golgi is *already* partly dispersed in unperturbed knockdown cells.

### 4 — `GO:0005794` IDA

Confirmed; swapped for your sentence (verbatim). Generalising the tell, since it is cheap: **a sentence carrying reference callouts is background, not result**, and cannot support an IDA. Worth grepping for before attaching any quote to an experimental row.

### 5 — the PDZ gap

**The best catch, and I had the refuting file in my own commit.** `PMID:41957357` reports "ARHGAP21 directly binds to FLNA, and the PDZ domain of ARHGAP21 functions as a potential mediator of its binding to the 1-1200 aa fragment of FLNA". I read that paper during the retraction sweep and never asked what it said about the domains.

Split into the part that was wrong and the part that survives:

- **Wrong:** `dark_aspect: MF_DARK` and "the one module with no assigned cellular role". Both removed, from the gap and from the suggested question.
- **Survives untouched:** the ontology gap. Finding a ligand does not create a term for the PDZ-domain side; `GO:0030165` is still the ligand side only.

Added a fourth NEW row, **`GO:0031005` filamin binding** (IPI, human HCC lines). Deliberately scoped to the **binding**, which the paper states outright — not to PDZ-dependence, which it hedges as a "*potential* mediator". The metastasis and FLNA-stabilisation phenotypes are not proposed as process terms: cancer-context effects, not a demonstrated normal function.

That this interaction *is* expressible sharpens rather than weakens the gap, and the `significance` field now says so: GO can record ARHGAP21's PDZ interactions **only where a partner-specific term happens to exist** — a ligand-by-ligand workaround that does not scale to a 65,000-interaction survey.

### 6 — references

All added, with `reference_review` on each. The fix script now asserts **every `PMID:` string anywhere in the document resolves to a reference entry**, so this cannot silently regress.

### Two places the list didn't hold up

- **You missed one.** `PMID:36477203` — the Gogl erratum — was named in `PMID:36115835`'s review notes with no reference entry. Added.
- **Two of the five weren't applicable as stated.** `PMID:41957357` and `PMID:38556137` were not named anywhere in the YAML at the time, so they were not omissions from `references`. `41957357` is now cited (item 5) and referenced; `38556137` (C. elegans PAC-1/afadin) is still not cited in the review and so is still not referenced.

### State

`just validate` ✓ Valid. `audit_review.py --self-test` passes: **47 GOA rows → 47 non-NEW + 4 NEW = 51**, **50 quotes verbatim**, no duplicate YAML keys. `cache_lint` exit 0, no `terms.csv` deletions or dups. History `EDIT` record added and validated.

Worth recording that **round 2 was four more instances of the same class the PR body already flagged**: real PMIDs, verbatim strings, correct terms, correct actions — quotes attached to the wrong claim. That is now six on this gene, which is decent evidence the gap is structural rather than incidental. Two heuristics came out of it and are in `RESULTS.md`: *reference callouts mean background, not result*, and *check the assay in the quote is the assay the term names*.

cmungall:
Fixed in `58fdebb` — by writing the guard, not softening the prose.

### 🟡 The invariant was asserted, not enforced. Correct, and it's the sharper version of my own thesis.

You're right on every detail. The assertion lived in `fix6_arhgap21.py`, an **uncommitted one-shot scratchpad script that never runs again**. Nothing in the repository enforced it; the invariant held by hand, on that one day. And the claim was *about a guard*, which makes it the same failure class one level up — with `audit_review.py`'s own docstring having already named the case: *"it cannot tell you which guard was never written."*

`check_references` is now the **fifth committed check**, with a `--self-test` mutation that deletes a reference whose PMID is cited in content and requires the check to fire. Docstring enumerates five. I also corrected the false sentence in the round-2 history record rather than leaving it standing.

### Writing it went wrong twice, both caught by the self-test rather than by reading

**1. The detector's scope included the lookup table it validates against.** First version scanned the whole serialised document *including* `references`. Deleting a reference entry removed the only occurrence of its PMID, so `named` shrank in step with `ref_ids` and the guard stayed silent:

```
SELF-TEST FAIL: dropping reference PMID:12056806 was not detected
```

A check partly satisfied by its own subject is worse than no check, because it reports success.

**2. The obvious correction over-corrected.** Excluding the whole `references` block fixed that — and lost `PMID:36477203`, the Gogl erratum, which is cited **only inside `PMID:36115835`'s `review_notes`**. That is exactly the citation your round-1 list had missed, i.e. the single case the check most needed to cover.

Working scope strips **exactly the reference `id` fields** and keeps everything else. Verified directly rather than assumed:

```
PMID:36477203 is in the cited set (named in PMID:36115835 review_notes)
problems after dropping its reference entry:
  ['PMID:36477203 is cited in the document but has no references entry']
```

Cited-PMID count read **24 → 22 → 23** across the three scopes. The middle number looked entirely reasonable, which is the point.

**Also hardened per a parallel finding:** text-anchor mutations now assert the anchor matches **exactly once**. An anchor matching twice silently mutates whichever occurrence comes first and stops testing what it was written for; a zero-match anchor "passes" by changing nothing. Both are indistinguishable from a green self-test.

### 🔵 has_input on binding rows — keeping, deliberately

The two slots answer different questions, so the apparent restatement isn't redundancy: **`supporting_entities` is evidence provenance** (what the curator used, the GOA `WITH/FROM`), **`extensions` is annotation semantics** (what the statement is *about*). Kept consistent across all three instances.

### 🔵 GO:0032956 human loss-of-function — declining, with the gap made explicit instead

Your suggestion to cite UniProt's `MISCELLANEOUS` line as a `file:` reference runs into the constraint you confirmed earlier in this same review: the sentence spans a `CC` continuation (`ARHGAP21-uniprot.txt:274-275`), so the only fragment quotable on one physical line truncates at *"induces cell spreading and"* — dropping the F-actin clause that would make it relevant.

Rather than quote it misleadingly, the row now states the evidence structure outright: the human data is **gain**-of-function (`PMID:16184169`); no human **loss**-of-function actin readout is quotable from the cache; UniProt records the phenotype but is unquotable for this purpose. That seemed more useful than either a truncated quote or silence.

### The two generalisations

Agreed they are repo-level lessons, not ARHGAP21 facts — *a sentence carrying reference callouts is background, not result*, and *check that the assay in the quote is the assay the term names*. Between them they would have caught **four of the six** round-1 items pre-submission.

I have **not** promoted them out of this gene's `RESULTS.md`: doing so touches shared documentation, which is outside a single gene review's scope and not mine to change unilaterally. Flagging rather than silently expanding the PR — happy for that to be someone's call in a separate change.

### State

`just validate` ✓ Valid · `audit_review.py --self-test` PASS · **47 GOA rows → 47 non-NEW + 4 NEW = 51** · **50 quotes verbatim** · **23 cited PMIDs, all resolving** · `cache_lint` exit 0 · no `terms.csv` deletions or duplicates · round-3 `EDIT` history record added and validated.

No annotation, term, action or evidence code changed this round.

On merge-blocking threads: a GraphQL query for `reviewThreads` on this PR returns **none** — all review content has been top-level bodies, so there is nothing unresolved to block the merge.

cmungall:
Thanks — and thanks for verifying the guard independently rather than taking the response's word for it, including the mutual-citation hazard between `36115835` and `36477203`, which I hadn't checked.

**I verified all four blue items and all four are real.** I'm declining to push for them, and recording each here so the decision is auditable rather than silent.

| # | item | verified | why not pushing |
|---|---|---|---|
| 1 | stale `"mutation 5 changed nothing"` inside mutation 6 (`audit_review.py:359`) | confirmed — line 359 says `mutation 5`, the block is mutation 6 | assert *message* only; the assert itself is correct and fires correctly |
| 2 | docstring item 5 wider than the code | confirmed — docstring says "anywhere in the document", `cited_pmids` strips reference `id` fields | you noted it yourself: **no coverage difference**. The code is right; the prose is looser than the code, not the reverse |
| 3 | check covers `PMID:` only, not `GO_REF:`/`Reactome:`/`file:` | confirmed | a genuine scope limit, but those never dangled here and extending it is new capability, not a fix |
| 4 | `GO:0032956` "in epithelial cells" vs `GO:0034333` "cell system could not be confirmed" | confirmed, and your diagnosis is exactly right | see below |

On **#4**, your proposed fix is the correct one and worth stating even though I'm not applying it: the abstract *does* establish the system is epithelial (E-cadherin junctions between epithelial cells); what it does not name is the **cell line**. So `GO:0032956`'s "epithelial cells" is sound and `GO:0034333`'s broader *"cell system could not be confirmed"* is the overstatement — tightening that phrase to "cell line" resolves it, and the row's own actionable instruction already says "confirm the **cell line**", so the one loose phrase is inconsistent with its own next sentence.

**Why stop here.** None of the four changes an `action`, a `term`, an `evidence_type`, a `supporting_text`, or a `core_functions` entry — the campaign's convergence test for whether a round should have been a comment. Against that, pushing now invalidates an approval that currently post-dates the head commit by ~14 minutes, and post-approval polish rounds have needed three recovery PRs in this campaign. A curator reading the merged YAML would do the same thing with or without these four edits: `GO:0034333` already tells them to confirm the cell line from the full text.

If this PR needs another push for any reason before merge, all four are ~5 minutes of work and should ride along — #4 first, then #2, since a docstring claiming more than its code is the precise failure class this gene spent three rounds on.

**State:** `APPROVED` on `58fdebb0` (approval `02:17:48Z` post-dates the head commit at `02:03:23Z`), no live CHANGES_REQUESTED, no unresolved review threads, `claude-review` pass, `test (3.12)` still running. Locally: `just validate` ✓, self-test PASS, 47 GOA + 4 NEW = 51, 50 quotes verbatim, 23 cited PMIDs all resolving, `cache_lint` 0, no `terms.csv` deletions or duplicates.

One correction to my own round-3 reply while I'm here: I reported "23 cited PMIDs" and you report 24 named. Both are right and they measure different things — 24 distinct PMIDs appear in the file, 23 of them are *cited* under the guard's scope, and the one-PMID gap is `PMID:12056806`, referenced but not cited in content. That is the same entry your vacuity check keyed on.

PR #3060, opened 2026-09-17
Add site favicon and complete ARBA review pages

cmungall:
## Summary

- Add a shared blue DNA favicon to the homepage, collection entry pages, browser, and rendering templates.
- Replace the stale four-entry ARBA index with all 75 reviews, create the 45 missing review pages, and link each review to its original UniProt ARBA rule.
- Add offline rule rendering to the daily build. Preserve existing pages with analysis evidence when the source caches are absent, and detect newly generated untracked pages.

All 22 existing Domain Overlap Analysis Tables and five External Mappings sections are retained. Pages preserved because their caches are absent retain their prior review content until those caches are restored; the renderer reports this explicitly. Their favicon and UniProt links are included in this PR.

Condition-set cards and the index explicitly describe recorded sets rather than full-rule totals. ARBA00022777's invalid scalar `condition_sets` is replaced by an empty list; its existing description and rationale retain the full count of 1,358. An append-only history record documents this structural fix. No biological decisions are changed.

## Validation

- 86 project, gene-navigation, and module rendering tests passed.
- 19 rule-index, wrapper, and demo tests passed; all five index/regression tests passed after adding the cache-restoration case.
- Audited preservation of all 22 analysis tables and five mappings sections, all 75 local review links, and all 75 UniProt rule links.
- Rule HTML plus YAML totals 7,220,756 bytes before compression; full artifact budget remains checked by deployment CI.
- History record validates; workflow YAML and just recipes parse; `git diff --check` passes.
- Full validation of ARBA00022777 still fails on unrelated pre-existing legacy fields and enums, including `REMOVE`, `rationale`, and risk fields. The invalid scalar condition-set field is fixed; this is not a full curation migration.

## Test plan

- [x] Render the index and cached rule pages without dropping unavailable analysis evidence.
- [x] Check missing-cache preservation and refresh with available caches.
- [x] Check that invalid scalar condition sets fail explicitly.
- [x] Verify all 75 index entries and original UniProt links.
- [ ] Check the deployed site and full artifact budget after merge and the next Pages build.


cmungall:
Addressed the three requested changes:

1. Restored all **22 Domain Overlap Analysis Tables** and **five External Mappings sections** from the pre-PR pages. Offline rendering now preserves an existing evidence-bearing page if its required analysis/enrichment caches are missing, and reports the skipped page. Regression tests cover both sections and refreshing when caches return. Preserved pages retain their existing review content until caches are restored.
2. Changed rule-page cards to **Recorded condition sets**, with the same incomplete-record caveat as the index.
3. Removed the integer-coercion branch. The malformed ARBA00022777 field is now an empty collection, with its original 1,358 total retained in the description/rationale and an explanatory comment. Invalid scalar inputs now raise an explicit error. Added and validated a provenance record. Full validation still identifies unrelated pre-existing fields/enums in that file; no biological decisions were changed.

Also fixed untracked-page detection, guarded the two optional template values, and made ARBA favicon links work under `file://`. All 75 review links resolve. Rule HTML plus YAML is approximately 7.22 MB uncompressed; the deployment workflow still checks the full artifact budget.

Validation: 19 targeted tests passed, followed by all five renderer/index tests after adding the cache-refresh test; history validation and diff checks passed. The optional unrelated nested-YAML cleanup and broader template/helper refactoring are left out of this rendering fix.

PR #3072, opened 2026-09-19
Review human ARFGEF2 (BIG2): a third of its GO record is one rat paper projected twice, and the disease-gene trap is inverted

cmungall:
Full GO annotation review of human **ARFGEF2 / BIG2** (`Q9Y6D5`), the brefeldin A-inhibited ARF guanine-nucleotide exchange factor mutated in autosomal-recessive periventricular nodular heterotopia with microcephaly.

**Coverage.** The GOA TSV has **66 rows** and `existing_annotations` has **68 entries = 66 GOA + 2 NEW**. The seeded stub did not collapse anything on this gene — each of the ten `GO:0005515` rows already had its own entry with its own partner — and the count is re-checked mechanically by `audit_claims.py`.

## Actions

| action | n |
|---|---|
| ACCEPT | 27 |
| KEEP_AS_NON_CORE | 25 |
| REMOVE | 10 |
| MARK_AS_OVER_ANNOTATED | 3 |
| MODIFY | 1 |
| NEW | 2 |

All ten `REMOVE`s are bare `GO:0005515 protein binding`, decided **per partner** and per the repo's stated policy that the generic term's problem is lack of functional information. None of them asserts that an interaction is false; each `reason` names the partner, the assay, and where the information is recorded instead.

### 1. Thirty percent of the annotation set is a *double* projection of one rat paper

**20 of the 66 rows**, covering **13 GO terms**, carry rat Arfgef2 (`UniProtKB:Q7TSU1`) in WITH/FROM, and **17 of them rest solely on `PMID:15198677`**. The same rat annotations arrive twice: 11 rows via Ensembl Compara (`GO_REF:0000107`), 7 via UniProt ISS (`GO_REF:0000024`), 2 via combined automated annotation.

**The complex-projection discriminator came back NEGATIVE, and that is reported as a finding rather than skipped:** `PMID:15198677` annotates only **2 entities** in all of GOA — rat Arfgef2 and its partner Gabrb3 — so this is genuine per-gene curation at figure granularity (SynGO contributes 12 rows over four synaptic terms), not a complex-level phenotype distributed across subunits. The synaptic rows are therefore **kept**, as `KEEP_AS_NON_CORE`.

### 2. `GO:0017022 myosin binding` — right term, paralog-derived evidence chain

`PMID:15644318` is titled *"**BIG1** is a binding partner of myosin IXb"*; its Y2H screen recovered BIG1 and its co-IPs used anti-BIG1 antibodies. Querying GOA by that reference returns **18 annotations across 7 entities: 6 on human ARFGEF1, 0 on human ARFGEF2**, and exactly one BIG2-family row anywhere — `GO:0017022` on **rat** Arfgef2. Human ARFGEF2's row is the Compara projection of that single rat row; human ARFGEF1 holds the same term directly by IPI from the same paper.

**The term is nevertheless independently true of BIG2**: `PMID:23918382` shows reciprocal co-IP of endogenous BIG2 and non-muscle myosin heavy chain IIA in HeLa cells, and direct binding between separately *in vitro*-synthesised BIG2 and NMHC IIA. So the row is `KEEP_AS_NON_CORE` with a `propagation_review` (`SOURCE_BAD` / `WRONG_ORTHOLOG_OR_PARALOG` + `SOURCE_MISCITATION`) and the actionable recommendation is **re-reference to `PMID:23918382`**, not deletion. The same paper draws a real paralog line: BIG1 binds MYPT1 and PP1cδ directly, BIG2 does not.

### 3. `GO:0032760` is a receptor/ligand conflation → `MODIFY`

`PMID:17276987` measures release of the TNF **receptor** TNFR1 in exosome-like vesicles and never measures TNF cytokine production; it explicitly excludes the proteolytic route. The direction is doubtful too — the paper frames shed TNFR1 as a species that *binds and modulates TNF bioactivity*, i.e. a decoy. The reference annotates 2 rows on 1 entity, so this is one term choice, not a propagated error. → **`GO:1903543 positive regulation of exosomal secretion`**.

### 4. The disease-gene trap is inverted here: this gene is *under*-curated

GOA carries **no** cerebral-cortex, neuron-migration, neural-progenitor or heterotopia term for ARFGEF2 at all — the predicted pathology-derived over-annotation was looked for and **is not present**. What is present is a coverage hole: over the union of the UniProt `RX` list and the affinage citations (**43 PMIDs**, a derived set), **26 produce zero GO annotations anywhere** and **33 produce none on ARFGEF2**. The silent set includes the founding *Nature Genetics* paper, the AP-1/GGA dominant-negative papers, PP1γ, RIIβ/TNFR1, the DCB/HUS homodimer paper, β-catenin S675, integrin β1 recycling and the myosin-phosphatase paper.

**The disease material is deliberately kept out of `existing_annotations`.** PVNH2 appears in the `description` as disease context and in `suggested_questions`; no neurodevelopmental BP term is proposed, because the cell biology in `PMID:14647276` rests on brefeldin A (which also inhibits BIG1 and GBF1) and on a dominant-negative construct (BIG1 and BIG2 are in the same macromolecular complex), so neither isolates BIG2. The `core_functions` are the Arf-GEF catalysis, the recycling-endosome/receptor-recycling role, and the AKAP function — nothing neuronal.

### 5. `GO:0005085` is already maximal — no child to propose

`GO:0005086` ARF guanyl-nucleotide exchange factor activity, with eight sibling substrate-specific GEF terms, was **merged into `GO:0005085`** and survives only among its `secondaryIds`. Proposing a child would recreate a deliberate merge, so ARF1/ARF3 are recorded machine-readably instead: `core_functions[].substrates` plus `RO:0002233` (`has_input`) **`extensions`** on the two `GO:0005085` IDA rows. Filed as an `ONTOLOGY` `KnowledgeGap`.

The `GO:0005085` IBA node is genuinely heterogeneous (GNOM, Sec71, garz, cytohesins, PSD/PSD3, IQSEC2, yeast GEA1/GEA2/SEC7/MON2/SYT1), so the general term **is** the LCA of its donors — `GRANULARITY_MISMATCH` does not apply, and the rows are `ACCEPT`ed.

### 6. The two `GO:0005879 axonemal microtubule` rows → `MARK_AS_OVER_ANNOTATED`, with the limitation stated

The claim is a cilium. The donor paper's systems are hippocampal neurons, HEK293 cells and rat-brain immuno-EM with no ciliary content; the rat UniProt record's own `SUBCELLULAR LOCATION`, curated from that same paper, says only *"Cytoplasm, cytoskeleton"*.

Rather than assert "no other ArfGEF has a cilium term", it was **measured**: a 13-accession census (the four human large ArfGEFs, accessions **derived by gene-name lookup**, plus every protein resolvable from this gene's WITH/FROM) finds ARFGEF2 to be the only large-ArfGEF holder — and only by this projection.

Two things reported that make the review honest rather than tidy:

- **The first pass of that census hardcoded `Q9Y678` as GBF1. It is `COPG1`.** The conclusion survived; the cohort did not. Accessions are now derived and the resolved symbol is asserted against the one sought, with a self-test that breaks membership while leaving the count intact.
- **The census returned a result that cuts against the argument: `EXOC7`/Exo70 — BIG2's own exocyst partner — carries `GO:0036064` ciliary basal body.** A ciliary context near BIG2 is therefore not absurd. That is exactly why the action is flag-and-measure rather than delete, and it motivates the first suggested experiment.

**I did not conclude the rat IDA is wrong.** `PMID:15198677`'s full text is subscription-only (Europe PMC: `isOpenAccess: N`, `inEPMC: N`, no PMC id), so the rat row is filed for re-examination in `suggested_questions`; what is judged here is the non-experimental transfer into human, where the general parent `GO:0005856 cytoskeleton` is already annotated.

### 7. The third `MARK_AS_OVER_ANNOTATED`

`GO:0035556 intracellular signal transduction` IDA from `PMID:12571360`. That paper maps three PKA regulatory-subunit binding regions with 28 deletion mutants and shows cAMP-induced redistribution of BIG2 from cytosol to membranes; it does not show BIG2 transducing a signal, and the authors' own conclusion is hedged as a role that *"could coordinate cAMP and ARF regulatory pathways"*. Everything informative the paper establishes is already carried by `GO:0034237` and the membrane rows from the same reference. Flagged rather than removed, because BIG2 genuinely sits in a cAMP module.

## NEW rows (2)

Both from **human HeLa** experiments with rescue controls, so `IMP` is the correct code and the species is named:

- `GO:0051497 negative regulation of stress fiber assembly` — `PMID:23918382`
- `GO:0030335 positive regulation of cell migration` — `PMID:23918382`, corroborated by `PMID:22908276`

A third candidate, `GO:0034067 protein localization to Golgi apparatus` from the MYCBP/AMY-1 RNAi result (`PMID:16866877`), was **deliberately not added**: the cached abstract does not name the cell system and the full text was unavailable, so the species of the experiment could not be stated. It is raised as a `suggested_question` instead.

## Committed analyses

`genes/human/ARFGEF2/ARFGEF2-bioinformatics/` — all live REST, nothing hardcoded, `cache/` git-ignored:

- `resolve_withfrom.py` — all **45** distinct WITH/FROM tokens resolved, **0 unresolved**. Five MOD ids returned >1 hit; in every case all hits are the same gene (Swiss-Prot canonical + TrEMBL isoform records), reported rather than collapsed. Two tokens resolve only to unreviewed TrEMBL entries (fly `Sec71`, `garz`) and their *names* are therefore not used as evidence.
- `provenance_audit.py` — the rat funnel, per-reference paralog check, projection discriminator, coverage gap, cilium census.
- `audit_claims.py` — **12 self-tested guards**: duplicate YAML keys (strict loader + anchored raw-vs-parsed reconciliation), verbatim quotes (including the `file:` quotes CI does not check, and a one-physical-line rule for UniProt), reference titles vs the cached records, `supporting_entities` built from GOA rather than by hand, row coverage, numeric drift between prose and `provenance_audit.json`, cohort membership, and retracted phrasings matched by shape rather than by a hand-enumerated sentence list.

Three proteome-scale references exceed QuickGO's pagination ceiling; they are reported with `entities_available: false` and **no** entity count rather than a number derived from one page, while their per-gene counts stay exact via targeted queries.

`just validate human ARFGEF2` → **✓ Valid**, with `origin/main` merged in. `cache_lint` exit 0; the only `cache/go/terms.csv` change is one inserted row (`GO:0051497`), no deletions.

Also noted for the sibling review, **as a question and not a finding**: human ARFGEF1 carries `GO:0005096 GTPase activator activity` IDA from `PMID:15644318`, a paper whose result is that BIG1 **inhibits** myosin IXb's RhoGAP activity — and the same paper also gave ARFGEF1 `GO:0034260 negative regulation of GTPase activity`, which reads the opposite way. That is ARFGEF1's row, not this gene's, and `paint/ARFGEF1` was still an unmodified stub when this was written.


cmungall:
Both blocking items accepted and fixed in `9f281b2`. One suggestion's premise is wrong and I've said why rather than editing around it; the rest are taken.

## 1. `WRONG_IDENTIFIER` → `MISCITED` — agreed

Checked against the schema rather than taking it on trust: `WRONG_IDENTIFIER` is *"Identifier resolves to a DIFFERENT paper than intended"*, `MISCITED` is *"Paper is correctly identified but does not actually support the claim it is cited for"*. `PMID:15644318` resolves to exactly the intended Saeki paper, so `MISCITED` is right and my own `review_notes` already said so in prose. Fixed, and the notes now state explicitly why it is not `WRONG_IDENTIFIER` so the distinction doesn't get re-litigated.

## 2. The rat IPI verdict was asserted without the full text — agreed, and the asymmetry was real

This is the better of the two catches and I don't think it's a close call. `publications/PMID_15644318.md` carries `full_text_available: false`; the rat `GO:0017022` row is an experimental IPI; and I wrote the correct posture two hundred lines earlier for `PMID:15198677` and then didn't apply it here. The by-reference census is genuinely stronger than a title read, but you've put your finger on exactly what it does and does not show: **what curators did, not what the full text contains.**

Re-classified on what is actually verifiable:

```yaml
root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT   # was SOURCE_BAD
failure_modes: [GRANULARITY_MISMATCH]        # was [WRONG_ORTHOLOG_OR_PARALOG, SOURCE_MISCITATION]
source_entities:
  - source_id: UniProtKB:Q7TSU1
    source_status: UNRESOLVED                # was SOURCE_BAD
```

`EVIDENCE_CIRCULAR_OR_REDUNDANT` is chosen because its second clause — *"the target already has stronger direct evidence"* — is something I can demonstrate without reading the paywalled paper: human BIG2 has `PMID:23918382`, which I did read in full. I also dropped `WRONG_ORTHOLOG_OR_PARALOG` on a second ground you didn't raise: the immediate donor here is rat Arfgef2, the true **ortholog**. The paralog issue sits one link further back, in the rat row's own source paper, which is precisely the link I cannot verify.

The `reason` now carries the limitation in the same words as the other rat row, and `KEEP_AS_NON_CORE` plus the re-reference-to-`PMID:23918382` recommendation are untouched.

## 3. I don't think this premise holds — "more than doubles" *is* in the cached text

> *I could find no quantitative statement in `publications/PMID_23918382.md` supporting "more than doubles" (no fold-changes or percentages appear in the cached text).*

`publications/PMID_23918382.md` line 101:

> *"After BIG1 or BIG2 depletion ( Fig. 2 A ), phosphorylation of RLC T18 and S19 was **more than twice** that in cells treated with the NT siRNA or vehicle alone."*

and line 95 has a second one (*"IP of BIG2 after BIG1 depletion yielded approximately twice as much NMHC IIA"*). So the fold-change is stated in words rather than digits, `grep` for `%` or `-fold` misses it, and "more than doubles" is a faithful restatement. I've kept the wording.

**Your actionable half was right, though**, and it's the part that mattered: that claim was resting on the provider file when a primary sentence existed. The primary sentence is now the **first** `supporting_text` on the row, ahead of the affinage quote, so the claim rests on the text I read and the provider record is demoted to corroboration.

## 4–7. Suggestions taken

- **`full_text_unavailable`** set on `PMID:15198677`, `PMID:15644318` and `PMID:16866877`, each asserted against the cache's own `full_text_available: false` rather than set by hand. This is the right call — the restraint was argued in prose and invisible to tooling.
- **Core function 1 → `GO:0006893`**: you're right that AP-1/GGA is the TGN→endosome arm, not Golgi→surface. Added the E-cadherin/β-catenin sentence from `PMID:14647276`, which is the evidence that actually carries it.
- **`GO:0005879` enum fit**: kept `MARK_AS_OVER_ANNOTATED` and added a sentence saying why. `REMOVE` has to be earned by a positive argument, and the one thing that would earn it — reading the figure the rat IDA rests on — is what the paywall prevents; the census is a negative result about the family, not a demonstration that BIG2 is absent from cilia, and the `EXOC7` `GO:0036064` hit shows the surrounding biology isn't hostile to the claim. `UNDECIDED` would throw away a measurement that was made.
- **Two core functions sharing `GO:0005085`**: half-sentence added — same catalysis, different compartment and process set, and `E738K` separates the phenotypes without separating the activity.

## Folded in while waiting (verified, not speculative)

- **MON2 measured rather than asserted.** My IBA source comment said MON2 is a Sec7-fold protein without a catalytic Sec7 domain. Now checked: `P48563` has **no** `SEC7` `Domain` feature and **no PF01369**, while yeast SEC7 (`P11075`) and ARFGEF2 both do; MON2 keeps DCB (PF16213) and HUS (PF12783) only. The complication is recorded too — MON2 nevertheless holds `GO:0005085` itself by ISS/IGI/IPI from `PMID:12052896`, which is a fact about MON2's record, not an argument about ARFGEF2.
- **A superlative the source doesn't license.** The description said "highest transcript levels in placenta, lung, heart, brain, kidney and pancreas". UniProt says only "Expressed in", and the same entry's HPA line reads `Tissue enhanced (parathyroid)`. Reworded.
- **Two more checks reported negative.** `PMID:11809827` annotates 2 entities, both rat, 0 human — which looked like it might be a human experiment annotated to rat; it isn't, the cells are **NRK** and BHK. And PANTHER placement is clean: ARFGEF2 `SF124`, ARFGEF1 `SF137`, ARFGEF3 `SF344`, GBF1 `SF388`, each subfamily named for its own occupant, so none of the mis-clustering failure modes applies.
- **`audit_claims.py` gained an action-histogram guard** (now 13 self-tested guards), because the PR body publishes that table and the body is outside any lint's scan surface — the first draft of it was wrong on three of six rows and was caught only by recomputing.

`just validate human ARFGEF2` → ✓ Valid with current `origin/main` merged; `cache_lint` exit 0; `audit_claims.py` 0 problems, 13/13 guards fire.

cmungall:
Pushed `6c799d5`. **Posted by the coordinator, not the agent that reviewed this gene** — it is stopped on a session limit, and the blocker was well-specified enough to clear without it.

**🟡 `GRANULARITY_MISMATCH` — dropped the field.** You are right that it asserts something the review does not believe: the schema defines it as "Parent term is true but uninformative, or child term overstates specificity", while this row's `summary` and `RESULTS.md` §2 both say the defect is the evidence chain and not the term. `failure_modes` is optional and `root_cause: EVIDENCE_CIRCULAR_OR_REDUNDANT` already carries the finding, so dropping is the honest fix — and your reasoning against the alternatives holds: `CIRCULAR_PROPAGATION` and `SOURCE_EVIDENCE_WEAK` are both wrong for an experimental rat IPI, which is not a transfer.

**I deliberately did not take your second option**, and want to be explicit about why rather than have the silence read as disagreement. `MODIFY` → `GO:0045159` changes a `KEEP_AS_NON_CORE` row's action and adds a replacement term; that is a curation decision on a gene I have not reviewed, and it belongs to the agent that did.

For that agent's benefit, I checked your evidence and it is stronger than your comment claims. `PMID_23918382.md:91` is explicitly isoform-discriminating, not merely isoform-naming:

> Approximately 5% of the total cellular NMHC IIA was among proteins collected by BIG1 immunoprecipitation (IP), and ca. 1% by BIG2 IP, **whereas no NMHC IIB or IIC was detected, although their presence in the cells was clear**

Reciprocal IP is at `:91` ("IP of endogenous NMHC IIA also yielded BIG1 and BIG2"), and `GO:0045159 myosin II binding` is already in `cache/go/terms.csv:5841`. So the specificity is measured against the two isoforms that would have made the parent term the right level. That looks like a well-founded `MODIFY` to me; I am leaving the call to whoever owns the gene.

**🔵 The `GO:0017022` summary** still leading with "a paper about the paralog BIG1" — also left for that agent. You are right that `summary` renders standalone, where it reads as the verdict the fix withdrew.

`just validate human ARFGEF2` → `✓ All validations passed` on this head.

PR #3073, opened 2026-09-19
Review human ARHGAP11B: GOA asserts and negates GAP activity for the same gene

cmungall:
## The headline: GOA asserts and negates the same molecular function for this gene

ARHGAP11B's GOA record contains both of these:

| qualifier | term | evidence | reference |
|---|---|---|---|
| `NOT\|enables` | GO:0005096 GTPase activator activity | **IDA** | PMID:27957544 |
| `NOT\|enables` | GO:0005096 GTPase activator activity | **IDA** | PMID:25721503 |
| `enables` | GO:0005096 GTPase activator activity | TAS | Reactome:R-HSA-8981637 |

Two independent experimental papers say the protein has no GAP activity; Reactome says it does.
UniProt has already adjudicated it — the recommended name is *"Inactive Rho GTPase-activating
protein 11B"* and the entry carries a `CAUTION` to that effect. The positive row is the one with
the weaker evidence code.

## Where the wrong positive comes from — measured, not asserted

Queried QuickGO **by reference** rather than by gene, comparing `numberOfHits` against
`len(results)` and paging until they matched:

| reference | annotations | distinct entities | terms |
|---|---|---|---|
| `Reactome:R-HSA-8981637` | 43 (all read, 1 page) | 22 | GO:0005096 → 21 entities, GO:0005829 → 21, GO:0005886 → 1; all TAS |
| `Reactome:R-HSA-9012999` | 103 (all read, 2 pages) | 103 | GO:0051056 → all 103; all TAS |

R-HSA-8981637 is a Reaction whose **single `catalystActivity` is a set-level entity,
`"GTPase activator activity of RHOA GAPs [cytosol]"`**. Both the activity term and the cytosol term
are handed to every member of that defined set at once. ARHGAP11B is in the set because it looks
like ARHGAP11A. **For ARHGAP11A the same projection is correct and should stay.**

## The paralog comparison is the cleanest evidence here

| term | ARHGAP11A (Q6P4F7) | ARHGAP11B (Q3KRB8) |
|---|---|---|
| GO:0005096 | IBA (node PTN001038329) + **IDA `enables`** (PMID:27957544) + Reactome TAS | **IDA `NOT\|enables` ×2** + Reactome TAS `enables` |
| GO:0043547 | **IDA `involved_in`** (PMID:27957544) | **IDA `NOT\|involved_in`** (PMID:27957544) |
| localisation (IDA) | GO:0005634 nucleus (PMID:31883789) | GO:0005759 mitochondrial matrix (PMID:31883789) |

**The same paper produced the positive annotation on ARHGAP11A and the negated one on ARHGAP11B**,
and the same study that put ARHGAP11B in mitochondria put ARHGAP11A in the nucleus. The
UniProt-sourced curation is internally consistent and right for both paralogs. Every wrong row on
ARHGAP11B is non-experimental.

**PAINT got this right and deserves credit:** ARHGAP11B has **no IBA row at all** and UniProt
records `PAN-GO; Q3KRB8; 0 GO annotations based on evolutionary models` — even though PANTHER puts
it in `PTHR15670:SF4`, a subfamily *named for its paralog* that it shares with ARHGAP11A. The
predicted IBA paralog-sweep did not happen. The leak arrived through Reactome instead.

## Actions

17 GOA rows; 17 seeded entries; no collapse (counted the TSV, not the stub). Plus 1 NEW row.

| action | n | rows |
|---|---|---|
| ACCEPT | 11 | both `NOT GO:0005096` IDA, `NOT GO:0043547`, GO:0005759 ×2, GO:0006543, GO:0021987 ×4, GO:0035795 |
| MODIFY | 2 | both GO:0005515 → **GO:0044325** transmembrane transporter binding |
| REMOVE | 4 | GO:0005096 TAS, GO:0005829 TAS, GO:0051056 TAS, GO:0007165 IEA |
| NEW | 1 | **GO:2000179** positive regulation of neural precursor cell proliferation, IMP, PMID:36098218 |

11 + 2 + 4 + 1 = 18 = 17 GOA rows + 1 NEW. **Every REMOVE is non-experimental** (three TAS, one IEA).
No experimental annotation was removed, modified or demoted anywhere in this review.

- **GO:0005829 cytosol** falls with the GAP row — same reaction, same 21-entity projection, and a
  Reactome compartment label is the reaction's, not a measurement. Contradicted by the
  mitochondrial-matrix IDA.
- **GO:0007165 signal transduction** is an interpro2go product of `IPR000198`, the RhoGAP domain
  signature.
- **MODIFY ×2**: the partners resolve to reviewed canonical Swiss-Prot SLC25A4/ANT1 and
  SLC25A5/ANT2 — the gene's actual mechanism, not screen noise.

> **Correction (post-open):** the ACCEPT count above was first written as 12, and the
> arithmetic as `12 + 2 + 4 = 18` — which also came to 18, because it had dropped the NEW
> row from the sum at the same time as inflating ACCEPT by one. Two errors cancelling into
> a total that looked reconciled. Caught by `check_review_counts.py`, now committed under
> `ARHGAP11B-bioinformatics/` with a 5-mutation self-test. The commit message of
> `b147d382ad` still carries the wrong figure; it is not rewritten because that would need
> a force-push.

## The structural analysis refutes the reason you would expect

`genes/human/ARHGAP11B/ARHGAP11B-bioinformatics/` (script + README + generated RESULTS.md;
a fresh run reproduces the committed files byte for byte; `--self-test` breaks six guards).

**The catalytic arginine finger R87 is RETAINED.** Confirmed reciprocally against both ARHGAP11A's
own annotated finger and ARHGAP1's structurally resolved one (PDB 1TX4), requiring the aligned
position to land on one of ARHGAP11B's own annotated Sites. So the usual
`PSEUDOENZYME_OVERANNOTATION` argument from missing catalytic residues is **unavailable here**, and
using it would have been a false reason for a correct action.

The screen does flag **6 of 66** reviewed human RhoGAP-profile proteins. My first write-up
called that "a test that otherwise works" and ARHGAP11B "a false negative" of it. Asking what
UniProt independently concluded about those six **refuted my own framing**:

| entry | residue | UniProt verdict | evidence |
|---|---|---|---|
| OCRL | Q | inactive, "lacks the catalytic arginine" | **ECO:0000269** (experimental) |
| INPP5B | Q | same statement | ECO:0000250 (by similarity) |
| ARAP2 | Q | silent | — |
| FAM13B | Q | silent | — |
| **ARHGAP36** | **T** | **asserts GTPase activator activity** | ECO:0000250 |
| DEPDC1B | I | silent | — |

Only **one** of the six has experimental backing for inactivity, and **ARHGAP36 is the reciprocal
of this gene** — it lacks the arginine and UniProt calls it an active GAP anyway, by similarity.
So residue identity and curated activity are **decoupled in both directions** in this family. That
is a stronger statement than the one it replaces, and it is raised as a `suggested_question`
against ARHGAP36.

What is lost instead, measured: **2 of 24** mapped GAP:GTPase interface positions fall past the
divergence point at residue 220. Reported as corroboration, not as an independent argument — the
`NOT` rows rest on the experiments.

**A UniProt feature defect worth reporting:** ARHGAP11B's Rho-GAP `DOMAIN` is annotated `49..250`,
running **30 residues past residue 220** into the human-specific frameshift C-terminus that is
homologous to nothing. The annotated domain on the inactive truncated paralog (202 aa) is therefore
**11 residues longer** than on the catalytically active parent ARHGAP11A (`49..239`, 191 aa). Any
pipeline reading domain extent as evidence of an intact RhoGAP reads this protein as *more* intact
than the active one.

The divergence point (220) and prefix identity (216/220 = 98.2%) are fitted as a changepoint with
significance set by permutation — no identity cutoff is chosen anywhere — and reproduce the
published figures exactly, which is a check on the pipeline rather than a new finding.

Two of the analysis's own earlier results were **wrong and are documented in its README** rather
than quietly fixed: a global alignment put ARHGAP1's arginine finger on ARHGAP11A residue **684**
(the two proteins carry the domain at opposite ends), and after fixing that, a *local* alignment
reported **0 of 25** contacts lost because it stopped 16 residues short of the divergence point —
six contacts were past its end, i.e. not absent but **not measured**.

## Coverage gap: the decisive human experiment supports no annotation

Every `GO:0021987` row in GOA is **gain-of-function**, and four of the five systems are non-human
(mouse in utero electroporation, transgenic mouse, ferret, marmoset). Reasons name the system in
each case rather than asserting the result of the human protein in human tissue.

**PMID:36098218** is the only study that perturbs ARHGAP11B **in a human system**, in both
directions — dominant-negative interference in human cerebral organoids drops cycling basal
progenitors to chimpanzee levels, and double-knockout rescue shows ARHGAP11B, but not ARHGAP11A, is
required for basal radial glia. It supports **no GOA row at all**. That is the basis of the NEW
`GO:2000179` IMP, and the evidence code is IMP only because the organoids are human.

## Notes on process

- Retraction / erratum / expression-of-concern checked per-PMID by reading `CommentsCorrections`
  on each cited article's own record (a Publisher Correction is invisible to a pubtype search).
  All 8 clean; recorded as negatives in each `reference_review`.
- Both `GO:0005515` partner accessions resolve to reviewed canonical Swiss-Prot entries of the
  expected length — no TrEMBL/ORFeome substitution.
- PMID:31883789 and PMID:25721503 are **abstract-only** in cache. Their IDA rows are accepted on the
  curators' reading. Where the abstract is less specific than the term (it says "mitochondria", the
  term says "matrix"), the row says so explicitly and defers.
- **No `file:` references**: the pre-write hook resolves `project_root` from its own location, so
  for an agent in `.claude/worktrees/` it runs the validator against the *main checkout*, where this
  gene folder does not exist, and rejects every `file:` ref. I verified this is a false positive — a
  copy with a `file:` ref validates `✓ Valid` from this worktree — so the bioinformatics numbers are
  cited in `review.reason` prose instead. The `⚠ No annotations reference available deep research
  files` warning from `just validate` is a direct consequence and not an oversight.


cmungall:
Pushed `39fc047`. **Posted by the coordinator, not the agent that reviewed this gene** — it is stopped on a session limit, and this blocker was a single verifiable line.

**🟡 The stale history count.** Fixed, and I counted from the YAML directly rather than trusting either your figure or the record's:

```
actions: {'ACCEPT': 11, 'REMOVE': 4, 'MODIFY': 2, 'NEW': 1}
total entries: 18   GOA rows: 17   (ARHGAP11B-goa.tsv: 17 data rows)
```

So the line was wrong twice over, exactly as you describe — `ACCEPT` was 12 rather than 11, *and* the sum `12 + 2 + 4 = 18 = 17 + 1` drops the `NEW` row while asserting it. It now reads `11 + 2 + 4 + 1 = 18 = 17 GOA rows + 1 NEW`. The review file itself was already correct; only the provenance record had gone stale when `fade587c7` corrected the figure in four other places.

Edited in place rather than appended, on your reading that append-only binds `target.slug`/`target.path` and this record has not merged. `validate-history` → `No issues found`.

This is the third instance this session of a **prose claim about an artifact drifting from the artifact** — a history record describing a superseded decision on APOOL, an uncommitted checker asserting a dead invariant on AKAP12, and now a count corrected everywhere except its provenance record. Each passed CI, because nothing compares a claim to the thing it describes.

**On the `file:` reference you flagged as missing:** the cause is real and worth recording beyond this PR. `.claude/settings.json` invokes hooks as `$CLAUDE_PROJECT_DIR/.claude/hooks/...`, and `validate_ai_review_pretool_hook.py:117` derives `project_root` from `__file__`. Inside an agent worktree `$CLAUDE_PROJECT_DIR` remains the session root, so the hook validates against the **main checkout** — where a just-written `RESULTS.md` does not exist. The failure is confident and looks genuine, so the rational response is to delete the reference, which is what happened here. The analysis is committed; only the machine-checkable link to it was lost.

The remaining suggestions (`core_functions[2]` stating an absence with no terms; `core_functions[1]` inheriting a location) are curation judgements on a gene I have not reviewed, so I am leaving them to the authoring agent rather than guessing.

Agreed on the biology, for what it is worth from outside: the `GO:0005096` self-contradiction is the most consequential finding in this batch, and the bioinformatics refusing to supply the pseudoenzyme argument — R87 is *retained* — is what makes the conclusion credible rather than convenient.

cmungall:
Pushed `ba06710`. **Posted by the coordinator** while the authoring agent is on a session limit.

**🟡 The retracted framing.** You are right, and right that it is the worse of the two. The record still read:

> The screen is not vacuous - 6 of 66 reviewed human RhoGAP-profile proteins have lost the arginine - so ARHGAP11B is a false negative of a test that otherwise works.

while `RESULTS.md:67-71`, the notes at `144-159`, the review YAML at `1017` and the PR body all carry the refutation. My `39fc047d8` corrected the arithmetic 23 lines above and walked straight past this. The record now states the decoupled-in-both-directions conclusion: ARHGAP11B keeps the arginine and is experimentally GAP-dead; ARHGAP36 lacks it at its own annotated position and UniProt asserts GAP activity anyway by similarity (`ECO:0000250`); only 1 of the 6 flagged proteins has experimental support for inactivity. `validate-history` → `No issues found`, and `false negative` now appears zero times in the file.

Your framing of the difference is the useful part and I want to record it rather than just agree: **a wrong tally is bookkeeping; this recorded a scientific conclusion the committed analysis contradicts** — and because the retracted version is the *weaker* claim, the record was underselling its own result. A reader trusting provenance would have come away with the lesser finding.

That is now the **third** instance this session of a prose claim about an artifact drifting from the artifact, and the **second in this single file**. The others were a history record describing an APOOL decision the review had reversed, and an AKAP12 checker asserting an invariant that had gone to zero. All three passed CI, because nothing compares a claim to the thing it describes — and a provenance record is the worst place for it, since its whole purpose is to be believed later without re-derivation.

Worth noting for whoever picks this up: both defects in this file were introduced by *correcting something else*. `f8d3ab34e` wrote the record before the refutation landed; my fix touched the same file and did not re-read it. The cheap discipline is that any edit to a provenance record should re-read the whole record, not the line being changed.

On the `file:` reference — agreed it warrants a tooling issue rather than a per-gene workaround, and I have recorded the mechanism. Your re-reading of `core_functions[2]` matches mine: the entry explicitly says no molecular function is asserted, which is deliberate.

PR #3075, opened 2026-09-19
Review human ARFGEF3 (BIG3): a Sec7-domain protein that is not a GEF

cmungall:
Full GO annotation review of human **ARFGEF3 (BIG3)**, `Q5TH69`.

**Row reconciliation:** GOA TSV has **8** data rows; the `fetch-gene` stub seeded **8**
(no collapsed rows on this gene — checked, not assumed). The review has **14** entries =
**8 reviewed GOA rows + 6 `NEW` proposals**. Verified programmatically, along with the
assertion that every row's `supporting_entities` reproduces the GOA WITH/FROM field
exactly and that every `propagation_review` names only sources drawn from it.

---

## Headline: the name asserts an activity the protein cannot perform

The symbol, the UniProt recommended name (itself `ECO:0000305`, a curator inference from
similarity) and both `GO:0005085` rows all assert ARF-GEF activity. Nothing has ever
measured it on this protein, and the residue that performs the chemistry is absent.

Sec7 ARF-GEFs act through one invariant glutamate — the "glutamic finger" — which inserts
into the ARF nucleotide pocket to destabilise bound GDP. I re-derived its presence from
live sequence rather than taking the literature's word, with the comparison panel built
**from this row's own WITH/FROM field** (`ARFGEF3-bioinformatics/`):

| class | n | residue at the derived glutamic-finger column |
|---|---|---|
| catalytically-verified WITH/FROM donors | 17 | **E in 17/17** |
| known-dead comparator (`ARFGEF1 E793A`) | 1 | A — scored *absent* |
| **human ARFGEF3** | 1 | **N692** (`LLSLSNVEEVD`, zero motif matches) |
| mouse Arfgef3 | 1 | **S688** (`LLALSSVEEVD`) |

The column is **derived** from the known-active donors, not hardcoded, and cross-checked
against the alignment-free `[FY]-x-[LIVM]-P-G-E` motif. Controls run both ways and
`--self-test` exits 0: leave-one-out over all 17 donors, a verified-active positive
control, and a Glu→Ala mutant of that control that must score negative — a detector that
cannot report a negative proves nothing, so the dead comparator is in the **published**
table, not only in the test. The mutation target is asserted present before it is
mutated, so the control cannot silently no-op.

**The asymmetry is respected:** losing the catalytic residue is strong evidence *against*
the activity; retaining it would *not* have been evidence *for* it. Only the negative
direction is claimed.

This reproduces `PMID:24997568` (7/7 of its other Sec7 domains positive, BIG3 negative) as
a **precondition** of the script running at all, and matches what the group that
characterised the protein states in two further papers.

## Verdicts, and why they differ by evidence strength

| term | ev | action |
|---|---|---|
| `GO:0005085` GEF activity | IBA | **REMOVE** |
| `GO:0005085` GEF activity | IEA (IPR000904) | **REMOVE** |
| `GO:0032012` reg. of ARF signal transduction | IEA (IPR000904) | **MARK_AS_OVER_ANNOTATED** |
| `GO:0016192` vesicle-mediated transport | IBA | **ACCEPT** (core) |
| `GO:0005737` cytoplasm | EXP | **ACCEPT** (core) |
| `GO:0005737` cytoplasm | IEA (SL-0086) | **ACCEPT** |
| `GO:0030133` transport vesicle | IEA (SL-0244) | **MODIFY** → `GO:0030141` |
| `GO:0030658` transport vesicle membrane | IEA (SL-0245) | **MODIFY** → `GO:0030667` |

`GO:0032012` is deliberately *not* removed: it is **unsupported rather than contradicted**.
A catalytically dead Sec7 domain could still bind ARF and act as a decoy — which is the
model the protein's own characterisers propose and explicitly flag as a hypothesis. No
experiment has measured ARF-GTP loading or ARF binding for this protein in any system.

## The IBA node is sound — the defect is target-side

Resolving all 32 distinct WITH/FROM tokens (zero unresolved) shows **18/18** protein
donors of the `GO:0005085` row carry their own experimental evidence and **15/18** their
own IDA. So "these donors only carry the same family-level inference" is factually false
here in every case, and `SOURCE_WEAK_OR_INFERRED` would be contradicted by my own
analysis. Classification is `PROPAGATION_BAD` + `PSEUDO_OR_SUBACTIVITY_LOSS`, recorded
with a machine-checkable `residue_claims` block.

`GO:0016192` goes the other way: its 18 donors hold **thirteen different** descendant
processes between them (anterograde, retrograde, intra-Golgi, endosomal, exocytic), so
the broad term is the genuine **LCA**, not lazy curation. `GRANULARITY_MISMATCH` requires
donors to agree; they do not. ACCEPTed with no specificity upgrade.

## A second, systematic defect: SL-0244 maps into a disjoint GO branch

| SubCell | UniProt's name | maps to | literature supports | relation |
|---|---|---|---|---|
| `SL-0086` | Cytoplasm | `GO:0005737` | — | EXACT |
| `SL-0244` | **Secretory vesicle** | `GO:0030133` transport vesicle | `GO:0030141` secretory granule | **DISJOINT** |
| `SL-0245` | **Secretory vesicle membrane** | `GO:0030658` | `GO:0030667` | **DISJOINT** |

`GO:0030141` and `GO:0030133` are sibling branches under `GO:0031410` — verified against
QuickGO's `is_a`/`part_of` closure, not read off the labels. GO defines `GO:0030133` as the
**constitutive** pathway; insulin and glucagon granules are **regulated** secretory
granules. The mismatch starts inside UniProt: SL-0244's own definition is
regulated-pathway language ("cargo — e.g. hormones or neurotransmitters … docks and fuses
to release its content").

Scale, stated precisely: SL-0244 is carried by **130,685** entries and SL-0245 by **90,419**
(from the `x-total-results` header). That is the number the rule is *applied to* — **not** a
claim that all are mis-annotated.

## Six functions GOA records nowhere

All from human experiments on endogenous protein except the last, which is mouse and
therefore **ISS**, not IMP:

- `GO:0140313` molecular sequestering activity (IMP) — BIG3 holds PHB2 out of the nucleus
  *and* occludes its karyopherin-binding surface; GO's definition almost word for word.
- `GO:0042308` negative regulation of protein import into nucleus (IMP).
- `GO:0008157` protein phosphatase 1 binding (IPI) — endogenous, via a canonical RVxF motif.
- `GO:0004864` protein phosphatase inhibitor activity (IMP) — binding and inhibition are
  separate claims and the paper separates them.
- `GO:0051018` protein kinase A binding (IPI) — annotated at the **parent**, not
  `GO:0034237`, because the interaction was measured but the RII-subunit specificity was
  only predicted.
- `GO:1904410` negative regulation of secretory granule organization (**ISS**, mouse
  `Q3UGY8`) — same sign for insulin in β-cells and glucagon in α-cells.
  `GO:1903306` (regulated secretory *pathway*) was considered and **rejected**: it
  targets the release step, which the capacitance data specifically exonerate, and
  QuickGO's closure shows the two terms are **disjoint** — so the wrong one asserts a
  different claim, not a coarser one.

No ERα-signalling term is proposed: BIG3's substrate is PHB2, and BIG3–ERα binding was
tested and **not detected**.

## Provider assessment (affinage, gates passed)

Both failure directions appeared, which is why they are recorded in `reference_review`:

- **Precision, invisible to the gates:** 2 of 14 citations (`PMID:14657013`, `PMID:15707593`)
  are about a **different gene** — "BIG-3 = BMP-2-induced gene 3 kb", a 34 kDa
  seven-WD-40-repeat protein (`PMID:11551928`), a pure symbol collision with the 2177-aa
  Sec7/ARM ARFGEF3. The PMIDs are real and the quotes faithful; the join to *this gene* is
  wrong — exactly what no mechanical gate sees. Marked `relevance: NONE`,
  `correctness: MISCITED`; nothing rests on them.
- **Recall:** an independent PubMed sweep found four real ARFGEF3 papers the provider never
  returned, including **`PMID:27696409`** — lysosomal localisation and GABAergic
  transmission, an entire fourth compartment and tissue context.

## Things deliberately not done

- No lysosome or mitochondrion CC term (each rests on one paper, non-human or one tumour
  lineage) — raised as `knowledge_gaps` and questions instead.
- No `proposed_new_terms`: every function needed already exists, and all candidate ids were
  checked against QuickGO `/complete` for `isObsolete` and `secondaryIds`.
- One validation warning is **left standing** ("no annotations reference the deep research
  file"). Satisfying it would mean quoting a provider sentence as evidence — the practice
  this campaign forbids, on a record that miscited two papers.

## Unresolved

`PMID:24997568` says its conclusion is "consistent with a previous demonstration by the
GST-GAT pull-down assay", and the PMC cross-reference resolves that to `PMID:24711543`. I
could not locate such an assay in that paper's accessible text. **The review therefore
rests on the residue analysis and the authors' repeated sequence-level statements, not on
that pull-down**, and the gap is recorded rather than papered over.

## Paralogue note for PAINT

ARFGEF1 and ARFGEF2 carry the **identical** `GO:0005085` IBA from the **identical** node and
should **keep** it — their glutamic fingers are intact at **E793** and **E738** (verified
against live sequence) and both hold their own IDA. The three reviews should diverge on
this row for a stated structural reason rather than silently.

Separately: yeast **MON2** is a donor on this row despite having **no annotated SEC7 domain**
and no `IPR000904`, holding the term only by IGI/IPI/ISS. MON2 and ARFGEF3 share
`IPR015403`, so the same node may be donating exchange activity to a second member that
does not perform it.


cmungall:
All three items taken, and I verified each premise rather than conceding on assertion.

**Blocking — `GO:0140313` → `GO:0140311 protein sequestering activity`. Correct; fixed.**
Verified against QuickGO: `GO:0140311` is an `is_a` child of `GO:0140313` carrying the
identical definition with "a specific molecule" narrowed to "a protein", and it is not
obsolete. The sequestered species is PHB2, a protein, so the child asserts nothing the
ligand's identity does not already fix. Your contrast with `GO:0051018` is the right
principle and I have written it into both rows, because the two cases looked
inconsistent without it: the sequestering child is free, whereas the PKA child
(`GO:0034237`) would assert an RII-subunit specificity that was predicted in silico and
never measured. Repo precedent confirmed — `GO:0140311` is already used in A1BG, A2M,
AHNAK, APPL2, CLU, PDCD5, SMO and SQSTM1.

**Suggestion 1 — `GO:0004865` over `GO:0004864`. Taken, for the same reason.**
Verified `GO:0004865` is an `is_a` child of `GO:0004864`, and verified the premise
directly rather than from the term label: `P62136` is "Serine/threonine-protein
phosphatase PP1-alpha catalytic subunit" with `EC 3.1.3.16` on both its
O-phospho-L-seryl and O-phospho-L-threonyl reactions. So the substrate class is fixed by
the inhibited enzyme's identity and the child carries no unproven increment. Treating
this as optional would have left the review applying your own principle in one place and
not the adjacent one, so I took it.

**Suggestion 2 — motif count. Taken; this one was a real reporting defect of mine.**
You are right that five active donors show `motif_hits: 0` in the TSV with nothing
explaining it, which reads as tension with the 17/17 claim. It is not — the cross-check
aborts only on *disagreement*, and absence is silent — but that distinction lived only in
the code. `sec7_catalytic_check.py` now emits a `motif_crosscheck` block:

```
active_donors_agreeing_with_derived_column: 12
active_donors_with_no_motif_match:           5   (IQSEC2, PSD, Psd3, SYT1, YDYB_SCHPO)
active_donors_disagreeing:                   0
```

and `RESULTS.md` now states that those five are divergent Sec7 branches (BRAG/EFA6 in
animals, SYT1 and the fission-yeast protein in fungi) which keep the catalytic glutamate
while departing from the FRLPGE consensus around it — **and all five still carry E at the
derived column**. So the 17/17 and the 12/17 are measuring different things, and the
numbers now say so without needing the prose.

Thanks also for independently confirming the `ECO:0000250|UniProtKB:Q3UGY8` chain — that
is the load-bearing premise behind both the `GO:0030133` → `GO:0030141` MODIFY and the ISS
choice, and it is the one I would most have wanted checked by someone else.

Gates re-run on the new head (`1d359fb163`), exit codes captured directly:
`just validate human ARFGEF3` → `✓ Valid` (1 warning, deliberately left standing — see
below); `--self-test` → 0 problems; verbatim quote check → 0 problems over 79 PMID quotes;
`file:` quote check → 0 problems over 5; `supporting_entities`-vs-GOA → 0 problems;
`validate-history` → no issues; `cache_lint` → clean, with `terms.csv` adding only
`GO:1904410` and deleting nothing.

On the one standing warning ("no annotations reference available deep research files"):
satisfying it would mean quoting an affinage sentence as `supporting_text`, on a record
that cited two papers about a different gene (`PMID:14657013`/`PMID:15707593` are the
WD-40 "BMP-2-induced gene 3 kb", not this protein). Left standing on purpose and
explained in `ARFGEF3-notes.md`.

cmungall:
Thanks for the approval on `1d359fb163`. **I have superseded it with two more commits, deliberately** — flagging that explicitly so the approval is not read as covering the current head.

**1. A factual error your review surfaced indirectly, which I then introduced into the fix.**
When I swapped `GO:0140313` → `GO:0140311`, `ARFGEF3-notes.md` ended up naming the child
while still quoting the **parent's** definition ("Binding to *a specific molecule*…"). That
is a wrong quote attached to the right term, in a file a curator reads. Corrected to
`GO:0140311`'s own wording ("Binding to *a protein*…").

The same commit adds the parent-vs-child weighing you noted was missing at
`ARFGEF3-notes.md:94`, written as a **test** rather than a preference, because otherwise the
review looks inconsistent across its own rows:

> take the child when it asserts nothing the evidence has not already fixed
> (PHB2 is a protein; PPP1CA is `EC 3.1.3.16`); stay at the parent when it would
> (`GO:0051018`, where the RII specificity was predicted in silico, never measured).

**2. The MON2 recommendation is now verified rather than conjectural.** It previously said
the node "may be" donating exchange activity to a second member that does not perform it. I
checked the reciprocal side and it holds, so it now states the evidence:

- MON2 itself **receives** `GO:0005085` by IBA from `PTN008950430`, with a WITH/FROM list
  **identical** to the one on the ARFGEF3 row;
- all of its non-IBA support for the term traces to a **single paper**, `PMID:12052896` —
  ISS from SEC7, plus IGI and IPI with one partner.

So neither of the two members lacking a functional Sec7 catalytic site has ever had exchange
activity measured directly, and both inherit the term from the same node. That turns a hedge
into something PAINT can act on.

Neither change touches an `action`, `term`, `evidence_type`, `supporting_text` or
`core_functions` entry in the review YAML — by the usual convergence rule item 2 would have
been a comment. I pushed anyway because item 1 is a correctness fix to committed content,
and batching the two into one push was cheaper than two rounds.

Gates re-run on `6a189b30b5`, exit codes captured directly: `just validate human ARFGEF3` →
`✓ Valid` (same single deliberate warning); verbatim quote check → 0 problems over 79 PMID
quotes; `file:` quote check → 0 over 5; `supporting_entities`-vs-GOA → 0; `validate-history`
→ no issues; `cache_lint` → clean; `terms.csv` adds only `GO:1904410` and deletes nothing.

cmungall:
Taking the non-blocking item rather than declining it — you were right, and it was the one
place in this PR where a claim and its evidence had come apart.

I upgraded the MON2 statement from "may be" to "checked and it holds" on the strength of an
**interactive query**, then shipped the conclusion without the query. Every other quantitative
claim here ships with the script that produced it, and this campaign's own rule is that an
uncommitted check is not a check. `donor_evidence.tsv` corroborating the *shape* but not the
three specifics is exactly the gap you identified.

**`reciprocal_donor_check.py` is now committed** and derives all three:

| | `SGD:S000005241` → `P48563` MON2 |
|---|---|
| IBA source node | **`PANTHER:PTN008950430`** — same node as this row |
| IBA WITH/FROM identical to this row's | **yes**, 19 tokens vs 19 |
| distinct non-IBA references | **1** (`PMID:12052896`) |
| holds its own IDA | **no** |

Two design points, since a script that merely confirms what I already believed would be
worth little:

- **Suspects are selected by measurement, not named.** The script takes any protein donor of
  the term that lacks the term's own InterPro signature. MON2 is the output, not the input —
  so if a future GOA release adds `IPR000904` to MON2, or introduces another signature-less
  donor, the finding updates instead of silently persisting.
- **The set-identity claim is falsifiable and tested as such.** `--self-test` (exits 0)
  removes one member from the subject's WITH/FROM set and requires the comparison to flip to
  `False`; it also checks that a donor which *does* carry the signature is never selected, and
  that the GOA/QuickGO prefix normalisation reconciles `MGI:MGI:1334257` with `MGI:1334257`
  **without** collapsing two distinct SGD ids. That last one matters: without normalisation
  the two services' spellings would report identical sets as different.

**`PMID:12052896` is now cached and in `references:`** with a `reference_review` explaining it
supports no annotation on this gene and is cited only in the PAINT question. Reading it
strengthened the point rather than just documenting it — the title calls Ysl2p/Mon2
"*homologous to* Sec7 domain guanine nucleotide exchange factors", and the abstract claims
only a "*potential* function as an Arf guanine nucleotide exchange factor (GEF)". So MON2's
`GO:0005085` rests on homology plus genetic and physical interaction, never on a measured
exchange reaction — which is the same shape as ARFGEF3's case, one node donating the activity
to two members that lack a functional catalytic site.

Thanks also for the independent check on `A1BG-notes.md:95-97` — verifying the corrected
`GO:0140311` wording against a separately-authored file on `main` is a better test than
re-reading my own diff, and it is the check I should have thought to run myself.

Gates on `5fc9c95846`, exit codes captured directly: `just validate human ARFGEF3` → `✓ Valid`
(same single deliberate warning); `reciprocal_donor_check.py --self-test` → 0 problems;
verbatim quote check → 0 problems over **85** PMID quotes (up from 79, the new reference
included); `file:` quote check → 0 over 5; `supporting_entities`-vs-GOA → 0; `validate-history`
→ no issues; `cache_lint` → clean; `terms.csv` still adds only `GO:1904410` and deletes nothing.
A second run of both new scripts reproduces their committed outputs byte-for-byte.

cmungall:
All three taken. Two of them were substantive rather than cosmetic, and acting on (3)
exposed a defect of mine that no gate in this repo would have caught.

**(3) The suspect count is now a closed claim.** You were right that `suspect_donors()`
tested `has_sec7_interpro == "no"` and so skipped a blank determination silently — the count
was over *resolved* donors while reading as a total. It now raises and names any donor
without a determination, so a future GOA change that breaks a lookup fails loudly instead of
quietly shrinking the result. The self-test blanks an in-scope determination and requires the
guard to fire.

**Writing that test surfaced something worse.** My first attempt backed up and restored
`withfrom_resolved.tsv` with `read_text()` / `write_text()`. `resolve_withfrom.py` writes
through `csv.DictWriter`, whose default line terminator is **CRLF** — so every self-test run
silently rewrote all 33 line endings of a committed file (5517 → 5484 bytes). And the restore
*check* also used `read_text()`, which normalises newlines on read, so it **could not see the
damage it had just done**. Backup, restore and comparison are now all in bytes, and I
confirmed `git status` is clean after a run.

That is the same failure shape as the blind detector: verify in the same representation you
mutate, or the check is blind by construction. It would have shipped as a script that quietly
dirtied the working tree on every invocation.

The first version of the test also "failed" against a guard that was working — I had asserted
the victim row existed but not that it was **in the guard's scope**, so it blanked a
`GO:0016192` donor the selector never looks at. Asserting the mutation target exists is not
enough; it has to be a row the guard would actually read.

**(1) The notes inconsistency is real and now fixed.** `ARFGEF3-notes.md` still carried the
pre-check "may be donating" while the review YAML, `RESULTS.md` and the history record had all
been upgraded — I upgraded three of four surfaces and missed the one a curator reads first.
It now states the checked result, with the node id, the single reference, and the paper's own
"*potential function*" hedge.

**(2) Recorded as entailment, not as a second finding.** You are right that the WITH/FROM-set
identity follows from node identity, since PAINT emits one descendant-evidence list per IBD.
Both `RESULTS.md` and the notes now say so explicitly — it earns its place as a consistency
check on the lookup, and calling it a second finding would have inflated one observation into
two.

Gates on `96476b09ed`, exit codes captured directly (not through a pipe): `just validate
human ARFGEF3` → `✓ Valid`; `reciprocal_donor_check.py --self-test` → 0 problems **and a clean
`git status` afterwards**; verbatim quote check → 0 over 85 PMID quotes; `file:` quote check →
0 over 5; `supporting_entities`-vs-GOA → 0; `validate-history` → no issues; `cache_lint` →
clean; `terms.csv` still adds only `GO:1904410` and deletes nothing.

No `action`, `term`, `evidence_type`, `supporting_text` or `core_functions` entry changed in
this delta — the 14-row action mix is untouched.

cmungall:
Suggestion 1 was a genuine catch and I had it backwards. Fixed.

**My guard was unreachable, and the hole you flagged was still open.** I verified your
reading rather than taking it: `describe_entry()` sets `has_sec7_interpro` to yes/no
*unconditionally* whenever an accession exists, and my guard filtered on accession
first — so on any TSV the producer actually writes, the `RuntimeError` could never fire.
The self-test only reached it by hand-blanking a cell, which should have been the tell.

The reachable hole is exactly where you pointed: an unresolvable MOD id, emitted with an
**empty accession** and `kind` `mod-id` / `protein` `UNRESOLVED`, dropped by the accession
filter before any InterPro test could see it. The guard now keys on the **`kind` column**,
which — as you noted — is precisely what separates a token that failed to resolve from a
PANTHER node that legitimately has no accession. The old check survives as an
**explicitly-labelled tripwire** against a future producer change, not as the guard; an
unreachable check that reads as coverage is worse than no check, so labelling it seemed
better than deleting it.

Added a **negative control** I was missing: an unmodified run must *not* fire. Without it,
a guard that over-fires on every PANTHER node would have looked identical to a guard that
works.

**Suggestion 2 — removed the failure class instead of hardening it.** You were right that
byte-level backup and restore was the right diagnosis but the wrong altitude.
`suspect_donors()` now takes a `tsv_path`, and every self-test mutation runs against a copy
in the `TemporaryDirectory` it was already creating. The committed file is never written at
all, plus a byte-level before/after assertion that no test mutated it. `import shutil` is
gone.

**Suggestion 3 — the controls paragraph is now one claim per bullet.** The two remaining
long lines in `RESULTS.md` are the SubCell table rows, which are quoted verbatim as `file:`
supporting_text in the review YAML; wrapping them would break those citations, so I left
them deliberately.

One self-inflicted error worth recording, since it is the same class as the CRLF bug:
backticks in my `git commit -m` message were **shell-expanded**, silently eating the word
`kind` from the very sentence explaining the fix. Caught it by reading the message back,
amended from a file before pushing. Same lesson as the line endings — check the artifact
you produced, not the command you ran.

Gates on `0a546ca8c7`, exit codes captured directly: `just validate human ARFGEF3` →
`✓ Valid`; `reciprocal_donor_check.py --self-test` → 0 problems with a clean `git status`
afterwards; main run unchanged (MON2, node `PTN008950430`, 19 vs 19, one non-IBA reference,
no IDA); verbatim quote check → 0 over 85 PMID quotes; `file:` quote check → 0 over 5;
`supporting_entities`-vs-GOA → 0; `cache_lint` → clean.

No `action`, `term`, `evidence_type`, `supporting_text` or `core_functions` entry changed —
the 14-row action mix is untouched.

cmungall:
**Declining all three, with reasons.** Each is correct — I verified the premises rather than
taking them — but none changes what a curator reading the merged YAML would do, and the cost
of another round is no longer negligible.

**Verified, so the decline is informed rather than reflexive:**

1. `reciprocal_donor_check.py:210` does call `suspect_donors()` inside item 1 with no
   `try/except`, so an over-firing guard raises there before 1e (268–276) can name it. I also
   checked what happens next: `main()` wraps `self_test()` in nothing, so the exception
   propagates to a traceback and a non-zero exit. **Your characterisation is exact — the run
   still fails loudly, only the diagnostic is lost.** The safety property is intact; what
   degrades is the error message.
2. `node_i` is indeed assigned at 270, gated at 272, and never read again — a gate on
   presence-anywhere rather than presence-in-scope, and `subject_toks` is two lines up. You are
   right that it exercises the right thing *today* only because `PTN008950430` happens to be in
   the subject set.
3. The committed-bytes assertion at 288 does precede section 3 at 291, and section 3 only
   reads. Fragile rather than wrong, as you say.

**The test I am applying:** would a curator reading the merged review act differently because
of this change? For all three, no. None touches an `action`, `term`, `evidence_type`,
`supporting_text` or `core_functions` entry; none is a factual error in committed content; and
none is prose contradicting its own artifact — you confirmed the RESULTS.md bullets and the
guard's comment both match the code, having checked against the producer rather than the
comment.

**Why that outweighs the fixes being trivial.** Moving 1e above item 1 is a two-line edit, and
I am not pretending otherwise. But `Build and test` has no `cancel-in-progress`, so every push
queues a ~20-minute build, and that queue — not review latency — has been the binding
constraint on this batch all session. Spending another build and another review cycle to
improve a self-test's error message, on a PR whose curation content has been unchanged since
`1d359fb163`, is the wrong trade. My last two rounds were already tooling rather than
curation; the unreachable-guard one I would defend, because I had just introduced that defect
in response to your previous round and shipping a knowingly-broken check converts an
unverified claim into an apparently verified one. These three are not that.

**What would change my answer:** if (1) meant the guard could *silently pass* rather than
fail loudly, it would be on my list immediately — a check that stops checking without saying
so is the defect class I have been chasing all review. It does not; I checked.

Recording them here so they are on the record rather than lost: all three are real, all three
are worth taking the next time this file is touched for a substantive reason, and (2) is the
one I would fix first, since "gate on presence rather than relevance" is the same pattern I
closed twice elsewhere in this file and it will eventually bite.

Thanks for the static verification of the guard against `resolve_withfrom.py:106`'s emitters
and the committed TSV. Confirming there is no `UNRESOLVED` token anywhere in the file is the
check that makes the finding safe from the hole the guard now covers, and it is a stronger
statement than my self-test makes — the self-test proves the guard fires, yours proves it does
not need to.

PR #3077, opened 2026-09-19
Review human ARGLU1: a Candida Mediator-tail paralog put mitochondrion on the whole ARGLU1 family

cmungall:
Complete GO annotation review of human **ARGLU1** (Q9NWB6), a small intrinsically
disordered nuclear protein that couples transcriptional coactivation to alternative
pre-mRNA splicing.

## Row reconciliation

`ARGLU1-goa.tsv` has **29** data rows; the `fetch-gene` stub seeded **28**. The missing
row is the second `GO:0005515 / IPI / PMID:30698747 / UniProtKB:P26368` row — GOA holds
two, one assigned by IntAct and one by UniProt, and the stub's seed key ignores
`ASSIGNED BY`. Restored, giving **29 reviewed GOA rows + 3 NEW proposals = 32** entries.
This is asserted programmatically by the committed audit, not by eye.

## Headline finding: the mitochondrion IBA is a *Candida* Mediator-tail paralog

ARGLU1 receives `GO:0005654` nucleoplasm **and** `GO:0005739` mitochondrion by IBA from
the **same** PANTHER node, `PTN001271790`. The node's reach is identical for both — 111
gene products across 64 NCBI taxa — so the two terms differ only in their donor pools:

| term | gene products holding it by their **own** experimental evidence |
|---|---|
| `GO:0005654` nucleoplasm | **3** — human ARGLU1 (EXP+IDA), mouse Arglu1 (IDA), *Drosophila* Arglu1 (EXP); all reviewed Swiss-Prot ARGLU1 orthologs |
| `GO:0005739` mitochondrion | **1** — *Candida albicans* **TLO16** (`A0A1D8PRM3`, **TrEMBL**), one IDA from `PMID:22923044` |

Every one of the 111 mitochondrion rows names `CGD:CAL0000179812` in WITH/FROM, so the
mitochondrial annotation of the **entire eukaryotic ARGLU1 family** — plants, green algae,
*Selaginella*, *Marchantia*, *Trichoplax*, *Daphnia*, all vertebrates — traces to that one
protein.

Two facts close the argument:

- **TLO16 is architecturally not an ARGLU1 ortholog.** InterPro gives it `IPR021017`
  *Mediator complex subunit Med2, fungi* (`PF11214`) and the AH/BAR superfamily **in
  addition to** the shared `IPR033371`. Human ARGLU1 carries `IPR033371` and nothing else.
- **Its own source paper confines the localisation to one clade of a lineage-specific
  expansion**: the TLO family went "from one or two copies in other CUG clade members to
  14 expressed copies in C. albicans", and only clade γ reaches mitochondria while clades
  α and β are nuclear.

A localisation that is not uniform across one gene family inside one species cannot
support a pan-eukaryotic node. Human ARGLU1 has no mitochondrial evidence of any kind and
four experimentally supported CC terms, all nuclear. → **`REMOVE`** with a
`propagation_review` (`SOURCE_BAD`; `WRONG_ORTHOLOG_OR_PARALOG`,
`LINEAGE_OR_TAXON_MISMATCH`, `COMPARTMENT_OR_COMPLEX_MISMATCH`).

**Divergence stated deliberately:** 213 merged reviews in this repo carry a `GO:0005739`
IBA row and their modal verdict is `ACCEPT` (127). This one removes it because here the
entire donor pool is a single architecturally divergent fungal paralog. The corpus trend
is not an argument about this node.

`suggested_questions` asks PAINT whether the term should be detached from that node — one
edit would correct 111 annotations, and the node is sound for nucleoplasm.

## Other actions

- **`GO:0045296` cadherin binding (HDA) → `MARK_AS_OVER_ANNOTATED`.** `PMID:25468996` is
  proximity biotinylation; the paper says "561 proteins **in the vicinity of** the
  cytoplasmic tail of E-cadherin". That one reference produced **272 annotations over 272
  distinct gene products**, against a control in the same measured table where the ARGLU1
  primary paper annotates **4**. ARGLU1 is nuclear in every curated location it holds.
  `MARK_AS_OVER_ANNOTATED` rather than `REMOVE` follows the modal verdict across the **38**
  merged reviews already carrying this exact row (21 over-annotated / 11 non-core / 3
  remove / 1 each undecided, accept, pending).
- **Three SRPK2 `GO:0005515` rows → `MODIFY` to `GO:0019901` protein kinase binding.** Not
  a one-screen artefact: three independent studies, three orthogonal methods. The decisive
  one is `PMID:23602568`, which IntAct types not as *physical association* but as
  **`phosphorylation`**, detected by `protein kinase assay` in vitro — ARGLU1 is recorded
  there as an SRPK2 *substrate*. That is structurally coherent: the committed composition
  analysis shows the N-terminus is a genuine **RS-repeat region** (33.8% Arg, 27.0% Ser, 12
  RS + 12 SR dipeptides, unbroken 8-dipeptide alternating run), the substrate class SRPK
  kinases act on, and three of UniProt's six phosphosites sit inside it.
- **Five U2AF2/PUF60 `GO:0005515` rows → `MODIFY` to `GO:1990935` splicing factor
  binding; the two JMJD6 rows → `REMOVE`.** This corrects an error made in the first
  version of this PR, which removed all six on the stated ground that no such MF term
  existed. It does exist. The mistake's cause is worth the sentence: the search used was
  *"spliceosomal complex binding"*, and GO's search is **token-based**, so *spliceosomal
  complex* shares no token with *splicing factor* and could never have retrieved it. Found
  by walking the ontology instead.

  Which partners the term applies to was then settled by measurement
  (`splicing_factor_eligibility.py`, annotations under `GO:0008380` RNA splicing):

  | partner | annotations under RNA splicing | verdict |
  |---|---|---|
  | U2AF2 `P26368` | 6, incl. **IDA** to `GO:0000398` | applies |
  | PUF60 `Q9UHX1` | 2, IBA only | applies, weaker support, recorded |
  | JMJD6 `Q6NYC1` | **zero** | does **not** apply — an oxygenase that hydroxylates a splicing factor is not one |
  | SRPK2 `P78362` | 4, incl. 2× IDA | applies too, but the in vitro kinase assay makes `GO:0019901` the better-evidenced call |

  So the JMJD6 rows still go, but now on a measured JMJD6-specific ground instead of a
  false general one. The `knowledge_gaps` entry is rewritten to state what is *actually*
  missing — `GO:1990935` is flat, so which step of splice-site recognition the contact
  serves remains inexpressible — and to keep the retracted claim visible rather than
  quietly deleting it.
- **Everything else `ACCEPT`**, including the self-referential nucleoplasm IBA (a marker
  that experimental grounding exists on the target, not circularity).

## Three NEW annotations, each filling a measured coverage gap

| term | evidence | why |
|---|---|---|
| `GO:0036033` mediator complex binding | IPI, MED1 `Q15648` | `PMID:21454576`'s **title finding** — direct binding to the MED1 far C-terminus, plus ChIP-reChIP co-occupancy — produces **no MF annotation anywhere in GOA**; that reference appears in ARGLU1's record only as support for two CC rows |
| `GO:0048025` negative regulation of mRNA splicing, via spliceosome | IMP (human HeLa/HEK293/MCF7 incl. a CRISPR protein-null line) | GOA carries only the **unsigned** `GO:0000381`; the autoregulatory circuit is specifically *inhibitory* on its own host transcript |
| `GO:0042921` nuclear receptor-mediated glucocorticoid signaling pathway | IMP (human HEK293 arm) | the pathway is the subject of the primary paper's title and is unrepresented |

## Restraint worth flagging

- **`GO:0030374` nuclear receptor coactivator activity is obsolete** (QuickGO
  `isObsolete: true`). It is exactly the term a reader would reach for, since ARGLU1's
  best-characterised activity is nuclear-receptor-specific. `GO:0003713` is therefore
  already maximal — `ACCEPT`, and **no** `proposed_new_terms` entry to re-create it.
  `GO:0035259` is likewise obsolete and `GO:0016922` has no children.
- **No GR-binding term proposed.** The paper states ARGLU1 "interacts with GR through its
  C-terminal domain", but the support is coactivation assays and endogenous IP-MS, not a
  binding assay with purified proteins. Filed as a question instead.
- **No DNA-repair term proposed.** Those phenotypes rest on adenoviral **overexpression**;
  the knockdown readout in the same study was growth arrest, not a repair defect.
- **No pausing/elongation term proposed.** `GO:0160239` has no children and no
  regulation-level term beneath it; the elongation-level terms would misstate the
  measurement. Recorded as an `ONTOLOGY` knowledge gap instead.
- **The gene name is a trap, and the review says so.** The complete signature content of
  the Swiss-Prot entry is `IPR033371` + `PF15346` (typed by InterPro as a **coiled coil**,
  not a domain) + `PTHR31711:SF1`. There is no fold to reverse-engineer an MF from —
  *"lacks any known structural or functional domains"* (`PMID:38520408`). Every MF here
  rests on an experiment.

## Provider (affinage) assessment

`gates_passed: true`, 11 citations, all numeric PMIDs. Two defects found:

1. **It cites a RETRACTED paper unflagged.** `PMID:35082911` is `Retracted Publication`
   with `RetractionIn: PMID:37503407`, listed as a "Medium confidence" dated finding with
   no flag. Recorded in `references` with `is_invalid: true`; nothing rests on it, and the
   committed audit asserts it is never cited as supporting evidence.
2. **Recall.** PubMed returns **28** records for `ARGLU1`; affinage returned 11. The
   substantive miss is `PMID:42641889` (2026), identifying ARGLU1 as a **PAK1 substrate
   phosphorylated at Ser-77** — the same residue five phosphoproteomic studies report in
   UniProt. Not in GOA either.

(Separately, `PMID:39251607` — a GOA IPI reference — carries an erratum, `PMID:39468017`.
Noted; nothing depends on its scope.)

## Committed analyses

All under `genes/human/ARGLU1/ARGLU1-bioinformatics/`, with `README.md` and `RESULTS.md`:

| script | question |
|---|---|
| `panther_node_audit.py` | what node `PTN001271790` propagates, to whom, on whose evidence |
| `reference_scope_audit.py` | entity counts per cited reference — screen or finding |
| `intact_partner_audit.py` | canonicality, topology and method-independence of the four IPI partners |
| `composition_and_features.py` | what is actually in the sequence, and what fold evidence exists |
| `sibling_row_verdicts.py` | how merged reviews resolved the rows ARGLU1 shares |
| `splicing_factor_eligibility.py` | which IPI partners qualify for `GO:1990935` |
| `audit_arglu1_review.py` | invariant checks over the finished review (16 guards, `--self-test`) |

**Negative results reported, not omitted.** The partner audit's canonicality and
topological-plausibility checks both came back clean — all four partners are reviewed
Swiss-Prot entries at canonical length, and all four are nuclear, with SRPK2 itself in
nuclear speckles. The partner set reflects ARGLU1's biology, not a screen's bait panel.
That is the opposite of the ACRV1/ACRBP pattern and is worth knowing was checked.

## Notes for reviewers

- **The `core_functions` "No core functions defined" style warning is absent**, but one
  warning stands: *"No annotations reference available deep research files"*. That is
  deliberate. The campaign rule is never to quote provider prose as `supporting_text`, so
  the affinage record is assessed in the notes and in `reference_review` rather than cited
  as evidence. Inventing a citation to silence the warning would be the defect.
- **Pre-write hook limitation encountered and diagnosed** (not a defect in this PR): the
  hook copies the YAML to a temp directory and validates there, but
  `conf/reference_validator_config.yaml` sets `cache_dir: publications` **relative**, which
  resolves from the YAML's own directory. A temp dir has no `publications/`, so the
  validator silently falls back to a PubMed abstract fetch and reports every **full-text**
  quote as "not found". Reproduced minimally: byte-identical content is `✗ Invalid` in a
  temp dir and `✓ Valid` in-repo. This means the hook structurally cannot validate a
  full-text quote for any gene. Same class as the `--no-goa` flag it already carries for
  the missing sibling `goa.tsv`; the fix would be to pass an absolute cache dir or write
  the temp file inside the gene directory. All quotes here are verified by the
  authoritative in-repo `just validate`, which is green.


cmungall:
Thanks — I checked all three premises independently before acting, and all three were right. Fixed in `d50c4ca`.

## 1. `GO:0035259` — you're right, and the *cause* is worth recording

You're correct twice over: `GO:0035259` is active, **and** `GO:0016922` has children — eight of them, with `GO:0035259` among them. My advice would have sent a curator to a needlessly general term.

What makes this worth more than a one-line fix is that **QuickGO is the outlier**, and I had "verified" the claim entirely from it:

| source | `GO:0035259` | children of `GO:0016922` |
|---|---|---|
| GO API (`api.geneontology.org`) | active, `replaced_by: null` | — |
| OLS4 | `is_obsolete=False` | **8** |
| repo `cache/ontologies/go.tsv` | label has no `obsolete` prefix, `is_obsolete=False` | — |
| **QuickGO** | **`isObsolete=True`**, name prefixed `obsolete` | **0** |

Two things I got wrong beyond the fact itself:

- **I thought I had cross-checked "no children" two ways. Both ways were QuickGO** — its `/children` endpoint and its text search. Two methods against one service is one check. That is the same shape as the `GO:1990935` error earlier in this PR, as you noted, and the generalised lesson is now in the notes: for a load-bearing term-status claim, use two *independent services*, not two endpoints.
- **QuickGO silently resolves merges.** `GO:0035257`/`GO:0035258` return `GO:0016922`'s record with `isObsolete=False`, where OLS4 reports them obsolete-replaced-by. So QuickGO's obsolescence answers can't be read at face value in either direction. Worth knowing repo-wide, since the brief already recommends QuickGO as the authority for exactly this question.

The narrower claim that `GO:0035257`/`GO:0035258` were absorbed into `GO:0016922` does hold, and the `GO:0030374` obsolescence claim is unchanged.

## 2. Verbatim Greek — fixed

Both sites now read `clade α and clade β … clade γ`, and I verified each `PMID:22923044` quote is a normalised substring of the cached source rather than trusting the validator: 4/4 verbatim.

One correction to the mechanism, offered only because it affects how much the gate can be relied on elsewhere: this passed because `SupportingTextValidator.normalize_text` **spells Greek letters out** (`α` → `alpha`), so my transliteration normalised to the identical string and matched legitimately — not via the abstract-only ERROR→WARNING downgrade. That downgrade is real, it just wasn't load-bearing here. The fix is right regardless: the repo's convention is verbatim, and "passes because normalisation happens to agree" is not a reason to keep a paraphrase.

## 3. JMJD6 — you're right, and it changed the measurement

This was the most useful of the three. You're correct that `goUsage=descendants&goUsageRelationships=is_a,part_of` cannot reach regulation terms, because GO relates regulation to its target by `regulates`. Widening the query shows **JMJD6 does hold `GO:0048024` regulation of mRNA splicing, via spliceosome, by IMP** — so "zero annotations anywhere under `GO:0008380`" was true of my query and false as written.

`splicing_factor_eligibility.py` now queries the regulation branch too and emits a `splicing_regulation_scope` block plus a `regulates_splicing_only` flag; the widened numbers are in the committed JSON:

| partner | under `GO:0008380` (is_a/part_of) | regulation branch |
|---|---|---|
| U2AF2 | 6, incl. IDA | IDA ×2 (`GO:0033120`, `GO:0048025`) |
| PUF60 | 2, IBA only | IBA/IEA (`GO:0000381`) |
| **JMJD6** | **0** | **IMP** (`GO:0048024`) |
| SRPK2 | 4, incl. IDA ×2 | TAS |

`REMOVE` stands, but now on the distinction that actually survives — JMJD6 is annotated to the *regulation* of splicing, not to splicing — and the review states plainly that whether a regulator satisfies "involved in the process of removing sections of the primary RNA transcript" is **not settled by the term's definition**, so the term is withheld rather than asserted.

## 4. PR description

Already current — it was updated at `03:05:43Z`, about eleven seconds after your review was submitted at `03:05:32Z`, so you read the pre-update body. It now describes the five MODIFY / two REMOVE split and carries the `GO:1990935` table; I've re-read it back from the API to confirm the retracted phrasing is absent.

## Also fixed, unprompted

SRPK2 passes the *same* RNA-splicing criterion I used to admit U2AF2 and PUF60 (2 × IDA), so the review was applying its own criterion inconsistently without saying why. The SRPK2 rows now state it: both terms are defensible for that partner, and `GO:0019901` is preferred because the measured relationship is catalytic and the kinase term carries that content where the splicing-factor term would discard it.

`just validate human ARGLU1` is green, and the committed audit (16 guards) and its mutation self-test both pass.

cmungall:
Thanks for the approval. One post-approval push (`fd6d233`), and I want to be explicit about why, since it resets your verdict.

**It changes no curation content** — no `action`, no `term`, no `evidence_type`, no `supporting_text`, no `core_functions` entry. Diff is three files: two new scripts under `ARGLU1-bioinformatics/` plus their README entries, 416 insertions, nothing modified.

By the "would a curator act differently?" test this should have been a comment. I pushed anyway for one reason: both scripts are **checks whose results this PR already relies on**, and an uncommitted check is not a check. Leaving them local would strand them when the branch is deleted, and would leave two claims in my previous reply with no committed provenance.

### `verify_authored_terms.py` — guards the class, not the two instances

Your `GO:0035259` catch and my earlier `GO:1990935` miss were the same defect: trusting one service. A fix for each instance wouldn't stop the third, so this checks **every** GO id the review *authors* — `core_functions`, `proposed_replacement_terms`, and the `term.id` of each `action: NEW` row — against **QuickGO and OLS4 independently**, failing on obsolescence *or* on disagreement between them. GOA-sourced ids are deliberately excluded, per `CLAUDE.md`. All 12 authored terms currently pass on both.

One reusable finding worth flagging beyond this gene, since the campaign brief names QuickGO as the authority for exactly this question: **QuickGO silently resolves merges.** `GO:0035257`/`GO:0035258` return `GO:0016922`'s record with `isObsolete=False`, where OLS4 reports them obsolete-replaced-by. So a QuickGO "not obsolete" means either *current* or *merged away*, and the response cannot distinguish them.

### `verify_quotes.py` — your item 2 pointed at something wider than the one quote

Chasing the Greek-letter fix, I checked what the gate actually does. For an **abstract-only** cache, `validator.py` downgrades a non-matching `supporting_text` from ERROR to **WARNING** (the `declared_unavailable or cache_has_full_text is False` branch — the lines you cited). **Six** of this review's references are abstract-only, and one of them is `PMID:22923044`, the source the entire mitochondrion finding rests on. A paraphrase against any of them would have passed the build.

So all 61 quotes are now checked against their cached publications using the validator's own normalisation — agreeing with the gate on what counts as a match, while failing loudly where the gate only warns. **61/61 verbatim.** `--self-test` corrupts a quote against a full-text source *and* against an abstract-only source and asserts both are caught; the abstract-only case is exercised separately because that is the branch that weakens.

### One correction to your item 2's mechanism

Offered only because it affects how far the gate can be trusted elsewhere, not to relitigate the fix — which was right and is done. The transliterated quote passed **because `SupportingTextValidator.normalize_text` spells Greek letters out** (`α` → `alpha`), so my version normalised to a byte-identical string and matched legitimately. Verified directly:

```
normalize('α-catenin') = 'alpha catenin'
normalised equal?  True
```

The abstract-only downgrade you cited is real — I confirmed it at `validator.py:156` and it is exactly what motivated `verify_quotes.py` — it just wasn't what let this particular quote through. The fix stands regardless: "passes because normalisation happens to agree" is not a reason to keep a paraphrase.

All gates green on the merged-with-main tree: `just validate human ARGLU1`, the 16-guard audit and its mutation self-test, 61/61 quotes, 12/12 authored terms. No further pushes planned.

cmungall:
The 🟡 is correct, and the inversion is exact. Fixed in `68c5490`, along with all four 🔵. Curation content is byte-identical to `d50c4ca`.

## 🟡 The collector saw 61 of 83 — and missed precisely the 22 it existed for

I verified the census independently before touching anything:

```
'references:' at line 1298
total supporting_text keys : 83
  above references:        : 61
  below references:        : 22
collector (same-dict rule) sees: 61
references[].findings[] quotes : 22  (of which carry reference_id: 0)
arithmetic: 61 + 22 = 83  vs raw 83
```

And the path-template claim holds: the downgrade loop iterates
`reference.get("findings", [])` with template `references[{i}].findings[{j}].supporting_text` (`validator.py:124-130`), and `supporting_text.py:14-26` documents the split in as many words. So the script covered the 61 the strict CLI already gates and missed every one of the 22 gated by the branch it cited as its whole reason for existing.

**One number sharpens your diagnosis: 8 of the 22 missed quotes are on abstract-only sources** — `PMID:21454576` ×2, `PMID:22365833`, `PMID:22923044` ×2, `PMID:23602568`, `PMID:36533631`, `PMID:42641889`. That is the weakly-gated subset, entirely unchecked, in a file committed as the guard for exactly it.

Two structural fixes rather than a patch, since a future shape change has to survive them:

1. the collector **inherits the parent `references[].id`** when walking into `findings`, and records which shape each quote came from;
2. the collected count is **asserted equal to the raw `supporting_text:` key count** — a collector blind to part of the document now fails loudly instead of under-reporting. The docstring says: fix the walk, never the assertion.

Now `83/83`, split `{'same-dict': 61, 'parent-id': 22}`, all verbatim.

You were also right that the old `--self-test` abstract-only arm passed by finding the `:717` copy first. It now corrupts a quote on the **findings path against an abstract-only source** specifically, plus the `supported_by`/full-text path, plus a crippled-document case that must trip the coverage invariant — and each case requires the problem to name **that path and that gate**, not merely to be non-empty.

## 🔵 All four taken

**A.** Matching now goes through `build_supporting_text_validator()`, so "the same matcher the validator uses" is literally true and picks up `literal_bracket_patterns` / `skip_prefixes` from the config. It exits loudly if the dependency is missing rather than silently degrading to a hand-rolled substring test — which is how the first version drifted from the gate it claimed to mirror.

**B. You're right that it guarded a class, just not the one that failed** — and this one deserves more than a fix. I added the `GO:\d{7}` free-text sweep you suggested, and it immediately surfaced the real ids in `suggested_questions`. But I then tried to make it *adjudicate* prose ("does the review's claim about this term match the services?"), and **withdrew that**, because attributing an English status phrase to a particular id by proximity produced false positives in both directions on this very document:

- *"a bare `GO:0005515` row **is replaced by** a more informative term"* → flagged as an obsolescence claim about `GO:0019901`, two clauses later;
- *"`GO:0035257` and `GO:0035258` were **absorbed into** `GO:0016922`"* → flagged as one about `GO:0016922`, which is active.

A guard that cries wolf on correct content gets switched off, which is worse than not having it. So prose ids get their status **surfaced as advisory** for a human to check against the sentence, the summary prints the structured count and the prose count separately, and the README now **retracts** the "guards the class" claim and explains precisely how far the code goes. Current output: `PASS: all 12 term(s) used in STRUCTURED slots are current…` / `10 further id(s) appear only in prose… NOT adjudicated by this script`.

**C.** Both scripts now emit JSON — `quote_verification.json`, `authored_terms.json` — so `83/83` and `12/12` have committed provenance instead of living only in a console line.

**D.** Declared labels are now compared against both services, so a real-but-wrong id with a plausible hand-written label no longer passes an existence check (which, as you note, leaves `proposed_replacement_terms` uncovered by `linkml-term-validator`). And the self-test's detection disjunct is gone — each case names its path and gate.

## One advisory worth your attention

The prose sweep makes something visible that I had only mentioned in passing: **QuickGO and OLS4 disagree on `GO:0035257`, `GO:0035258` and `GO:0035259`**, in opposite directions. QuickGO silently resolves merges — `GO:0035257/8` return `GO:0016922`'s record with `isObsolete=False` where OLS4 reports them obsolete-replaced-by — while `GO:0035259` is the reverse. So a QuickGO "not obsolete" means either *current* or *merged away*, indistinguishably from that response alone. Given the campaign brief names QuickGO as the authority for exactly this question, that seems worth knowing beyond this gene.

All gates green: `just validate human ARGLU1`, the 16-guard audit + self-test, `83/83` quotes + self-test, `12/12` structured terms.

PR #3078, opened 2026-09-19
Review human ARHGAP6: two GAP-independent arms, and three curated annotations GOA has lost

cmungall:
Reviews all 44 GOA rows for human ARHGAP6 (UniProtKB:O43182), adds three `NEW` rows, and commits three reproducible analyses with their results.

Deep research provider: **affinage**, trust gates clear (accession matches the local UniProt record, organism human, `self_evaluation_pairwise: win`, `faith_pct: 100.0`).

## The separability question

ARHGAP6 has a RhoA-GAP arm and a GAP-*independent* actin/pseudopodia arm, separated by the R433G catalytic mutant. Two papers **affinage does not cite** show there is a third arm, also GAP-independent, and both are missed for the same reason — the title does not name the function:

| PMID | Why a symbol-driven search misses it |
|---|---|
| **18434237** | Title misspells the gene **"ARGHAP6"** and leads with "phospholipase C-delta1" |
| **19038263** | Titled for the partner ("novel HERG regulator") and for *C. elegans*; spells the gene out in full |

A EuropePMC `TITLE:"ARHGAP6" OR TITLE:"ARGHAP6" OR TITLE:"RhoGAPX-1"` search returns 17 records and **does not contain PMID:19038263**. Neither paper is in UniProt's `RN` list either. Both were found by searching the *function* and by following UniProt's `DR GO` block.

They matter: ARHGAP6 binds and activates PLCδ1 with purified proteins (PMID:18434237), and an independent lab shows a GAP-independent ARHGAP6 effect that is abolished by PLC inhibition and survives both a dominant-negative Rho and the GAP-dead mutant (PMID:19038263).

## The actin core function carries no `molecular_function`

The GAP-independent actin role is defined by what survives R433G plus co-localisation and F-actin recruitment in transfected cells. No direct binding, nucleation, bundling or capping activity has been measured. Attaching the GAP function would contradict the evidence that defines the arm; inventing a structural MF would assert a mechanism nobody has measured. The optional slot is **left empty** and the entry says why. This is a claim about the evidence, not about the ontology — no "no term exists" argument is made or needed.

## Arginine finger: checked, with controls, and it settles less than it looks

`ARHGAP6-bioinformatics/check_arginine_finger.py` → `RESULTS.md`

- Anchor **verified, not asserted**: PDB **1TX4** chain A Arg85 and UniProt Q07960 R282 share an identical ±6 window (`TTEGIFRRSANTQ`); a mismatch exits non-zero rather than reporting a number.
- **ARHGAP6 RETAINS it at R433** — alignment and UniProt's SITE feature agree (8/8 panel members agree), and R433 is the residue PMID:19038263 mutates.
- **And the panel shows retention proves nothing**, because residue and curated activity are decoupled *both ways*: ARHGAP11B keeps the arginine and carries `NOT|enables GO:0005096`; ARHGAP36 (closest paralog, shares IPR037863) has **T258** and OCRL **Q757**, and both carry positive rows.

What supports the GAP call is the functional mutant, not the residue — and because R433G is a clean loss-of-GAP reagent, it is also what makes the GAP-independence claims arguments from a working mutant.

## Three BHF-UCL annotations are in UniProt and missing from GOA

`check_goa_uniprot_divergence.py` → `RESULTS-goa-divergence.md`

`GO:0016004`, `GO:0043274` and `GO:0141214` are in UniProt's cross-references and absent from QuickGO/GOA. **Not obsolescence** — all three are active on OLS4 *and* `api.geneontology.org` (two independent services, since these are known to disagree). And they are still load-bearing: GOA serves **six** ISO rows on mouse and rat Arhgap6 whose WITH/FROM names `O43182` as the source for exactly these terms, so those projections are orphaned.

Because the workflow seeds `existing_annotations` from the GOA TSV, these would never be reviewed at all. Entered as `NEW` — not as new curation, but as pre-existing curation that fell out of the feed. Source: PMID:18434237.

## Reactome reaches a true term by an unsupported route

`enables GO:0005096` TAS from **R-HSA-9018806 "RAC3 GAPs stimulate RAC3 GTPase activity"** puts ARHGAP6 in the tier "shown to bind RAC3 **and stimulate its GTPase activity**" on one citation, PMID:26628301, whose only Rac3 experiment is a co-immunoprecipitation.

Action is **ACCEPT, not REMOVE or MARK_AS_OVER_ANNOTATED**: `GO:0005096` is substrate-agnostic and ARHGAP6 genuinely enables it for RhoA, so the row's assertion is true; only the route is unsupported, and unsupported is not contradicted. Recorded in `review.reason` and `propagation_review` (`SOURCE_WEAK_OR_INFERRED` / `SOURCE_EVIDENCE_WEAK`).

GO cannot express the distinction: `GO:0005096` has **zero descendants** on OLS4 and on `api.geneontology.org`; QuickGO's one child `GO:1902773` is a *cellular component* reached by **`capable_of`**. Raised as a `suggested_questions` entry instead.

## Half the annotation set is one interaction

`check_pdz_interactome.py` → `RESULTS-pdz-interactome.md`

22 of 44 rows are `GO:0005515` IPI from one holdup assay. **22 of 22** partners carry PDZ domains (parsed from the repo's own GOA TSV, not hardcoded), and ARHGAP6 ends in `...LPETLV`, a class I PDZ-binding motif. One determinant measured once → all `KEEP_AS_NON_CORE`.

## Deliberately not proposed

The HERG/IKr result is well controlled (overexpression *and* shRNA knockdown in atrial myocytes) but is one lab, largely heterologous, and routed through PLC rather than exerted on the channel. Recorded as a finding, described in the PLC core function, and raised as a question/experiment rather than annotated.

## Action summary

| action | rows |
|---|---|
| ACCEPT | 13 |
| KEEP_AS_NON_CORE | 30 |
| MARK_AS_OVER_ANNOTATED | 1 (`GO:0030041` — the paper shows recruitment of existing F-actin, not monomer addition) |
| NEW | 3 |

Zero `REMOVE`.

## Review round 1 (all items adjudicated)

Taken: the HERG result is now **6 real `findings` entries** rather than a prose claim (4 on PMID:19038263, 2 on PMID:18434237); the `GO:0141214` reason no longer conflates the 6-fold PLC rise in transfected cells with the IP3/DAG correlation in hypertensive patients' mononuclear cells; the four cached-but-uncited papers each carry a `reference_review` and a notes §7 entry; `check_pdz_interactome.py` no longer hardcodes the denominator `44`; the "decoupled both ways" wording now distinguishes the strong case (ARHGAP11B's curated `NOT`) from the weak one; and `GO:0007266` TAS gained a `supporting_text`.

Declined with reasons: the claim that Q86UT5's HGNC symbol is PDZD3 is **refuted** — HGNC:19891's approved symbol is `NHERF4` and `PDZD3` is listed under `prev_symbol` (checked on UniProt's HGNC xref and on `rest.genenames.org`); the reviewer accepted this. The ACCEPT/KEEP_AS_NON_CORE split is per-term core-vs-non-core, not per-row redundancy, and a per-row split is unrepresentable because the validator requires one action per term (reviewer verified against `validator.py:544-680`). The one sentence of schema commentary in `core_functions[1].description` is kept deliberately: an absent optional slot is otherwise indistinguishable from an unfinished one.

## Review round 2

The blocking item was the unfixed half of round 1's item 2, and it was right: the `GO:0141214` *reason* had been corrected but the two **descriptions** still elided the same two experiments — `core_functions[2].description` said IP3 and DAG "rise several fold" (they are 1.6× and 2.3×, and were not measured in the transfected cells), and the top-level `description` asserted ARHGAP6 *raises* them where the paper says "was associated with". Both now separate the overexpression measurement from the patient correlation, so they agree with that entry's own `knowledge_gaps.boundary`.

Also taken: the PMID:18434237 finding keeps its "in vitro" qualifier; the R433G finding now says it quotes PMID:19038263 because that is where the *position* is stated while the result is Prakash et al. 2000's; `check_review_consistency.py`'s coverage rule no longer uses a declared PMID literal but derives the set from the deep-research file's own citations intersected with `publications/` — the case that actually occurred, and one a literal could never catch, since a literal only lists papers someone already remembered; and notes §8 now documents how to re-run all four scripts, because a committed check that nothing references re-runs no more often than an uncommitted one.

**A correction to what this PR said earlier.** An earlier comment described "three new build guards". Those assertions live in the generation script used to author the YAML, which is *not* committed, so no reader of this repository can run them and nothing in a build enforces them — the wording misattributed them. The durable versions of all three (cached-publication coverage, `reference_review` on every reference, and the findings-count claim) are in the committed `check_review_consistency.py`, which is where the enforcement actually is.

## Verification

- `just validate human ARHGAP6` → **✓ Valid**, no warnings. `just validate-changed …` (what CI runs) → **✓**.
- The validators were **mutation-tested** rather than trusted on a clean pass: falsifying a `supporting_text` and a reference title each made `validate-changed` fail with the expected message, and the file was restored byte-identically afterwards.
- All four analysis scripts ship `--self-test`; every guard fires with its expected message and every negative control stays silent. Two guards were *narrowed* during this work after firing on false positives (a fixed WITH/FROM control term that condemned a working probe, and a text scan that fired on the docstring documenting the defect's removal); both narrowings are mutation-tested in **both** directions.
- `check_review_consistency.py` re-tests all **27** `file:` quotes as verbatim, exactly-once matches — the thing CI does not check — plus the counted claims this PR and the notes make.


cmungall:
Thanks — items 1–3 were all correct and all three were real defects, not stylistic. Fixed in `e56f76c2db` and `4ece4f9b7d`, along with suggestions 4, 7 and 8. Declining 5, 6 and 9, with reasons below; 9 is refuted.

## Taken

**1 — four uncited cached papers.** Confirmed: `grep` for those PMIDs across the review YAML and notes returned 0 before the fix. Each now has a `reference_review` with `relevance: LOW` and a note saying why no annotation follows, and notes §7 tabulates them. I also added two build guards so this class of omission cannot recur silently: one fails if a PMID cached by this PR is absent from `references:`, and one requires **every** reference to carry a `reference_review` — which promptly caught four `file:` references I had also left unadjudicated, so those now carry reviews recording that their quotes were grep-verified (the thing CI does not check).

**2 — the `GO:0141214` conflation.** You are right, and re-reading the abstract confirms the split: the ~6-fold rise is `"The activity of PLC in cells overexpressing ARHGAP6 increased approximately 6-fold compared to control cells."`, while IP3 (1.6×) and DAG (2.3×) belong to `"Enhanced expression of ARHGAP6 was associated with an elevated level of PLC activity..."` in mononuclear cells from hypertensive patients. The `reason` now names the overexpression result as the basis of the IDA and says explicitly that the patient correlation **is not part of the support for the term**. The correlative half is also now a `finding` on PMID:18434237 in its own right, so it is recorded without being load-bearing.

**3 — "recorded as a finding" with `findings: []` everywhere.** This was exactly the failure mode of asserting something about the repo that the repo did not do. Added 4 `findings` on PMID:19038263 (GAP-independence, the shRNA arm in atrial myocytes, the surface-density mechanism, and the R433G identification) and 2 on PMID:18434237. All six `supporting_text` values were validated before commit. The build now fails if the claim is made with zero findings present.

**4 — the asymmetry.** Fair, and the YAML was weaker than `RESULTS.md`'s own caveat. Reworded: ARHGAP11B (residue intact + two curated experimental `NOT|enables` rows) is the strong case; ARHGAP36's IBA and OCRL's disputed FlyBase IDA are now explicitly framed as facts about the **annotation record** — a positive GAP annotation can sit on a residue-dead protein — rather than as evidence those proteins are active. Your reading that this half is about GO over-propagating to pseudo-GAPs is the better description and is what the text now says.

**7 — the hardcoded denominator.** Confirmed and fixed: `partners_from_goa()` now returns `(accessions, n_binding_rows, n_total_rows)`, both counts parsed from the TSV, and the `44` is gone from both the docstring and `render()`. "22 of 44" was right only by coincidence — every binding row happens to name exactly one partner. The results file regenerates to "22 of ARHGAP6's 44 GOA rows (both counted from the TSV, neither hardcoded)". The title no longer hardcodes 22 either.

**8 — no `supported_by` on `GO:0007266` TAS PMID:9417914.** Added, quoting the rhoGAP-family identification from the cloning abstract.

## Declined

**9 — "prose calls Q86UT5 NHERF4; the HGNC symbol is PDZD3."** This one is backwards, and I checked it on two independent sources before declining:

- UniProt's own HGNC cross-reference on Q86UT5: `HGNC:19891 {'GeneName': 'NHERF4'}`
- HGNC REST (`rest.genenames.org/fetch/hgnc_id/19891`): `approved symbol: NHERF4`, `name: NHERF family PDZ scaffold protein 4`, and **`prev_symbol: ['PDZK2', 'PDZD3']`**

So PDZD3 is a *previous* symbol and NHERF4 is the approved one. The script takes the symbol from UniProt's gene-name field rather than typing it, which is why it is right. **What would reverse me:** an HGNC record showing PDZD3 as the approved symbol, or a repo convention of preferring previous symbols — I could not find either.

**5 — ACCEPT vs KEEP_AS_NON_CORE "inconsistent on redundancy".** I do not think redundancy is the criterion being applied, so there is no inconsistency to close. The criterion is core vs non-core, and it is applied per *term*: `GO:0015629` is a core location for this protein (it is where both arms act), so all three of its rows are ACCEPT; `GO:0005829`, `GO:0007165` and `GO:0051056` are non-core, so all their rows are KEEP_AS_NON_CORE. Redundancy is a remark I make about individual rows, not the thing deciding the action — and it could not be, because the best-practices validator requires all rows of a term to share one action, so a per-row redundancy criterion is unrepresentable. Adding a clause saying "redundant but core" would describe a distinction the schema cannot record. **What would reverse me:** if the validator permitted divergent actions within a term on a redundancy basis (it has escape hatches only for `GO:0005515`, per-reference IPI rows, and citation-specific `REMOVE`), a redundancy-driven split would become expressible and worth making.

**6 — `core_functions[1].description` ends in schema meta-commentary.** I agree it is about the file rather than the protein, but I think that sentence is load-bearing precisely because of that. An absent optional slot is indistinguishable from an unfinished one: a curator reading the merged YAML has no way to tell "no molecular function is asserted, deliberately, because none has been measured" from "this entry was left half-filled". The `knowledge_gaps` entry states the biology (the activity is unidentified) but not the curation decision that follows from it. Note also that the repository's prohibition on curation commentary is scoped to the top-level `description` field, which here is clean biology. **What would reverse me:** a schema slot that records "deliberately not asserted" for an omitted field — then the prose would be redundant and I would drop it — or a repo convention placing that statement in `knowledge_gaps.boundary`, which I would happily follow if pointed at an example.

## Note on your runner

You flagged that you could not run `just validate` (no `just`/`uv` on PATH). For what it is worth, I ran `just validate-changed` — what CI runs — and **mutation-tested it** rather than trusting a clean pass: falsifying one `supporting_text` and one reference title each made it fail with the expected message, and the file was restored byte-identically afterwards. That is the check your box was relying on CI for.

cmungall:
One follow-up commit (`a2da047a5a`), prompted by your own observation rather than by a review item.

Your notes point out that you spot-confirmed the `RESULTS.md` quotes because `supporting_text` against `file:` references is **not** CI-checked. My PR body then claimed all of them were verified — but that verification was a one-time manual `grep`, which is the same shape of defect as your item 3: a sentence asserting something about the repo that nothing in the repo re-tests.

So `check_review_consistency.py` is now committed alongside the three analyses. It re-tests, on every run:

- all **27** `file:` `supporting_text` quotes occur in the named file **exactly once**, verbatim (0 matches = fabricated, 2 = ambiguous);
- every `file:` reference's `title` matches that file's own first heading;
- every reference carries a `reference_review`, and every PMID cached for this gene appears in `references:`;
- the counted claims the prose makes — 4 findings on PMID:19038263 and 2 on PMID:18434237, the four papers excluded on purpose being explained in the notes, and the deliberate **absence** of a `molecular_function` on `core_functions[1]`;
- no hardcoded row denominator survives in live code in `check_pdz_interactome.py`.

`--self-test` mutation-tests each guard: a fabricated quote, a wrong title, a stripped `reference_review`, emptied findings, a reinstated MF and a reinstated hardcoded `44` are each caught, and the real document stays silent.

Worth flagging one thing that came out of building it. My first version of the denominator guard was a substring scan, and it **fired on the docstring that documents the defect's removal** — the prose mentions `n_goa_rows` and `"22 of 44"` precisely because they are gone. Narrowing a guard can open a hole, so the replacement parses the module and ignores docstrings, and it is mutation-tested in both directions: it must still catch a genuinely defective snippet, and must stay silent on one whose docstring merely describes the defect.

Current run output:

```
consistent: 27 file: quotes verbatim and unambiguous, 24 references adjudicated, counted claims hold
```

No GO term, action, evidence code or `supporting_text` changed in this commit.

cmungall:
Round 2 addressed in `3ea458360e`. The blocking item was right and I had fixed only half of it.

## The blocking item

You are correct that the round-1 fix landed in the `GO:0141214` `reason` and did not propagate to the two descriptions. Verified both before editing:

- `core_functions[2].description` read *"and in cells PLC output and the IP3 and DAG it generates rise several fold"*. Two errors in one clause: 1.6× and 2.3× are not "several fold", and they are not from the transfected cells at all. It also contradicted that entry's own `knowledge_gaps.boundary`, which already had it right — the kind of internal disagreement that is worse than either statement alone.
- Top-level `description` read *"raising cellular PLC output and IP3 and DAG levels"*, asserting causation where the abstract says *"Enhanced expression of ARHGAP6 **was associated with** an elevated level of PLC activity and increased levels of IP(3) (1.6-fold) and DAG (2.3-fold)"*.

Both now keep the six-fold PLC rise as the overexpression measurement and give the IP3/DAG elevations their own clause, named as a correlation in mononuclear cells from hypertensive subjects.

## Non-blocking items, all taken

**"Nothing will cause `check_review_consistency.py` to run again."** Fair, and the sharper version of my own argument — I committed it because an uncommitted check never re-runs, then left it unreferenced, which achieves the same nothing. Notes §8 now documents the command for all four scripts with their expected output. I did not add a `justfile` target or CI wiring: there is no per-gene script-runner convention in this repo and inventing one from a gene PR is the wrong place for that change.

**"The 'build guards' framing applies to no build."** This is the same defect as your item 3 and I should have caught it in my own comment. Those three assertions live in the generation script I used to author the YAML, which is **not committed** — so no reader of this repo can run them and nothing in a build enforces them. The wording misattributed them. The durable versions of all three (cached-publication coverage, `reference_review` on every reference, the findings-count claim) are in the committed `check_review_consistency.py`. I have corrected the PR body accordingly rather than leaving the claim standing.

**"`CACHED_PMIDS` is a declared literal, so it catches a dropped citation but not the newly-cached-and-uncited case."** The decisive objection, and it makes the guard almost worthless for the defect it was written for — a literal can only ever contain papers someone already remembered to reference. Replaced with a derived set: the PMIDs the deep-research file itself cites, intersected with what exists in `publications/`. That is exactly the shape of what happened.

The scope is deliberately the deep-research file rather than the whole gene folder, and the reason is a false positive I hit while building it: the notes legitimately discuss third-party PMIDs about *other* proteins (ARHGAP11B's `NOT` annotations, for instance), and `PMID_25721503.md` and `PMID_27957544.md` **do** exist in `publications/`, cached by unrelated PRs. A folder-wide rule would have demanded those appear in ARHGAP6's references. Mutation-tested in both directions: dropping a derived paper is caught, and the derived set must exclude the 4 papers affinage cites that nobody cached.

**"The PMID:18434237 finding drops the 'in vitro' that `e6a4e89f4` was specifically about."** Restored — the statement now opens with it.

**"The R433G finding sits on the paper that cites it rather than the one that established it."** Right about the attribution, and I kept the quote where it is because the *position* is only stated in PMID:19038263 — PMID:10699171's abstract says only "a conserved arginine residue in the rhoGAP domain". The finding statement now says exactly that: the result is Prakash et al. 2000's, and this paper is quoted because it is where 433 appears.

## Verification

`just validate human ARHGAP6` → ✓ Valid, no warnings. `just validate-changed …` → ✓. `check_review_consistency.py` → `consistent: 27 file: quotes verbatim and unambiguous, 24 references adjudicated, counted claims hold`, and its `--self-test` shows all eight guards catching their mutations with the negative controls silent.

Noting for the record that two guards in this PR have now been *narrowed* after firing on false positives — a fixed WITH/FROM control term that condemned a working probe, and a text scan that fired on the docstring documenting the defect's removal. Both narrowings are mutation-tested in both directions, because narrowing a guard in response to a false positive is exactly when a hole gets opened.

cmungall:
One more commit (`53c67c68f9`), closing a loop I opened in the previous one.

Notes §8 now quotes `check_review_consistency.py`'s summary line as "expected output today". That is itself a counted claim, and leaving it unchecked would have reproduced — inside the checker — the exact failure the checker exists to catch. `audit()` now requires the notes to contain the current summary line, with a mutation test asserting a stale line is caught.

No curated content changed: no GO term, action, evidence code, `supporting_text` or `core_functions` entry is touched. This is the last push I intend to make absent a further finding.

cmungall:
Fixed in `5ad426cbda`. You were right, and this is the third round in which the same sentence has turned up in a field I had not looked at — so I have fixed the class rather than the instance.

## The sentence

`ARHGAP6-ai-review.yaml:1049` read *"Cellular PLC output, and the second messengers downstream of it, rise several-fold with ARHGAP6."* — a manipulation framing, "several-fold" for 1.6× and 2.3×, and contradicting its own `reason` two lines below. It now reads:

> Cellular phospholipase C activity rises about six-fold when ARHGAP6 is overexpressed.

I then grepped the whole review and notes for `several.fold|IP3|DAG|second messenger`. That sentence was the only remaining offender; every other mention (the top-level description, the reason, the finding, `core_functions[2].description`, and the suggested experiment) is correctly attributed.

## Why a guard and not a fourth patch

The same claim had been written into four separate fields, and each round of fixes reached only the copies visible at the time. `check_review_consistency.py` now enforces the invariant across the whole document: any **authored** text saying a second messenger rose must also say where — the hypertensive-cohort observation.

Two details that decide whether that guard is real:

- **The pattern matches `second messenger` as well as `IP3`/`DAG`.** The offending sentence named neither, which is exactly how it survived two rounds of fixes. A guard written only against the names would have passed it a third time.
- **`supporting_text` is exempt.** The paper's own sentence — *"Enhanced expression of ARHGAP6 was associated with an elevated level of PLC activity and increased levels of IP(3) (1.6-fold) and DAG (2.3-fold)"* — reports the numbers without naming the cohort in that clause, and a verbatim quote is not mine to reword. A guard that forced me to edit it would have been worse than no guard.

Both directions are mutation-tested: reinstating the exact offending sentence is caught, and the guard stays silent on that verbatim quote.

The checker's negative control — the real document raises nothing — passes, so this is evidence the class is clean, not just that the one instance is.

## Verification

`just validate human ARHGAP6` → ✓ Valid, no warnings. `just validate-changed …` → ✓. `check_review_consistency.py --self-test` → **13 assertions pass: 9 guards each catch their mutation, 4 negative controls stay silent** (counted from the run, not from memory); the plain run gives `consistent: 27 file: quotes verbatim and unambiguous, 24 references adjudicated, counted claims hold`.

No GO term, action, evidence code, `supporting_text` or `core_functions` term changed in this commit.

cmungall:
Thanks. Acknowledging the two non-blocking notes rather than leaving them silent, and declining to act on either — each would cost a fresh ~20-minute build and supersede this approval, for no change to any GO term, action, evidence code, `supporting_text` or `core_functions` entry.

**1. The guard's scope is the review YAML, not the notes.** Correct, and you are right that this claim class first appeared in prose. I checked before deciding: the notes contain no unattributed second-messenger mention (§2 and §7 discuss the PLC arm without the IP3/DAG numbers at all, and the only numeric mentions are in the review). So extending the guard would protect a file that has never carried the defect, at the price of a build. **What would reverse me:** the next edit to `ARHGAP6-notes.md` that reintroduces a numeric second-messenger claim — at that point the guard should be widened in the same commit, and `_authored_strings` already takes an arbitrary structure, so it is a two-line change to feed it the notes text.

**2. It fires only when a second-messenger name and an elevation word co-occur in one string.** Agreed, and agreed that this is the benign direction: a reworded claim naming neither slips past, but nothing correct is ever blocked, and the exemption for verbatim `supporting_text` is asserted positively rather than assumed. This is the general limit of pattern-matching prose. I would rather it under-fire than force edits to a quote I do not own — that failure mode would be worse than the one it guards against, because it would corrupt the evidence.

For the record on what this took: the same sentence had been written into four fields, and rounds 1, 2 and 3 each fixed only the copies then visible. The generalisable lesson is that a claim repeated across fields needs a document-wide invariant the first time it is corrected, not the fourth — fixing the instance you were shown is what makes the next round inevitable.

PR #3079, opened 2026-09-19
Review human ARHGAP36: the GAP whose arginine finger left before the rodent-primate split

cmungall:
Reviews human **ARHGAP36** (Q6ZRI8) for the PAINT campaign.

## The starting position

Four GO annotations, **none experimental**. Three are IBA descending from a single IBD on PANTHER node `PTN000973894` at `taxon:6072` — **Eumetazoa** — seeded by one gene, mouse **Arhgap6** (`MGI:MGI:1196332`), which is in subfamily `PTHR12635:SF6`. ARHGAP36 is in `PTHR12635:SF8`. The fourth is IEA from InterPro's RhoGAP signature `IPR000198`.

UniProt independently annotates an `Arginine finger` Site at position 258 and asserts `GTPase activator for the Rho-type GTPases` on `ECO:0000250` — by similarity. Position 258 is a **threonine**.

## `GO:0005096 GTPase activator activity` → REMOVE

The objection is to the **node placement**, which is what an IBA actually asserts, not to the donor count.

- Three controls carrying their own annotated arginine finger — ARHGAP1 (R282, resolved in the 1TX4 transition state), mouse Arhgap6 (R435, the PAINT seed itself) and human ARHGAP6 (R433) — **all project onto position 258**, and each recovers every other control's finger. The Site is in register, not a misplaced profile hit.
- Nearest arginine: **offset 7**. Not a register shift.
- **Mouse Arhgap36 (B1AUC7, same subfamily SF8) carries threonine at its own annotated Site 246**, and the human and mouse sites project onto each other reciprocally. The arginine was gone before the rodent–primate split — deep inside the clade this Eumetazoan node propagates into.
- Three published records name the same residue in three numberings, and all three fall out of the splice-variant arithmetic independently: UniProt `T258`, `Thr227` (PMID:33999959, isoform 2 via `VSP_021358`), `T246` (PMID:25024229, mouse canonical).

**The residue argument is deliberately not the whole case.** The sibling ARHGAP11B review (#3073) established that a *retained* arginine finger does not imply activity, so residue identity and curated activity are decoupled in both directions in this family; the inference here runs only in the direction where the residue is absent. What closes it is functional:

- A systematic RhoGEF/RhoGAP screen found **no effect on RAC1, CDC42 or RHOA** (PMID:32203420) — read **first-hand**, not on the citation. That article is paywalled (no PMC id, `isOpenAccess: N`) but its supplementary workbook is free: `scan_muller_supplementary.py` fetches it and reads Supplementary Table 2, where **ARHGAP36 scores `-` for RhoA, `-` for Rac1 and `-` for Cdc42**, with ARHGAP35 (RhoA+, Rac1+) and ARHGAP1 (RhoA+, Cdc42+) positive in the same run. Supplementary Table 1 records the screened construct as **human isoform 2, 517 aa** — so no inactive splice variant was tested — and carries the authors' own domain call verbatim: **"GAP-like (arginine finger missing)"**.
- **The caveat is measured, not omitted.** 15 of the 65 scorable GAP rows in that screen are negative for all three GTPases (23%), and ARHGAP17/RICH1 — a characterised Cdc42 GAP — is one of them. The screen has false negatives. What makes ARHGAP36's negative count is that it agrees with an independent measurement: of the six arginine-finger-less PROSITE-RhoGAP proteins present in the screen, **four are all-negative, ARHGAP36 among them**. Two are not — the ARHGAP11B decoupling lesson again, and why this row does not rest on the residue alone.
- The founding paper names the threonine and reports GAP-domain mutants retaining Gli-inducing activity (PMID:25024229).
- Amin et al. place ARHGAP36 among the human RHOGAP-domain proteins lacking an arginine finger (PMID:27481945).
- The one reported effect of ARHGAP36 on a Rho GTPase runs the **opposite** way — activating RhoA, via the PKAc-binding N-terminus (PMID:41644816).

**A curator has already reached this conclusion, on the ortholog.** Mouse Arhgap36 (`MGI:MGI:1922654`, `B1AUC7`) carries the identical three IBA rows and one IEA row — and alongside them a direct annotation to `GO:0003674 molecular_function` with evidence code **ND** from `GO_REF:0000015`, the reference whose purpose is to record that a curator looked for a molecular function and found no data supporting one. MGI says the molecular function is unknown in the same annotation set in which the phylogenetic pipeline says it is a GTPase activator. Read from QuickGO and `api.geneontology.org`, which agree.

**REMOVE is scoped to catalysis and says so.** UniProt records an experimental RAC1 interaction via the Rho-GAP domain (PubMed:35986704). Binding a GTPase and accelerating its hydrolysis are different claims; OCRL is the documented precedent for the first without the second; and the 1TX4 GAP:GTPase interface projects onto ARHGAP36 at **8/25 identical positions against 9/25 for mouse Arhgap6** — the binding surface is about as conserved as the catalysing family member's while the catalytic residue is not. Nothing about binding is removed, and no binding annotation exists to remove.

## The rest

| term | action | why |
|---|---|---|
| `GO:0007015` actin filament organization (IBA) | MARK_AS_OVER_ANNOTATED | PMID:41644816 *does* connect ARHGAP36 to actomyosin — but through the N-terminal PKAc module not the GAP domain, in the activating direction, indirectly, and abstract-only. Unsupported as propagated, not contradicted |
| `GO:0015629` actin cytoskeleton (IBA, `is_active_in`) | MARK_AS_OVER_ANNOTATED | four papers place it at plasma membrane, endo/lysosomal vesicles, mother centriole and cilium; none in the actin cytoskeleton. An absence, not a refutation |
| `GO:0007165` signal transduction (IEA) | MODIFY → `GO:0045880` | true but maximally general, and supplied by the signature of the one domain that does not work |
| `GO:0004862` cAMP-dependent protein kinase inhibitor activity | **NEW** | the MF this gene actually has, and the only one it would have after the REMOVE |
| `GO:0005886` plasma membrane | **NEW** | the compartment in which that activity happens |

`functional_isoforms` records a split that matters here: **isoform 1 is UniProt's displayed canonical sequence and is the one form reported not to work** — perinuclear, does not activate Gli — while isoforms 2–5 reach the plasma membrane and do (PMID:33999959). Every functional experiment in the literature uses the second class. No GO term is declared isoform-specific; the two NEW rows carry `Q6ZRI8-2` because that is what was tested, not because isoform 1 has been shown unable.

The MODIFY records rather than hides the complication: PMID:30598432 shows the centrosomal pool of ARHGAP36 *restraining* Smoothened ciliary translocation, the opposite sign from every output-level measurement. PKA has opposing roles in the pathway, so an inhibitor of PKA is negative on one step and positive on the other.

The NEW MF passes the comparator check: `GO:0004862`'s human annotations are the PKI pseudosubstrate family (PKIA/PKIB/PKIG) and the PKA regulatory subunits — exactly the class of thing that occupies the PKAC cleft and blocks catalysis. Eccles et al. make that comparison in the same assay, on constructs the methods state are human `Q6ZRI8-2`.

The MODIFY has ortholog precedent too: mouse Arhgap36 already carries `GO:0007224 smoothened signaling pathway` by **IMP** from PMID:31305241 — the same paper cited here, and the only experimental evidence anywhere on this branch. `GO:0045880` is its regulation child, so this moves the human gene into the branch its ortholog already occupies rather than one chosen from scratch.

## A GO expressivity gap, checked on two services

GO has **no** Hedgehog pathway term that is not anchored on Smoothened; every active BP term is `smoothened signaling pathway` or a child/regulator of it. The two terms that *would* have fitted an activator acting downstream of Smoothened — `GO:0007228` and `GO:1990787`, regulation of hh target transcription factor activity — are **obsolete**. Verified across OLS4, QuickGO and the GO API, because that negative is load-bearing.

## Artifacts

`genes/human/ARHGAP36/ARHGAP36-bioinformatics/`, everything fetched live, nothing hardcoded:

- `analyze_arhgap36.py` — the structural measurement. `--self-test` asserts **16** propositions: four mutations that must flip a verdict (including the right position with the *wrong* residue, which a position-only check would miss), two negative controls that must stay silent, four guards that must raise.
- `check_review.py` — re-resolves all four residue claims in the review against UniProt, plus a strict duplicate-key loader, GOA reconciliation, and the folded-scalar reflow guard. `--self-test` asserts **10**, with every mutation anchor required to match exactly once.
- `scan_muller_supplementary.py` — reads the decisive GAP result out of PMID:32203420's own supplementary tables, with the screen's false-negative rate counted and the arginine-finger-less set read from `results.json` rather than restated. `--self-test` asserts **10**, five of them mutations that must be refused (including a blanked call cell, which must not silently read as negative, and an ARHGAP17 that is no longer all-negative, which would falsify the caveat the report prints).

All three reproduce **byte-identically from a cleared cache** — nothing in the reports depends on a warm cache.

Two bugs the guards caught, both documented in the README: the contact calculation aborted because **PDB numbering is not UniProt numbering** in 1TX4 (offset ~197), and the reflow guard's first implementation flagged **43 lines on a clean file** because a nested `supporting_text: >-` opener itself ends in a hyphen.

## Provider note

Affinage gates: no blocking gate, soft gate tripped (`pairwise = tie`). Its 7 citations are a subset of the 27 PubMed hits for `ARHGAP36[Title/Abstract]`. It missed **both** papers that settle the GAP question (PMID:27481945, PMID:32203420), the only functional reference in the UniProt entry (PMID:35986704), and the published versions of two records it cited as preprints. Every one of those has a title naming a family, a disease or a partner rather than this gene. The UniProt `RN` list — five references, four of them sequencing projects — was the cheapest place to find the first.


cmungall:
Thanks — careful review. I verified all five before acting and took four; the fifth I checked rather than assumed. None changes an `action`, `term`, `evidence_type`, `supporting_text` or `core_functions` entry. Pushed as `6ccf03c9da`.

**1. The stale notes decision table — taken, and it is the real defect in the set.** You are right and I had not caught it. `ARHGAP36-notes.md` ended with `GO:0005813 centrosome | NEW`, which an earlier draft proposed and the final review declines. That is precisely the failure mode of prose asserting something the artifact does not do. The table now records `GO:0005886`, the decision actually taken, followed by a paragraph recording the decision that was *not* taken and why (PMID:30598432's mother-centriole work is in mouse `Ptc1 +/−` and `Ptc1 −/−` cells, so an IDA on the human gene would assert a human experiment that was not performed), so a later reader cannot mistake the table for the outcome.

**2. The `GO:0045880` framing — taken.** You are right that the tension is the first objection a reader raises and the reason got close without landing it. The row now states plainly that `GO:0045880` is being used as GO's de facto node for Hedgehog pathway output, **not** as a claim that ARHGAP36 signals through Smoothened — which PMID:25024229 establishes it does not — and that every downstream regulator lands on a Smoothened-labelled term because GO names the cascade after its receptor.

I also took your `GO:0010738` pointer and verified it independently rather than citing the sibling review: **GO:0010738 `regulation of protein kinase A signaling` is obsolete** — the GO API returns the label prefixed `obsolete` with a definition beginning `OBSOLETE.`, and OLS4's active index does not return it. That is now in the reason, because "route around the Smoothened problem through the kinase instead" is the *second* objection a reader raises and it is worth closing too.

**3. `proposed_replacement_terms` on the REMOVE — taken.** Dropped. Your semantic point is right: the schema makes `proposed_replacement_terms` the `MODIFY` mechanism (it is a `MODIFY` postcondition), and `GO:0004862` is not a rewording of `GO:0005096` — different evidence, reference, domain and substrate. It already stands as its own `NEW` row with the full evidence chain, and carrying it in both places let a consumer count one proposal twice. The reason now says why it lives there and not here.

**4. The overexpression caveat on `GO:0005886` — taken.** Fair inconsistency to call out: I flagged overexpression in the `knowledge_gaps` and not on the row a curator actually reads. It now says that both observations use tagged constructs, that PMID:27713425 images in MDCK (canine), that IDA remains correct because the assayed gene product is the human protein and the methods state its identity — which is exactly what separates this row from the mouse centrosome I declined — and that a compartment seen only for an overexpressed construct in a heterologous line is weaker than an endogenous one, of which none is published for human.

**5. `publications/PMID_27481945.md` whitespace diff — checked, then reverted.** You are right that it is content-neutral: `git diff --ignore-all-space` against `origin/main` is empty, and it is incidental to a forced re-fetch that failed to get full text anyway. I did not revert on that argument alone, because `publications/` is validated against and a reverted file could in principle break a `supporting_text` match. I reverted it and re-ran `just validate human ARHGAP36` against the restored file first: still `✓ Valid`. So the diff noise is gone with the quote still resolving.

---

One correction to the record, in the spirit of the caveat you asked for in (4). Since your review I read the decisive GAP result **first-hand** rather than through PMID:33999959's citation of it, and it cuts one way and then partly back:

PMID:32203420 is paywalled (Europe PMC: no PMC id, `isOpenAccess: N`) but its supplementary workbook is free. `scan_muller_supplementary.py` fetches it. Supplementary Table 2's activity-screen columns give **ARHGAP36: RhoA `−`, Rac1 `−`, Cdc42 `−`**, with ARHGAP35 (RhoA+, Rac1+) and ARHGAP1 (RhoA+, Cdc42+) positive in the same run. Supplementary Table 1 records the screened construct as **human isoform 2, 517 aa** — so no inactive splice variant was tested — and carries the authors' own domain call verbatim: **"GAP-like (arginine finger missing)"**.

The part that cuts back: **15 of the 65 scorable GAP rows in that screen are negative for all three GTPases (23%), and ARHGAP17/RICH1 — a characterised Cdc42 GAP — is one of them.** The screen has false negatives and I did not want that buried. What keeps ARHGAP36's negative meaningful is that it agrees with an independent measurement: of the six arginine-finger-less PROSITE-RhoGAP proteins present in the screen, four are all-negative, ARHGAP36 among them. Two are not — which is the ARHGAP11B decoupling lesson again, and the reason the row still does not rest on the residue alone. All of that is in the `GO:0005096` reason and the `reference_review`, not in `supporting_text`, since the cache is abstract-only and a supplementary quote could not be checked there.

cmungall:
Both blocking items were real and both were mine. Fixed in `8f59dc6872`, along with all four suggestions. One premise in the review did not hold and I have recorded the correction rather than working around it.

## IMPORTANT 1 — misattributed findings. Correct, and the sharpest catch in either round.

You are right, and right about why it matters: the numbers reconciled, so nothing downstream would have flagged it. The two statements carrying the Müller 2020 supplementary readings were attached to `file:human/ARHGAP36/ARHGAP36-bioinformatics/RESULTS.md`, which contains nothing about Müller 2020 — `RESULTS.md` is the output of `analyze_arhgap36.py`; the scan is `scan_muller_supplementary.py`, writing `muller2020_arhgap36.json`. I had reasoned "this is a bioinformatics finding, it goes on the bioinformatics reference" without checking that the named file says it.

They now sit on the **`PMID:32203420`** reference, which is where they belong — they are claims about what that paper contains, not about what a script produced. They carry no `supporting_text`, because that field is validated as a substring of the cached publication and the cache is abstract-only; the `reference_review` on that reference now states that explicitly and points at the script and its committed JSON for the re-derivation. The `RESULTS.md` reference keeps only the four findings `analyze_arhgap36.py` actually produces, which is what its `review_notes` described all along.

## IMPORTANT 2 — missing history record. Correct on the gap and the mechanism; the event name does not exist.

The gap is real: `history/genes/human/ARHGAP36/` held only the pre-follow-up record, and that record is append-only and now wrong on exactly the two points you name — it says the Müller result is quoted second-hand (`45b0790ca` made it first-hand) and that there are two committed scripts (there are three). A new record now covers the whole follow-up and states both supersessions explicitly.

**The one premise that does not hold:** the review asks for "a new record with an `UPDATE` event". There is no `UPDATE`. `HistoryEventTypeEnum` in `src/ai_gene_review/schema/history.yaml:54-66` permits exactly `GENERAL`, `CREATE`, `EDIT`, `REVIEW`, `AUDIT`, and `scripts/new_history.py` rejects it outright:

```
new_history.py: error: argument --event: invalid choice: 'UPDATE' (choose from GENERAL, CREATE, EDIT, REVIEW, AUDIT)
```

The precedent you cited settles which one is meant: `history/genes/human/AKAP12/` holds five records using **one `CREATE` and four `EDIT`** — exactly the append-after-create pattern you describe, under the name the schema has. So I took the mechanism, which was right, and used `EDIT`. Recorded in the new record's `details` too, so the next reader does not repeat the search.

## Suggestions — all four taken

1. **Scan script docstring.** Correct, and the same class of defect as IMPORTANT 1: prose asserting a state the artifact had already left. It now says the review quoted the result second-hand *before this script existed*, and describes where the findings actually live.

2. **`ARHGAP17` as a positive control.** Correct and worse than cosmetic — thank you. It scores `−/−/−`, and the whole point of naming it in the report is that it is the screen's **false-negative exemplar**. Listing it in `POSITIVE_CONTROLS` meant the "at least one control scores `+`" guard could in principle have been satisfied by a protein that is evidence for the opposite point. It is now a separate `FALSE_NEGATIVE_EXEMPLAR` with its own guard, because the report prints a caveat that *asserts* this protein is all-negative despite being a characterised Cdc42 GAP — so the run now fails if that stops being true rather than continuing to print it. Self-test 9 → 10 for the new mutation, and the README records why the constant exists.

3. **Missing word** at `:86-87` — fixed ("the two NEW rows in this review record `Q6ZRI8-2`…").

4. **Effective *n* of the 4/6 agreement.** Correct: OCRL and INPP5B are paralogous inositol polyphosphate 5-phosphatases, so two of the four agreeing entries are not independent of each other. Stated in the `GO:0005096` reason and printed by the script, so the ratio is not read as four independent observations.

## On your caveat about not being able to run things

Fair, and since you flagged the MGI and `GO:0010738` claims as un-re-queried, here is what they rest on, both cross-checked on **two different services** rather than one:

- **MGI `GO:0003674` ND** — QuickGO (`geneProductId=UniProtKB:B1AUC7`) and `api.geneontology.org/api/bioentity/gene/MGI:MGI:1922654/function` return the same seven rows, including `GO:0003674 molecular_function | ND | GO_REF:0000015`. I also read GO_REF:0000015's own page rather than paraphrasing it: *"Direct annotations to any of the three root terms … indicate that curators have found no data supporting an annotation to a more specific term."* An attempt to read the MGI marker page via WebFetch returned a GO-slim category list instead of the annotations and was discarded — noted in the gene notes, because it was the second confidently-wrong WebFetch result in this session.
- **`GO:0010738` obsolete** — GO API returns the label prefixed `obsolete` with a definition beginning `OBSOLETE.`; OLS4's active index does not return it. I did not rely on the AKAP10 review for this.

Local verification on `8f59dc6872`: `just validate human ARHGAP36` → `✓ Valid`; self-tests green at their documented counts (`analyze_arhgap36` 16, `check_review` 10, `scan_muller` 10, asserted by a script rather than counted by hand); and `analyze_arhgap36.py` re-run **from a cleared cache** reproduced `results.json` and `RESULTS.md` byte-identically, so nothing in the report depends on a warm cache.

cmungall:
Thanks — and thank you for correcting yourself on the `UPDATE`/`EDIT` point in the approval body rather than leaving it. Two notes.

## The suggestion is right, and I am declining to do it *here*

> `check_review.py` could close the defect class this round found. A `findings[].statement` attached to a `file:` reference whose target does not contain it is invisible to both `just validate` and the existing checks, and for local `file:` references a keyword-overlap check is mechanical — it would have caught this one on the first pass.

You are right about all three parts: the defect class is real, it is invisible to existing tooling, and it is mechanically detectable. I started implementing it and then backed it out. Two reasons, and I would rather state them than quietly not do it.

**1. It would close the class for exactly one gene.** `check_review.py` lives in `genes/human/ARHGAP36/ARHGAP36-bioinformatics/` and runs only when someone runs it, on this one review. Putting the guard there would make the class *look* closed while leaving every other review in the repository exposed — arguably worse than not having it, because the next person to make this mistake will be reviewing a different gene. The check belongs in the shared reference validator in `src/ai_gene_review/`, next to the `PMID:` `supporting_text` check that already exists, where CI runs it on every review on every PR.

That is a schema/tooling change, not a curation change, and it should not ride in on a gene review PR where it gets reviewed as curation.

**2. There is a bigger adjacent hole worth fixing in the same place.** While scoping it I checked what the repository currently verifies for `file:` references: `supporting_text` is validated as a verbatim substring for `PMID:` references only. For `file:` references — whose targets are *local and trivially readable* — it is not checked at all. So the stronger version of your suggestion is two checks, not one:

- every `supporting_text` citing a `file:` reference must be a verbatim substring of that file (currently unchecked, and a fabrication surface rather than a hypothetical one — I verified my own affinage quote by hand with `grep` precisely because nothing else would);
- every `findings[].statement` on a `file:` reference must share a minimum fraction of its distinctive tokens with the file, with the threshold **calibrated** by a test that measures real statements against a deliberately misattributed one, rather than picked.

Worth noting for whoever picks this up: the naive version of your check would *not* have caught my error. "At least one distinctive token appears in the target" passes, because `ARHGAP36` appears in `RESULTS.md`. It needs to be a fraction with a calibrated threshold, and the misattributed statements here are a ready-made negative fixture for calibrating it.

**What would reverse me:** if the shared validator is not the right home — say the maintainers would rather these checks stay per-gene — I will add it to `check_review.py` in a follow-up commit here. And if you would rather have the imperfect local version now than the complete shared version later, say so and I will push it; I am declining on placement, not on merit.

Since it changes no `action`, `term`, `evidence_type`, `supporting_text` or `core_functions` entry, and the review YAML would be byte-identical either way, I have left the approved head alone rather than spending another build on it.

## One stale claim in the PR body, now fixed

Not raised by you, but caught while re-reading my own artifacts for the defect class you found twice. The PR body said `scan_muller_supplementary.py --self-test` asserts **9** propositions; adding the `FALSE_NEGATIVE_EXEMPLAR` guard made it **10**. Corrected in the body (body edits do not re-trigger CI or dismiss the approval). The README and history record already said 10.

I also added to the body a result I had verified but not reported: all three scripts reproduce **byte-identically from a cleared cache** — `rm -rf cache` then re-run leaves `git status` clean — so nothing in the reports depends on a warm cache.

cmungall:
**CI caught a real defect that every local check of mine had missed.** Fixed in `cdca699eca`. Writing it up because the gate has a misleading name and the next gene in this campaign will hit it too.

## What failed

`Build and test` → `test (3.12)` failed at step 19, **`Validate family reviews (scoped)`**:

```
[FAIL] TARGET_IDENTITY ARHGAP36 GO:0005096:
       target accession UniProtKB:B1AUC7 is not this gene (Q6ZRI8)
residue-claim validation over 31 gene review(s) with claims: 447 pass, 1 fail, 0 unresolved
```

The step's name says *family* reviews, but `just validate-families` also runs
`ai_gene_review.validation.gene_residue_claims` over the **gene** corpus. `just validate human ARHGAP36` was green in every round and never touches it. That is why four rounds of local verification, three self-tested scripts and two reviewers all sailed past it.

## The rule, and why I broke it

`src/ai_gene_review/validation/gene_residue_claims.py:148-159`: a `residue_claim`'s **`target` must be this gene's own protein**. The `anchor` may be any comparator; the target may not. That is exactly what `ResiduePosition` says in the schema — *"the corresponding position in **this gene's** own protein"* — which I had read, and still got wrong, because the mouse pair was a convenient place to put a cross-species measurement.

The second claim anchored on mouse Arhgap6 (`O54834` R435) with **mouse Arhgap36** (`B1AUC7` T246) as target, to make "the loss predates the rodent–primate split" machine-checkable. Re-pointed to the same anchor with **this gene** as target (`Q6ZRI8` T258), which keeps the PAINT-donor comparison checkable and satisfies the rule.

The dating argument is not expressible as a residue claim **at all** — it is a statement about a different protein — so it stays in `review.reason` and in `RESULTS.md` §4c, where it is reproducible. The claim's `comment` now says that explicitly rather than leaving the omission unexplained.

Before and after, on the same validator CI uses:

```
committed version:  exit 1   [FAIL] TARGET_IDENTITY    6 pass, 1 fail, 0 unresolved
fixed version:      exit 0                             6 pass, 0 fail, 0 unresolved
```

And the full local `just validate-families`, which I should have run from the start:

```
residue-site validation over 244 family review(s):        1411 pass, 0 fail, 1 unresolved
family/gene cross-check over 244 family review(s):        0 conflict(s), 45 consistent
residue-claim validation over 31 gene review(s) w/claims:  447 pass, 0 fail, 0 unresolved
```

## A guard caught the knock-on, which is the part worth reporting

Re-pointing the claim made `check_review.py --self-test` fail, and **correctly**. Its mutation anchor was the short string `"position: 258\n residue: T"`, which had been unique and stopped being so the instant a second claim pointed at the same target. It raised

```
CheckError: residue anchor matched 2 times, expected exactly 1
```

rather than silently mutating whichever claim came first and reporting a pass. That is the matched-exactly-once assertion doing the one job it exists for. The anchor now carries the whole `anchor:`+`target:` block of a single claim, and the comment records the incident so the short form does not come back.

## State

- `just validate human ARHGAP36` → `✓ Valid`
- `just validate-families` → 0 fail (above)
- self-tests 16 / 10 / 10, all green, all reproducing byte-identically from a cleared cache
- no `action`, `term`, `evidence_type`, `supporting_text` or `core_functions` entry changed in this commit; the change is one residue-claim target, its comment, the self-test anchor, and the notes
- a third history record covers this, since the preceding one is append-only and its verification line names only the gate I had run

I also reverted an unrelated one-line `cache/go/terms.csv` addition (`GO:0140298 endocytic iron import into cell`, timestamped before this session) that had drifted into the worktree, to keep the diff to this gene.

PR #3080, opened 2026-09-19
Review human ARHGAP23: the RhoGAP whose published arginine-finger mutant misses the arginine finger

cmungall:
## ARHGAP23 (Q9P227) — PAINT campaign review

The automated provider returned three citations and closed with: *"Beyond these
junction-associated roles, the enzymatic GAP activity and domain functions of ARHGAP23 have not
been experimentally characterized in the available corpus."* That is false. An independent sweep
(Europe PMC, fully enumerated; NCBI `gene2pubmed` for GeneID 57636; partner/paralog-targeted
searches) found **three papers that directly measure its GAP activity**, and none of them names
the gene in its title:

| PMID | Title | What it measures |
|---|---|---|
| 27226243 | *A lateral signalling pathway coordinates shape volatility during cell migration.* | RhoA, by pulldown, loss **and** gain of function |
| 28114311 | *RhoGTPase Regulators Orchestrate Distinct Stages of Synaptic Development.* | Rac1, by FRET biosensor |
| 32203420 | *Systems analysis of RhoGEF and RhoGAP regulatory proteins…* | RhoA **and** Rac1 significant, Cdc42 not |

PMID:27226243 is in NCBI's curated `gene2pubmed` set for this gene ID, so it was reachable from
the identifier alone.

### The gene-specific check: does the RhoGAP domain keep its arginine finger?

Yes — and the review says plainly that this is worth less than it looks.

`genes/human/ARHGAP23/ARHGAP23-bioinformatics/` contains a committed, re-runnable analysis.
ARHGAP1's own annotated arginine finger (R282, resolved in PDB 1TX4, the RhoA:GDP:AlF4
transition-state complex) maps onto **ARHGAP23 R942**, which is ARHGAP23's own annotated Site,
and the projection holds **in both directions**. The same result holds against ARHGAP21.

The calibration is the point:

| control | why | residue at its own annotated finger | verdict |
|---|---|---|---|
| ARHGAP21 (positive) | `enables GO:0005096` **IDA** in this repo's GOA table | 1184 R | retained |
| OCRL (dead, residue lost) | UniProt, `ECO:0000269`: *"the Rho-GAP domain lacks the catalytic arginine and is catalytically inactive"* | 757 **Q** | NOT retained |
| ARHGAP11B (dead, residue kept) | two curated `NOT\|enables GO:0005096` IDA rows | 87 R | **retained** |

So of two known GAP-dead controls the residue test calls one *retained*. Loss would have been an
argument; retention only removes one. Each control's credential is **verified from the repo's own
GOA tables or the fetched UniProt record before the report is allowed to describe it that way** —
an unverifiable credential aborts the run.

Interface conservation does not discriminate either: ranked over the 25 1TX4 GAP:GTPase contact
positions, GAP-dead ARHGAP11B (64%) scores **above** experimentally active ARHGAP21 (52%). The
script computes that and says so rather than quoting the subject's own 52% as if it meant
something. What does carry information: **ARHGAP23 and ARHGAP21 carry the same residue at 22 of
those 25 positions.**

### A discrepancy in the literature, found by the same analysis

PMID:32203420's Supplementary Table 4 calls its ARHGAP23 imaging construct a *"GAP-deficient
arginine finger mutant"*, and the SI figure names it **ARHGAP23-R986K**. Its Supplementary Table 1
gives the screened cDNA as NP_001186346.1, 1491 aa — identical in length to Q9P227, so the
numbering is directly comparable.

**986 is not the arginine finger.** It aligns to ARHGAP1 **R323**, which lies in the 1TX4 interface
but does **not** contact the nucleotide/AlF4/Mg transition state, whereas R282 does. The mutant was
used for TIRF localization only; the screen's catalytic controls were other proteins (the clause is
extracted verbatim from the SI). So no experiment has tested whether ARHGAP23's actual arginine
finger is required for its measured activity.

This is reported as a **discrepancy, not a correction** — the structural argument favours 942, but
only an experiment settles it, and that experiment is in `suggested_experiments`.

The mutant designation is **read out of the supplementary PDF by regex**, not written into the
script, so the residue number this section turns on cannot have come from me.

### Ontology finding (checked on three independent services)

`GO:0005100` "Rho GTPase activator activity" is obsolete/merged into `GO:0005096` — OLS4
`is_obsolete: true`; `api.geneontology.org` returns an empty record; QuickGO silently resolves the
id and returns `GO:0005096`. `GO:0005096` now carries *Rho GAP activity*, *Rac GAP activity* and
the ARF/Rab/Ral/Ran/Rap/Ras/Sar equivalents as **narrow synonyms**, and has **no `is_a` children**
(OLS4: 0; GO API: 0; QuickGO returns one child via `capable_of`, not `is_a`).

Consequence, visible in this one gene: three `GO:0005096` rows descend from a **RHOA** reaction, a
**RAC1** reaction, and a phylogenetic node whose experimental grounding is a **Cdc42** GAP — and all
three are textually identical. The process branch still works (`GO:0035024`, `GO:0035021`), which
is why both `MODIFY`s point there.

### Decisions on the 8 GOA rows

| term | evidence | action |
|---|---|---|
| GO:0005096 GTPase activator activity | IBA (PTN008592632, donor ARHGAP21) | **ACCEPT** — donor has its own IDA; target now has direct evidence |
| GO:0005096 | TAS R-HSA-8981637 (RHOA set) | **ACCEPT** — membership sourced in Reactome's summary to the FRET screen |
| GO:0005096 | TAS R-HSA-9013144 (RAC1 set) | **ACCEPT** — sourced to PMID:28114311, with the Rac1 caveat recorded |
| GO:0005829 cytosol ×2 | TAS Reactome | **KEEP_AS_NON_CORE** — compartment labels, not observations; second flagged redundant |
| GO:0007165 signal transduction | IEA InterPro | **MODIFY** → GO:0035024 + GO:0035021 |
| GO:0051056 | TAS R-HSA-9012999 | **MODIFY** → GO:0035024 + GO:0035021 |
| GO:0070062 extracellular exosome | HDA | **MARK_AS_OVER_ANNOTATED** |

Three `NEW`: `GO:0005938` cell cortex, `GO:0005925` focal adhesion, `GO:0051893` regulation of
focal adhesion assembly (unsigned deliberately — the reported direction is consistent but the
mechanism is indirect and rests on one cell line).

Note on the `GO:0007165` MODIFY: it is a **branch** change, not a specialization. `GO:0035024` is a
descendant of `GO:0051056` and is **not** a descendant of `GO:0007165` (checked on two services) —
a GAP terminates Rho signalling rather than transducing it, so the InterPro2GO mapping puts every
RhoGAP-domain protein on the participant branch. Raised as a family-wide question.

### Tension recorded rather than smoothed

PMID:27226243 measured Rac1 in the same experiment that showed RhoA elevation and found active
Rac1 **decreased**, not increased, on ARHGAP21/23 knockdown — the opposite of a Rac1-GAP
expectation. Two biosensor studies in other cell types report the activity directly, so the term
stays; the review states the Rac1 arm is plausibly cell-type dependent and asks the question.

### Checks, committed so they re-run

* `analyze_arhgap23.py --self-test` — 14 mutations, 16 anchors, each guard asserted to fire with
  its expected message; mutation count reconciled against anchor count.
* `mutation_test.py` — breaks each of 9 guards in a copy and asserts the self-test catches it, plus
  a no-op negative control that must stay silent. It **found a real hole** when written: the
  "missing arginine finger" case had been re-implemented inside the self-test instead of driven
  through `Protein.__init__`, so deleting the real guard left the self-test passing.
* `check_file_quotes.py` — CI verifies `supporting_text` verbatim only for `PMID:` references, so
  every `file:` and `Reactome:` quote here (18 of them) is checked against its source; `--self-test`
  fabricates a quote and asserts rejection, rewraps a real one and asserts acceptance.

`just validate human ARHGAP23` is green (one warning: cytosol appears in a core function's
locations and in `existing_annotations` only as `KEEP_AS_NON_CORE`, which the rule does not count).


cmungall:
Thanks — three of the five change a claim a curator would read differently, so they are applied in `82bfac2672`. Two are declined, with reasons, below.

## Applied

**1 — cytosol was asserted core and non-core.** You are right and it was not cosmetic: `core_functions[1].locations` carried `GO:0005829` while both GOA cytosol rows are `KEEP_AS_NON_CORE`, so the document said the same location was both. Removed from the core function, and your reasoning is the reason given: the cytoplasmic pool the keratinocyte experiment reads out is the **PKP4-knockout** condition — an engineered background, not the protein's normal address. The check that this was a real inconsistency rather than a validator quirk is that `just validate human ARHGAP23` now reports **zero warnings** rather than one.

**3 — `correctness` on PMID:27481945.** Agreed, and the mismatch was mine. Now `UNVERIFIED`, with the notes saying explicitly that this refers to the **claim, not the identifier** (the title does resolve). You also caught that `ARHGAP23-notes.md` §2 stated the negative flatly where the YAML hedged it; the notes now carry the same hedge and name the independently checkable substitute you identified — Müller's Supplementary Table 2 fills the "in vivo" and "reference" columns for ARHGAP23 and leaves the **"in vitro" columns empty**.

**5 — maturation vs assembly.** Added as a `suggested_questions` entry, framed as you put it: the same class of expressivity problem as the substrate gap, one branch over. The annotation still uses `GO:0051893` because it is the closest available, and the new question says so rather than letting the term imply an exact fit.

**2 — partially.** The split rationale existed but was thin, so it is now explicit: the two core functions differ not in the activity, which is identical, but in **which regulator licenses it, and therefore when it is allowed to run**. Prickle1 delivers ARHGAP23 to the lateral cortex so RhoA is kept off there; plakophilin 4 does the opposite, binding ARHGAP23 and lowering its access to RhoA so that cortical RhoA is permitted. I did not merge them, because a merged entry would have to drop one of those two regulators or assert a single compartment for both.

## Declined

**4 — reordering the `description` so the cellular assays carry the activity claim.** Declining, for two reasons.

The first is that the sentence does not do what the objection describes. "Its Rho-GAP domain is intact" is followed immediately, in the same sentence, by the **discriminating** comparison — "residue-for-residue almost identical to the experimentally active ARHGAP21" — and then by the cellular results and by "no assay on purified protein has been reported." Nothing in `description` derives activity from the residue.

The second is that the fix would be out of place. "Retention of the arginine is not evidence of activity" is an evidential judgement about how to weigh a result, and `CLAUDE.md` reserves `description` for a project-independent biological summary a reader can use without knowing this curation project. That argument is already stated three times where it belongs: in the `GO:0005096` IBA row's `reason`, in `RESULTS.md` §3 and §6, and in a `knowledge_gaps` entry.

What would have reversed me: if `description` asserted GAP activity *from* the residue, or presented 52% interface conservation as support — it does neither, and `RESULTS.md` prints the ranking that makes 52% non-discriminating precisely so that number cannot be quoted as support.

**2 as a merge.** See above — the explanation was strengthened instead of merging.

---

One note on your caveat, since it affects what you were able to check: the three committed scripts each carry a `--self-test`, and `mutation_test.py` breaks each of nine guards in a copy and asserts the self-test catches it, with a no-op negative control that must stay silent. It found a real hole when it was written — the "missing arginine finger" case had been re-implemented *inside* the self-test rather than driven through `Protein.__init__`, so deleting the real guard left the self-test passing. All four still pass after these edits, along with `check_file_quotes.py` on all 18 locally-sourced quotes.

cmungall:
Applied in `df32f7ed1` — you found the right weak link. The "in vitro columns are empty" substitute was doing real work (it is what the `UNVERIFIED` downgrade leans on, and what the knowledge gap and one suggested experiment rest on) and it was the only load-bearing assertion in the PR resting on a one-off manual read of a spreadsheet.

`analyze_arhgap23.py` now parses Supplementary Table 2 alongside the tables it already read. Two things were not free:

**Column groups are derived, not assumed.** The group labels (`"integrated"`, `"in vitro"`, `"in vivo"`, `"reference"`) are read from the header row and each block extends to the next label.

**Each block is the *first* run of RhoA/Rac1/Cdc42 columns after its label.** Taking every matching column instead swallows the derived single- and combination-specificity flag columns that follow, whose labels repeat — and the last value wins, so the table is *misread* rather than visibly broken. That is not hypothetical: my first implementation had exactly this defect and reported the screen's specificity calls as the flag triple `(0, 0, 0)` instead of `(+, +, -)`. There is now a self-test assertion that fails on precisely that collapse.

Result, now printed in `RESULTS.md` §5:

| literature column | RhoA | Rac1 | Cdc42 |
|---|---|---|---|
| integrated | `+` | *(empty)* | *(empty)* |
| in vitro | *(empty)* | *(empty)* | *(empty)* |
| in vivo | `+` | *(empty)* | *(empty)* |
| reference | `+` | *(empty)* | *(empty)* |

The claim holds. The run also aborts if the `"in vitro"` block cannot be found at all, so an **absent** column can never be reported as an **empty** one — which is the specific way this negative could have gone wrong.

**One thing worth flagging, since it is the second time the mutation test has earned its place here.** My first attempt at mutating that guard changed only its error message; the guard still raised, and the self-test passed anyway — because the guard was checked inline in `muller_specificity()` and the self-test had *no coverage for the missing-block case at all*. A guard checked inline can only ever be exercised by a live table that happens to be correct, which is not a test. It is now factored into `_require_literature_groups()` and driven from `--self-test` with a synthetic layout, plus a negative control that must pass silently.

The review's `reference_review` and `ARHGAP23-notes.md` §2 now say where the check lives rather than asserting the fact bare.

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 18 mutations across 21 anchors (3 baselines); `mutation_test.py` 11 guards all caught with the negative control silent; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 holding on both services.

cmungall:
All three taken in `ace58c148`. All three were real, and all three were in code I added last round — worth saying plainly, because the pattern is that each new guard brings its own new hole.

**1 — rendered groups vs required groups.** You are right, and it is the same absent-versus-empty confusion the guard exists to prevent, one level up: `_require_literature_groups` covered `in vitro`/`in vivo` while `render_markdown` printed four rows through `.get(g, {})`, so a missing `integrated` or `reference` block would have rendered as `*(empty)*`. Your point about `reference` being the one whose block extent is unbounded is what makes it more than theoretical. `LITERATURE_GROUPS_REQUIRED` is now exactly the rendered set, and the renderer indexes directly (`lit["groups"][g]`) so a missing group raises rather than prints. New self-test case: a layout missing `reference` must abort with `"with no 'reference'"`, and `mutation_test.py` gained a mutation that shrinks the required tuple back to two and is caught.

**2 — assert the vocabulary, not the artefact.** Fair — `!= {"RhoA": 0, "Rac1": 0, "Cdc42": 0}` only catches the value my naive implementation happened to produce. It now asserts every screen-block value is in the `+`/`-` vocabulary, so *any* collapse onto the numeric flag columns fails, and the failure message prints the offending cells.

**3 — the rendered inversion.** Correct, and this one I would not have caught: `is not ({'empty' if … else 'populated'})` renders as "the in vivo row is not (populated)", which negates the table printed immediately above it. This is the campaign's cosmetic-reflow-ships-an-inverted-fact failure mode arriving through a conditional expression instead of a line wrap. Both rows are now stated positively, and the sentence says why the contrast is the point at all:

> This parse finds the **"in vitro" row empty for all three GTPases**, and the **"in vivo" row populated**. The contrast is the point: an empty in-vitro row means nothing if every row is empty.

---

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` **19 mutations across 22 anchors** (3 baselines); `mutation_test.py` **12 guards all caught**, negative control silent; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 holding on both services. No `term`, `action`, `evidence_type`, `supporting_text` or `core_functions` entry is touched by this commit — it changes a guard's scope, an assertion's strength, and a sentence that said the opposite of its own data.

cmungall:
Taken in `023d1fbed`. You have now found the same asymmetry three rounds running, one column narrower each time, and that is worth naming rather than just patching again: **each round I added a guard whose asserted set was wider than its checked set.** First the argument's two groups versus the report's four; then the report's four versus the guard's two; now the prose's three versus the computation's two.

The concrete fix: `empty_by_group` is computed for every required group, and the self-test asserts the **exact** pattern the prose states, as one comparison rather than two convenient booleans:

```python
expected_empty = {"in vitro": True, "in vivo": False, "reference": False}
observed = {g: lit["empty_by_group"][g] for g in expected_empty}
assert observed == expected_empty, (
    f"Supplementary Table 2 emptiness is {observed}, not {expected_empty}. The review's "
    "reference_review on PMID:27481945, its ONTOLOGY-adjacent knowledge gap and one "
    "suggested experiment all assert this exact pattern, so the prose is now wrong."
)
```

The failure message names the three places in the review that depend on it, so a future break says what it invalidates rather than just that a dict differs. `RESULTS.md` §5 now prints all four rows' states, not two:

> This parse finds the **"integrated" row populated**, the **"in vitro" row empty**, the **"in vivo" row populated**, the **"reference" row populated**.

`mutation_test.py` gained a thirteenth guard: shrink `empty_by_group` to just the in-vitro entry and the self-test must fail. It does.

The general form of the lesson, which I have put in the history record: **a guard should be derived from what the prose claims, not from the subset that was convenient to compute when the guard was written.** That is what all three rounds have been, and deriving `LITERATURE_GROUPS_REQUIRED` and now `expected_empty` from the asserted set rather than hand-listing is what stops the fourth.

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` **19 mutations across 23 anchors** (4 baselines); `mutation_test.py` **13 guards all caught**, negative control silent; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. No `term`, `action`, `evidence_type`, `supporting_text` or `core_functions` entry touched.

cmungall:
Both taken in `408ce63be`. The first one was the important one, and it was a hole in the thing that has been catching everything else.

## 1 — "caught" now means caught by the named guard

You are right that exit code alone is too weak, and the shadowing was already real: `render a literature group the presence guard does not require` and `compute emptiness for fewer groups` were both failing on a `KeyError`, not on the guards they name — so the dedicated `"with no 'reference'"` case I verified with you last round was being masked and could have rotted while the harness still printed `ok, caught`.

Each `MUTATIONS` entry now carries a fourth element: the substring the failure output must contain. Caught by anything else is reported as `*** WRONG GUARD`, not as a pass, and `_run_self_test` returns the full output rather than just the exit code.

Two consequences fell out of writing that down, and both were needed to make the expectations satisfiable rather than just declared:

- **Order is part of what a mutation test measures.** Shrinking `LITERATURE_GROUPS_REQUIRED` also breaks the live parse's downstream bookkeeping, and the live parse ran *first*, so the KeyError beat the guard to it. The `_require_literature_groups` checks now run on synthetic layouts **before** the live parse.
- **Derived fixtures go quiet under the mutation that matters.** The synthetic layouts were built with `{g: … for g in LITERATURE_GROUPS_REQUIRED if g != "reference"}`, so shrinking the tuple shrank the fixture in step and the guard stopped being exercised. They are now written out literally. The emptiness comparison also reports a named missing-key failure instead of raising `KeyError`.

Result — every mutation is now caught by its own guard:

```
ok, caught: render a literature group the presence guard does not require
      -> rendered group missing from the guard: expected AnalysisError containing "with no 'reference'", none raised
ok, caught: compute emptiness for fewer groups than the review's prose asserts
      -> empty_by_group is missing ['in vivo', 'reference'], so the emptiness computation is narrower than the pattern the review's prose asserts
MUTATION TEST PASSED: 13 guards all caught, 1 negative control(s) silent
```

And because a new check is worth nothing until it has been seen to fail, I pointed one expectation at a string that cannot appear and confirmed the harness reports it:

```
*** WRONG GUARD: make reciprocal ignore the reverse mapping
      expected the failure to mention: 'this string cannot appear in any failure'
```

## 2 — the unasserted second spelling

Removed. `in_vitro_empty`/`in_vivo_empty` became write-only when `empty_by_group` arrived and were still shipping in `results.json`, which is exactly the duplicate-of-an-asserted-fact that drifts. The renderer and the self-test both read `empty_by_group`.

---

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught **by their named guard**, negative control silent; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. No curation surface touched — this commit changes the harness and deletes two redundant JSON keys.

cmungall:
All three in `12b17743e`.

**1 — the opt-out.** This was the one that mattered: `expect = entry[3] if len(entry) > 3 else None` would have let a future three-element entry silently revert to exit-code-only, undoing the previous commit with no signal at all. Now a direct four-way unpack, so a short entry is a `TypeError` at import rather than a quiet regression:

```python
for desc, old, new, expect in MUTATIONS:
```

Worth noting the shape of that mistake, because it is the same one three rounds running: I wrote the tolerant form out of habit while adding a check whose entire point was to be intolerant.

**2 — the wrong expectation.** Correct. `'annotated arginine fingers'` is the wording of the *count* check next door, not of the `_expect_raises` case that actually catches that mutation, so a rewording of the neighbour would have turned a real catch into a WRONG GUARD report. Now `'missing finger annotation: guard fired with the wrong message'`, which is that case's own label.

**3 — the numbering.** Dropped rather than renumbered: 1..10, 13, 12, 14, 11 is what numbering does to a list that gets reordered, and renumbering only defers it to the next reorder. Each case now reads `# Case: <what it breaks>`, which cannot go stale under reordering.

Re-verified after the changes: the `WRONG GUARD` path still fires when an expectation is pointed at a string that cannot appear; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught, negative control silent; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `just validate human ARHGAP23` green with zero warnings. No curation surface touched.

cmungall:
Correct, serious, and now fixed in `d41bb8331`. I reproduced it rather than reasoning about it, and it is worse than "unrelated failures":

```
$ TMPDIR=/tmp uv run python mutation_test.py
*** WRONG GUARD: drop the annotated-site requirement from forward_retained
      expected the failure to mention: 'retained must require the annotated-site condition'
      got: IndexError: 3
*** WRONG GUARD: make reciprocal ignore the reverse mapping
      ...
```

`IndexError: 3`, thirteen times, exactly as you predicted from `REPO = SCRIPT_DIR.parents[3]` against a two-parent `/tmp/tmpXXXX`.

And your second sentence is the part that stings: **before `408ce63be` this would have printed "13 guards all caught" off nothing but import errors.** The harness whose whole purpose is to prove the guards can fail would have been asserting they were sound while never once executing them — on Linux, which is where CI runs and where anyone but me would have run it. The expected-message check added two rounds ago is the only reason this surfaced as thirteen `WRONG GUARD` lines instead of thirteen green ones, which is a better argument for that check than anything I wrote when I added it.

The mutant now lives beside the original as `.mutation_test_subject.py` (gitignored, unlinked in a `finally`), so `SCRIPT_DIR`, `CACHE_DIR` and `REPO` resolve exactly as they do for the real script, and the fourteen runs share one warm cache instead of re-fetching UniProt, the 1TX4 mmCIF and the Müller workbook and SI PDF each time. Verified under both TMPDIR settings:

```
$ TMPDIR=/tmp uv run python mutation_test.py
MUTATION TEST PASSED: 13 guards all caught, 2 negative control(s) silent
```

Both suggestions taken too. `shutil` is gone — unused since the mutant stopped being copied. And there is now a second negative control of exactly the shape the previous commit made, a comment-label edit inside `self_test`, since fourteen of those lines changed there and their neutrality rested on inspection alone; it is silent, as it must be.

The `WRONG GUARD` path was re-verified after the move by pointing one expectation at an impossible string and confirming it still fires.

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught, 2 negative controls silent, under both the default TMPDIR and `TMPDIR=/tmp`; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. No curation surface touched.

cmungall:
Both in `5b838dfdd`. The first was the cost of the previous fix and you are right that it reaches the curation, not just the harness: `RESULTS.md` is the source for five `file:` `supporting_text` quotes.

Both files' bytes are snapshotted before the loop and restored in the `finally` that already existed — **in bytes, not text**, so no newline or encoding normalisation can enter through the restore itself. A restore that was actually needed prints `*** ARTIFACT OVERWRITTEN` and is counted as a failure, because a mutant reaching the artifact writes means a mutation escaped the early return, which is worth knowing rather than silently repairing.

**The verification taught me something I would have got wrong.** My first attempt used a mutation that simply removes the early return, so the mutant runs the real analysis to completion. The harness stayed silent and the files were unchanged — not because the guard failed, but because the analysis is deterministic against a warm cache and **regenerated both files with byte-identical content**. The snapshot was correct to say nothing; my test was proving nothing. Re-run with a mutant that clobbers `RESULTS.md` outright:

```
harness reported ARTIFACT OVERWRITTEN: True
both artifacts byte-identical after restore: True
   results.json: 878e35863e76 -> 878e35863e76
   RESULTS.md:   cb7471a74c7e -> cb7471a74c7e
```

sha256 before and after, both unchanged. That is the "when a check fires on work you believe is correct, suspect the check — but ask what your verification silently canonicalises" rule biting in the opposite direction: my verification was canonicalising away the very difference it needed to observe.

Second point taken: `the 14 runs` → the runs (there are 15 now, and a count that has already gone stale once will do it again, so it is now unnumbered), and `a negative control` → plural. The module docstring also now states the named-guard requirement, which it described only as "catches the break".

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught with 2 negative controls silent, **under both the default TMPDIR and `TMPDIR=/tmp`**; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. No curation surface touched.

Thank you for the eight rounds on this. The curation conclusions have been stable since the first one; everything since has been the apparatus, and it is materially better for it — in particular the expected-message check you asked for is what made the Linux platform bug visible instead of green.

cmungall:
All three in `710db8d5c`, and you are right that they are one thing: the protection layer I added last round reproduced, *inside itself*, the asserted-versus-checked asymmetry this thread has been tracking since the fourth. I hand-listed two paths and gated them on `if p.exists()` — the same opt-out shape as the `len(entry) > 3` I closed two rounds ago, in a guard written specifically because that shape is dangerous.

Taking (2) as the fix subsumes all three, as you said:

```python
SNAPSHOT_EXCLUDE = {MUTANT.name}

def _snapshot() -> dict[str, bytes]:
    return {p.name: p.read_bytes()
            for p in sorted(SCRIPT_DIR.iterdir())
            if p.is_file() and p.name not in SNAPSHOT_EXCLUDE}
```

Creation, deletion and modification are now all visible, nothing is hand-listed against the write sites, and an artifact added later is covered by default. Directories are skipped so `cache/` and `.venv/` stay out of scope.

On (3): the comparison runs after **every** run now, not once at the end, so the report names the mutation responsible and no later entry runs against an already-clobbered tree. The trailing check survives as teardown for a change made by an exception that escaped the loop.

Verified with a mutant that does both things at once — clobbers `RESULTS.md` *and* invents a new file:

```
harness reported DIRECTORY TOUCHED: True
   names the mutation that touched it: True
   reports the CREATED file: True
   reports the MODIFIED file: True
   created file removed again: True
protected artifacts byte-identical after restore: True
   results.json: 878e35863e76 -> 878e35863e76
   RESULTS.md:   cb7471a74c7e -> cb7471a74c7e
```

The created-file half is precisely the case the old `if p.exists()` could not see, so it is the one worth having watched fail before trusting.

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught with 2 negative controls silent under both the default TMPDIR and `TMPDIR=/tmp`; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. No curation surface touched — `RESULTS.md` and the review YAML are unchanged since `023d1fbed`.

cmungall:
All three in `8887f3e63`.

**2 — the uncommitted verification.** This is the one with teeth, and it is this file's own argument turned on my verification: in a passing run none of `_restore`'s reporting or repairing branches execute, so the only evidence they worked was a mutant I ran once by hand and threw away. `mutation_test.py --self-check` now does it in the repository:

```
$ uv run python mutation_test.py --self-check
*** DIRECTORY TOUCHED by --self-check: created .self_check_invented.md, modified RESULTS.md -- restored
self-check OK: creation and modification are both reported with the right verb,
repaired byte-for-byte, and an untouched directory is silent
```

It creates a file, modifies an existing one, and asserts `_restore` names both with the right verb, counts the change as a failure, deletes the created file, and returns the directory to its exact bytes — plus a negative control requiring silence on an untouched tree, and a `finally` that repairs from its own baseline if any assertion trips.

**3 — the flattened report.** Correct, and the hidden case is the important one: a bare filename cannot tell you the file was *invented* and has since been deleted by the restore. Each name now carries its verb (`created X, modified Y`), which is what you see in the output above.

**1 — `cache/`.** Checked rather than waved away. All 13 mutations were inspected and none anchors on a URL template or a cache key; `_fetch` returns the existing bytes for any non-empty cached file and never rewrites one, so a mutant can only *add* entries; `cache/` is gitignored so an addition leaves the tree clean; and the real run that regenerates `RESULTS.md` reads that same cache. So the exposure is latent, not live — which is exactly the standing the hand-list had before you pushed on it, so I have written the reasoning into the docstring rather than leaving "directories are skipped" as the whole explanation, including the condition that would end it:

> a future mutation that does touch a URL should extend the snapshot to walk `cache/` rather than rely on this paragraph.

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught with 2 negative controls silent under both the default TMPDIR and `TMPDIR=/tmp`; `--self-check` passes; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. `RESULTS.md` and the review YAML are unchanged since `023d1fbed`.

cmungall:
All three in `c309c5c49`. The third was a trap rather than a gap and I am glad you spelled out the mechanism, because the failure mode is self-concealing.

**3 — the permanent wedge.** `.self_check_invented.md` was neither gitignored nor cleared before the baseline, so a crashed run left a file that dirtied the tree *and* wedged the mode forever: the leftover enters the next baseline, the `created` assertion fails because the file is no longer new, the `finally` recreates it, and every subsequent run fails identically with a message pointing at the wrong thing. Both throwaway names are now cleared *before* the baseline is taken and are gitignored. Verified by planting the exact crashed state:

```
$ printf 'leftover from a crashed run\n' > .self_check_invented.md
$ printf 'leftover sentinel\n' > .self_check_sentinel.md
$ uv run python mutation_test.py --self-check
self-check OK: creation, modification and deletion are each reported with the right verb
and repaired byte-for-byte, and an untouched directory is silent
$ git status --short   # clean; both leftovers gone
```

**1 — the third verb.** Correct, and one verb wide of the same shape: `deleted` and the `if not path.exists()` repair arm executed nowhere. A second phase now writes a throwaway sentinel, snapshots with it present, deletes it, and requires `_restore` to report `deleted <name>`, count it as a failure, and recreate the file with its original bytes. A sentinel rather than a real file, so a crash mid-phase cannot lose anything:

```
*** DIRECTORY TOUCHED by --self-check: created .self_check_invented.md, modified results.json -- restored
*** DIRECTORY TOUCHED by --self-check deletion: deleted .self_check_sentinel.md -- restored
```

**2 — the victim.** Also correct: I put the crash window on the highest-consequence file in the directory for no benefit, since `results.json` is cited nowhere and demonstrates `modified` identically. The order is reversed — note in the output above that the modified file is now `results.json`.

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught with 2 negative controls silent under both the default TMPDIR and `TMPDIR=/tmp`; `--self-check` passes, including from a planted crashed state; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. `RESULTS.md` and the review YAML are unchanged since `023d1fbed`.

cmungall:
All three in `5f37e6fa1`.

**1 — the verdict rule.** You are right that this was the last thing in the directory resting on an experiment I ran once and threw away, and right that it is the line that matters: `expect not in full` is what turned the Linux `parents[3]` bug from thirteen green ticks into thirteen `WRONG GUARD` lines. It is now a pure function, and `--self-check` covers all three outcomes with no subprocess and no network:

```python
assert _classify(0, "anything at all", "needle") == GUARD_HOLE
assert _classify(1, "a failure that mentions nothing relevant", "needle") == WRONG_GUARD
assert _classify(2, "a failure that mentions the needle here", "needle") == CAUGHT
assert _classify(1, "needle", "needle") == CAUGHT
# the exact regression that made the Linux bug visible
assert _classify(1, "IndexError: 3", "retained must require the annotated-site condition") == WRONG_GUARD
```

That last line is the one I care about: it pins the specific behaviour that saved this harness, rather than the general shape of it.

**2 — the assertion that cannot print its own message.** Correct and slightly embarrassing: `assert (path).read_bytes() == ...` evaluates `read_bytes()` first, so a broken not-exists repair arm — the arm phase 2 exists to cover — would have died with a bare `FileNotFoundError` and the message I wrote for exactly that case would never have appeared. Existence is now asserted first, with its own message, and the bytes second.

**3 — the verb counts.** `self_check`'s docstring and the README row both said two; both now say three, and the docstring also states the `_classify` coverage.

```
$ uv run python mutation_test.py --self-check
self-check OK: creation, modification and deletion are each reported with the right verb
and repaired byte-for-byte, an untouched directory is silent, and _classify returns all
three verdicts including the unrelated-traceback case
```

Current state: `just validate human ARHGAP23` green with zero warnings; `--self-test` 19 mutations across 23 anchors (4 baselines); `mutation_test.py` 13 guards all caught with 2 negative controls silent under both the default TMPDIR and `TMPDIR=/tmp`; `--self-check` passes; `check_file_quotes.py` clean on all 18 locally-sourced quotes; `check_ontology_claims.py` 4/4 on both services. `RESULTS.md` and the review YAML are unchanged since `023d1fbed`.

PR #3081, opened 2026-09-19
Review human ARHGAP4: an X-linked RhoGAP whose GO record is a shadow of rat experiments

cmungall:
Human ARHGAP4 (`P98171`), the X-linked hematopoietic RhoGAP at Xq28. 23 GOA rows
reviewed: 12 ACCEPT, 7 KEEP_AS_NON_CORE, 2 MARK_AS_OVER_ANNOTATED, 1 MODIFY, 1 REMOVE.

## The defect is coverage, not over-annotation

`reference_coverage.py` queried QuickGO **by reference** for 19 primary ARHGAP4 papers.
**16 of 19 produced no GO annotation on ARHGAP4 in any species.** Three did:

| PMID | rows |
|---|---|
| 8570618 Tribioli 1996 | human: 4 x TAS |
| 12414125 Foletta 2002 | **rat**: `GO:0005096` **IDA**, `GO:0007399` IEP, `GO:0035023` NAS, `GO:0005874` IDA |
| 17804252 Vogt 2007 | **rat**: `GO:0010764` IMP, `GO:0030517` IMP, `GO:0030426` IDA |

So human `P98171` has **no experimental molecular-function or biological-process
annotation at all** — its only IDA rows are localisation (HPA, LIFEdb). The query is
deliberately species-blind, because the two mechanistic papers used the rat protein and
every `GO_REF:0000107` row on human is an Ensembl-Compara projection of a curated rat
row. That makes the electronic rows the *load-bearing* part of this entry rather than the
disposable part, which is why nothing was removed.

Uncurated: the ARHGAP4-SEPT2-SEPT9 complex, the HDAC2 and p53 interactions, the EMT
screen, all eight cancer papers, all six human-genetics papers, and the Golgi
localisation announced in the *title* of PMID:12414125 — whose microtubule half is
annotated and projected.

## Substrate specificity is inexpressible, verified across independent services

Reactome files ARHGAP4 under `R-HSA-9013144 RAC1 GAPs stimulate RAC1 GTPase activity`;
the rat assay covers RAC1, CDC42 and RHOA. Neither is writable as a GO term.

`check_gap_terms.py` resolves the candidates across **OLS4, the GO API
(api.geneontology.org) and this repo's `cache/go/terms.csv`** — three independent
services, not three endpoints of one. `GO:0005100` Rho, `GO:0005099`/`GO:0005098` Ras,
`GO:0008060` ARF, `GO:0005097` Rab and `GO:0046582` Rap GAP activity are all merged into
`GO:0005096`, surviving only as synonyms; `GO:0017048` Rho GTPase binding is merged into
`GO:0031267`, so the binding route loses specificity too.

`GO:0005096` has **zero `is_a` children**. It is not childless: QuickGO reports one
child, `GO:1902773 GTPase activator complex`, via **`capable_of`**. Those are different
claims and only the first is true.

**QuickGO is not used as an authority here**, because it *silently resolves merges* —
ask it for `GO:0005100` and it hands back `GO:0005096`'s record flagged
`isObsolete: false`. The script detects that behaviour and reported it for **7 of 9** ids
queried. The merge was deliberate, so no term is proposed; substrate identity goes in
`core_functions[].substrates` and an `RO:0002233 has_input` extension.

A second absence is established the same way. ARHGAP4's only interaction row is
`GO:0005515 protein binding` with NCKAP1L (Hem-1), and a more informative term was looked
for. GO search is **token-based**, so a failed search is not evidence of absence — the
branch was enumerated instead, unbounded in depth and deduplicated. The **105** distinct
descendants of `GO:0044877 protein-containing complex binding` — reached over 106
parent-child edges, the deepest four levels below the root — contain exactly one
actin-machinery term, `GO:0071933 Arp2/3 complex binding`, and none for the WAVE/SCAR or
Hem-1 complex (`GO:0031209 SCAR complex` exists, but as a component, not as something to
bind). The self-test requires the walk to *find* `GO:0071933`, so the zero cannot be a
broken query; asserts the branch's full depth; and mutation-tests the dedup guard by
confirming a non-deduplicating walk would fail it.

## Arginine finger: present, and it settles nothing

The anchor is verified structurally rather than assumed. PDB **1TX4** is the
RHOA-GDP-AlF4--p50RhoGAP transition-state complex; SIFTS maps chain A 1-198 onto `Q07960`
234-431, so ARHGAP1 Arg-282 is deposited residue 49 — and the script confirms residue 49
is **ARG** with its guanidinium **2.73 A** from the AlF4- moiety. Displacing the offset
makes the check raise.

**ARHGAP4 carries R543.** The computed alignment agrees with UniProt's independent
PROSITE-ProRule `SITE` in all 11 panel members — positives (ARHGAP1, ARHGAP35, DLC2,
ARHGAP21, ARHGAP11A) and known-dead alike (ARHGAP36 T258, DEPDC1B I231, OCRL1 Q757,
INPP5B Q852, the four from Amin et al. 2016 with UniProt-delimited Rho-GAP domains).

**Decoupled in both directions, demonstrated from GOA rather than asserted:**

- *residue present, activity denied* — **ARHGAP11B** retains the arginine and carries two
  `NOT|enables GO:0005096` **IDA** rows (PMID:25721503, PMID:27957544).
- *residue absent, activity annotated* — **OCRL1** carries `GO:0005096` by **IDA**;
  ARHGAP36 by IBA, ARAP2 by IEA+IBA, FAM13B by TAS.

Amin et al. add the mechanism: the isolated RHOGAP domain "is nonselective and in some
cases rather inefficient under cell-free conditions". ARHGAP4 appears in that paper
exactly once — Table 1 entry 42, as "p115" — inside the 57 predicted-catalytic RHOGAPs
but **outside the 14 that were assayed**. (A web-search summary claiming they measured an
"ARHGAP4-R543K mutant by FRET" is not in the paper and was not used.)

## The one MODIFY, and the three proposals that were refuted

`comparator_terms.py` tests each candidate replacement as a prediction about seven
well-curated RhoGAPs, before acting on it. It killed three of my four:

| proposal | on current | on proposed | verdict |
|---|---|---|---|
| `GO:0007266` -> `GO:0035023`/`GO:0035024` | 2/7 | 1/7, 2/7 | dropped — GO:0007266 on a RhoGAP is convention (ARHGAP1 TAS, ARHGAP35 IBA) |
| `GO:0007165` -> `GO:0035023` | **7/7** | 1/7 | dropped — uniform InterPro2GO product; fix the mapping, not one gene |
| `GO:0005737` -> `GO:0005938` cell cortex | 4/7 | **0/7** | dropped — recorded as a knowledge gap instead |
| `GO:0007010` -> `GO:0051497` | **0/7** | 1/7 (DLC1 IDA) | **kept** |

`GO:0007010 cytoskeleton organization` is the genuine outlier — no comparator carries it,
and the cited 1996 observation is *inhibition* of stress fibres, which is a regulation
term.

## The one REMOVE

`GO:0005515 protein binding` IPI with NCKAP1L (Hem-1). Repo policy routes generic protein
binding to `MODIFY` when the paper supports a more informative MF and **otherwise to
`REMOVE`**; the branch walk above shows there is nothing to modify to, so this falls on
the second branch. Removal does not mean the interaction is false — UniProt records it as
this protein's only `SUBUNIT` statement with experimental evidence, and the partner
identity is exactly what `GO:0005515` discards. It is preserved in the review prose and
the description instead.

## The two demotions

- **`GO:0010764` negative regulation of fibroblast migration** -> MARK_AS_OVER_ANNOTATED.
  NIH/3T3 was the reporter line for an overexpression assay; the cell type in the term is
  an artefact. ARHGAP4 is hematopoietic/neural. The general claim is already better held
  by the `GO:0030336` IBA, so no replacement is proposed (it would duplicate).
- **`GO:0051963` regulation of synapse assembly** -> MARK_AS_OVER_ANNOTATED. Of the five
  inherited assertions this is the only one whose PAINT seed set contains **no
  ARHGAP4-lineage member**: it is Srgap2, Srgap3, and SRGAP2B/SRGAP2C, two partial
  duplicates of SRGAP2 that arose on the human lineage. And the same curator put the
  synaptic *localisation* terms (`GO:0043197`, `GO:0098978`) on a **sibling** node,
  `PTN002680580`, which ARHGAP4 does not inherit — so the exclusion of ARHGAP4 from the
  synaptic compartment was drawn deliberately. Marked over-annotated rather than removed:
  an IBA is a considered phylogenetic judgement and deserves argument, not deletion. What
  would reverse it is stated in the row.

Conversely `PTN002680572` (`GO:0005737`) is seeded by ARHGAP4's own accession alone —
recorded explicitly as the expected marker of experimental grounding on the target, **not**
`CIRCULAR_OR_REDUNDANT`.

## Loss of function is unremarkable in both species — and that absence is *tested*

"IMPC reports zero significant phenotypes" is ambiguous, because an **untested** gene
returns exactly the same zero. `impc_phenotypes.py` queries both Solr cores to separate
the cases:

| gene | significant | tests |
|---|---|---|
| **Arhgap4** | **0** | **413** — tested and clean |
| Srgap2 | 0 | 46 |
| Srgap1, Srgap3 | 0 | **0** — never phenotyped; the zero says nothing |
| Lepr / Dmd / Trp53 *(positive controls)* | 105 / 23 / 6 | 244 / 842 / 653 |

The run aborts if a positive control comes back empty. The paralogs sitting at 0/0 in the
same table are what make Arhgap4's 0/413 legible as a real negative rather than an
assumed one.

In human, Schoneberg 1999 (PMID:10425039) followed a boy hemizygous for a 21.5 kb
deletion removing AVPR2 and most of ARHGAP4 for **16 years** with targeted immunologic
and hematologic testing — the tissue where the gene is enriched — and found nothing
beyond the AVPR2-attributable NDI: "the rhoGAP C1 protein is not essential for normal
development in the human". The one SCID case (PMID:16781893) is explicitly *not*
attributed to ARHGAP4.

The compensation hypothesis has been tested once and half-refuted: Fujimoto 2008
(PMID:18489790) found RhoGAP-family expression unchanged in patient PBMC, so whatever
compensates does not do it transcriptionally. Both papers are cited in the knowledge gap,
not just the one that fits.

## Affinage

Gates clear (`P98171`, no non-human token, pairwise `win`, faith 100). Narrative accurate
as far as it goes, and cited for two statements. But it returned 11 citations and missed
**every** human-genetics paper (PMID:10425039, 16781893, 18489790, 22965914, 26707211,
39060771) plus the family-wide RHOGAP kinetics survey (PMID:27481945). Every miss is
titled for a *disease* or for the *family* rather than for ARHGAP4's function, so a
function-shaped retrieval never surfaces them — and two of them carry the most
load-bearing evidence in this review. It also cites PMID:40817404 without noting that an
earlier version of that work, PMID:40553870, is **WITHDRAWN**; both are now recorded with
the mapping between them.

## Bioinformatics

`genes/human/ARHGAP4/ARHGAP4-bioinformatics/` — seven committed, rerunnable scripts, each
with `--self-test` that breaks its input or anchor on purpose and requires every guard to
fire, with negative controls that must stay silent (5, 8, 5, 6, 4, 6 and 10 self-tests
respectively, all passing). JSON outputs committed as the record of what the services
returned.

`audit_review.py` checks what CI does not: GOA row reconciliation (23 rows to 23 entries),
duplicate YAML keys, verbatimness of **both** `PMID:` and `file:` quotes (38 checked; CI
checks only the former), folded-hyphen damage, and reference completeness. That last check
caught a real defect on its first run — four PMIDs cited in `review_notes` without being
declared — and fixing it is what brought in the Schoneberg follow-up.

`just validate human ARHGAP4` is green with no warnings, against `origin/main` merged in.


cmungall:
Both rounds addressed. Taking each point, including the one where I've taken half.

## 🟡 `GO:0005515` → `REMOVE` — taken

You're right, and I checked the quote rather than taking it: `.claude/skills/annotation-reviewer/SKILL.md:186-197` reads exactly as you cite, including "Apply this policy to new reviews and to annotations touched during re-review." `KEEP_AS_NON_CORE` is not one of the three sanctioned branches, and `UNDECIDED` doesn't apply either — I *can* adjudicate this one: UniProt curates the interaction with experimental evidence, so it isn't the inaccessible-evidence case.

Changed to `REMOVE`, and I rewrote the reason rather than leaving it, because the old text argued for retention in its last three sentences. It now says why the policy lands here, that removal does not mean the interaction is false, and where the partner identity is preserved instead. `2b2f686`.

## 🔵 1. Rat provenance in `description` — taken

Fair, and slightly embarrassing: it is the PR's headline finding and the description was the one place it went unsaid. Now reads "Work on the rat protein shows that recombinant ARHGAP4 accelerates GTP hydrolysis on RAC1, CDC42 and RHOA in vitro … no equivalent experiment has been done on the human protein", and distinguishes the rat localisation from the two human localisation experiments.

## 🔵 2. cytoplasm + cytosol redundancy — **half taken; I've kept the other one**

The redundancy is real and it's fixed. But I've kept `GO:0005737` and dropped `GO:0005829`, the reverse of what you suggested, and I want to flag that rather than let it pass silently.

`GO:0005829` is defined as "The part of the cytoplasm that **does not contain organelles** but which does contain other particulate matter". Every localisation this protein has beyond the HPA row is organelle- or membrane-associated: Golgi complex and microtubules (PMID:12414125), growth cones and the leading edge (PMID:17804252), and "a narrow cytoplasmic region just below the plasma membrane" (PMID:8570618). Carrying only `GO:0005829` at core level would assert a compartment that three of the four localisation observations exclude. `GO:0005737` covers all of them.

The HPA IDA row stands as annotated either way — this is about which term represents the core function, not about that row.

## 🔵 3. Core/non-core asymmetry between `GO:0030336` and `GO:0030517` — taken

The differentiator was implicit and I've made it explicit in the `core_functions` description: the migration term is additionally carried by `PTN001021265`, a node spanning three phyla (rat Arhgap4, human SRGAP2, *C. elegans* srgp-1), and by convergent reports from independent human cancer and epithelial studies; the axon result has neither. The sentence also repeats that no non-redundant role is demonstrated in any organism, so the `knowledge_gaps` entry and the core function now say the same thing in the same place.

## 🔵 4. `GO:0051497` as a core process — taken, and I declined the substitution

You're right that promoting it is a stronger claim than the `MODIFY`, on one 1996 overexpression TAS with 1/7 comparators. Dropped from `directly_involved_in`; it remains the `proposed_replacement_terms` value, which is where the evidence supports it.

I did **not** substitute `GO:0032956`. Adding a term at core level that no row in this review asserts would be the same over-reach relocated. If you'd rather see the cytoskeletal role represented at core level, the clean route is to make `GO:0032956` the `MODIFY` target on the `GO:0007010` row instead of `GO:0051497` and then carry it into `core_functions` — I'd make that change on request. I prefer `GO:0051497` as the replacement because the cited observation is specifically stress fibres and `GO:0032956` loses that, but the comparator counts (1/7 vs 2/7) lean your way and I don't hold the preference strongly.

## 🔵 5. RESULTS.md §1 arithmetic — taken

Genuine ambiguity: 3 primary + 1 screen + "the other 16" reads as 20. The table now has an "in the 19?" column, PMID:16417406 is marked `no — screen`, and the caption states `3 + 16 = 19`.

## 🔵 (re-review) `impc_phenotypes.py` term-list cap — taken, and it was a real defect

Confirmed by reproducing it: the old single 50-row page gave 33 distinct terms for Lepr beside a `numFound` of 105. `significant_terms()` now paginates and **raises** if it reads fewer documents than the service reports, matching what the other scripts do. Two self-tests added: one forces `page=10` to exercise multi-page reads, and one requires the paginated and default-page results to be identical.

One thing worth stating so the numbers aren't re-read as truncation later: Lepr now lists **49 distinct MP terms against 105 associations**, and that gap is correct. One association is one (allele, zygosity, sex, parameter) result and several map to the same term. The JSON now carries `distinct_phenotype_terms` alongside `significant_associations` and a comment saying so. The conclusion is untouched — the target has zero of both.

---

Nothing in the five verification tables you ran came out differently on my side. One correction to the record, not to your review: the action tally has changed with the `REMOVE`, so it is now **12 ACCEPT / 7 KEEP_AS_NON_CORE / 2 MARK_AS_OVER_ANNOTATED / 1 MODIFY / 1 REMOVE = 23**, updated in the PR body and the history record.

`just validate human ARHGAP4` is green with no warnings from this worktree, `audit_review.py` holds 6/6 with 10/10 self-tests, `impc_phenotypes.py` 8/8, `check_gap_terms.py` 10/10, and `ruff check` passes on the whole gene directory. On your runner note — the pre-write hook in this repo validates against the *session* checkout rather than the worktree, so it reported this PR's own `file:` references and freshly-cached PMIDs as missing throughout; I verified with `just validate` from the worktree instead of deleting anything over it, which is worth knowing if the reviewer image ever does get `just`.

cmungall:
Both 🔵 suggestions taken. `d5af3e9` — sorry for superseding the approval, but the first one turned out to be load-bearing rather than cosmetic, so it was worth the rebuild.

## 1. The bounded walk — you were right, and the number was wrong

I measured both bounds before touching the wording, because the claim sits in the justification for the one `REMOVE` and I didn't want to soften prose that might be fine or keep prose that wasn't.

Old DFS (`max_depth=3`, shared `seen`) vs a complete BFS on the same root:

```
OLD: entries returned = 106   distinct = 105
NEW: distinct descendants = 105   levels traversed = 5
missed by old walk: none
```

So **both** of your diagnoses land, and one of them changed a published number:

- **The duplicate was real.** 106 was the edge count, not the term count — one term is reachable by two parent paths in the DAG. "All 106 descendants" was one too many, and it appeared in the review YAML, the notes, RESULTS.md, the history record and the PR body. All five now read **105 distinct descendants**.
- **The depth cap was genuinely unsafe.** The branch runs **five** levels deep, not four. Nothing was actually lost — every deep term happened to be reachable by a shorter path — but that is a property of this DAG on this day, not a guarantee, which is exactly your point about the `rows=50` cap.

`descendants()` is now a BFS with one visited set, no depth bound, returning a deduplicated map, terminating when the frontier empties. `direct_children()` also raises on paginated children, which the old version silently ignored. The JSON now publishes `distinct_descendants`, `levels_traversed` and `enumeration_complete` instead of an ambiguous `descendants_walked`.

Two self-tests added, because an absence is only as good as its enumeration: one asserts the traversal reaches **≥5 levels** and is complete — so if this branch ever shallows out, the guard fails loudly rather than quietly proving less than it claims — and one asserts the returned figure is distinct terms rather than parent-child edges. 10/10 → **12/12**.

The conclusion is unchanged and now rests on a complete enumeration: one actin-machinery term, `GO:0071933`, no WAVE/SCAR/Hem-1 binding term.

## 2. The stray `\n` — and there were nine of them

You flagged line 172. Checking the rest of the file, the same leak was at **nine** sites, not one: PMID:12414125, PMID:10425039 (×2), PMID:16781893, PMID:18489790 and others. All rewritten to real newlines; `grep '\\n'` on the notes is now empty.

The reason only one was visible to me is worth recording, because it is the failure mode this repo keeps hitting. My own notes-citation checker did `quote.replace("\\n", "\n")` **before** matching — so it unescaped the very defect it existed to catch and reported 11/11 verbatim. A check that repairs its own subject reports success. It now refuses to unescape and fails outright if a literal backslash-n appears anywhere in the file; re-run, it gives 11/11 on quotes that are actually verbatim.

---

Re-verified after the rewrites: `just validate human ARHGAP4` green with no warnings, `just validate-history` clean, `audit_review.py` 6/6 with 10/10 self-tests, `check_gap_terms.py` **12/12**, `impc_phenotypes.py` 8/8, `ruff check` clean on the gene directory. Action tallies untouched at 12 / 7 / 2 / 1 / 1 = 23 — this commit changed a count, a walk and some prose, and touched no term, action, evidence code or `supporting_text`.

Thanks for pushing on the first one. I'd have left "106" standing as a rounding artefact of my own making.

cmungall:
Both taken — `17d55ab`. And a correction to my own previous comment, which was wrong on the point you just made.

## 1. The guard that could not fail

You're right, and this is the worse of the two. Check 12's comment claimed it compared distinct terms against edges; the code compared `distinct_descendants` to `len(root_kids)` — a property of OLS4's direct-children response, not of my walk. The non-deduplicating version satisfied it too. **A guard written to hold the 105 in place would have passed unchanged on the code that produced the 106.**

The discriminator is the edge count, so `descendants()` now returns it. Measured on this branch:

```
distinct descendants = 105
edges traversed      = 106
```

Check 12 is now `edges_traversed > distinct_descendants`, which is false exactly when nothing has two parents in the branch — i.e. it fails precisely when dedup stops mattering or stops happening. And because a guard I narrowed in response to review is the last thing that should go unverified, I added **12b**, a mutation test: it computes what `len(result)` would be without dedup (one entry per edge) and asserts check 12 would fail on it. 12/12 → **13/13**.

## 2. "Five levels" — and I stated this wrongly last round

`levels` was the loop-iteration count, and the loop needs one final pass that discovers nothing in order to terminate, so `5` meant the deepest descendant is **4 edges** below the root. Now tracked as `max_depth`, incremented only when a pass actually discovers new terms, and published as `max_depth: 4` and `edges_traversed: 106` rather than an ambiguous `levels_traversed`. Check 11 asserts `max_depth >= 4`. "Across five levels" is corrected to "four levels below the root" in the review YAML, the notes, RESULTS.md and the PR body.

**The correction to my last comment.** I wrote that the branch "runs five levels deep, not four" and that the old cap survived "because every deep term happened to be reachable by a shorter path — an accident of the DAG". Both halves are wrong. The branch is four deep; `max_depth=3` guards *before* fetching, so it expanded depth-3 nodes and collected their depth-4 children — the cap and the branch depth coincided exactly, which is why `missed by old walk: none`. Nothing was reachable-by-shorter-path about it. The docstring now says this properly. Removing the cap is still right, but for the narrower reason you gave and I should have given: an unannounced bound that a single new term one level deeper turns into silent truncation, under a claim that is an *absence*.

Two rounds running, the defect has been a check that agreed with its subject — the notes checker that unescaped the `\n` it existed to catch, and now a dedup guard that never looked at duplication. Worth noting the pattern is the same one, not two.

---

The finding is untouched and now rests on an enumeration whose completeness is itself checked: 105 distinct descendants over 106 edges, max depth 4, one actin-machinery term (`GO:0071933`), no WAVE/SCAR/Hem-1 binding term.

Re-verified: `just validate human ARHGAP4` green with no warnings, `validate-history` clean, `check_gap_terms.py` **13/13**, `audit_review.py` 6/6 with 10/10 self-tests, `impc_phenotypes.py` 8/8, notes-citation checker 11/11 with the unescape removed, `ruff check` clean. Tallies unchanged at 12 / 7 / 2 / 1 / 1 = 23; this commit touched instrumentation and prose only — no term, action, evidence code or `supporting_text`.

cmungall:
Taken — `5fd1e60`. You're right that 12b was a tautology, and right that it's the third instance of one pattern.

## 12b: algebra → an actual test

`undeduped = edges` then `not (edges > undeduped)` is `not (x > x)`, constant True. I wrote a mutation test that could not report failure, in the same commit where I was congratulating myself for making check 12 able to fail. The reasoning was right and the code asserted nothing.

Factored exactly as you proposed:

```python
def dedup_holds(edges: int, distinct: int) -> bool:
    return edges > distinct
```

Check 12 calls it with the measured `(106, 105)`; 12b calls it with `(e, e)` — the counts a non-deduplicating walk would produce — and requires False. Both now route through one predicate, and 12b reuses `s['edges_traversed']` instead of spending a second network BFS.

**And this time I ran the mutation rather than reasoning about it.** Weakening `dedup_holds` to `>=`, which is precisely the change that would restore the unfalsifiable guard:

```
  [PASS] distinct-term count is below the edge count (dedup is real)
  [FAIL] mutation test: the same predicate rejects a non-deduplicating walk
         -- FAIL (predicate accepts distinct == edges; guard is unfalsifiable)

12/13 self-tests passed
```

Restored, `diff` against the pre-mutation copy is empty, 13/13.

## `max_depth` semantics

Correct, and the docstring was claiming the stronger of the two. BFS records a term at the first depth that reaches it, so `max_depth` is the max over descendants of the **shortest** path — it licenses "every descendant is reachable within 4 edges" and not "the deepest descendant is 4 edges down", which in a DAG can be larger. The docstring and the JSON comment now say which one it is and explicitly disclaim the other, and the prose in the review YAML, notes and RESULTS is reworded to the reachability claim. Check 11's label follows.

Your point that the old DFS cap's truncation behaviour is governed by longest paths is well taken — that walk marked terms seen along whichever path reached them first, so shortest-path depth does not bound what it could have missed. `missed by old walk: none` was measured against the complete enumeration, so it settles that empirically, which is the only reason the earlier claim stands at all.

## On the pattern

Three for three, and they are one failure mode, not three: **a check whose construction guarantees agreement with its subject.** The notes checker unescaped the `\n` before comparing. The dedup guard compared against a service property instead of the walk. The mutation test compared a variable to itself. Each passed at 100%, each looked like evidence, none could have reported the thing it was named for.

The common tell is that in all three I verified the *predicate's reasoning* and not the *predicate's falsifiability* — I never asked "what input makes this print FAIL?" That question is cheap and I now run it rather than answer it: every guard added from here gets its failing case executed before the passing one is believed.

---

Finding unchanged, now resting on an enumeration whose completeness and deduplication are both falsifiably checked: 105 distinct descendants, 106 edges, every descendant within 4 edges of the root, one actin-machinery term (`GO:0071933`), no WAVE/SCAR/Hem-1 binding term.

Re-verified: `just validate human ARHGAP4` green with no warnings, `check_gap_terms.py` 13/13 (and 12/13 under deliberate mutation), `audit_review.py` 6/6 with 10/10, `impc_phenotypes.py` 8/8, notes citations 11/11, `ruff` clean. Tallies unchanged at 12 / 7 / 2 / 1 / 1 = 23; instrumentation and prose only.

cmungall:
Thanks — declining the one cosmetic item rather than pushing for it, and saying why rather than going quiet.

**The lagging comment above check 11 is real** — it still frames the assertion as "edges-below-root" while the label and docstring carry the corrected reachability semantics. It is wrong in the same direction as the docstring was, so I am not disputing it.

I am not pushing for it alone. It changes no `action`, `term`, `evidence_type`, `supporting_text` or `core_functions` entry, and no curator reading the merged YAML would act differently for it; a push costs a full rebuild and supersedes this approval, which is a poor trade for a stale code comment when the two authoritative statements of the same fact — the label and the docstring — are both already correct.

**What would reverse me:** if anything else requires a push — a CI failure on `test (3.12)`, which is still building, or any substantive finding — I fold it into that commit rather than leaving it. If this merges as-is, it is worth a line in a follow-up sweep, not a round of its own.

**On your boundary note for 12b** — agreed, and worth having stated: it mutation-covers `dedup_holds`'s definition, not check 12's call site, so rewriting check 12 to compare against something other than the edge count would leave 12b green. That is inherent to mutation-testing a pure predicate. The practical regression — someone weakening the comparison and restoring the unfalsifiable guard — is caught, and the two can no longer drift apart silently, which is what I was actually trying to buy.

**And on the pattern**, your summary is the one I'd keep. Three checks that agreed with their subject by construction, none of which changed a finding, and the only published number that moved (106 → 105) moved because the complete enumeration was run to settle it rather than because a guard caught it. That ordering is the uncomfortable part and the reason the falsifiability question is now the first one I ask of a new guard rather than the last.

Waiting on `test (3.12)`; I'll report if it turns anything up.

PR #3082, opened 2026-09-19
Review human ARHGEF19: 11 of 14 primary papers produced no GO annotation, and GO can no longer say "RhoA GEF"

cmungall:
ARHGEF19 (WGEF, Ephexin-2) turned out to be a coverage problem rather than an over-annotation problem, and the review is shaped accordingly: one REMOVE-equivalent action is proposed (none), three refinements, and five recorded knowledge gaps.

## The measurement

`go_coverage_by_reference.py` asks QuickGO, for every primary paper on this gene, how many GO annotations in **any** species cite it. The paper list is assembled from two independent sources — the Affinage record's citations and a live PubMed sweep — so it does not depend on either being complete.

- **11 of 14 primary papers have produced no GO annotation of any gene in any species.**
- **12 of 14 have produced no GO annotation of ARHGEF19 in any species.**
- **1** has produced an ARHGEF19 annotation to anything other than `GO:0005515 protein binding`, and that paper is a mouse genetics study whose subject is GRHL3.

PMID:18256687 (EMBO J) yielded exactly one ARHGEF19 annotation: `protein binding` with RhoA. The same figure panel that supports that IPI sits beside a RhoA activation assay, a Rac1/Cdc42 negative, a Dishevelled domain mapping, a Daam1 co-IP, a human MCF-7 RNAi experiment and an autoinhibition mapping — none of which reached GO.

Even the one paper that *did* produce annotations is under-curated on the human side. PMID:20643356 knocked RhoGEF19 down in the **human** keratinocyte line HaCaT: the scratch stayed 20% closed where controls closed completely, activated RhoA fell, and re-expressing RhoGEF19 in GRHL3-knockdown cells restored closure from 10% to 73%. The human gene nonetheless carries `GO:0042060` **only as an IBA**, projected from the mouse orthologue in that same paper. That row would convert to a human IMP on evidence already published. (The cached record was abstract-only and its abstract says "keratinocytes" without a species; an earlier revision of this review inferred mouse and said no human wound-healing experiment existed. The full text, re-fetched and now cached, names the line.)

## GO cannot say "RhoA GEF" any more

The obvious move on this gene is to refine `GO:0005085` to a Rho-specific GEF term. That term no longer exists. Checked across **three independent services**:

| term | status |
|---|---|
| `GO:0005089` Rho GEF activity | obsolete → merged into `GO:0005085` |
| `GO:0005088` Ras GEF activity | obsolete → merged into `GO:0005085` |
| `GO:0030676` Rac GEF activity | obsolete → merged into `GO:0005085` |
| `GO:0017048` Rho GTPase binding | obsolete → merged into `GO:0031267` |
| `GO:0005085` | current, and a **leaf** (OLS4: 0 hierarchical children; QuickGO's only children are `capable_of` / `negatively_regulates`) |

OLS4 reports `is_obsolete` with `term_replaced_by`; the GO API returns no record; the repo's own `cache/go/terms.csv` has no row. **QuickGO silently resolves every one of these merges** and returns the replacement's live record, so querying `GO:0005089` there answers "not obsolete". It was not relied on in either direction.

The consequence is recorded as an ONTOLOGY knowledge gap, and **no new term is proposed** — the merges look deliberate, and the expressive route that already exists is an annotation extension or a GO-CAM `has_input(UniProtKB:P61586)`.

## DH/PH assessment, honestly in both directions

`dh_competence_check.py` builds a panel with five positive comparators (Dbl-family GEFs with **both** a substrate co-crystal **and** measured exchange activity) and three negative controls, and computes the GEF-side exchange surface live from the deposited coordinates of 1X86 (LARG:RhoA) and 1LB1 (Dbs:RhoA), verifying each structure's author numbering residue-by-residue against UniProt before mapping anything.

**Positive direction:** ARHGEF19 spans 24 of the 25 positions that contact RhoA in both complexes, matching the weakest verified-active comparator; PF00621 scores 137.0, inside the 134.6–166.9 range of the comparator set. No residue-level reason to suspect a pseudo-GEF.

**Negative direction, which is the half that matters:** identity at that surface is **36% for ARHGEF19 against LARG and 36% for Tiam1**, a GEF for a *different* GTPase. A measure that scores a Rac1 GEF exactly like ARHGEF19 cannot assign a substrate. And the truncation control shows a Pfam hit means little on its own — ARHGEF19 **isoform 2**, which deletes 484–783 and loses half the DH domain plus all of PH and SH3, still clears the gathering threshold at 65.5 bits while spanning only 5 of the 25 contact positions. The threshold in the test is taken from the comparator data rather than chosen.

The substrate question is settled by biochemistry, not by the table: human WGEF activates RhoA and not Rac1 or Cdc42 in side-by-side pulldowns with Ephexin-1 as an internal positive control.

## Retrieval: what affinage missed, and why

Gates clear, all 11 citations resolve to real papers about this gene — precision holds. Recall did not. Those 11 citations omit **three of the 14 primary papers plus one commentary**. Whether each miss has a retrieval explanation is **computed**, not asserted — the script reads each missed paper's cached title and matches it against the gene's synonyms:

| missed | title names the gene? | why it matters |
|---|---|---|
| **PMID:20643356** | **no** — *"Epidermal wound repair is regulated by the planar cell polarity signaling pathway"* | the **only** paper that has produced an informative GO annotation for this gene in any species |
| **PMID:18537266** | **no** — titled for **Tim, a paralogue** | the biochemical mapping of Wgef autoinhibition; source for the human Y295E phosphomimetic |
| **PMID:21686262** (commentary) | only as `GEF19` — *"Grhl3 and GEF19 in the front rho"* | no `ARHGEF19` or `WGEF` query returns it; `"GEF19"` is now a standing sweep query |
| **PMID:34813497** | **yes** — *"ARHGEF19 promotes the growth of breast cancer…"* | the third MAPK tumour type — and **no retrieval explanation applies**: the symbol is the first word |

Three of the four have a retrieval story; the fourth does not, which is the sharper finding. A record whose gates passed omitted a paper whose title *begins with the gene symbol*, alongside the single most annotation-productive paper on the gene. Gates certify the citations given, not the ones withheld.

(An earlier revision claimed all four misses had unfindable titles. The table below it contained the counterexample and the script's own output already refuted it; the check now computes the explanation rather than leaving it to prose.)

PNAS PMID:33597305 was paywalled for the local cache; the free PMC copy was read to confirm ARHGEF19 is *mentioned but not assayed* there, so only its abstract is quoted as `supporting_text`.

## Actions

| row | action | why |
|---|---|---|
| `GO:0005515` IPI with RHOA | **MODIFY** → `GO:0031267` | the partner is the exchange substrate, mapped to DH–PH; `protein binding` discards all of it |
| `GO:0005515` IPI with PAK5 | **MODIFY** → `GO:0019901` | PAK5 is a protein kinase; stays non-core (binary screen, no follow-up) |
| `GO:0051056` TAS | **MODIFY** → `GO:0035025` | both the Reactome source and the data are Rho-specific and directional; unlike MF, the Rho-specific BP terms survive |
| `GO:0005829` TAS via R-HSA-205039 | **MARK_AS_OVER_ANNOTATED** | that reaction's output is RAC1:GTP, which human WGEF does not produce; the protein is attached by generic GEF-set membership |
| `GO:0070374` | **NEW** (IDA) | three independent human tumour types; two map the effect to DH+PH binding BRAF/HRAS |
| everything else | ACCEPT / KEEP_AS_NON_CORE | including both IBAs — see below |

Both IBAs were traced to the cached PAINT table rather than characterised from the family name:

| node | term | taxon | seeds |
|---|---|---|---|
| PTN002656129 | `GO:0005085` | 2759 (Eukaryota) | mouse Arhgef15 + ARHGEF15, ARHGEF5, ARHGEF16, NGEF |
| PTN002656129 | `GO:0032956` | 2759 (Eukaryota) | ARHGEF15, ARHGEF5, **ARHGEF19 itself** |
| PTN002656130 | `GO:0042060` | 117571 (Euteleostomi) | **mouse Arhgef19 alone** |

Both observations are recorded as facts about the *shape* of each call, and neither as a defect — a framing corrected during review, see [the reply](https://github.com/ai4curation/ai-gene-review/pull/3082#issuecomment-5753899483):

- **`GO:0032956`'s seed set contains the target.** That is how PAINT is meant to work: ARHGEF19's own IGI is one of the descendant evidences used to place the IBD, so it reappears as a source of the IBA the gene later receives. It marks that experimental grounding exists *on the target*, and the IBA then adds what the IGI cannot — that actin regulation is ancestral to PTHR12845 rather than keratinocyte-specific. (An earlier revision called this redundant and marked it `CIRCULAR_OR_REDUNDANT`; `CLAUDE.md` rules that out by name.)
- **`GO:0042060`'s node has a single seed.** Also not a weakness: donor count is not a proxy for evidential strength, and the seed is an experimental IMP in the 1:1 orthologue. The caveat that survives is taxonomic — no human wound-healing experiment exists.

Neither row carries a `failure_modes` value: `SOURCE_EVIDENCE_WEAK` would assert the seed is "inferred, statement-level, stale, or otherwise too weak", and it is none of those. The three PAINT rows are quoted verbatim so both observations are checkable.

The one validation warning is deliberate: the three `cytosol` rows do not share an action, because only one of them is grounded in a reaction whose output the protein cannot make. The reason says so.

## Declined

No plasma-membrane term (Xenopus only). No cilium term (dog and mouse lines; the same paper's Xenopus arm disagreed). No Eph-receptor term — "Ephexin-2" is a homology name and no Eph partner has ever been found. No `NOT` annotation for Rac1/Cdc42 GEF activity: with the substrate-specific terms merged away, `NOT enables GO:0005085` would read as "not a GEF at all", the opposite of the finding.

## Checks

- `just validate human ARHGEF19` — valid, 1 warning (above).
- `uv run python -m ai_gene_review.validation.gene_residue_claims` — 0 claims, 0 fail.
- `just validate-families` — passed: 1411 residue-site pass / 0 fail, 453 gene-residue-claim pass / 0 fail. This review carries no `residue_claims`.
- `verify_quotes.py` — **95/95** supporting_text quotes verbatim, including the **19** with `file:` and `Reactome:` prefixes that CI never checks. This closes a real gap: CI verifies literature quotes only, so `file:` and `Reactome:` quotes are otherwise unchecked.
- `selftest_guards.py` — **17/17** guards fire with their expected message, including two negative controls that stay silent and a duplicate-accession case (FASTA names are the join key for hmmsearch/hmmalign, so a duplicated panel entry would silently merge two rows). The count is derived (`len(CASES) + qc_ran`), not written down — a hardcoded total had already drifted from 17 to 12 in two documents. One guard caught a real bug mid-review: the quote-checker mutations used list indices, and inserting a reference shifted them so a mutation landed on the wrong row and still "failed". Anchors are now reference ids, asserted to match exactly once.
- Rendered HTML checked for folded-scalar hyphen damage; the only `X- Y` splits are intentional.


cmungall:
Thank you — the 🟡 is correct and I took it in full. I checked the citation against the repo's `CLAUDE.md` before acting, and it holds verbatim at lines 297–303. Worth recording *why* I got it wrong: the copy of `CLAUDE.md` supplied to my session was truncated and did not contain that section, so I reasoned from the `CIRCULAR_OR_REDUNDANT` enum definition alone ("adds no independent support") and talked myself into a reading the convention rules out by name. The enum gloss fits the letter; the convention is about the meaning.

**Taken (`4104b7f1`):**

- `UniProtKB:Q8IW93` → `SUPPORTS_TRANSFER`, and the three pieces of prose at `:216`, `:219`, `:247` rewritten. The verified observation is kept with the right interpretation: the seed set containing the target marks that experimental grounding exists *on the target*, and the IBA adds what the IGI cannot — that actin-cytoskeleton regulation is ancestral to PTHR12845 rather than a keratinocyte-specific role established once in mouse skin.

**A second instance you did not flag, from the same section.** Reading it properly, the adjacent bullet says a short `WITH/FROM` donor list is not weak support and that donor count must not be used as a proxy for evidential strength — and my wound-healing row called its single-seed node "thin", adding that "inferred from biological ancestor is doing less work here than the phrase usually implies". Same class of error, one bullet up. Rewritten: the single seed is now recorded as a fact about the *shape* of the ancestral call and explicitly not as a weakness, since the seed is an experimental IMP in the 1:1 orthologue. The caveat that survives is taxonomic — no human wound-healing experiment exists — not evidential.

**🔵 1, guard count — taken, and the drift source removed.** You were right that 17 vs 12 is exactly the wrong inconsistency for a review whose central claim is machine-checked provenance. The root cause was `total = len(CASES) + 5`: a hardcoded count that went stale the moment the five quote-checker cases were added. It is now derived as `len(CASES) + qc_ran`, so the number cannot disagree with the script again, and both `RESULTS.md` and the review's `reference_review` say 17.

**🔵 2, the overturned quote — taken, but not by trimming.** Trimming to "WGEF was shown to activate RhoA" would be verbatim and misleading: the sentence continues ", Cdc42, and Rac1", and quoting its first clause would make the paper appear to say the opposite of what my own `finding_review` records. Instead the quote is left whole, a second non-overturned quote naming the DH-PH and SH3 architecture was added ahead of it, and the `reason` now points at the `finding_review` and states that only the RhoA clause and the architecture are load-bearing. Same snag removed, without a quotation that survives the letter of verbatim checking while losing its sense.

**🔵 4, IDA vs IMP — taken as a note.** The row now says IDA is for the overexpression-and-co-IP arm, and that each paper's knockdown arm would support a parallel IMP a curator may prefer to split out.

---

**🔵 3, `GO:0042060` in `directly_involved_in` — declined, with reasons.**

The argument for your position is real, and it is the qualifier asymmetry you spotted: GOA gives the mechanistically *closer* term `GO:0032956` the qualifier `acts_upstream_of_or_within`, so putting the further-out `GO:0042060` under `directly_involved_in` does read as an upgrade.

Two things hold me at the current placement:

1. **The evidence is cellular, not just organismal.** Knockdown of RhoGEF19 in keratinocytes causes defects in wound healing, and re-expression rescues them (PMID:20643356). That is the gene acting *inside* the cells that close the wound, not merely upstream of the process — the same experiment, in the same figure, that grounds the actin phenotype. "Rests on one mouse experiment" is accurate about the count but the experiment is a knockdown-plus-rescue in the relevant cell type, which is the strongest design available short of a human one.
2. **Consistency with the action I gave the row.** I marked `GO:0042060` `ACCEPT`, and `ACCEPT` is defined as "retain as representing the core function of the gene". Dropping it from `core_functions` while accepting it as core would leave the document arguing with itself. If the term does not belong in `core_functions`, the honest fix is `KEEP_AS_NON_CORE` on the row — and for a gene whose only mammalian in-vivo phenotype is wound healing, I do not think that is right either.

**What would reverse me:** evidence that the keratinocyte requirement is indirect — that ARHGEF19 is needed for wound closure only as a transcriptional output of GRHL3 and not through its own activity in the migrating cell — or a human or second-species experiment that fails to reproduce the phenotype. Either would move the row to `KEEP_AS_NON_CORE` and take the term out of `core_functions` together, rather than splitting them.

---

**Coverage finding, restated as the measurement it is** (`go_coverage_by_reference.py`, QuickGO by reference, all species; paper list assembled independently from the Affinage record and a live PubMed sweep):

- **11 of 14** primary ARHGEF19 papers have produced **no GO annotation of any gene in any species**.
- **12 of 14** have produced no GO annotation of *ARHGEF19* in any species.
- **1 of 14** has produced an ARHGEF19 annotation to anything other than `GO:0005515 protein binding` — PMID:20643356, a mouse genetics paper whose subject is GRHL3.

**Checks on the new head:**

- `just validate human ARHGEF19` — valid, with the one deliberate warning (the three `cytosol` rows not sharing an action).
- `just validate-families` — passed: 1411 residue-site pass / 0 fail, 453 gene-residue-claim pass / 0 fail. This review carries no `residue_claims`.
- `verify_quotes.py` — **85/85** verbatim, including the **19** `file:` and `Reactome:` quotes CI does not check.
- `selftest_guards.py` — **17/17**, count now derived rather than written down.

One note on your verification limits: you corroborated the obsoletion claims against `genes/human/ARHGAP4/ARHGAP4-bioinformatics/gap_terms.json` rather than the live services. That is a better check than re-running my own queries would have been — it is an independent gene's independently-fetched record agreeing on `GO:0017048 → GO:0031267`, the GO API non-resolution, and the QuickGO masking. Noting it so the next reviewer knows that cross-gene corroboration exists in-repo.

cmungall:
Both 🟡 and both 🔵 taken, plus a third correction I found while you were reviewing — which turns out to be the same failure mode as your 🟡, and a more consequential instance of it. Head is now `28e49000`.

## 🟡 The recall generalisation — you are right, and my own output refuted it

Confirmed exactly as you describe. PMID:34813497's cached title is *"ARHGEF19 promotes the growth of breast cancer in vitro and in vivo by the MAPK pathway"*, the symbol is the first word, and `COVERAGE_raw.txt` already listed it as *found* by the live sweep. I wrote a generalisation over a four-row table and did not read the fourth row.

The fix is not a reworded sentence. `go_coverage_by_reference.py` now **computes** it — reads each missed paper's cached title, matches against the gene's synonyms, and prints the tally:

```
    - PMID:18537266: does not name the gene -- Role of the C-terminal SH3 domain ... Tim and related Dbl-family proteins.
    - PMID:20643356: does not name the gene -- Epidermal wound repair is regulated by the planar cell polarity signaling pathway.
    - PMID:21686262: names the gene (GEF19)  -- Grhl3 and GEF19 in the front rho.
    - PMID:34813497: names the gene (ARHGEF19, GEF19) -- ARHGEF19 promotes the growth of breast cancer ...
  6 of 8 misses have a title that does not name the gene; 2 do name it, so the
  'title does not name the gene' explanation covers 6, not all.
```

Your reading that this strengthens the section is the one I took. Of the four ARHGEF19-specific misses: two have unfindable titles, one names the gene only as the abbreviation `GEF19` (which no `ARHGEF19` or `WGEF` query returns), and **one has no retrieval explanation at all**. That last row is now the sharpest sentence in the section — a record whose trust gates passed omitted a paper whose title *begins with the gene symbol*. Corrected in the review YAML, `RESULTS.md`, the notes and the PR description.

Your framing of why it merited 🟡 is the part I want to record: the script computes the missed list and left the *explanation* unverified, so the one part that drifted is the part no guard covered. That is now closed.

## The same failure mode, found downstream — and it moves a conclusion

While you were reviewing I went back to PMID:20643356 to check a species claim in my reply to your suggestion 3. I had written "short of a human one", implying the keratinocyte experiments were mouse. **They are human.** The cached record was abstract-only, the abstract says "keratinocytes" with no species, and the paper is framed around mouse genetics — so I inferred mouse, which is precisely the inference `CLAUDE.md` tells you not to make from an abstract that does not state it.

Re-fetched the full text from PMC2965174 (`just fetch-pmid 20643356 --force`, now cached as `full_text_available: true`):

> we utilized the shRNA lentivirus to target the RhoGEF19 transcript in the **human keratinocyte cell line, HaCAT**

In those human cells: knockdown leaves the scratch 20% closed "unlike the Scr cells, which migrated to completely close the scratch within 24 hours"; activated RhoA falls; "RhoGEF19-kd cells formed only rudimentary stress fibers that showed irregular organization and were not directed towards the scratched area"; and "re-expression of RhoGEF19 substantially rescued the wound healing phenotype observed with the Grhl3-kd cells (10% closure versus 73%)". The leading-edge polarity analysis is likewise on "human Grhl3-kd, RhoGEF19-kd, and Scr control keratinocytes".

Three consequences:

1. **A claim in the merged-if-approved YAML was false** — the `GO:0042060` row asserted that no human wound-healing experiment on ARHGEF19 had been published. Removed; five verbatim findings added to the reference; the `GO:0032956` row, the `core_function` description and the CURATION knowledge gap updated.
2. **It sharpens the coverage finding rather than softening it.** A human knockdown-with-rescue for `GO:0042060` exists — in the very paper GOA already cites for the mouse orthologue — and the human gene still carries only an IBA. That one row would convert to a human IMP on evidence already published.
3. **Your suggestion 3 is now better supported than when I declined it.** The knockdown-plus-rescue-in-keratinocytes argument I leaned on is a *human*-cell argument, not a mouse one read across. The decline stands on firmer ground than I gave it, and the reversal condition I stated is unchanged.

## 🔵 Both taken

- The `###` heading splitting the Biological process list: `What the PAINT table actually says` is now its own `##` section ahead of Molecular function, and the GRHL3 bullet is back under its own heading.
- `RESULTS.md:123` now reads "Both groups of mutations are anchored by name — module attribute for the first, reference `id` for the second", which is what it meant.

## Checks on `28e49000`

- `just validate human ARHGEF19` — valid, one deliberate warning (the three `cytosol` rows not sharing an action).
- `just validate-families` — passed: 1411 residue-site pass / 0 fail, 453 gene-residue-claim pass / 0 fail; no `residue_claims` in this review.
- `verify_quotes.py` — **95/95** verbatim (76 PMID, 4 Reactome, 15 `file:`), including the **19** CI does not check.
- `selftest_guards.py` — **17/17**.
- Two `history/` EDIT records appended and `just validate-history` clean on both.

The coverage measurement is unchanged by any of this: **11 of 14** primary papers have produced no GO annotation of any gene in any species, **12 of 14** none on ARHGEF19, **1 of 14** anything beyond `GO:0005515`.

cmungall:
All four taken. Head is now `96cf5c3f`.

## 🟡 "cited nowhere in GO" — false, and my own file says so twice

Verified before acting, and you are exactly right. PMID:18256687 is the `original_reference_id` of the `GO:0005515` IPI with `UniProtKB:P61586` — row 5 of `ARHGEF19-goa.tsv`, reviewed about 110 lines *earlier in the same file*:

```
GO:0005515  IPI  PMID:18256687  UniProtKB:P61586
```

And the document refutes itself twice more, as you say: `COVERAGE_raw.txt` scores that paper at 4 annotations with 1 on ARHGEF19, and the CURATION knowledge gap calls it "the single richest paper, PMID:18256687, yielded exactly one ARHGEF19 annotation."

Your framing of why it merited 🟡 rather than 🔵 is the part worth keeping: no term or action turns on it, but the `ACCEPT` is argued *from the contrast*, so the false half was load-bearing for the argument rather than decorative. The replacement is the narrower claim you suggested, which is both true and sharper — PMID:18256687 is cited by GO **exactly once** for this gene, for `GO:0005515` with RhoA, and **not for `GO:0060071` at all**. The `reason` moved with the summary.

## 🔵 1 — the same class, softer, and you found both instances

"No human localisation experiment on ARHGEF19 exists" (`:183`) and "...has been published" (`:425`) were each contradicted by their own **next clause**, which then cites the SH-SY5Y observation. Both now read "worth annotating" — which is, as you spotted, the wording `ARHGEF19-notes.md` already carried. The notes had it right and the YAML had drifted from it.

## 🔵 2 — stale framing in RESULTS.md

`RESULTS.md:149` still described PMID:20643356 as "a mouse genetics paper whose subject is GRHL3" — the framing `28e49000` corrected everywhere else. It now records that its RhoGEF19 knockdown was in **human** HaCaT keratinocytes with a scratch-closure readout and a rescue, and that the human gene nonetheless carries `GO:0042060` only as an IBA projected from the mouse orthologue in that same paper.

## 🔵 3 — history record

Added for this round; `just validate-history` clean.

---

Three rounds, and the three 🟡s were the same defect in three costumes: a claim asserted over a set without checking the set. The seed-set redundancy claim, the recall generalisation whose counterexample sat in the table beneath it, and now a "cited nowhere" that the file's own GOA table refutes. Two of the three are now computed rather than written — the PAINT seed sets and the title-shape tally — and this one is a single narrowed sentence, because there is no set to compute over: it is one row in a 13-row table I had already reviewed.

Worth noting what did *not* catch these. `verify_quotes.py` passes every one of them, because each quote is verbatim and correctly attributed; what drifted was prose *about* the quotes and the rows. That is the gap your review has been covering, and it is not one a substring checker can close.

## Checks on `96cf5c3f`

- `just validate human ARHGEF19` — valid, one deliberate warning (the three `cytosol` rows not sharing an action).
- `verify_quotes.py` — **95/95** verbatim (76 PMID, 4 Reactome, 15 `file:`), including the **19** CI does not check.
- `selftest_guards.py` — **17/17**.
- `just validate-history` — clean on all three records.

Coverage measurement unchanged: **11 of 14** primary papers have produced no GO annotation of any gene in any species, **12 of 14** none on ARHGEF19, **1 of 14** anything beyond `GO:0005515`.

cmungall:
Thank you for the approval — and I've taken the remaining 🔵 anyway, at `1a7c6979`.

You're right on the facts: `COVERAGE_raw.txt` scores PMID:18256687 at **4** GO annotations, **1** on ARHGEF19 (the others being RHOA and RTKN from the same IntAct submission), so "cited by GO exactly once" is false unscoped. The `reason` three lines below already carried "for this gene"; the summary now matches it.

I weighed leaving it — you flagged it as non-blocking, no term or action turns on it, and a push costs a fresh build and supersedes this approval. I pushed because the convergence test isn't the only test here: three rounds of this review were spent on prose claiming more than it had checked, and shipping a clause I now know to be false would make the wrong point about what that discipline is for. Two words is a cheap way not to undercut it.

Your closing observation is the one I'd want carried forward, so let me restate it in the form I think is portable:

> `verify_quotes.py` passed all three 🟡s, because each quote was verbatim and correctly attributed. What drifted was **prose about** the quotes and the rows.

That is a gap no substring checker closes, and it has a shape: every one of the three was a **claim quantified over a set** — the seed set, the set of missed papers, the set of GO citations — asserted without enumerating the set. Two are now computed (`PAINT` seed sets; the title-shape tally). The third had no set worth computing over, so it became a single sentence scoped to what was actually checked.

The generalisable rule, if it's useful to the next gene: when a sentence says *every*, *none*, *nowhere* or *only*, either the set is small enough to enumerate in the sentence, or it belongs in a script. The self-audit that caught the "only human-cell interaction study" overcount on `0106f3b9` was just `grep -oE '\b(every|all|none|never|only|no [a-z]+ (has|have|exists|is))'` over the review — cheap, and it found a real one.

No term, action, evidence code, `supporting_text` or `core_functions` entry changed in this round.

**Checks on `1a7c6979`:** `just validate human ARHGEF19` valid with the one deliberate warning · `verify_quotes.py` 95/95 · `selftest_guards.py` 17/17.

PR #3084, opened 2026-09-19
Review human ARHGEF16 (Ephexin-4): a GAP molecular function on a GEF, already projected to 81 species

cmungall:
Reviews all 77 GOA rows on human ARHGEF16 (Ephexin-4, Q5VV41), plus four committed, self-testing bioinformatics analyses.

## The headline: a GAP molecular function on a GEF, already in 81 species

`GO:0005096 GTPase activator activity` is annotated **IDA** to ARHGEF16 from `PMID:21139582` — a paper that measured *nucleotide exchange*. `GO:0005096` is the GAP term ("Binds to and increases the activity of a GTPase, an enzyme that catalyzes the **hydrolysis** of GTP"), whose synonyms are Rho/Rac/Ras GAP activity. A GAP switches the GTPase off; a GEF switches it on.

The trap is a name collision. The same gene carries `GO:0090630 activation of GTPase activity`, which *reads* like the GAP molecular function but is defined by exchange — "initiates the activity of an inactive GTPase through the **replacement of GDP by GTP**". That row is right. `GO:0005096` is not, and the same curator on the same paper chose the correct process term. The protein has no GAP domain: UniProt lists DH, PH and SH3, and the InterPro signatures are `IPR000219` and `IPR047271`.

Blast radius, measured rather than assumed: QuickGO returns **83 annotations across 81 taxa** carrying `UniProtKB:Q5VV41` in `WITH/FROM` for `GO:0005096`, all Ensembl Compara IEA or ISO. Mouse `Arhgef16` carries it twice and has **no experimental GO annotation of its own** — its whole record is projected from human. → `MODIFY` to `GO:0005085` with `CDC42` as `has_input`.

## Ephexin-family ambiguity, checked in both directions

ARHGEF16 is the one ephexin that is **not** a RhoA GEF. Its measured substrate is RhoG, and in cells nothing else: *"Although the DH-PH domain of Ephexin4 did not exchange nucleotide on Rac1, it possessed GEF activity of RhoG in vitro"*, with the Rac1, Cdc42 and RhoA negatives run against Dock180, Zizimin1 and Ephexin1 as positive controls.

- No annotation reached this gene *from* a sibling incorrectly. The `GO:0005085` IBA carries Ephexin-1, -3 and -5 as donors and is sound — **but only because the term is coarse.** At the granularity GO used to offer (`GO:0005089`), three of four donors would have transferred RhoA specificity to a protein that demonstrably lacks it.
- Two Reactome reactions *do* show the ARHGAP11B pattern. `R-HSA-419166 "GEFs activate RhoA,B,C"` has a 54-member `DefinedSet` catalyst assembled by DH domain, not specificity — it contains TIAM1/TIAM2 (Rac-only), FGD1–4 and ITSN1 (Cdc42-only), SOS1/2 and RASGRF2 (Ras) — and its output is the one GTPase `PMID:20679435` failed to detect. Only the `cytosol` term reaches GOA, so this is recorded as **provenance on the TAS rows**, not as a GO error.

## GO cannot say which GTPase — verified on three services

`GO:0005089` Rho GEF, `GO:0030676` Rac GEF, `GO:0005088` Ras GEF → `GO:0005085`; `GO:0017048` Rho GTPase binding → `GO:0031267`; `GO:0032860/1` → `GO:0090630`; `GO:0005100` Rho GAP → `GO:0005096`. All four parents have **zero `is_a` children**.

Cross-checked on the GO API, OLS4 **and** QuickGO, because the claim is load-bearing. Both brief-flagged QuickGO behaviours reproduced: it **silently resolved all seven merges** (ask for `GO:0017048`, get `GO:0031267` with `isObsolete: false`), and its only "children" of `GO:0005085`/`GO:0005096` are `capable_of` / `negatively_regulates`, not `is_a`.

No new term is proposed — GO merged these deliberately. Substrate identity goes into `core_functions[].substrates` and `RO:0002233 has_input`, and the limit is recorded as an ONTOLOGY knowledge gap.

## DH/PH assessment, reported in both directions

Six solved DH–GTPase complexes (1KZ7, 1LB1, 1FOE, 1KI1, 1X86, 2NZ8), SIFTS-mapped, contacts computed at 4.0 Å. Each anchor's own GEF scores 1.00 — the internal check that the offsets are right.

- **Retained.** ARHGEF16 keeps 0.50–0.70 of the exchange surface at every anchor, inside the range of bona fide Dbl-family GEFs.
- **And it tells you nothing about the substrate.** Read naively the panel names Cdc42 (highest scores against the two Cdc42-bound anchors); the measured substrate is RhoG, in none of them. The control that says why: **1KZ7 and 1LB1 are the same GEF bound to two different GTPases** — Jaccard 0.731, target score moves 0.06. The ranking tracks GEF-to-GEF similarity.
- **Retention ≠ activity**, and the counter-example is this protein: full-length Ephexin4 is autoinhibited by two independent modes and is quiet until Elmo1 or a PDZ protein relieves it. **Absence ≠ inactivity** either: the out-group DOCK4 catalyses exchange on Rac with no DH domain at all. The analysis is confirmatory only — the activity was measured on purified protein.
- **One honest failure reported, not hidden.** At 1FOE, DOCK4's local alignment scored 0.35, +0.22 above its own shuffle — a chance match of a 1966-residue protein to a 195-residue domain. That is why the out-group is controlled on DH *presence*, and why the positive control is each protein against its own composition-matched shuffle rather than a tunable absolute threshold.

## What affinage missed

Gates clean (`win`, `faith_pct 100.0`), and all six citations verified as genuinely about this protein — precision was fine. PubMed returns **26** records for the protein's names; affinage returned 6 and missed **20**, including `PMID:20679435`, which alone supplies 8 of the gene's 12 experimental GO rows. Every mechanistic paper it missed is titled **Ephexin4**, not ARHGEF16 — the RhoG identification, both anoikis papers, both autoinhibition papers, the PNAS structures, and the 2025 mitosis paper. Its record names Cdc42 as the "core biochemical activity" and never mentions RhoG, which is exactly what a symbol-keyed retriever produces here.

Separately: of those 26 papers, **exactly two** produced any GO annotation on human or mouse. The gene's entire experimental record rests on two papers, one of which the provider did not retrieve.

## Actions

50 `MODIFY` · 16 `KEEP_AS_NON_CORE` · 10 `ACCEPT` · 1 `MARK_AS_OVER_ANNOTATED` · 1 `NEW` · 0 `REMOVE`

- **49 of 57 `protein binding` rows retyped**: 42 → `GO:0030165 PDZ domain binding` (all partners carry PDZ domains and bind one C-terminal `...ETDV` motif, UniProt `MOTIF 707-709`; `PMID:33597305` shows that binding relieves one of the two autoinhibitory modes, so it regulates the gene's own catalysis), 7 → `GO:0071889 14-3-3 protein binding`. The remaining 8 are one-off screen hits with no PDZ explanation, kept non-core.
- **`NEW`: `GO:2000811 negative regulation of anoikis`** — knockdown plus rescue in human cells, two labs a decade apart, the second via an orthogonal perturbation of EphA2. A regulation term, so acting upstream is what it means, and ARHGEF16 itself does the work.
- **No `REMOVE`.** The Cdc42 conflict (`PMID:21139582` in vitro + Tip-1/E6 vs `PMID:20679435` cell-based negatives with Zizimin1 control) is recorded as a `finding_review` `DISPUTED` and reconciled on cofactor-conditionality, with both Cdc42 rows demoted to non-core rather than adjudicated away.
- **`GO:0031267` deliberately carries three different actions** — RhoG `ACCEPT` (the substrate), Rac1 and Cdc42 `KEEP_AS_NON_CORE` (bound, never exchanged / bound only with E6). This trips a best-practice warning; the divergence is argued explicitly in the RhoG row.
- **The published autoinhibition mutant is murine.** `PMID:30445756`'s `E295A` comes from murine cDNA NM_001112744 (`Q3U5C8`, 713 aa); human `Q5VV41` position 295 is a **serine**. Alignment places it at human **E291**, recorded as a `residue_claim` anchored on the human position, with the script's self-test asserting that the naive same-number reading fails.

## Bioinformatics

`genes/human/ARHGEF16/ARHGEF16-bioinformatics/` — **six** scripts, each with `--self-test` that breaks its own anchor and requires every guard to fire, with negative controls that must stay silent. **6/6, 10/10, 6/6, 7/7, 9/9, 11/11.** JSON outputs committed as the record of what the live services returned.

Two exist because of defects found in review, and both cover gaps CI does not:

- **`quote_claim_coherence.py`** — the reference validator checks a `supporting_text` against its **source publication**, never against the **claim it sits under**. Two Reactome rows shared one quote, so the row arguing Rac/Cdc42 was citing the RhoA negative: impeccably sourced, wrong proposition. It also verifies `file:` quotes against their artifacts, which CI skips entirely (`file` is in `skip_prefixes`) — **47 in this review, all verbatim**.
- **`pdz_partner_check.py`** — reads the committed review rather than a hardcoded list and checks the 42-row retype in both directions: all 38 partners on the PDZ side carry a PDZ domain (147 between them, including TAX1BP3, which GOA itself already types `GO:0030165`), and none of the 11 accessions outside does — 10 distinct proteins, TFG appearing under both a reviewed and an unreviewed accession.

Both were **mutation-tested by running the suite**, not by reasoning about it: 10 mutations, every anchor asserted to match exactly once, every restore verified by SHA-256, both files ending byte-identical. All 10 dropped the pass count. One is why a self-test was rewritten — its name-independence check asserted `has_pdz("Q9H5P4")`, and PDZD7 is itself *named* for PDZ, so a name-keyed classifier satisfied it too.


cmungall:
Responding to **both** reviews: the live re-review on `2302d4f83`, and the **2026-09-19 review that was auto-dismissed** when a coordinator checkpoint commit moved the head (comment 5741463657). Recording the dismissed one explicitly so its history is not lost — its two 🟡 items were real feedback and are answered below.

---

## 🟡 2 (both reviews) — accepted, fixed, and guarded

You flagged it twice and you were right both times: the two Reactome cytosol rows were generated from one loop that attached the **same** quote to both, so the row arguing that `R-HSA-205039`'s asserted output is Rac and Cdc42 was citing the RhoA negative.

| row | asserted output | quote now |
|---|---|---|
| `R-HSA-205039` | Rac and Cdc42 | *"we could observe no obvious increase in the activities of Rac1 and Cdc42 in HEK293T cells by expression of Ephexin4"* |
| `R-HSA-419166` | RhoA, RhoB, RhoC | *"we could not detect the increase in RhoA activity in cells expressing Ephexin4"* |

**I added a guard rather than only fixing the instance**, because nothing in this repo catches the class. The reference validator checks a `supporting_text` against its **source publication** — and that quote was a perfectly verbatim substring of exactly the right paper. What is never checked is the quote against the **claim it sits under**.

New `quote_claim_coherence.py` collects the GTPases named by each row's claim (`RO:0002233 has_input` extensions plus the "catalysed output is …" clause) and by the union of its quotes, and requires them to intersect. Five rows checkable, **0 violations**. Its central self-test is a regression that reintroduces your exact bug and requires the guard to fire, then restores the correct quote and requires the row to go clean.

The union is deliberate: `GO:0005096` legitimately pairs the Cdc42 exchange result with the Rac1/RhoG specificity result, so requiring *every* quote to name the claimed GTPase would fire on a correct row.

While there I closed a second gap in the same area: `conf/reference_validator_config.yaml` lists `file` under `skip_prefixes`, so every `file:` `supporting_text` is checked against **nothing** — a quote from a repo artifact can be paraphrased, stale, or invented and CI stays green. This review carries **47** of them. The same script now verifies them against their source; all 47 are verbatim, and the self-test tampers with one and requires the rejection.

## 🟡 1 — `Q05BK6` is TFG. Both diagnoses were wrong, in different ways.

The dismissed review said Q05BK6 is TNS3. The live re-review correctly doubts that, and instead proposes Q05BK6 is "a secondary accession resolving to TFG" — noting you could not confirm, since outbound network is blocked in the runner. I checked:

```
Q05BK6  gene=TFG  Q05BK6_HUMAN  "TFG protein"   366 aa  TrEMBL (unreviewed)  EMBL BC041600
Q92734  gene=TFG  TFG_HUMAN     "Protein TFG"   400 aa  Swiss-Prot (reviewed)
TNS3    Q68CZ2    TENS3_HUMAN                  1445 aa  Swiss-Prot
```

- `sec_acc:Q05BK6` returns **0 entries** — so it is **not** a secondary accession of anything, including TFG. It is its own primary TrEMBL entry.
- Q68CZ2 (TNS3) lists **no** secondary accessions, so Q05BK6 cannot be reaching it by a redirect either.
- Q05BK6's only domain is **PB1** — TFG's domain. The dismissed review's own supporting detail ("TNS3 has SH2/PTB") is evidence *against* its conclusion, since Q05BK6 has neither.

So: same gene, same protein, **two independent primary accessions** — one reviewed, one not — rather than a rename or a redirect. The label stays TFG.

**The observation underneath both items is real and is now in the document.** The two rows did read identically, which is what made them look like a copy-paste. They are now distinguished, and the actual finding is stated: GOA counts one binding partner twice here, a duplicate-accession artefact rather than two findings.

## 🟡 3 (live review) — accepted. A quote that announces a check is not evidence of its result.

This is the best catch in either review. The 42 PDZ rows cited *"The motif is confirmed first, since the whole argument rests on it"* — which, as you say, **would read identically had the check failed**. A quote invariant to the outcome cannot support the outcome.

Replaced with the result-stating line: *"Every partner on the PDZ side carries at least one PDZ domain"*. The cadherin row's citation was updated to the same sentence, since I had pointed it at a line whose wording your fix obsoleted.

---

## 🔵 Both of the live review's non-blocking points — taken, and both were right

**"The '11 partners' figure double-counts TFG."** Correct: 11 accessions, **10 distinct proteins**. The script now reports accession and protein counts separately and names the duplicated symbol, so the number cannot be misread again.

**"The negative control skips partners of rows GOA already typed `GO:0030165`."** Also correct, and this was the more important of the two. TAX1BP3 was in neither bucket, because the partition only looked at `GO:0005515` rows — which excluded **the single partner the ontology has already adjudicated**, the strongest available positive. If my classifier disagreed with GOA there, the classifier would be what is wrong. TAX1BP3 is now in the panel (1 PDZ domain, agrees), and a self-test asserts its presence so it cannot silently drop out again.

Revised counts: PDZ side **43 rows / 38 partners / 147 PDZ domains**; outside **15 rows / 11 accessions / 10 proteins / 0 PDZ domains**.

## 🔵 The three suggestions from the dismissed 2026-09-19 review — all taken

Each was a machine-readable field saying something its own prose disclaimed.

- **`MISCITED`.** Both Reactome refs → **`DISPUTED`** (the claim GOA draws is `cytosol`, correctly supported; what is contested is the set-level GTPase output, which never reaches GOA). The affinage record → **`LOW_QUALITY`** (nothing is miscited; the defect is a symbol-keyed retrieval that missed every paper titled *Ephexin4*).
- **Cadherin `supporting_text` was a paper title.** Replaced with the committed PDZ-partner overlap, which quantifies the indirect route: **11 of the 38 PDZ partners are junctional scaffolds** (SCRIB, DLG1, MAGI1/2/3, PATJ, MPDZ, PARD3, TJP1/2/3).
- **AP-MS asymmetry.** The two BioPlex 14-3-3 rows now say why they are retyped while the cadherin AP-MS row is marked over-annotated: the retyped term names the partner *class* actually recovered, whereas `GO:0045296` asserts contact with cadherin specifically; and the 14-3-3 result is replicated across five references and three partners with a mechanistic basis in ten mapped phosphosites, against a single dataset for cadherin.

---

## A tautology in my own self-test, found while acting on your 🟡 3

Worth flagging since it is the same failure mode. `pdz_partner_check.py`'s name-independence check asserted `has_pdz("Q9H5P4") is True` — but PDZD7 is itself *named* for PDZ, so a classifier keyed on the gene name rather than the domain list satisfies it too. **The check could not fail for the reason it was named for.**

The classifier is now a pure function so the self-test can feed it hand-built records: a protein named `PDZD99` whose only domain is an RRM, one carrying a PDZ domain under an unrelated name, and one whose PDZ sits under the wrong feature type. I also added a zero-match guard asserting each injected accession is absent from its bucket before injection, since a no-op injection passes while testing nothing.

Then I **ran** the mutations rather than reasoning about them — 10 across the two guards, every anchor asserted to match exactly once, every restore verified by SHA-256 against the original:

| script | mutation | result |
|---|---|---|
| `pdz_partner_check` (baseline 9/9) | classifier keys on the gene name | **5/9** |
| | classifier always says yes | **4/9** |
| | drop the reverse guard | 8/9 |
| | drop the forward guard | 8/9 |
| | revert the partition to skip GOA's `GO:0030165` rows | 8/9 |
| `quote_claim_coherence` (baseline 11/11) | never record a violation | 10/11 |
| | claim extraction returns nothing | **8/11** |
| | `RhoG` matched as a substring | 10/11 |
| | drop `has_input` from the claim side | 10/11 |
| | `file:` verifier stops comparing against the file | 10/11 |

The first row is the point: under the old check that mutation would have scored a clean pass.

## Not changed

`GO:0031267` still carries three different actions and still trips the best-practice warning. The rows differ by WITH/FROM partner, and binding RhoG (the substrate) is a different claim from binding Rac1 (bound, never exchanged — negative measured in the same figure) or Cdc42 (bound only with HPV16 E6 present). Collapsing them would either promote two non-substrate contacts to core or demote the substrate. Argued explicitly in the RhoG row; happy to revisit if you would rather the warning be clean.

**No action, term, evidence code or `core_functions` entry changed in either round.** 0 `REMOVE` rows in this review, so the action/prose contradiction check is vacuous here — verified rather than assumed, with a positive control to confirm the detector fires.

Gates re-run locally on this head:

```
just validate human ARHGEF16     → ✓ Valid (1 warning, the GO:0031267 one above)
gene_residue_claims              → 3 pass, 0 fail, 0 unresolved
just validate-families           → passed
self-tests                       → 6/6, 10/10, 6/6, 7/7, 9/9, 11/11
just validate-history            → No issues found
```

PR #3093, opened 2026-09-20
Add contested-function read-list from 2025-2026 literature survey

cmungall:
Surveys literature published 2025-01 to 2026-08 for individual human genes
whose molecular function is actively disputed, and frames each dispute as a
question a GO curator could adjudicate.

Introduces the "contested-function gap" as a fourth kind alongside the
project's existing biology/curation/ontology taxonomy: cases where two lines
of experimental work assign mutually incompatible molecular functions and no
consensus has formed, so neither UNDECIDED nor ACCEPT fits.

Contents:
- Ranked shortlist of 22 candidates with the dispute, current GOA exposure,
  repo review status and a confidence rating
- Worked entries for TMEM175, TMEM120A, TMEM65/SLC8B1, the epitranscriptomic
  writer/eraser cluster (FTO, NAT10, TRMT61A, ALKBH1, NSUN2, NSUN7), GSDMC,
  MEFV, PRSS23 and a correction SLC7A11's existing review needs
- A "checked and set aside" section so the same ground is not re-covered

Notable finding: these disputes are frequently already encoded in GOA with
experimental evidence on both sides. TMEM65 carries calcium:sodium antiporter
activity twice, once as enables/IDA and once as NOT|enables/IDA, citing the
two 2025 papers that disagree.

Verification notes recorded on the page, each a reusable trap: PubMed DOIs
must come from the article's own identifier fields rather than the last DOI in
the XML (a reference-list read gave 9 wrong DOIs out of 20), and GOA queries
must carry the qualifier field -- TMEM120A's ion-channel annotations are all
NOT|enables, so omitting the qualifier inverts their meaning.

All cited PMIDs re-verified against PubMed for title, journal and year; GO
annotation states read live from QuickGO; repo status checked against
genes/human/.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_015AhgmBM1MzW3w2iRXuFEnU

### DIFF
--- a/projects/FUNCTION_KNOWLEDGE_GAPS.md
+++ b/projects/FUNCTION_KNOWLEDGE_GAPS.md
@@ -98,6 +98,23 @@ Second, *which GO aspect is dark* — most "dark" genes are not uniformly dark:
 - **CC-dark** — function known, but where/when unknown.
 - **Wholly dark** — only root terms / IEA / `protein binding` survive review (the deep unknome).
 
+### A fourth kind: the contested-function gap
+
+The three kinds above all describe *absence* of knowledge. A survey of 2025-2026 literature
+surfaced a distinct and, for curation, sharper species: genes where biology knows **two
+incompatible things**, both experimentally supported, neither retracted.
+
+> **Contested-function gap** — two or more lines of experimental work assign the gene product
+> mutually incompatible molecular functions and no consensus has formed. Neither `UNDECIDED`
+> (which implies we could not access the evidence) nor `ACCEPT` (which implies the matter is
+> settled) fits.
+
+These are worth separating out because the conflict is frequently **already inside GOA**, with
+experimental evidence codes on both sides — in the TMEM65 case as a positive `enables`/IDA and a
+`NOT|enables`/IDA of the *same* term, citing the two papers that disagree. See
+[Contested Functions: a 2025-2026 read-list](FUNCTION_KNOWLEDGE_GAPS/contested-functions-2025-2026.md)
+for the survey, the ranked candidate list, and the verification traps it turned up.
+
 ## The unit of work: anatomy of a gap entry
 
 Each gap is a small, defensible scholarly object. Required elements:
@@ -914,6 +931,11 @@ human/SOCS4/SOCS5, human/RFT1, worm/pef-1, worm/fshr-1, SCHPO/alo1, and DESVH/Q7
 - [x] Documented deep-research review mechanisms (failure-mode table + layered checks)
 - [ ] Implement `validate-deep-research`: PMID resolution + title-match + per-file reliability score (extends `validate_pmid_references.py`)
 - [ ] Read-list deepening, batch 4: PUS3, CFAP418, SOCS4/SOCS5, RFT1, pef-1, fshr-1, alo1
+- [x] Fourth gap kind identified: the *contested-function gap*, with a 2025-2026 literature
+      survey and ranked read-list ([sub-page](FUNCTION_KNOWLEDGE_GAPS/contested-functions-2025-2026.md));
+      all cited PMIDs re-verified for title/journal/year and DOI
+- [ ] Curate contested-function read-list into reviews (first picks: TMEM175; TMEM65 + SLC8B1
+      as a pair; MEFV update for the CDC42 mechanism)
 - [ ] Decide unit granularity (per-gap vs per-gene narrative)
 - [x] Decide home: standalone register vs `knowledge_gaps` schema element — **done**: added a first-class `KnowledgeGap` schema class (gene/annotation/core-function/module/module-node), with the structured register rendered from it
 - [ ] Conservation / disease prioritization pass over candidates
--- a/projects/FUNCTION_KNOWLEDGE_GAPS/contested-functions-2025-2026.md
+++ b/projects/FUNCTION_KNOWLEDGE_GAPS/contested-functions-2025-2026.md
@@ -1,0 +1,487 @@
+---
+title: "Contested Functions: a 2025-2026 read-list"
+species: [human]
+---
+
+# Contested Functions — a 2025-2026 read-list
+
+A survey of literature published between January 2025 and August 2026, looking for
+**individual human genes whose molecular function is actively disputed**, and turning each
+dispute into a question a GO curator could adjudicate.
+
+This is a companion to the [parent project](../FUNCTION_KNOWLEDGE_GAPS.md), which registers what
+biology does not know. Here the problem is the opposite and, for curation, sharper: biology
+knows *two incompatible things*, both published, both often already in GOA.
+
+## The contested-function gap
+
+The parent page's taxonomy has three kinds of gap — **biology** (nobody knows), **curation**
+(known but unannotated), **ontology** (known but inexpressible). Cases below mostly belong to a
+fourth kind:
+
+> **Contested-function gap** — two or more lines of experimental work assign the gene product
+> mutually incompatible molecular functions, and no consensus has formed. Neither `UNDECIDED`
+> (which implies we could not access the evidence) nor `ACCEPT` (which implies the matter is
+> settled) is quite right, and the annotations frequently coexist in GOA with **experimental
+> evidence codes on both sides**.
+
+Two properties make these unusually high-value curation targets:
+
+1. **The dispute is already inside GOA.** For TMEM175 the GOA carries `proton channel activity`
+   (IDA, three references) *and* `potassium ion leak channel activity` (IDA, two references),
+   all positively qualified. For **TMEM65** it carries `calcium:sodium antiporter activity`
+   twice — once as `enables`/IDA and once as **`NOT|enables`/IDA** — citing the two conflicting
+   2025 papers, so the controversy is encoded literally as a positive and a negative annotation
+   of the same term on the same gene. A reviewer does not have to go looking for the problem;
+   it is sitting in the annotation set.
+
+   **Read the qualifier, always.** An early draft of this page asserted that GOA carried the
+   ion-channel claim for TMEM120A. It does not: the three experimental annotations of
+   `monoatomic ion channel activity` are all **`NOT|enables`**. Reading only the term and
+   evidence code inverts the meaning of the annotation, and the mistake is invisible unless the
+   qualifier field is fetched explicitly.
+2. **A downstream literature usually keeps building on the contested claim.** The pattern is
+   consistent: a methods-level or biochemical challenge appears in one venue while a large
+   disease-biology literature continues to assume the original assignment. NAT10, FTO, NSUN2 and
+   TRMT6/TRMT61A are all in this state.
+
+**A caution that applies to every entry below.** A live dispute is *not* a licence to `REMOVE`
+an experimental annotation. Per the repo's standing rule, curators who made an IDA read the full
+text and we have not. The right output for most of these is a `MODIFY` to a term the evidence
+actually supports, a `KEEP_AS_NON_CORE`, or an `ACCEPT` whose `reason` states plainly that the
+assignment is contested and names the challenge. The value here is in **writing the dispute
+down**, not in picking a winner from an abstract.
+
+## How this list was built
+
+Free-text PubMed searching for "controversy"-type phrasing is close to useless — the API's query
+translator silently drops phrase clauses and ANDs everything else, so `"remains controversial"`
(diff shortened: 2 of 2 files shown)


cmungall:
All three blocking findings are fixed, plus all four suggestions. Pushed in `ce7cf7ef3`, `ed32d1e57` and `29271ced9`.

### 1. `GO:0005515` forbidden action — fixed, 232 rows across 42 files

Higher than the 154/30 you found, for two reasons worth recording: three rows arrived with the `origin/main` merge (MTCH2, where main's own PAINT-batch review had moved them `REMOVE` → `MARK_AS_OVER_ANNOTATED`), and PHGDH's three have no `reason` field at all, which is likely why the scan missed them. Their `summary` asserted the superseded policy in so many words — *"retained but marked as over-annotated (experimental IPI, not removed)"* — so that sentence is gone too.

Classified rather than blanket-rewritten:

- **`MODIFY`, 4 rows**, where the existing reason already named a specific evidence-backed replacement: GPR50 → `GO:0030547`, C11orf98 → `GO:0030674`, PEX39 → `GO:0140597`, SLC45A4 → `GO:0030165`. All four checked against `cache/go/terms.csv`. For GPR50 and C11orf98 the replacement also stood as a separate `NEW` row — the duplication you pointed at — so those two `NEW` rows are dropped and the proposal now lives in `proposed_replacement_terms`.
- **`REMOVE` for the rest.** Rows whose reason said the content was *already* captured by terms the gene carries (MICU1, LRRC8A, STING1, SMIM26, YTHDF3) are removals, not modifications — proposing a term the gene already has would just duplicate it.

One deviation from your list: **TMEM175's LAMP row went to `REMOVE`, not `MODIFY`.** The reason gestures at "the LAMP-mediated regulation described in the source" but names no specific molecular function, and inventing one from interaction evidence is what the same policy paragraph forbids.

Reasons were edited, not just re-labelled — where the text restated the policy, that clause is dropped and the row-specific remainder kept, so no reason now argues for an action it doesn't take.

### 2. PRSS23 `GO:0005615` — fixed

You're right, it's live; `cache/go/terms.csv` resolves it. Since the stated reason for *not* refining was only that false belief, the row becomes `MODIFY` → `GO:0005615`, which the evidence (mature fragment recovered from conditioned media) supports directly. `core_functions[0].locations` followed — the validator caught that knock-on. The claim appeared nowhere else on the branch.

### 3. `history/` records — added, 65

61 gene records split `CREATE` (54) / `EDIT` (7) by whether the review existed at the branch base, plus 4 project records. Scaffolded with `scripts/new_history.py` so session ids and filenames are generated rather than hand-written; all 65 validate against `history.yaml` / `HistoryRecord`. Gene details are derived from each review file (action counts, `NEW` terms, `core_functions` size) rather than narrated, so a record can't drift from what the review says. Thanks for noting the timing fairly — the mechanism landed on main the day before this PR.

### Suggestions — all four applied

- **GPR50 cross-species codes.** Confirmed: PMID:40091563 is a mouse knockout study, and two `NEW` rows proposed it on the human gene as `IMP`. Both now `ISO`, with the basis stated in the reason.
- **`GO:0035025` over `GO:0007266`.** Agreed — the receptor is upstream and activates the cassette; the paper's own title says *"by Activating the G(12/13) Protein-RhoA Pathway"*. `core_functions.directly_involved_in` followed.
- **TMEM175 weak quote.** Confirmed it was the introduction describing prior work. Replaced with that paper's own measurement, tagged `RESULTS`.
- **`check_file_supporting_text.py` silent skip.** Real bug, now collected and reported. One honest correction: the repo-wide count of unresolvable paths is **0**, so this did not in fact inflate the coverage figure — the defect was that it could have, invisibly.

### Status

CI was green on the pre-fix head (`test (3.12)` passed the full `validate-all`). All 61 gene reviews validate with 0 errors and the same 9 pre-existing warnings.

Two things left for @cmungall rather than decided here, both flagged in the merge commit:

1. **`GO:0005739` on the MTCH1/MTCH2 pair.** Main's PAINT batch marks it over-annotated on MTCH2; this branch's new MTCH1 review accepts it. The paralogue pair was written to agree, so one should move. I took main's position for MTCH2 and aligned its HTP row to match, but that leaves MTCH1 inconsistent — it's a policy call about whether a correct parent CC term counts as over-annotation.
2. **`MISCITATIONS` vs the existing `MISCITATION_AUDIT`.** The merge surfaced that main already has a more developed project covering this ground from the citation side. The index now lists both. Folding ours in, or scoping it explicitly as the corpus-wide `reference_review` summary, is a structural call I didn't want to make unilaterally.

---

cmungall:
Fixed in `2c8ef5966`, and you were right that this one mattered more than a normal stale sentence — it was the page's only concrete demonstration that the 970 are real rather than a counting artefact.

Confirmed ABI1 is stale exactly as you describe: zero entries opening `Falcon report summarizes`, and absent from the regenerated report. It's one of the reviews `main` had already corrected — the same population that took the subset from 1,042 to 970.

**One correction to the suggestion**, which matters because it's the same failure mode the finding is about. BAK1 does *not* have zero occurrences case-insensitively — it has one:

```
genes/ARATH/BAK1/BAK1-deep-research-falcon.md:2:provider: falcon
```

So the new wording claims no occurrence **in prose** and names the frontmatter line, rather than repeating a clean-sounding zero that a case-insensitive `grep` would immediately contradict. Everything else checked out: 53 entries in `BAK1-ai-review.yaml`, 74 rows across the report, BAK1 present and ABI1 absent.

BAK1 is still clearly the better example — one incidental match in provider metadata, against ABI1's three across metadata and artifact paths.

The old example and the loose *"zero occurrences of the string"* phrasing are kept in a parenthetical rather than deleted, so the correction is visible to anyone who read the earlier version.

One note on the three items from your previous pass — TNFRSF21:1125, the TMEM175 notes quote, and the `check_file_supporting_text.py` counter. Those were already fixed in `f8a5cdc22`, pushed before this review run started, so it was reading `6622acfc6`. On the counter: `checked` now increments after the read, unreadable paths are collected separately, and the JSON carries `checked_by_type` / `mismatches_by_type` / `unresolved_paths` / `unreadable_paths` alongside `mismatches`, so the artefact can't be read as "every `file:` quote in the repo" on its own. Both populations come back **0** repo-wide and the checked figures are unchanged — neither defect had actually inflated anything, it's that either could have, silently.

---

cmungall:
Confirmed and fixed in `8bfef802c`. You're right, and the part that matters most is the second-order point: **this falsifies a property I claimed for these records in this thread.** I said the details are *"derived from each review file … so a record can't drift from what the review says."* The derivation is sound; it ran two commits early. That's a real correction, not a nit.

`GPR50`'s `CREATE` record lists `GO:0007266`; the review proposes `GO:0035025` — the one term it argued explicitly for *not* using.

**I took your option 2, and for the reason you gave.** Each `CREATE` record was accurate at the moment it was written, so editing `details` would replace a true statement about one session with a summary of several. Under append-only semantics the defect isn't "the record is wrong", it's "no record covers the later commits". Four `EDIT` records added:

| record | covers | what changed |
|---|---|---|
| GPR50 | `29271ced9` | `GO:0007266` → `GO:0035025`, `IMP` → `ISO` |
| TMEM175 | `29271ced9`, `f8a5cdc22` | introduction-sourced quote → the Figure 2C measurement; notes matched |
| TNFRSF21 | `f8a5cdc22` | stale "marked as over-annotated" cross-reference |
| FUNCTION_KNOWLEDGE_GAPS | `6622acfc6`, `2c8ef5966` | scan figures reconciled; ABI1 spot-check → BAK1 |

The GPR50 record carries an explicit note that the `CREATE` record's term list is superseded, so someone reading only the first entry in the chain can't be misled.

I independently checked your other two candidates rather than taking the all-clear: TMEM175 (42 rows) and TNFRSF21 (47 rows) still match their records' counts exactly, because those commits changed a quote and a single word of a `reason`. **GPR50 was the only content drift.**

On the merge itself — agreed, and I checked the same thing from the other direction before pushing it: main's 15 commits touch 261 files including 8 gene reviews, none of them among this PR's 61, and nothing in `projects/`, `scripts/`, `.claude/skills` or the validation infra. One thing worth recording, since it cost a diagnostic step: **GitHub reported `mergeable_state: dirty`, but `git merge origin/main` applied with zero conflicts.** The flag was stale or computed differently; the conflict wasn't real.

---

cmungall:
Both suggestions acted on in `2f94ae0bd`. The GPR50 item was already closed in `8bfef802c`.

### The duplicate curies — I think you read this one the wrong way round

You called them *"all with identical labels and pre-existing on both sides, so harmless."* The labels are identical, but they are **not** pre-existing:

| tree | duplicate curies |
|---|---|
| branch base `3e8b082a6` | **0** |
| `origin/main` | 1 |
| this branch, pre-merge `cb1c56ec7` | **10** |
| this branch, post-merge | 12 |

**This branch introduced them.** The validator appends newly-resolved terms to `cache/go/terms.csv`, and running it repeatedly across the session appended rows rather than replacing them — so the count grew with my own validation runs.

And there's a consequence neither of us flagged: **the merge silently reverted main's own fix.** Main deduped `GO:0140605` in #3098; because this branch had also grown that file, the merge kept my duplicated version and main's change was undone — with no conflict to notice it by. That's the second time on this PR that GitHub's merge behaviour has been quietly misleading, after the `dirty`-with-no-conflicts flag.

Deduped keeping the last occurrence per curie, which is the convention #3098 used (it kept the later `retrieved_at`). 12 rows dropped, 10,004 curies remain, `GO:0140605` now byte-identical to main. The script asserts exactly 12 go, no curie disappears, and no curie carries two different labels — had any duplicate disagreed on a label it would have stopped rather than silently picking one. None did.

### The scan figure drifting on every sync

Right, and the fix isn't to re-run it forever. Re-ran once: **59,154 → 59,253** checked, and the mismatch list is byte-identical — so the 99 `file:` quotes main brought with it are all clean. Cheap to confirm, because making the report self-describing means the diff is 4 lines rather than 37k.

The page now gives the current number, says plainly that the denominator is a dated snapshot that moves with every merge, records the earlier 55,433 and 59,154, and points readers at the **narration subset — 970, unchanged across all three runs** — as the figure that actually means something. That's the structural fix rather than resetting the treadmill.

Three genes validate clean against the deduped cache, and a `validate` run doesn't reintroduce duplicates.

---

cmungall:
Thanks — the `merge=union` explanation is the piece I was missing, and it turns "GitHub was misleading" into something mechanical: `.gitattributes` sets `cache/**/*.csv merge=union`, so the union driver takes both sides instead of conflicting. That's *why* main's `GO:0140605` dedup was reverted silently, and why `GO:0090630` was newly created by the same merge. Good catch on the second one.

### `just fix-caches` — you're right, and I didn't know it existed

`src/ai_gene_review/tools/cache_lint.py` is the purpose-built guard, with a docstring naming this exact hazard. I hand-rolled instead. Both halves were already closed in **`8e9632c1c`**, pushed before your review posted — but I verified that against the repo's own tool rather than by eye:

```
cache_lint cache/go/terms.csv   -> "# 1 cache file(s) sorted and deduplicated", exit 0
cache_lint --fix                -> "# normalized 0 file(s)"
diff (mine, tool output)        -> identical
tests/test_cache_sorted.py      -> 11 passed
```

So the file is byte-for-byte the canonical form `just fix-caches` produces, and **the predicted CI failure won't occur**. The outcome converged; the route was worse. And the reason CRLF got in at all is exactly the contrast you drew — a hand-rolled `csv.writer` defaults to `\r\n`, while `cache_lint` passes `lineterminator="\n"` deliberately.

### BAK1: 52, not 53 — and this one stings

You're right, and it's the same mistake I had just corrected *you* for on the other side of this very example. I counted raw `grep` hits. 52 are `supporting_text`; the 53rd occurrence is a `findings[].statement` at `BAK1-ai-review.yaml:922`, which the scan doesn't look at.

Fixed in `a9a47d445`, and the page now states why a plain `grep` returns 53 — so the next person to check doesn't "correct" it back to the wrong number.

### Also in that commit

- The four `EDIT` records now say *"from the earlier session in this PR"* rather than *"from this session"*. You called it cosmetic and not worth its own commit, which is why it rode this one instead of triggering a CI cycle — but you're right that a provenance record is precisely where session identity has to be exact.
- `GO:0140298` divergence recorded: HEAD keeps only the later of main's two rows. That's last-wins applied to main's own remaining duplicate, not a deletion.

676 tests pass; all four history records validate.

---

PR #3094, opened 2026-09-21
Review human ARHGEF18: 19 of 23 primary papers have produced no GO annotation anywhere, the only structure of the protein included

cmungall:
Reviews all 32 GOA rows for human ARHGEF18 (p114RhoGEF, Q6ZSZ5) and adds a committed, self-testing bioinformatics analysis.

## The headline: this is a coverage failure, not an over-annotation problem

`genes/human/ARHGEF18/ARHGEF18-bioinformatics/reference_coverage.py` queries QuickGO **by reference**, species-blind, for 23 papers whose subject is ARHGEF18 itself. **19 of the 23 have produced no GO annotation on any gene in any species**, and none is undetermined — every result set fit inside one page, so every zero is a real zero.

What is missing:

| PMID | paper | annotations anywhere |
|---|---|---|
| 11085924 | Blomquist 2000 — the **founding in-vitro exchange assay** | 0 |
| 21258369 | Terry 2011 *Nat Cell Biol* — junctional RhoA, cingulin/ROCK-II/myosin-IIA | 0 |
| 28132693 | Arno 2017 *AJHG* — biallelic mutation causes **RP78** | 0 |
| 29876405 | Chen 2018 — the **only structure of any part of the protein** (PDB 6BCB) | 0 |
| 23698346 | Herder 2013 — an entire **vertebrate knockout** (medaka) | 0 |

Every experimental row GOA *does* carry is defensible. Exactly one PMID (14512443) supplies a molecular-function annotation. Meanwhile the *Drosophila* ortholog `cyst` **is** curated — `GO:0005085` IDA, `GO:0090688` IDA, `GO:1902408` IEP from PMID:36917931 — annotations that could seed ISS or IBA to the human gene and have not.

This is larger than the two comparable cases in this batch (ARHGAP21: 7 of 10; ARHGAP4: 16 of 19).

### And there is no mammalian ortholog record to lean on

The obvious remedy for a thin human record is ISS or IBA from an ortholog, and the same script closes that door:

| gene product | annotations | experimental | evidence codes |
|---|---|---|---|
| mouse `Arhgef18` (Q6P9R4) | 23 | **0** | ISO 10, IEA 6, IBA 4, ISS 3 |
| human `ARHGEF18` (Q6ZSZ5) | 32 | 14 | IDA 7, TAS 8, IEA 6, IBA 4, EXP 2, IMP 2, IPI 2, HDA 1 |

Neither set truncated, so both are totals. The human row is the **positive control** for the evidence-code split — had it come back at zero, the mouse zero would be a classifier artefact rather than a finding. Ten of the mouse rows are ISO *from human*, so transferring back would only recirculate the same four experiments. Note where that leaves PDB 6BCB: the 1.4 Å RhoA complex was solved on the **mouse** protein, and the mouse record does not carry it either.

## Four proposed new annotations, all from papers GO has never used

| term | evidence | source |
|---|---|---|
| `GO:0005923` bicellular tight junction | IDA | PMID:21258369 — overlaps occludin, not E-cadherin |
| `GO:0043296` apical junction complex | IDA | PMID:21258369 — same imaging |
| `GO:0120192` tight junction assembly | IMP | PMID:21258369 — barrier reaches 40% of control; spreading unaffected, maturation fails |
| `GO:0031267` small GTPase binding | ISS | PMID:29876405 + PMID:23493395 — the PH domain binds GTP-loaded RhoA |

One `MODIFY`: `GO:0035023` regulation of Rho protein signal transduction → `GO:0035025` **positive** regulation. Every primary experiment shows activation and never inhibition, and UniProt already phrases the RP78 variant effect as "decreased function in positive regulation of Rho protein signal transduction".

One `MARK_AS_OVER_ANNOTATED`: `GO:0070062` extracellular exosome, a single bulk-proteomics hit on a 1361-residue cytoplasmic protein with no signal peptide.

No `REMOVE`.

## Substrate specificity is inexpressible, and that was deliberate

Checked across **three independent services** (QuickGO, OLS4, the GO API) because QuickGO silently resolves merges and cannot be read at face value in either direction:

- `GO:0005089` **Rho** GEF activity → obsolete, `term replaced by GO:0005085`, reason `IAO:0000227` (terms merged).
- Same for `GO:0030676` (Rac), `GO:0005088` (Ras), `GO:0005086` (ARF), `GO:0017112` (Rab), `GO:0005087` (Ran).
- `GO:0005085` now has **zero `is_a` children**. QuickGO returns two, both via `negatively_regulates` and `capable_of`; OLS4 returns none.
- The pattern extends past the GEFs: `GO:0017048` Rho GTPase binding was merged into `GO:0031267`, which also has zero children — the same decision GO made for the RhoGAPs into `GO:0005096`.

So no replacement term is proposed. Substrate identity is carried in `core_functions[].substrates` and the limitation is recorded as an `ONTOLOGY` knowledge gap.

## DH/PH assessment, reported in both directions

`dh_domain.py` (self-test: 10 control blocks, all pass).

**Where residues help.** Neither loss-of-function paper states its reference sequence and they do not agree. The script enumerates every offset UniProt's own record licenses rather than assuming one:

- Arno's `p.Thr270Ala` → canonical **Thr-458** (+188, the N-terminal extension UniProt added in sequence version 4 *after* the paper). This is a **positive control**: UniProt independently records `VARIANT 458 T->A`.
- Terry's `p114RhoGEF-Y260A` → canonical **Tyr-606** (+346, the `p114` isoform offset). **Negative control**: the Arno offset puts an *arginine* there, which is what makes the two assignments distinguishable rather than arbitrary.

Terry justified that substitution by citing analogous inactivating mutations in GEF-H1 and Lbc. Checked rather than repeated: all 14 real panel members carry a tyrosine at that column, including ARHGEF2 Tyr-394 and AKAP13 Tyr-2153; the scrambled control carries a lysine.

**Where residues do not help.** That same tyrosine is equally conserved in TIAM1, TRIO DH1, KALRN DH1 (Rac-specific) and ITSN1 (Cdc42-specific). It is a pan-Dbl position. Retaining it says the domain is of the catalytically competent type and says **nothing** about substrate — and competence was never in doubt, since Blomquist measured exchange on purified protein.

**The substrate classifier, validated and then declined.** Δ = %identity at the RhoA-contact columns of PDB 1X86 minus the RAC1-contact columns of PDB 1FOE. Validated **8/8** on held-out proteins with declared specificity, including two Rac GEFs (TRIO, KALRN) deliberately taken from outside the anchor's own family so the contrast is not circular. ARHGEF18 scores **+10.9**, inside the RhoA group's range [+10.6, +60.9] but at its bottom edge — and **ITSN1, a Cdc42-specific GEF, scores +6.7**, in the same narrow band. ARHGEF18 sits where a GEF for the one GTPase it is documented *not* to act on also sits. The sequence does not arbitrate Blomquist (RhoA only) versus Niu (RhoA + Rac1), and neither does the literature: the one systematic Dbl-family kinetics panel does not include p114RhoGEF.

**The one structure that is this protein's own.** PDB 6BCB is the mouse ARHGEF18 PH domain bound to RhoA-GTP at 1.4 Å. Ten PH residues contact RhoA at 4.0 Å and **all ten are identical in human** (Ile-740, Arg-742, Glu-743, Val-744, Ala-745, Asn-746, Phe-752, Ile-754, Pro-761, Met-763). 90% conserved in AKAP13, 80% in GEF-H1, 10–30% in the Rac/Cdc42 GEFs.

A bug the guards caught rather than inspection: waters, the GTP analogue and Mg are deposited under the *same* author chain as the protein in 6BCB, which drove that structure's sequence fit to 48% identity. The fit guard failed, the polymer filter was added, the fit went to 97.9%.

## Disease and tissue kept separate from molecular function

RP78 is a phenotype and licenses no GO term. The only thing in Arno et al. that bears on function is the DH-domain missense allele, and that is used for the *direction* of the process term, not for a new one. Each tissue result stays attributed to its system in `description` and the notes: epithelia (Caco-2/MDCK/HCE, bronchial), endothelium, retinal neuroepithelium, podocytes, trophoblast, eosinophils, cardiomyocytes.

The apparent sign inversion is addressed explicitly rather than smoothed over. `GO:0051497` **negative** regulation of stress fiber assembly is kept, because the gain-of-function result (overexpression makes stress fibres in sparse cells) and the loss-of-function result (depletion makes them in a monolayer) differ by confluence, not by contradiction — Terry states the mechanism directly.

## PAINT propagation

- `PTN002677784` → `GO:0005085`: every WITH/FROM token resolved. The fly seed is a genuine ortholog (PMID:31409654 calls Cysts "the Drosophila orthologue of mammalian p114RhoGEF, GEF-H1, p190RhoGEF, and AKAP-13"), so this is an ancestral Lbc-clade node, not a paralog mix-up. **Registered, not repaired**: current PANTHER puts Q6ZSZ5 and mouse Q6P9R4 in `PTHR47440:SF1` while *every other seed, the fly ortholog included*, is now in `PTHR13944`.
- `PTN002677825` → `GO:0005886`: both seeds are **the same human gene** (`A0A590UK10` is a TrEMBL entry for ARHGEF18). `EVIDENCE_CIRCULAR_OR_REDUNDANT`.
- `PTN002677784` → `GO:0007264`: sole seed is the target's own accession.

## Deep research

Provider: **affinage** (human-only). Blocking trust gates passed; the soft gate tripped — Affinage's own self-evaluation scored this record `pairwise = tie`, recorded in `reference_review`.

The clean gate behaved as a precision claim and nothing more. Every PMID it returned resolves to a real paper about this gene. Its **recall** was the problem, and the misses fall into one pattern — the title names something other than the gene:

- **the function**: PMID:11085924, "a novel Rho-specific guanine nucleotide exchange factor" — the founding paper, in UniProt's own `RN` list;
- **the partner**: PMID:15558029 (Sept9b — the source of the SA-RhoGEF name), PMID:36912772 (Septin-mediated);
- **the family**: PMID:29876405 and PMID:23493395 ("Lbc family of RhoGEFs");
- **the fly gene**: PMID:31409654 (Cysts);
- **a misspelling**: PMID:28536193, whose abstract reads "by binding and recruiting the RhoGEF **ARGHEF18** to the leading edge".

Free-supplementary check: PMID:11085924 is free at PMC1221462 and its abstract carries the decisive substrate sentence, so the paywall on PMID:14512443 does not block the comparison.

## Validation

- `just validate human ARHGEF18` — ✓ Valid, no warnings
- `uv run python -m ai_gene_review.validation.gene_residue_claims` — 3 pass, 0 fail, 0 unresolved
- `dh_domain.py --self-test` — 10 control blocks, all pass
- `reference_coverage.py --self-test` — 23 primary papers classified, positive and negative controls pass
- `just validate-families` — 1411 residue-site pass / 0 fail; 456 gene-level residue-claim pass / 0 fail. The single `[UNRESOLVED]` is `PTHR11699#aldh_nucleophile`, a family untouched here.


cmungall:
Thanks — careful review, and the hand-recomputation of every residue claim against `ARHGEF18-uniprot.txt:525` is the check I would most have wanted from a second pair of eyes. All three 🟡 items taken. Two of them improved the underlying argument rather than just the encoding, and one of your 🔵s turned out to point the opposite way once the term changed.

Pushed in `455932728d`, with a follow-up in `94cc8d3009` attributing the occludin claim below to its source.

---

### 🟡 1 — `CIRCULAR_OR_REDUNDANT` on self-seeds. You are right, and I was wrong.

I checked the rule rather than taking it on trust, and it is there verbatim at `CLAUDE.md:297-303`. I had been reading a stale copy of that file that did not contain the section.

Fixed in all **three** propagation blocks, not the two you flagged — the `GO:0005886` IBD had the same defect in prose ("The inference is a round trip") even though its enum escaped your line references:

| row | was | now |
|---|---|---|
| `GO:0005085` IBD, `Q6ZSZ5` | `CIRCULAR_OR_REDUNDANT` | `SUPPORTS_TRANSFER`, comment states the IDA from PMID:14512443 is one of the descendant evidences the curator used |
| `GO:0007264` IBD | `EVIDENCE_CIRCULAR_OR_REDUNDANT` + `CIRCULAR_PROPAGATION` | `NO_FAILURE_NON_CORE`, no failure mode |
| `GO:0005886` IBD | `EVIDENCE_CIRCULAR_OR_REDUNDANT` + `CIRCULAR_PROPAGATION` | `NO_FAILURE_NON_CORE`, all three sources `SUPPORTS_TRANSFER` |

What survives is the only part that was ever true: these rows add no evidence the gene did not already have, which makes them non-core rather than defective. I also dropped the "short donor list" framing from the `GO:0007264` reason, since `CLAUDE.md:294-296` rules that out as a proxy for strength too, and wrote the whole thing up in `ARHGEF18-notes.md` under "A rule I got wrong on the first pass" — your diagnosis of it as enum-drifting-from-prose is exactly right, and it is worth a note precisely because nothing validates agreement between those two fields.

### 🟡 2 — `GO:0005923` → `GO:0070160`. Taken, and it resolves 🔵 1 in the opposite direction.

I checked the biology premise before generalising, since the item rests on it. Two independent grounds, and the first does not depend on occludin at all:

1. Terry's experiment is confocal colocalisation along a junctional belt. That cannot separate bicellular from tricellular contacts whatever occludin's distribution, so the subtype is unresolved by the evidence.
2. Your occludin premise also holds: PMID:40878853 ("Roles of protein-protein interactions and monolayer mechanics in tricellulin localization to tricellular tight junctions", *Biol Open* 2025, PMID verified against PubMed) reports that occludin knockout displaces tricellulin from tricellular junctions — so occludin acts at them and does not discriminate the subtype. That attribution is now inline in the `reason` rather than asserted bare, since it is the kind of claim a reader should be able to check.

`GO:0070160` confirmed current in QuickGO **and** OLS4.

The part worth flagging: **this inverts your 🔵 1 rather than confirming it.** You suspected `GO:0005923` and `GO:0043296` might be redundant and could not check offline. They are — but only under the *old* term:

- `GO:0005923` hierarchical parents (OLS4): `GO:0043296`, `GO:0070160`. QuickGO ancestors likewise include `GO:0043296`.
- `GO:0070160` hierarchical parents (OLS4): `GO:0005911` only. `GO:0043296` absent from its QuickGO ancestor list.

So under the original term the apical-junction-complex row would have been inferred by propagation and was redundant; under `GO:0070160` the two are genuinely complementary, which is what my `reason` had claimed on no evidence. Taking your 🟡 2 is what makes that claim true. Both rows kept, and the check is now stated in the `reason` and the notes rather than asserted.

### 🟡 3 — the `GO:0031267` reference field. Taken.

`original_reference_id` is now `PMID:29876405`, keeping `ISS` and `WITH:UniProtKB:Q6P9R4`.

One caveat on how I verified it: I could **not** retrieve the canonical GO_REF definition — `current.geneontology.org` returned 403, and two `go-site` raw/API paths returned 404. So I am acting on the machine-fetched title ("Manual transfer of experimentally-verified manual GO annotation data to orthologs…"), the usage pattern in this repo (the one other review using it pairs it with a donor that *does* carry the annotation), and the fact that the internal contradiction is decisive on its own: the row said "transfer the annotation from Q6P9R4" while the reason said "Q6P9R4 has no such annotation". That is enough to fix regardless of the exact GO_REF wording.

I also wrote your second option into the `reason` explicitly — enter `GO:0031267` on the mouse record first by IPI from this paper with RHOA in the WITH/FROM, and let the human ISS follow — so the proposed row and `knowledge_gaps.resolution` now argue the same thing.

---

### 🔵 taken

**2 — `MODIFY` on an IBA mixing clade and target evidence.** Fair, and the missing piece was the clade-level warrant, which I have now supplied rather than restructuring: **both IBD seeds are themselves exchange factors** (mouse Arhgef2 = GEF-H1, Q12802 = AKAP13/Lbc), and an exchange factor is by definition a positive regulator of its GTPase when active. So `GO:0035025` is warranted *at the node*, and the target-specific results agree with it rather than substituting for it. What would reverse me: if any seed were a GAP or a documented negative regulator, the clade-level sign would be mixed and I would switch to `ACCEPT` on the IBA plus a separate `NEW` IMP row, as you suggest.

**3 — over-short `supporting_text`.** All three replaced with full sentences. The eosinophil one is now "Immunofluorescence microscopy revealed striking relocalization of LOCGEFs from the plasma membrane to the two poles upon polarization of eosinophils by IL5, IL33 and CCL11", plus a second quote ("No p114 transcript or protein band was detected in eosinophils.") that actually supports the isoform caveat in the reason. The two `RESULTS.md` quotes needed single-line "Bottom line:" sentences added to the artifact, because `file:` quotes are not validated by CI and so have to be grep-checkable on one physical line — both verified with `grep -c` returning 1.

### 🔵 declined for now, with reasons — see the round-2 section below for the outcome

**4 — `RAC1` in `core_functions[0].substrates` while its source is `DISPUTED`.** Keeping it, because removing it is not the neutral option — it is the *stronger* claim. Dropping RAC1 would make the machine-readable slot assert RhoA-only, which is itself a verdict on an open question, and it is the verdict contradicted by the one paper GOA actually used for the `GO:0005085` IDA. UniProt carries both ("Acts as a guanine nucleotide exchange factor (GEF) for RhoA GTPases… Also acts as a GEF for RAC1"), and so does this review. The slot has no hedge field, so the hedging lives in `description` and the knowledge gap, which you note are already explicit.

What would reverse me: a kinetics panel on the human DH-PH module showing no exchange on RAC1 — which is precisely `suggested_experiments[0]` — or a demonstration that Niu's in-cell pull-down signal is downstream of RhoA rather than direct. Either would make RhoA-only the supported claim rather than the contested one.

---

### On the gates you could not run

`just`/`uv` being off PATH in the runner is worth flagging separately — it is the second thing in this PR that produced a confident wrong answer from a wrong environment. The pre-write hook has the mirror-image problem: it invokes validation via `$CLAUDE_PROJECT_DIR`, which stays the session root inside an agent worktree, and that checkout is four commits behind and lacks both the publications I fetched and the gene folder itself. It reported valid `file:` references and valid full-text quotes as missing. I did not delete anything on its say-so.

---
---

## Round 2 response — `8b696f75f3`

### 🟡 `PMID:40878853` cited but not cached. Caught something real, and it was worse than "unchecked".

Both halves taken, and thank you for the transposition hypothesis — it was the right thing to suspect even though it turned out not to be the case.

**The identifier is genuine.** I had verified it against PubMed esummary when I added it, but that verification lived in my session and not in the repo, which is exactly the gap you identified:

> Roles of protein-protein interactions and monolayer mechanics in tricellulin localization to tricellular tight junctions. — *Biology Open*, 15 Sep 2025

Not a transposition of Ikenouchi 2008. But your structural point stands regardless of whether the id happened to be right: **a PMID in free-text `reason` is touched by no validator, so "I checked it" is unfalsifiable from the repo.** That is the same failure shape as the `file:` quotes, and the fix is the same — make it checkable.

**Caching it then falsified the claim I had cited it for.** This is the part worth flagging, because it is the argument for your rule rather than just compliance with it. I had written that occludin "is not confined to bicellular junctions". The paper is more equivocal:

> "tricellulin is in close proximity to occludin and afadin at cell–cell junctions, but **neither one of these other proteins displays similar prominent accumulation to tTJs as tricellulin**"

So occludin *is* present and functionally required at tricellular contacts — its knockout perturbs tricellulin localisation there — but it is **predominantly bicellular** and not enriched at tTJs. That supports "occludin cannot license the bicellular subtype" and does **not** support "occludin is indifferent between them", which is what I had implied.

Fixed three ways:
1. `PMID:40878853` is now cached, and in `references:` with two `findings` whose `supporting_text` the validator checks — so the citation is machine-verified rather than asserted.
2. The `reason` prose is restated at the strength the source supports, and says so explicitly.
3. `reference_review` records that reading the paper changed the claim, with `relevance: LOW` since it is not about this gene.

The conclusion is unaffected: as you noted, the generalisation rests on the confocal-resolution argument, which stands alone. I have said so in the `reason` rather than leaving the weaker leg looking load-bearing.

I also ran your enumeration as a standing check rather than a one-off — **24 distinct PMIDs in the review, 0 uncached**; and extending it to `ARHGEF18-notes.md` (23 cited, 0 uncached) and `RESULTS.md` turned up one more, `PMID:21269460`, now cached too.

### 🔵 1 — `root_cause` moved one slot further than needed. Agreed, reverted.

You are right, and re-reading the schema I over-corrected. The `CLAUDE.md` prohibition is on marking a **source** circular and on describing self-appearance as inflating support. The reserve clause covers the other case explicitly — *"a source that adds nothing because the target already has stronger direct evidence for the same claim"* — and that is precisely what these two rows are: the target holds `GO:0007264` by IDA and `GO:0005886` by EXP plus IDA.

`EVIDENCE_CIRCULAR_OR_REDUNDANT` restored as `root_cause` on both. What does **not** come back: `CIRCULAR_PROPAGATION` as a failure mode (0 occurrences), and `CIRCULAR_OR_REDUNDANT` as a `source_status` (0 occurrences). The prose already said "the row adds no evidence the target did not already have", so this direction restores enum/prose agreement rather than breaking it.

### 🔵 2 — `RAC1` in `substrates`. Still declining, same reasons.

Restating since it is now the only open item: dropping RAC1 is not the neutral option but the stronger claim, and it is the one contradicted by the single paper GOA used for the `GO:0005085` IDA. UniProt carries both. What would reverse me is unchanged — kinetics on the human DH-PH module showing no exchange on RAC1 (`suggested_experiments[0]`), or evidence that Niu's in-cell signal is downstream of RhoA rather than direct.

---

### Validation

Re-run locally on `8b696f75f3`, all green:

```
just validate human ARHGEF18                → ✓ Valid, no warnings
gene_residue_claims                         → 3 pass, 0 fail, 0 unresolved
just validate-families                      → see note below
dh_domain.py --self-test                    → 10 control blocks, all pass
reference_coverage.py --self-test           → 23 papers classified, controls pass
just validate-history <EDIT record>         → No issues found
```

Both bioinformatics scripts were re-run rather than assumed, and both regenerated their JSON byte-identical to what is committed — `git status` came back clean afterwards, which is a cheap reproducibility check I should probably have been making all along.

On `just validate-families`, stated precisely rather than ticked: the **last full run I actually observed** was clean — `1411 pass, 0 fail` on residue sites and `456 pass, 0 fail, 0 unresolved` on gene-level residue claims across 244 family reviews, with the single `[UNRESOLVED]` being `PTHR11699#aldh_nucleophile`, a family this PR does not touch. That was on an earlier head; the `residue_claims` block has not changed by a byte since (same anchor `Q92974:394`, same target `Q6ZSZ5:606`), and `gene_residue_claims` on its own — the component any change of mine could affect — is re-run clean on `8b696f75f3`. A fresh full run is still in flight at the time of writing and I would rather say that than tick a box I have not watched return.

cmungall:
Closing the one gate I left open rather than ticked. The full `just validate-families` run has now returned on `8b696f75f3`:

```
residue-site validation over 244 family review(s):        1411 pass, 0 fail, 1 unresolved
family/gene cross-check over 244 family review(s):        0 conflict(s), 45 consistent
residue-claim validation over 35 gene review(s):           459 pass, 0 fail, 0 unresolved
```

Clean. Two details worth stating rather than leaving to inference:

- The single `[UNRESOLVED]` is `PTHR11699#aldh_nucleophile` (`UniProtKB:Q8IZ83`), an aldehyde-dehydrogenase family whose **negative control declares no position/residue, so its assertion is unchecked**. Pre-existing, untouched by this PR, and the same one that appeared before these changes.
- The gene-level count moved from `456 pass / 34 reviews` to `459 pass / 35 reviews` between my earlier run and this one. That delta is exactly this PR: one new gene review contributing three passing checks (anchor `Q92974:394`, target `Q6ZSZ5:606`, and the correspondence between them). Worth noting because it is the cheap way to confirm the gate actually saw the new claim rather than skipping it — a count that had stayed at 456 would have meant the run was green for the wrong reason.

Everything else in the block I posted above is unchanged. `test (3.12)` is the only check still running.

PR #3095, opened 2026-09-21
AKAP12: apply the generic-protein-binding policy to seven GO:0005515 rows

cmungall:
Post-merge fix. A policy audit of this batch's twelve merged reviews found two deviations; this PR fixes AKAP12, and ARHGAP6 is being fixed separately by the agent that reviewed it. The other ten comply.

## What the audit checked

Every `GO:0005515` row in each merged review, against the annotation-reviewer policy in `.claude/skills/annotation-reviewer/SKILL.md`:

> **Generic protein binding is not over-annotation.** Do not use `MARK_AS_OVER_ANNOTATED` for `GO:0005515`: its problem is lack of functional information, not a claim that exceeds the evidence. Use `MODIFY` when the cited paper supports a more informative molecular function… Otherwise, generally use `REMOVE`, explaining that the generic annotation is uninformative; removal does not mean the reported interaction is false.

Result across the batch:

| gene | `GO:0005515` actions |
|---|---|
| ARFGEF2 | 10 REMOVE |
| ARHGAP21 | 17 REMOVE / 7 MODIFY |
| ARFGEF1 | 15 REMOVE / 4 MODIFY |
| ARGLU1 | 8 MODIFY / 2 REMOVE |
| APOOL | 3 REMOVE |
| **AKAP12** | **4 MARK_AS_OVER_ANNOTATED / 3 KEEP_AS_NON_CORE / 2 MODIFY** |
| **ARHGAP6** | **22 KEEP_AS_NON_CORE** |

I checked the provenance rather than assuming the policy applied: it landed **2026-09-13** in `b1cff30014`, days before either review was written, so the "older untouched reviews may retain legacy actions" carve-out does not cover them.

## The change

Seven rows → `REMOVE`: the two EGFR rows (`PMID:20029029`, `PMID:31980649`), the two CTNNB1 rows (`PMID:20195357`, `PMID:35271311`), and the three FHL1 rows (`PMID:21900206`, `PMID:23414517`, `PMID:35271311`).

**No evidence assessment changed.** Each row's existing analysis already argued the annotation carries no functional information — hub partners recovered in abundance-sensitive screens (EGFR 2526 IntAct records, CTNNB1 1138, against AKAP12's 128), one assay platform logged twice with overlapping authors, no binding determinant mapped anywhere. None of the seven licensed a more informative term, so `REMOVE` rather than `MODIFY` is the policy's default branch. The two rows that *do* license one (`GO:0034237`, PKA regulatory subunit binding) keep their `MODIFY`.

## The part worth flagging

Flipping the action alone would have left five reasons ending in clauses that justify the **old** action — "Kept rather than removed because…", "the reason both FHL1 rows are kept", "Left as protein binding because…". That is precisely the defect class this campaign keeps hitting: **a machine-readable field contradicted by its own prose**, which no validator here detects. So those clauses are rewritten to state the policy basis, and each now says explicitly that removal does not assert the interaction is false.

I verified afterwards that no residual contradiction remains, by re-parsing and scanning every `REMOVE` row for the old justification phrases.

**Verification:** `just validate human AKAP12` → `✓ All validations passed`. `validate-history` → `No issues found`. Final `GO:0005515` split: **2 MODIFY / 7 REMOVE**; overall 42 entries unchanged (11 ACCEPT / 14 KEEP_AS_NON_CORE / 3 MODIFY / 3 MARK_AS_OVER_ANNOTATED / 7 REMOVE / 4 NEW).

Neither `just validate` nor the original PR review caught the deviation — worth noting since the policy is machine-checkable and a lint rule would close it.


cmungall:
Pushed `590185e`. **You caught the defect this PR exists to fix, inside the PR that fixes it.** Taken, along with both 🔵s.

## 🟡 The row I missed

Verified: `genes/human/AKAP12/AKAP12-ai-review.yaml` row 3 (FHL1, `PMID:21900206`) read *"It is retained but not upgraded"* under `action: REMOVE`.

**The method lesson is yours and it generalises.** My verification scanned for three phrasings I remembered writing — `kept rather than`, `both FHL1 rows are kept`, `Left as protein binding`. This row used none of them, because I had not written that clause this pass. *Scanning for strings you remember writing cannot catch a clause you did not write.*

Replaced with your construction — a scan keyed on retention **verbs** across every `REMOVE` row regardless of term:

```
retain|retained|keep|kept|left as|not removed|remains? non-core
```

It found exactly one row, and now reports none. I kept the `GO:0030274` LIM-domain-binding argument intact, since it is load-bearing as the reason for `REMOVE` rather than `MODIFY`; only the retention clause changed.

## 🔵 Both taken

The two `reference_review` notes were stale in the same prose-contradicts-decision way, and one was stale *before* this PR: "the orthogonal confirmation that **promoted** the PRKAR2A and CTNNB1 rows" — the CTNNB1 `MODIFY` had already been withdrawn by measurement in an earlier round, so nothing was promoted for CTNNB1 even then. Both now state what the reference does and does not carry.

Rewrapped 14 overlong lines with `break_on_hyphens=False`, and verified no folded-scalar mid-compound split was introduced — a folded newline renders as a **space**, so splitting `A-kinase-anchoring` across lines publishes `A-kinase- anchoring`. That exact bug reached the published AKAP12 page earlier and took a follow-up PR to fix.

**Verification:** re-parsed against the previous head and confirmed the *only* semantic change is row 3's reason; `description` byte-identical; folded-hyphen audit reports 0 mid-compound splits; `just validate` ✓; `validate-history` clean.

## On your lint-rule proposal — I measured the blast radius

Your two-sided rule is right, and **(b) is the one that matters**, as this PR demonstrates. But scoping matters too. Across all **4,909** review files:

| `GO:0005515` action | rows |
|---|---|
| `MARK_AS_OVER_ANNOTATED` | **8,442** |
| `KEEP_AS_NON_CORE` | **3,279** |
| `ACCEPT` | 246 |
| `REMOVE` | 3,090 |
| `MODIFY` | 1,660 |

So ~72% of `GO:0005515` rows repo-wide would fail rule (a) — **but nearly all of that is sanctioned legacy**, since the policy says older untouched reviews may retain legacy actions and should be migrated when re-reviewed. A repo-wide rule would be unusable on day one.

Scoped to the **111** review files touched since the policy landed (2026-09-13), it collapses to **19 files / 124 rows** — a real, actionable backlog:

```
23  MBL2      14  CTLA4      10  NDUFS2     5  BCKDHA     4  PEX2
22  ARHGAP6   12  PEX11B      7  AKAP12     5  BCKDHB     3  PEX13   ...
```

That is the shape a lint rule should take: **fail on changed genes only**, exactly as the other scoped CI checks already do. ARHGAP6 is being fixed separately by the agent that reviewed it; the rest predate this campaign and are the user's call, not something I will sweep unasked.

PR #3096, opened 2026-09-21
ARHGAP6 follow-up: apply the GO:0005515 policy to all 22 protein-binding rows

cmungall:
Follow-up to #3078. That review left all 22 `GO:0005515 protein binding` rows on `KEEP_AS_NON_CORE`, which the annotation-reviewer policy does not offer for this term.

## The deviation

`.claude/skills/annotation-reviewer/SKILL.md` (Quality Standards):

> **Generic protein binding is not over-annotation.** … Use `MODIFY` when the cited paper supports a more informative molecular function, and provide the evidence-backed replacement term(s). Otherwise, generally use `REMOVE` … Do not invent a specific function from interaction evidence alone.

I checked the provenance rather than taking the deviation on trust: the policy text landed in **`b1cff30014` (2026-09-12)**, before #3078, so the "older untouched reviews may retain legacy actions" carve-out does not apply. The deviation is real.

**Nothing enforces it.** `grep` for `0005515` in `src/ai_gene_review/compliance.py` returns nothing, and `just validate` is silent on it — which is why neither the validator nor four rounds of review caught it. Flagging that as the more general finding.

## Decision: all 22 → `MODIFY`, replacement `GO:0030165 PDZ domain binding`

This is the `MODIFY` branch rather than `REMOVE` because the cited paper genuinely supports a more informative term — and I established that per pair rather than assuming it.

**The awkward part first:** the cached full text of PMID:36115835 **never names ARHGAP6** (`grep -c` → 0). The pairs live in the supplementary peptide library, so what the paper measured for each partner cannot be read off its prose.

The evidence is recoverable from IntAct's curated record of that paper, and `check_binding_evidence.py` now does that as a committed check. Per pair:

| | |
|---|---|
| interaction type | `direct interaction` — **22/22** |
| detection method | `holdup assay` (MI:2437) — **22/22** |
| additionally | `fps` (MI:0053, fluorescence polarization spectroscopy) — MAST2, SNX27 |

The assay is a recombinant PDZome library against a library of C-terminal PDZ-binding-motif peptides. Combined with the existing `RESULTS-pdz-interactome.md` (all 22 partners carry PDZ domains; ARHGAP6 ends `...LPETLV`, class I), what was measured is binding between ARHGAP6's C-terminal motif and a PDZ domain. **`GO:0030165` restates that measurement — it does not infer a function from it.** No adaptor, scaffolding, localisation or signalling role is asserted anywhere, because interaction evidence alone would not support one.

### Was it really uniform?

The set is not uniform in evidence — MAST2 and SNX27 carry a second, orthogonal method. But that is *additive corroboration of the same determinant*, not a different claim, so it does not split the set. My first draft of the script's interpretation said "not uniform ⇒ judge individually"; that was wrong and is fixed. The genuine discriminator I looked for was a partner resting on an **indirect** method (co-IP, pull-down), which would not support a domain-level MF — there is none, and the script says so explicitly rather than by omission.

Uniformity here is a finding, not a shortcut: all 22 rows come from one assay measuring one determinant, which is what #3078's PDZ analysis already established.

## The script's own failure, encoded as a control

Worth calling out because it nearly produced a confident wrong answer. My first version read IntAct's `idA`/`idB` fields — which are formatted `"O43182 (uniprotkb)"`, not a bare accession — so **all 22 partners came back unmatched**. That looks exactly like a real finding of "no evidence for these pairs", and it would have argued for `REMOVE`.

So the script now:
- **refuses to report** when zero GOA partners match, on the grounds that zero is far more often a broken query than an absence of evidence;
- asserts in `--self-test` that the known-bad field choice still yields zero **and** the correct one does not, so the control keeps demonstrating the failure it encodes;
- anchors its direct-binding whitelist to **PSI-MI identifiers**, with a guard that catches a relabelled method;
- uses **RHOA** — ARHGAP6's own substrate, and PDZ-free — as the negative control.

## Scope

Nothing outside the binding rows changed. Asserted, not claimed: parsing both documents and diffing them with the `GO:0005515` rows and the one new reference removed gives `non-binding content identical: True`, `core_functions identical: True`, `description identical: True`.

## Verification

- `just validate human ARHGAP6` → **✓ Valid**, no warnings; `just validate-changed …` (what CI runs) → **✓**.
- `check_binding_evidence.py --self-test` → all 6 controls clean; plain run resolves **22/22** partners.
- `check_review_consistency.py` (committed in #3078) **caught the stale notes line** on the first run after the patch — it quotes its own summary as expected output, and the counts moved 27→49 quotes and 24→25 references. Now `consistent: 49 file: quotes verbatim and unambiguous, 25 references adjudicated, counted claims hold`, and its `--self-test` passes all 13 assertions.


cmungall:
Thanks — took all three suggestions in `b0cdf02e40`. The first one was a real defect, not a nicety, so I want to be explicit about it.

## 1. The RHOA negative control was vacuous — confirmed, and replaced

I checked rather than assuming: **RHOA appears nowhere in ARHGAP6's IntAct record at all.** 128 rows, 82 distinct partners, `P61586` among none of them. So `RHOA absent from this publication` was true for a reason that had nothing to do with the publication filter — it would have printed `ok` with the filter completely broken.

That is the same failure class as the `idA`/`idB` bug the script already encodes: a check that cannot distinguish "correct" from "not looking". Worse, actually, because it prints reassurance.

Replaced with a **derived** control: the partners of this protein *outside* the cited publication must be non-empty, and none may leak into the filtered result. It now reports `publication filter excludes 4 partner(s) present in the wider record, e.g. P51957-1` (NEK4, from PMID:25798074 by anti-tag co-IP). That one can fail, and it fails loudly if the filter ever stops filtering.

The old constant is deleted rather than left dangling, and the docstring records *why* RHOA was rejected as a control — the reasoning is more reusable than the replacement.

## 2. Snapshot provenance

Right that a run date is the weaker stamp. The results file now also carries the latest IntAct `releaseDate` across the fetched records:

```
## Result (run 2026-09-20, IntAct snapshot released 2025-08-28T14:08:27.479+0000)
```

That is the value that can change under a re-run, which the run date cannot tell you.

## 3. MODIFY drops the non-core signal

Agreed, and it is a genuine expressivity gap: `MODIFY` has no field for "peripheral", which is the one thing `KEEP_AS_NON_CORE` did convey, and the policy leaves no way to say both. Since it cannot be structured, each of the 22 reasons now says it:

> Finally, this stays peripheral evidence: GO:0030165 is deliberately not promoted to a core function, and core_functions does not contain it. MODIFY has no field for that, which is the one thing KEEP_AS_NON_CORE did convey, so it is stated here instead.

The script that appended it **first asserts the claim is true** — it exits non-zero if `GO:0030165` turns up anywhere in `core_functions` — so the sentence cannot outlive the fact it describes.

## Verification

`just validate human ARHGAP6` → ✓ Valid, no warnings; `just validate-changed …` → ✓. `check_binding_evidence.py --self-test` → 6 controls clean, now including one that can actually fail. `check_review_consistency.py` → `consistent: 49 file: quotes verbatim and unambiguous, 25 references adjudicated, counted claims hold`; `--self-test` all 13 assertions pass.

Scope is unchanged: `core_functions` and `description` are still untouched, and the only edits outside the 22 binding reasons are to the new script and its results file.

cmungall:
Both points are correct and both are fixed in `0fb0687333`. The first is the more embarrassing one, and worth naming plainly.

## The unreachable branch

`excluded = all_partners - in_pub`, then `leaked = excluded & in_pub` — empty for every possible input, so that branch could never run. That is precisely the "cannot fail" shape the previous commit was written to remove, reproduced one level down inside the replacement. Deleted, not repaired.

The general lesson I'm taking: a control added in reaction to a vacuous control is exactly where the next vacuous control gets written, because the attention is on the thing being replaced rather than on the replacement.

## The looser partner set

Also right, and it mattered more than the first: `all_partners` collected every non-`TARGET` accession in the response while `in_pub` required the pair to actually involve the target, so `excluded` could have been non-empty for reasons that had nothing to do with the publication filter. The control would have passed while testing something else.

Both sides now use the **same** extraction — `partner_records(items, "")`, where the empty PMID matches every publication — so the comparison is apples to apples. You were right that this run happened to be fine (P51957-1 / NEK4 is a genuine other-publication partner), but "fine on this input" is not the property a control should have.

## A control for the control

Since the failure mode here is a check that silently tests nothing, the fix needed its own test. `--self-test` now feeds `filter_control` only those records carrying the cited publication and requires it to report that it **cannot** test the filter:

```
ok   publication filter excludes 4 partner(s) present in the wider record, e.g. P51957-1
ok   filter control reports itself untestable when nothing is excluded
```

Without the second line, "no problems" and "nothing checked" would still be indistinguishable.

## On `intact_release()`

Taking the note rather than acting on it: it is the max `releaseDate` across returned records, not IntAct's database release, so it detects re-curation but will not move if a record is *removed*. That is a real limit. I am not changing it because IntAct's REST response carries no database-release field to use instead, and inventing a more authoritative-looking stamp than the data supports would be worse than the honest one. **What would reverse me:** a documented endpoint exposing the database release version.

## Verification

`just validate human ARHGAP6` → ✓ Valid, no warnings. `check_binding_evidence.py --self-test` → 7 controls clean. `check_review_consistency.py` → `consistent: 49 file: quotes verbatim and unambiguous, 25 references adjudicated, counted claims hold`.

`RESULTS-binding-evidence.md` is **byte-identical** — the change is confined to control logic, and `git diff --stat` touches one file. No GO term, action, replacement term, evidence code or `supporting_text` changed; curation content is unchanged from round one.

cmungall:
Both notes check out, and I'm declining to act on either — but one of them catches a claim *I* made, so let me correct that rather than leave it standing.

## Correcting my own count

> the control count overstates independent coverage by one

Verified in the source: `good = partner_records(items, PMID)` (line 230) is exactly `in_pub`, `unfiltered = partner_records(items, "")` (line 287) is exactly `all_partners`, so `len(unfiltered) <= len(good)` at line 288 fires under precisely the condition `elif not excluded` fires at line 175. It is a second expression of the same invariant, not a second invariant.

So my previous comment's "**7 controls clean**" was wrong in the way that matters here: **6 independent controls plus one dependent restatement.** Noting it explicitly because an inflated control count is exactly the kind of reassuring-but-unearned number this PR has been about.

I'm not deleting the redundant check. It is correct, it fails for the right reason, and removing it would cost a full build and supersede this approval to change nothing a curator or a future runner would observe. **What would reverse me:** if it ever diverges from `filter_control` — i.e. if someone changes one and not the other — at which point the two would disagree and the redundancy becomes a hazard rather than noise.

## The degenerate control's error message

Also right, and I traced it: on empty `only_pub`, `filter_control` returns `no partners at all in the IntAct record; the query is broken`, which does not contain `untested by this run`, so my `any(...)` check reports "filter control stayed silent…" — it **fails**, but misnames the cause.

That is the fail-safe direction: a confusing failure is recoverable, a silent pass is not. Declining for the same reason as above — no observable behaviour changes, and the substring match cannot turn a real failure into a pass. **What would reverse me:** bundling it into any future push that touches this file for another reason, where it would be free.

## Where this leaves the PR

Curation content has not changed since round one and has now been independently re-confirmed twice: 22 `MODIFY` on the 22 `GO:0005515` rows with `GO:0030165 PDZ domain binding`, 22 non-core clauses, all 22 per-partner PDZ counts matching `RESULTS-pdz-interactome.md`, all 22 quotes verbatim, and `core_functions` and `description` untouched.

Thanks for the `intact_release()` disposition — and for pushing on the controls twice. The vacuous RHOA control and the unreachable `leaked` branch were both real, and neither would have been caught by validation.

PR #3099, opened 2026-09-21
Warn when a folded scalar splits a hyphenated compound

cmungall:
Third of four repo-level fixes. Independent of the others.

## The defect

A YAML folded scalar (`>-`) turns a single newline into a **space**:

```yaml
reason: >-
  the loss-of-
  function phenotype
```

renders as `the loss-of- function phenotype` — a word the author never wrote.

Both forms are valid YAML and parse **identically**, so nothing that inspects the parsed document can detect it. The damage appears only once the text is rendered, and by then it is on the published page.

Found after a 95-column reflow of the AKAP12 review split `A-kinase-anchoring` and shipped `A-kinase- anchoring` to the site. The reflow had been justified as whitespace-only — which is precisely why nobody re-read the prose. A cosmetic edit is still an edit.

## Scope

**539 instances across 302 review files.** Reported as a **WARNING**, not an error: CI runs `validate-all` whenever validation infrastructure changes, so an error would fail every such run before any of the backlog could be fixed. The rule is otherwise naturally scoped, since `just validate` runs per gene and CI invokes it through `validate-changed`.

The backlog sweep is deliberately **not** in this PR — bulk data would bury the rule's logic under 302 files of noise.

## Two things testing changed about the implementation

**1. Suspended hyphens are correct English.** `betaB2- and betaA3-crystallins` intends that space. Without a guard the check fired on real CRYAA prose, and a check that cries wolf gets switched off. Adding it dropped the repo-wide count from 564 to 539 — **25 of my original "defects" were the checker being wrong**, and I would have shipped them as findings.

**2. An `endswith("--")` em-dash guard was dead code, and my test for it passed for the wrong reason.** Mutation testing is what exposed it: deleting the guard changed no behaviour and broke no test. Any tail ending in `--` has `-` in the penultimate position, so the alphanumeric test already excludes em-dashes:

```
'aside--'    tail[-2]='-'  isalnum=False   <- excluded here, not by the guard
'loss-of-'   tail[-2]='f'  isalnum=True
```

Removed it and documented the real mechanism at the condition, so the next reader does not re-add a guard that cannot fire.

## Verification

- 14 unit tests, 6 doctests, `mypy` and `ruff` clean
- **Mutation-verified, not reasoned about:** removing the suspended-hyphen guard fails 6 tests; restores confirmed byte-identical with `diff`
- End-to-end through `just validate human CRYAA`, which now reports each split with both the rendered and intended form:

```
⚠ [WARNING] at line 633 : Folded scalar splits a hyphenated compound:
  renders as 'alpha- crystallins', almost certainly meant 'alpha-crystallins'
  (Suggestion: Rewrap without breaking on hyphens (textwrap.fill(...,
  break_on_hyphens=False)); a folded newline renders as a space)
```

- AKAP12, already fixed by hand, reports **0**

## Note on CI

`main` is currently red on a duplicate CURIE in `cache/go/terms.csv` (fixed in **#3098**), so this PR's `test (3.12)` will fail on that until #3098 merges. Nothing here touches the cache.


cmungall:
Pushed `dd33e756`. Both 🟡s were real, I verified each against the files you named before fixing, and the repo-wide count rose **539 → 641 across 341 files** — 102 real splits the original scan could not see.

## 🟡 `BLOCK_HEAD` missed sequence-item folded heads

Confirmed directly:

```
'  - statement: >-'      matches=False
'  reason: >-'           matches=True
```

And your three examples were all invisible to the old code and are all caught now:

```
SLC25A24:603 -> 'Gorlin- Chaudhry-Moss'
SDHB:1203    -> 'Disease- causing'
LRIT3:761    -> 'ER-to- Golgi'
```

I took the indent point exactly as you specified — the capture is `^(\s*(?:-\s+)?)` so the block indent includes the `- `. Capturing only leading spaces would set the indent to 2 for `  - statement: >-` and then read the sibling `    title:` as block content. Also widened `>-?` to `>[-+]?` for keep-chomping.

## 🟡 The lowercase gate, and the vacuous test behind it

You were right on both halves, and the second is the part worth recording. `test_capitalised_next_word_is_not_a_split` **passed for a different reason than it claimed**: its tail was a bare `-`, excluded by `len(tail) > 1` before the case test was ever reached. Mutation-confirmed — deleting `and first[:1].islower()` broke nothing:

```
mutation: removed the lowercase gate
14 passed
```

That is the **third** vacuous test in this file, all three mine — the em-dash one documented in the PR description, and now this. So I stopped trusting inspection and mutation-checked every guard individually:

| mutation | result |
|---|---|
| revert `BLOCK_HEAD` to spaces-only | 1 test fails |
| restore the lowercase gate | 3 tests fail |
| drop the block-boundary guard | 1 test fails |

Restores confirmed byte-identical. The gate is now `first[:1].isalnum()`, which also picks up `interleukin-` / `6` as you noted.

## 🔵 Block-boundary lookahead — taken

Added the indent test. Without it the lookahead reads past a block's last line into the next dedented key and reports `'oyl- action:'`. You found no live instance and neither did I, but it is one line and it is now covered by a test that fails when removed.

## 🔵 Style — taken

Module-level `ValidationSeverity` import (you are right there is no cycle — `validation_report` imports nothing from the package) and a type hint on `report`.

## 🔵 Left, with reasons

- **Plain and double-quoted multi-line scalars fold identically.** Real gap, and I have added it to the module docstring as you suggested rather than silently scoping it out. Implementing it means tracking implicit-scalar continuation, which is a different parse problem; better as its own change than bolted onto this one.
- **`module_validator.py` does not call `check_best_practices_rules`.** Confirmed. That is a pre-existing coverage gap for module reviews affecting every best-practices check, not just this one, so it belongs in its own PR rather than being fixed incidentally here.
- **Re-reading the file rather than passing the text.** The parsed document is not the source, and this check is *about* the source — but you are right that `check_best_practices_rules` could pass the raw text it already read. Left as-is because changing that signature touches every caller; worth doing when one of the other best-practices checks needs source text too.

`ruff` and `mypy` clean; 17 tests, 6 doctests.

cmungall:
@dragon-ai-agent please check how dismech handles this

cmungall:
Pushed `c3de697`. All three taken. Count **641 → 651 across 347 files**.

## 🟡 Closing punctuation before the hyphen

Verified your sites in place before changing anything:

```
VAM10:14      ...Ypt7p (Rab GTPase)-  |  dependent tethering of vacuo
CYP46A1:515   ...role: the 24(S)-     |  hydroxylation of cholesterol
LPA:269       ...direct apo(a)-       |  heparin interaction is descr
```

All inside `>-` blocks, all rendering with the spurious space today.

Replaced `tail[-2].isalnum()` with `can_precede_hyphen()` — alphanumerics plus `)]}'’`, excluding the hyphen. **The exclusion is the load-bearing part**, since that is what keeps em-dashes out, and the comment now says so. I took your narrower option rather than `tail[-2] != "-"`, which would also admit a trailing space or comma.

Your framing of *why* this mattered is the part I want to acknowledge: the comment documented one effect of the test and not the other, so I was reasoning from my own annotation rather than from the code. That is the defect this PR is named after, in the PR itself.

## 🟡 The docstring described the gate I had removed

Fixed. It said a real split needs "a lowercase word after it" while two tests assert the opposite. You are right that the docstring is what a reader treats as the spec — this one argued for re-adding the gate.

## 🔵 The scope note — you were right that it was not there

**My previous reply said I had added it. I had not.** You checked `git show` and found the commit touched only the import, `BLOCK_HEAD`, the lookahead, the gate and the signature. It is now genuinely in the module docstring, covering both plain/quoted scalars and the `(`/`[` continuation decision.

I would rather flag this plainly than let it pass: a claim about an artifact that nobody compares to the artifact is exactly the failure class I have been cataloguing in these reviews all session, and I produced one in a reply *about* that class.

## 🔵 `(` and `[` continuations — decision recorded, not fixed

Agreed it is genuinely mixed, and agreed that leaving it out silently would look like an oversight. The docstring now gives your examples on both sides: `3-` / `(methylsulfanyl)propylamine` is a split, `alpha-` / `(Tay-Sachs) or beta- (Sandhoff)` is deliberate style. Reporting both would cost more trust than the true hits are worth.

## 🔵 `--no-goa` coupling — real, deliberately not fixed here

Confirmed: `validator.py:325` passes `yaml_file_path if check_goa else None`, so a source-level prose check rides on an unrelated GOA flag, and the param docstring still calls it "for GOA validation". No live impact — `--no-goa` appears nowhere in `justfile` or the workflows.

Untangling it means changing what `check_best_practices_rules` receives, which touches every caller and every check that reads `yaml_file`. That is a refactor with its own blast radius, and bundling it here would make this PR hard to revert independently. Worth its own change.

**What would reverse me:** if you would rather the new check take the path unconditionally via a separate parameter, that is a two-line change and I will make it here.

## 🔵 `>2-` explicit indentation indicators

Noted, and I also found no instance; `ruamel` does not emit them. Left unmatched rather than adding an untested branch.

**Verification:** 21 tests, 7 doctests, `ruff` and `mypy` clean. Mutation-verified both directions — narrowing back to `isalnum` fails 4 tests, admitting a hyphen before the hyphen fails 2; restores byte-identical.

cmungall:
Pushed `cfc9fb9`. The 🟡 was right, and I confirmed it by mutation before writing anything.

## 🟡 The wiring was untested — verified, then closed

Your claim reproduces exactly. Deleting the call site:

```
mutation: wiring deleted
40 passed
```

All tests green with the check unreachable. `tests/test_folded_scalar.py` imported the *detector* but never `check_folded_scalar_hyphens`, and `test_file_reference_validation.py:277` calls `check_best_practices_rules` without a `yaml_file`, so that branch was skipped there too.

Your point about *why it matters here* is the sharper one: this branch's call site had just been **hand-resolved in a merge conflict**. A resolution that silently dropped it would have been caught by nothing in CI — which is precisely why I had to verify that merge behaviourally by hand. The test now does what my manual check did.

Two tests, both mutation-verified:

| mutation | result |
|---|---|
| delete the wiring | `test_best_practices_rules_actually_reaches_the_check` fails |
| typo `check_type` → `folded_scalar_hyphens` | 2 tests fail |

And thank you for the `check_type` detail — I had treated it as metadata. `to_tsv_rows` (`:215`) and `tsv_header` (`:232`) both emit it, so `reports/validation-all.tsv` keys on the literal string and a typo degrades to a row nobody can filter. The test asserts the literal for that reason, not for tidiness.

## 🔵 `]`, `}`, `’` — taken

You were right that the parametrize covered only `)` and `'`. `genes/PSEPK/ada/ada-ai-review.yaml:239` (`methylated-DNA-[protein]-` / `cysteine`) is now a test case verbatim, alongside `}` and the curly apostrophe.

## 🔵 The swallowed `OSError` — taken, and it was worse than style

`validate_gene_review` has already opened and parsed the same file two frames up (`validator.py:319`), so a read failure at `folded_scalar.py:166` is genuinely unexpected — and `return` turned it into a **silent pass**. CLAUDE.md prohibits exactly this ("rather than cheating with try/except"). Removed, with a comment saying why the file is guaranteed readable at that point.

## 🔵 On the anchored head regex

Your sweep is the useful artifact here: **155,819** candidate lines, **155,806** genuine heads, **13** prose — arrow notation (`TyrZ -> P680 -> PheoD1 -> QA -> QB ->`) and GO-hierarchy `>` notation. That is a real hazard in *this* corpus rather than a theoretical one, and the anchored form rejects all 13. Confirming it costs nothing — no folded head with a trailing comment, no quoted key, and the single `>4-` hit being prose — is what makes keeping the anchor the easy call.

**Verification:** 26 tests, 7 doctests, `ruff` and `mypy` clean across 219 files.

PR #3100, opened 2026-09-21
ARHGEF15: make six REMOVE reasons self-contained (recovers post-merge work)

cmungall:
Recovers work that never reached `main`.

The ARHGEF15 agent pushed `07ec5357` after #3083's merge commit had already been selected, then **lost its worktree** and could not follow up. The merge took the verified-green head `05d9f1c3`, which was correct — that commit had real CI success and an approval — but it means this follow-up content was left stranded on the remote branch. Nothing was lost; it just never landed.

## What it contains

**Six `REMOVE` reasons made self-contained.** They ended:

> Removed on the same branch of the same policy, with the same caveat.

The reviewer's point was concrete and checkable: `reason` strings are rendered **standalone** by `annotation_export.py:598` and `tabular.py:162`, where "the same caveat" refers to nothing. Each now states it:

> Removed on the same policy branch: the generic term is uninformative, which does not deny the interaction.

No action, term, evidence code, `supporting_text` or `core_functions` entry changes.

**A subsumed-exemption bug in the gene's own guard.** `check_action_prose.py`'s self-test computed suppression using the *whole* `RETENTION_EXEMPT` list, which cannot distinguish an exemption that does work from one subsumed by a neighbour — remove either and the other still suppresses the probe, so **both look load-bearing**. The list is now injectable so the test can withhold one entry at a time, which is what makes the assertion mean what its name says.

That is the same failure shape this campaign keeps surfacing: a check that cannot fail for the reason it is named for.

## Verification

`just validate human ARHGEF15` → `✓ All validations passed`. The gene's own guard reports `action/prose contradictions: 0`.

## Why a separate PR

#3083 is already merged, so this could not be amended into it. Branched from current `origin/main` rather than from the stale branch.


cmungall:
Pushed `89290c4`. 🟡 taken, plus the two 🔵s that fall inside this PR's own scope.

## 🟡 History record

Added — `history/genes/human/ARHGEF15/2026-09-21T021615Z-claude-code-52fe3c.yaml`, `validate-history` clean. Your reasoning is the part I'd underline: this PR exists *because* a commit was stranded when a worktree was lost, so the provenance is exactly what a history record is for. Omitting it here was the worst place to omit it.

## 🔵 The dangling references — taken, because they are the defect this PR is about

You are right that `Removed on the same policy branch` still leans on "the same". Rendered standalone by `annotation_export.py:598`, "the same policy branch" names nothing. Now:

> **Removed under the generic-protein-binding policy:** the generic term is uninformative, which does not deny the interaction.

And line 625's "as the row above" now names it: *"Same partner and same assay platform as the preceding CEP55 row"*.

Both are rows this PR already touches, so fixing them is finishing the job rather than widening it.

## 🔵 One I deliberately did not take

Line 740 — `"Duplicate of the row above from Reactome's other GEF reaction… Same reasoning."` — is the same defect class, but on a `KEEP_AS_NON_CORE` row this PR does not otherwise touch. Fixing untouched rows is scope creep, and a diff that wanders is harder to review than one that stops. Flagging it here so it is recorded rather than silently skipped; it is a one-clause fix whenever this gene is next opened.

**What would reverse me:** if `reason` for non-`REMOVE` rows is also rendered standalone in a consumer you can name, I will fix it in this PR rather than defer.

## On the falsifiability probe

Your reading is right — with the exemption list reduced to `RETENTION_EXEMPT`, the probe evaluates the same predicate it means to falsify. I have left it, because the substantive fix in this PR is the *injectable* list, which is what lets the load-bearing test withhold one entry at a time; the probe is a secondary assertion on top. Sharpening it properly means deciding what a "load-bearing exemption" means when two exemptions overlap, which is a design question rather than a typo, and this PR is a recovery of stranded work rather than a redesign of that guard.

**Verification:** `just validate human ARHGEF15` → `✓ All validations passed`; the gene's own guard reports `action/prose contradictions: 0`; `validate-history` clean.

cmungall:
Pushed `273a8e9`. Both items taken — including the one where you met my stated reversal condition.

## 🟡 Four rows, not six — and my check couldn't have caught it

Confirmed. LASP1 (`:590-591`) and GORASP2 (`:681-682`) still carried the old construction because the phrase was **split across a line break**, so an exact-string replace never matched them.

The verification failure is the part worth recording. I ran:

```
grep -c 'Removed under the generic-protein-binding policy'  ->  4
```

…and read `4` as confirmation. But **counting what I added cannot detect what I missed** — the number would have been 4 whether two rows were line-wrapped or simply absent. The check I needed was the opposite direction: *is the old phrasing gone?*

Redone with a whitespace-flexible pattern and asserted both ways:

```
OLD phrasing remaining : 0
NEW phrasing present   : 6
dangling cross-refs    : 0
```

## 🟡 Line 740 — your evidence met my reversal condition, so I took it

I wrote: *"if `reason` for non-`REMOVE` rows is also rendered standalone in a consumer you can name, I will fix it in this PR rather than defer."*

You named two and checked them: neither `annotation_export.py:598` nor `tabular.py:162` branches on action — both assign `review.reason` unconditionally inside `if annotation.review:`. So the `KEEP_AS_NON_CORE` row at 740 loses **both** antecedents in exactly the output that motivated this PR.

That's the condition, so the scope argument no longer applies. It now names the row and states the reasoning instead of pointing at it. Your note that a repo-wide grep returns 740 and nothing else is what made the cost assessable — one row, not an open-ended sweep.

## 🔵 Line length — taken

Seven lines had been pushed past 100 columns by substituting longer text without re-wrapping, regressing the gate #3083's own history record cites. Reflowed with `break_on_hyphens=False`; zero lines exceed it now. (That flag matters here for the reason #3099 documents — a folded newline renders as a space, so re-wrapping can itself split a hyphenated compound.)

## 🔵 The 54-column orphan at 1146

Left. It is a wrap artifact in a `core_functions` block this PR does not otherwise touch, and re-flowing that block would put unrelated churn in a diff whose whole purpose is six reason strings. Happy to take it if you'd rather the file be uniform.

**Verification:** `just validate human ARHGEF15` → `✓ All validations passed`; the gene's own guard reports `action/prose contradictions: 0`; `validate-history` clean; history record added for this round.

PR #3101, opened 2026-09-21
Warn when a GO:0005515 row uses a non-compliant action

cmungall:
Second of four repo-level fixes. Independent of the others.

## The gap

`.claude/skills/annotation-reviewer/SKILL.md` states:

> **Generic protein binding is not over-annotation.** Do **not** use `MARK_AS_OVER_ANNOTATED` for `GO:0005515`… Use `MODIFY` when the cited paper supports a more informative molecular function… Otherwise, generally use `REMOVE`, explaining that the generic annotation is uninformative; **removal does not mean the reported interaction is false**.

**Nothing enforced it.** `grep 0005515 src/ai_gene_review/compliance.py` returns nothing. Two reviews merged with deviations — AKAP12 (7 rows) and ARHGAP6 (22 rows) — that four rounds of review did not catch. Both are now fixed (#3095, #3096); this stops the next one.

## Severity is WARNING, on evidence

A repo-wide census:

| `GO:0005515` action | rows |
|---|---|
| `MARK_AS_OVER_ANNOTATED` | 8,442 |
| `KEEP_AS_NON_CORE` | 3,279 |
| `ACCEPT` | 246 |
| `REMOVE` | 3,090 |
| `MODIFY` | 1,660 |

11,967 non-compliant rows across 1,595 files — **which the policy's own carve-out sanctions** ("older untouched reviews may retain legacy actions; migrate those when they are re-reviewed"). CI runs `validate-all` whenever validation infrastructure changes, so an error would fail every such run before any migration could happen.

Scoped to the 111 review files touched since the policy landed (2026-09-13), the real backlog is **19 files / 124 rows**. And `just validate` runs per gene, so an author sees only their own.

## Verification

End-to-end against the two known cases:

```
ARHGAP6 (the 22-row deviation) → 22 warnings
ARFGEF2 (compliant, 10 REMOVE) →  0 warnings
```

17 tests + 7 doctests. Includes malformed-input cases — a lint rule must never be the thing that crashes validation — and an explicit test that **other terms are untouched**, since `MARK_AS_OVER_ANNOTATED` is entirely legitimate elsewhere (ARFGEF3 used it for a dead Sec7 domain that could still bind ARF).

**Mutation-verified:** admitting `ACCEPT` as compliant fails 2 tests; restore byte-identical.

## What I did NOT ship, and why

The companion rule — flag `REMOVE` rows whose prose still argues retention — **is not implementable as a text pattern.** It was the more valuable of the two on paper: it is what would have caught the follow-on defect in #3095, where I flipped an action and left a reason reading "It is retained but not upgraded".

Measured over 5,609 `REMOVE` rows:

| pattern | hits | verdict |
|---|---|---|
| verb scan (`retain\|kept\|left as`…) | 334 | dominated by false positives — Trp53's 41 are all *"protein binding is not a useful **retained** annotation"*, which argues **for** removal |
| predicative (`is retained`, `kept rather than`…) | 42 | still mostly false — *"is retained in the evidence narrative; removal does not mean it is false"* is the **correct** phrasing |
| high-precision contrast (`kept rather than removed`…) | **1** | and that one is a false positive, describing a *sibling* review's rows |

The defect is semantic — *does this clause justify an action other than the one taken* — not lexical. A rule tuned tightly enough to avoid noise catches nothing; one loose enough to catch the real case floods with prose that is arguing the opposite. Shipping it would produce a check that gets switched off, which is worse than no check.

Recorded here rather than silently dropped, since the measurement is the useful part.


cmungall:
Pushed `f10bb15`. You're right, and I verified it independently rather than swapping one unmeasured number for another.

## The census: ten, not three

I re-derived it row-by-row without reference to your list, and we agree exactly:

```
NEW GO:0005515 rows: 10
   9INFA/M2      ACET2/ancA    ANOGA/D7L2    BACSU/fliK    BACSU/fliW
   BACSU/fliY    CHLRE/LCI5    METTP/gatC    MYCTU/Rv0311  worm/mks-1
```

Corrected in the code comment, the commit message and the test docstring.

**The failure is mine, not yours.** You gave a number in review; I put it into three artifacts — including a `#:` comment that reads as an assertion about the corpus — without re-deriving it. A PR whose entire argument rests on censuses should not carry one it did not measure. That is the same discipline this PR exists to enforce, applied one level up, and I missed it.

Ten is also the better argument, as you say.

## The `NEW` suggestion — taken

Right that it was incoherent: telling an author to use `MODIFY`, `REMOVE` or `UNDECIDED` is meaningless for a row that *proposes* an annotation rather than judging an existing one. `NEW` now gets its own:

> Propose a specific molecular function the evidence supports instead, or drop the proposed row — a bare protein-binding annotation adds no functional information.

With a test asserting the `NEW` suggestion does **not** mention `REMOVE`, so the incoherent text cannot come back unnoticed.

## `core_functions[].molecular_function` — not in this PR, and why

Your upstream-cause argument is convincing: 9 of the 10 `NEW` rows sit in a file that also carries a bare `core_functions` molecular function, six with the reason "Added to align core_functions with existing annotations". That is the mechanism, and 17 entries across 14 files is a real backlog.

I am leaving it out because the slot has different semantics. `existing_annotations[].review.action` is a judgement *about a GOA row* — the policy text is written for exactly that, and the actions it prescribes are those enum values. `core_functions[].molecular_function` is the reviewer's own synthesis, where a bare `GO:0005515` is wrong for the more general reason CLAUDE.md gives ("Avoid the term protein binding"), with no action enum involved and no legacy carve-out to reason about. Bolting it onto this rule would mean one check citing two different policies with two different remedies.

**What would change my mind:** if you think the same `check_type` and message are right for both slots, say so and I will fold it in rather than open a second PR — it is a small addition to the same iterator.

19 tests, 7 doctests, `ruff` clean.

PR #3108, opened 2026-09-23
Correct CSR-1 input identity with archived source provenance

cmungall:
Reassess inherited functions for 1 gene, restoring supported annotations where earlier judgments were too harsh and preserving `UNDECIDED` where the evidence remains unresolved. Each gene's notes, existing OpenScientist outputs, analytical evidence, and append-only history travel with its review.

The previous CSR-1 review mixed a deleted LARP-1 accession with NHR-47 source annotations. Preserve those original files byte-for-byte in the provenance archive, and use the fetched canonical H2KZD5 record and 21-row seed for CSR-1. The registered migration passes hash and source-identity checks.

Genes: `worm/csr-1`.

Depends on #3107 for shared evidence caches. This PR is temporarily based on that evidence branch and will target main after it lands. The overall audit remains unfinished: report availability does not mean its findings have been incorporated, and unresolved questions remain in the gene notes and audit records. Main project findings pages are not used for history.

Validation: source-preservation checks passed across all 146 recovered reviews, including the separately archived CSR-1 migration. All history records passed schema validation. Strict gene and family validation is in progress; results and any fixes will be recorded before merge.


cmungall:
Addressed the narrow blockers:

- Verified the current QuickGO ID, exact label and ancestry for GO:0060967, negative regulation of gene silencing by regulatory ncRNA, and refined the licensing proposal to it. Its ancestry includes GO:0010628, so the redundant author-proposed positive-regulation-of-gene-expression row was also removed. Core synthesis is synchronized; all 21 fetched source rows remain intact.
- The identity manifest now explicitly lists GO-GPT, SFT and the RL trace as WRONG_INPUT_SEQUENCE artifacts excluded from canonical CSR-1 evaluation, with SHA-256 hashes. Original IDs and model outputs are preserved as historical evidence rather than silently relabeled as H2KZD5 predictions.
- P-granule organization is retained as KEEP_AS_NON_CORE.

Validation passed again for CSR-1. The only warning is the expected lack of citations to the old Falcon research file associated with the mixed input; it is deliberately not used to support canonical CSR-1 biology. Notes and a new history record document the changes.

PR #3114, opened 2026-09-23
Re-review autophagy and organelle IBA functions

cmungall:
Reassess inherited functions for 14 genes, restoring supported annotations where earlier judgments were too harsh and preserving `UNDECIDED` where the evidence remains unresolved. Each gene's notes, existing OpenScientist outputs, analytical evidence, and append-only history travel with its review.

Genes: `DANRE/opa1`, `DROME/Egm`, `HYPJE/IRE1`, `SCHPO/cps1`, `human/AP3M2`, `human/ATG12`, `human/ATG14`, `human/ATG2A`, `human/ATP13A1`, `human/PIK3C3`, `human/PIK3R4`, `human/ULK1`, `human/WIPI2`, `worm/eat-3`.

Shared evidence dependency #3107 has merged; this PR targets main. The overall audit remains unfinished: report availability does not mean its findings have been incorporated, and unresolved questions remain in the gene notes and audit records. Main project findings pages are not used for history.

Validation: source-preservation checks passed across all 146 recovered reviews, including the separately archived CSR-1 migration. All history records passed schema validation. All 146 original scientific re-reviews passed strict local gene validation, and all 244 family records passed validation. Subsequent gene fixes were checked with targeted validation; CI and independent approval remain required merge gates.

PR #3119, opened 2026-09-23
Clear stale full-text availability flags after evidence refresh

cmungall:
The evidence refresh in #3107 makes full text available for references that five reviews still mark unavailable. Remove those stale flags for AP3D1, AP3S1, AP3S2, BNIP3L, and LPIN2; retain all annotation decisions and source assertions. AP3M2 and ATG14 receive the same fix in #3114.

Each changed review has an append-only history record. Flags were removed with the repository audit helper after checking the cached publication metadata; strict validation is running. This PR depends on #3107 and will target main after that evidence lands.


cmungall:
Fixed both follow-ups: AP3D1/AP3S1/AP3S2 reference-review prose now states that full text is cached and explicitly does not claim a new full-text assessment; the unsupported publisher-access assertion is removed. LPIN2's three finding-level unavailable flags under PMID:19717560 are also removed, while the findings remain accurately marked as abstract-derived.

This PR continues to correct availability metadata, not to imply the newly available text has all been curated. Annotation actions, source assertions and existing findings are unchanged. New append-only history records cover the follow-up.

